PR #1, opened 2015-01-19
Setup building in travis. Fixed some and added new tests around attaching to channels

thevixac:


mattheworiordan:
@thevixac I have added Travis support now, so will try and kick off a build to see what happens.  One thought, should we be testing against multiple versions of iOS, I believe Travis does support this?

PR #12, opened 2015-05-14
ably-ios 0.8 started. ARTRest and ARTRealtime constructor changes.

thevixac:
-ARTRest and ARTRealtime constructors now return an error callback for bad construction.
-ping and close work correctly
-extra tests for realtime connection state transitions
-extra tests for realtime presence enter and update


mattheworiordan:
@thevixac I have added comments throughout, however few bigger thoughts:

* I am concerned that the API naming is quite different to other client libraries.  Why is this? Can we not,  where feasible, keep the naming & arguments similar?
* It seems callbacks are used with error states as opposed to exceptions. Whilst we use callbacks to for any asynchronous failures, for failures that we can spot immediately, we raise exceptions.  Is there any reason we cannot do this?
* We are often not using an ErrorInfo or Ably Error Exception like class with any consistency with our common error states, see https://github.com/ably/ably-common/blob/master/protocol/errors.json

PR #15, opened 2015-06-11
Presence Map, Fallback and Exceptions

thevixac:


mattheworiordan:
@thevixac I have reviewed most of this, although not looked at the test code in detail.  However, this looks very good, really great progress.

Is the spreadsheet of features / tests up to date?

I have sent quite a bit of feedback in your commits directly

thevixac:
@mattheworiordan , yes the spreadsheet is currently up to date. Will work on the feedback next.

mattheworiordan:
Ok, I'll wait for your review of the feedback and lets take it from there.

PR #16, opened 2015-08-10
BDD specs

fealebenpae:
This is a large pull request, but it sets up the foundation I need. There are two main things it does: some small improvements in the way the library is consumed (adds an umbrella header and a modulemap which all frameworks should have) and the initial specs for the RestClient initialization and logging, along with the attendant test infrastructure.

I've chosen BDD so I can more closely align the tests to those of the Ruby library. Going forward, I have to migrate the existing tests from the XCTest framework to the Quick framework, as well as annotate each test with the code of the part of the spec it verifies.


mattheworiordan:
Looking very good @fealebenpae, great progress.

PR #17, opened 2015-08-11
RestClient#stats

fealebenpae:
Add a new -[ARTRest stats:callback:] method that replaces the old stats API. Along with it comes ARTStatsQuery that describes query options such as direction, limit, etc.

Add tests for it all.

### DIFF
--- a/ably-ios/ARTHttpPaginatedResult.h
+++ b/ably-ios/ARTHttpPaginatedResult.h
@@ -10,13 +10,23 @@
 #import <ably/ARTHttp.h>
 #import <ably/ARTPaginatedResult.h>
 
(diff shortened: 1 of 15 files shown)


mattheworiordan:
This looks really good.  :+1: 

There seems to be a slight deviation on the spec naming convention.  If the language warrants it because it's idiomatic then that is fine, but in this case I am not sure it is.  For example, it seems you support `getNext` whereas we simply use `next`, and I don't see any `isLast` function either.  See http://docs.ably.io/client-lib-development-guide/features/, section PaginatedResult under Types.

fealebenpae:
Oh, yeah, I forgot about `isLast`. I'll add it and rename the methods on `ARTPaginatedResult` as well.

PR #23, opened 2015-09-28
Merged Yavor work, refactoring and RSA1 tests

ricardopereira:


mattheworiordan:
Thanks for the WIP update.

ricardopereira:
**Little update**
I was implementing the test for `RTC1e`(realtimeHost and restHost changes) and I faced some problems:

1. Weird test results;
2. No error when the WebSocket tries to connect an invalid host.

**tl;dr**
⚠️ **Important rule for async tests:**
We must ensure the test does not finish until the asynchronous call is done and the best solution for those cases is: `XCTestExpectation`.

----

I suspected the problem was the asynchronous code. So I created a new project and made some tests.

**First test:**

```
describe("quick") {
  it("does it performs well with background work") {
    dispatch_async(dispatch_get_main_queue()) {
        expect(false).to(equal(true))
        NSLog("Done")
    }
  }
}
```

And my surprise was: `Test Case '-[ProjectTests.ProjectTests quick__does_it_performs_well_with_background_work]' passed (0.002 seconds).`

I need something in order to block the test from finishing before the async code does

**Second test:**

```
describe("async") {
  it("does it performs well with background work") {
    var done = false

    dispatch_async(dispatch_get_main_queue()) {
        expect(false).to(equal(true))
        NSLog("Done")
        done = true
    }

    // Wait until finished
    while !done {
        CFRunLoopRunInMode(kCFRunLoopDefaultMode, CFTimeInterval(0.1), false)
    }
  }
}
```

Result: `[ProjectTests.ProjectTests quick__does_it_performs_well_with_background_work] : failed - expected to equal <true>, got <false>`

Success!
Now the test works with asynchronous code but I think using a `while` loop isn't the best approach.

I investigate more about it and Xcode testing suite has a nice feature for those cases 🙈.
The code:

```
describe("async") {
  it("does it performs well with background work") {
    let expectation = self.expectationWithDescription("async")

    CFRunLoopPerformBlock(CFRunLoopGetMain(), kCFRunLoopCommonModes) {
        expect(false).to(equal(true))
        NSLog("Done")
        
        expectation.fulfill()
    }

    self.waitForExpectationsWithTimeout(10.0, handler: nil)
  }
}
```

mattheworiordan:
@ricardopereira FYI, in Ruby we had similar issues in regards to ensuring the test waits before completion.  Fortunately, Ruby is quite flexible, and what I was able to do was wrap our test blocks / closures with an EventMachine block that will wait until the EventMachine reactor (evented system) is explicitly killed, see https://github.com/ably/ably-ruby/blob/master/spec/support/event_machine_helper.rb#L67-L104 if it means anything to you.

I also recently realise that we were not closing connections explicitly automatically at the end of tests, so added https://github.com/ably/ably-ruby/commit/1abbc8c1bd67790d67c13e2209027837bc17ffc5 to ensure a connection is closed from one test before the next test starts

ricardopereira:
@mattheworiordan merge please.

PR #25, opened 2015-10-09
Converted tests to Swift 2

ricardopereira:
- New Quick version
 - New Nimble version
 - New stable version of SocketRocket
 - No change on Objective-C lib for compatibility with older versions
 - Now it's possilbe to use Asynchronous Expectations

**Important:**
 - Main workspace had a different name: *ably-ios.x* VS *ably.x*.
 - I changed it for consistency and future updates.


ricardopereira:
🚩 Travis error:
`xcodebuild: error: 'ably-ios.xcworkspace' does not exist.`.

Should run `ably.xcworkspace` instead.

**_Why the change?_**
R: The workspace should only be managed by CocoaPods so when we run `pod install` it will, by default, use the `ably.xcodeproj` to generate a new workspace file. The generated file corresponds to `ably.xcworkspace`. It is possible to force CocoaPods using the current workspace but from my experience it is better to build a new and clean one.

ricardopereira:
This is an important PR for future work because the tests depends on it but it brings a lot of advantages.
@mattheworiordan merge it when it is possible. Thank you.

mattheworiordan:
Looks good, should I merge this first?

ricardopereira:
Yes, please.

On 12 October 2015 at 22:14:07, mattheworiordan (notifications@github.com) wrote:
Looks good, should I merge this first?

PR #26, opened 2015-10-12
Auth Token specs

ricardopereira:
- `RSA3` Token Auth:
	- [x] `RSA3a` Can be used over HTTP or HTTPs
	- [x] `RSA3b` For REST requests, the token string is Base64 encoded and used in the Authorization: Bearer header
	- [ ] `RSA3c` For Realtime connections, the querystring param access_token is appended to the URL endpoint
- [x] `RSA4` Token Auth is used if useTokenAuth is set to true, or if useTokenAuth is unspecified and any one of the following conditions cause token auth to be selected as the default: a clientId is specified; authUrl or authCallback is provided; an explicit token or TokenDetails is provided
- [x] `RSA14` If Token Auth is selected, an exception will be raised if a token is not provided or there is no means to generate a token. For example, if only the option useTokenAuth is specified, and a key is not provided, then the client library is unable to authenticate or issue a token
- `RSA15` If Token Auth is selected and clientId has been set in the ClientOptions when the library was instanced:
	- [x] `RSA15a` Any clientId provided in ClientOptions must match any non null clientId value in TokenDetails or connectionDetails of the CONNECTED ProtocolMessage, where applicable
	- [x] `RSA15b` If the clientId from TokenDetails or connectionDetails contains only a wildcard string '*', then the client is permitted to be either unauthenticated (effectively anonymous without a clientId) or authenticated providing a clientId when communicating with Ably
	- [x] `RSA15c` Following an auth request which uses a TokenDetails or TokenRequest object that contains an incompatible clientId, the library should change the connection state to FAILED and emit an error
- [x] `RSA5` TTL for new tokens is specified in milliseconds and defaults to the REST API default (1 hour)
- [x] `RSA6` The capability, if not specified, defaults to allow all operations permitted for the key associated with the token
- `RSA7`clientId and authenticated clients:
	- [ ] `RAS7a` If a clientId is provided in the ClientOptions, or is present in the current authentication token, then the client is considered to be authenticated (it has a clientId that is implicit in all operations). Note that an authentication token clientId wildcard value of '*' is the exception where the client is not necessarily authenticated and any clientId is permitted. The following applies to authenticated clients:
		- [x] `RAS7a1` All operations (such as message publishing or presence) will have an implicit clientId. The Ably service automatically updates the clientId attribute (when empty) for all Message and PresenceMessage messages received from that authenticated client, and any messages then published from the Ably service, will have the clientId attribute populated. It is therefore expected that Ably client libraries do not explicitly set the clientId field on messages published when clientId is implicit in the connection or authentication scheme
		- [x] `RSA7a2` If clientId is provided in ClientOptions, and an API key is provided along with no other means to generate a token, the client library will authenticate with Ably and obtain a token using the provided clientId ensuring the token is restricted to operations for that clientId
		- [x] `RSA7a3` Auth#clientId attribute returns a string value for the authenticated client’s clientId
	- [x] `RSA12` Auth#clientId attribute is null, when following authentication, the clientId attribute of the TokenDetails is a wildcard '*', indicating that any clientId can be used by this client
	- `RSA7b` Auth#clientId is not null when:
		- [x] `RSA7b1` A clientId is provided in the ClientOptions. clientId should be a string
		- [x] `RSA7b2` Token authentication is being used, and the TokenRequest or TokenDetails object, used for authentication, has a clientId value that is not null or a wildcard string '*'
		- [ ] `RSA7b3` Following a realtime connection being established, if the CONNECTED ProtocolMessages contains a clientId that is both not null and not a wildcard string '*'. clientId is an attribute of ProtocolMessage#connectionDetails within a CONNECTED ProtocolMessage
	- [x] `RSA7c` A clientId provided in the ClientOptions when instancing the library must be either null or a string, and cannot contain only a wilcard '*' value which is reserved


ably-ios/ARTAuth.m
```diff
@@ -38,12 +40,17 @@ - (instancetype)init:(ARTRest *)rest withOptions:(ARTClientOptions *)options {
         } else if (options.tokenDetails) {
             [self.logger debug:@"ARTAuth: setting up auth method Token with supplied token only"];
             _method = ARTAuthMethodToken;
+        } else if (options.authUrl && options.authCallback) {
+            [NSException raise:@"ARTAuthException" format:@"Could not setup authentication method with given options."];
```
mattheworiordan: Can this rather say "Incompatible authentication configuration: please specify either authCallback and authUrl"

ably-ios/ARTAuth.m
```diff
@@ -29,6 +29,8 @@ - (instancetype)init:(ARTRest *)rest withOptions:(ARTClientOptions *)options {
         _options = options;
         _logger = rest.logger;
         
+        //options.clientId
```
mattheworiordan: Why is this commented out?

mattheworiordan:
Looking good. Can you confirm in what order I should do the merges please?

ably-ios/ARTAuth.m
ricardopereira: Sure, I will change it. The first PR to merge is the conversion into Swift 2.

About the `clientId`... the attribute is missing on Auth class. I don't know why and I'm trying to figure it out.

On 12 October 2015 at 22:04:20, mattheworiordan (notifications@github.com) wrote:
In ably-ios/ARTAuth.m:

> @@ -38,12 +40,17 @@ - (instancetype)init:(ARTRest *)rest withOptions:(ARTClientOptions *)options {
>          } else if (options.tokenDetails) {
>              [self.logger debug:@"ARTAuth: setting up auth method Token with supplied token only"];
>              _method = ARTAuthMethodToken;
> +        } else if (options.authUrl && options.authCallback) {
> +            [NSException raise:@"ARTAuthException" format:@"Could not setup authentication method with given options."];
Can this rather say "Incompatible authentication configuration: please specify either authCallback and authUrl"

ably-ios/ARTAuth.m
```diff
@@ -206,6 +225,7 @@ - (void)createTokenRequest:(ARTAuthTokenParams *)tokenParams options:(ARTAuthOpt
 }
 
 - (BOOL)canRequestToken {
+    // FIXME: not used?!
```
paddybyers: When will this get resolved Is it not possible to do it with this PR?

ably-ios/ARTAuthTokenParams.m
```diff
@@ -15,8 +15,8 @@
 #import "ARTPayload.h"
 #import "ARTAuthTokenRequest.h"
 
-//X7: NSArray<NSString *>
-static NSArray *decomposeKey(NSString *key) {
+// FIXME: reuse
```
paddybyers: When will this get addressed?

ablySpec/TestUtilities.swift
```diff
@@ -167,5 +222,5 @@ class MockHTTPExecutor: ARTHTTPExecutor {
     func executeRequest(request: NSMutableURLRequest!, callback: ((NSHTTPURLResponse!, NSData!, NSError!) -> Void)!) {
         self.requests.append(request)
         self.executor.executeRequest(request, callback: callback)
-    }
+    }
```
paddybyers: trailing whitespace?

ricardopereira:
@paddybyers Wow. Some code is duplicated on `Auth.swift`. I think it was the last commit where I tried to amend a commit message. I will fix it and change the code that you specified.

ablySpec/TestUtilities.swift
ricardopereira: Strange. I have that option activated on Xcode: `Automatically trim trailing whitespace`. Maybe a bug 😅

PR #29, opened 2015-10-16
Auth.canRequestToken and Auth.validate revision

ricardopereira:
I removed `canRequestToken`. It isn't necessary because the constructor does the validation.

Fixed `isBasicAuth` with:
> _It defaults to basic auth if the `keyValue` is provided, and there is no `clientId` specified. In this case, however, `useTokenAuth` forces token auth to be used._


ricardopereira:
First merge!

paddybyers:
LGTM

PR #30, opened 2015-10-16
Client manage token requests

ricardopereira:
- Auth.authorise was missing the possibility of receiving just an access token.
 - Auth.options.clientId can be nil.
 - Sign TokenRequest with clientId=* when the user is anonymous.
 - Rest.calculateAuthorization was missing the Auth.authorise for token reusability.

⚠️
`RAS7a1` and `RAS7a2` are failing. Will check that on the next PR.

### DIFF
--- a/ably-ios/ARTAuthTokenParams.m
+++ b/ably-ios/ARTAuthTokenParams.m
@@ -22,7 +22,7 @@
         _ttl = 60 * 60;
         _timestamp = [NSDate date];
         _capability = @"{\"*\":[\"*\"]}"; // allow all
-        _clientId = @"";
+        _clientId = nil;
     }
     return self;
 }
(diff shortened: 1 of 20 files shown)


ricardopereira:
Second merge!

ably-ios/ARTAuthTokenParams.m
```diff
@@ -143,7 +145,7 @@ - (ARTAuthTokenRequest *)sign:(NSString *)key {
     NSString *keyName = keyComponents[0];
     NSString *keySecret = keyComponents[1];
     NSString *nonce = generateNonce();
-    NSString *clientId = self.clientId ? self.clientId : @"";
+    NSString *clientId = self.clientId ? self.clientId : @"*";
```
ricardopereira: @paddybyers Correct?

paddybyers: > @paddybyers Correct?

No, I don't think so. A `TokenRequest` only contains the wildcard `clientId` if explicitly requested by the caller.

If the caller does not specify a `clientId`, and the Ably library was not initialised with a `clientId`, then the token request does not contain any `clientId`. For the purposes of canonicalisation, if there is no `clientId` then an empty string is used in place of the `clientId` to construct the canonical text.

ricardopereira: Done. The authentication is working.

ricardopereira:
@paddybyers Please merge when it's possible.

Sorry for the long change. It was really necessary because the `Auth.authorise` wasn't handling the authentication well.

Both authentication methods are stable now.
I think it's a good opportunity to work on the `RealtimeClient` considering that it's a high priority.

paddybyers:
Thanks

PR #31, opened 2015-10-21
Realtime transport auth fixed (RTC1)

ricardopereira:
- `RTC1` is passing.

Next on new PR:
 - `ARTError` need to be a `NSError`.
 - Channels are not working.

### DIFF
--- a/ably-ios/ARTWebSocketTransport.m
+++ b/ably-ios/ARTWebSocketTransport.m
@@ -100,7 +100,7 @@ enum {
                 return;
             }
 
-            NSURLQueryItem *accessTokenParam = [NSURLQueryItem queryItemWithName:@"access_token" value:decodeBase64(tokenDetails.token)];
+            NSURLQueryItem *accessTokenParam = [NSURLQueryItem queryItemWithName:@"access_token" value:(tokenDetails.token)];
             [selfStrong setupWebSocket:@[accessTokenParam] withOptions:selfStrong.options];
             // Connect
             [selfStrong.websocket open];
--- a/ably-ios/ARTWebSocketTransport.m
+++ b/ably-ios/ARTWebSocketTransport.m
@@ -40,8 +40,12 @@ enum {
 @property (readonly, strong, nonatomic) dispatch_queue_t queue;
 @property (readwrite, strong, nonatomic) SRWebSocket *websocket;
 @property (readwrite, assign, nonatomic) BOOL closing;
+
+// From RestClient
 @property (readwrite, strong, nonatomic) id<ARTEncoder> encoder;
 @property (readonly, strong, nonatomic) ARTLog *logger;
+@property (readonly, strong, nonatomic) ARTAuth *auth;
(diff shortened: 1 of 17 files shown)


ably-ios/ARTWebSocketTransport.m
```diff
@@ -127,7 +80,81 @@ - (void)send:(ARTProtocolMessage *)msg {
...
+            ARTWebSocketTransport *selfStrong = selfWeak;
+            if (!selfStrong) return;
+
+            if (error) {
+                [selfStrong.logger error:@"ARTWebSocketTransport: token auth failed with %@", error.description];
+                [selfStrong.delegate realtimeTransportFailed:selfStrong];
+                return;
+            }
+
+            NSURLQueryItem *accessTokenParam = [NSURLQueryItem queryItemWithName:@"access_token" value:decodeBase64(tokenDetails.token)];
```
ricardopereira: @paddybyers Correct? I'm decoding the access token (`decodeBase64`).

ably-ios/ARTWebSocketTransport.m
```diff
@@ -127,7 +80,81 @@ - (void)send:(ARTProtocolMessage *)msg {
...
+            [self.logger error:@"ARTWebSocketTransport: recovery string is malformed, ignoring: '%@'", options.recover];
+        }
+    }
+    else if (options.resumeKey != nil) {
+        NSURLQueryItem *resumeKeyParam = [NSURLQueryItem queryItemWithName:@"resume" value:options.resumeKey];
+        queryItems = [queryItems arrayByAddingObject:resumeKeyParam];
+
+        NSURLQueryItem *connectionSerialParam = [NSURLQueryItem queryItemWithName:@"connection_serial" value:[NSString stringWithFormat:@"%lld",options.connectionSerial]];
+        queryItems = [queryItems arrayByAddingObject:connectionSerialParam];
+    }
```
ricardopereira: @paddybyers Please review.

ricardopereira:
Merge when it's possible. Thank you,

ably-ios/ARTWebSocketTransport.m
paddybyers: There shouldn't be any need to decode the token

paddybyers:
Can you check that comment above about the token? You should treat the token string as opaque.

ricardopereira:
But the received token is encoded and then the it's not possible to establish connection to the realtime server.

SimonWoolf:
@ricardopereira realtime provides it as a base64-encoded string, and expects back the same base64-encoded string. If it's not working unless you have a decode in there, I'd hazard a guess that there's code elsewhere that's mistakenly encoding it before sending it.

(The client lib can't do anything with it decoded anyway -- it won't be valid ascii when decoded, it's a binary)

ricardopereira:
@SimonWoolf Ok, I will investigate. Thanks.

ricardopereira:
@paddybyers @SimonWoolf Fixed. Encoder was misplaced on ARTAuth.

paddybyers:
Thanks

PR #32, opened 2015-10-22
RTC1a, RTC1b, RTC1c

ricardopereira:
RTC1c: not passing because the channels are not working.


ablySpec/RealtimeClient.swift
```diff
@@ -43,22 +52,60 @@ class RealtimeClient: QuickSpec {
...
+                                break
+                            }
+                        }
+                    }
                 }
                 
                 //RTC1d
                 it("should modify the realtime endpoint host if realtimeHost is assigned") {
-                    let options = ARTClientOptions()
+                    //let options = ARTClientOptions()
```
paddybyers: Are these addressed in later PRs?

PR #35, opened 2015-10-28
RTC1 done

ricardopereira:
All specs from `RTC1`.

### DIFF
--- a/ablySpec/RealtimeClient.swift
+++ b/ablySpec/RealtimeClient.swift
@@ -112,13 +112,7 @@ class RealtimeClient: QuickSpec {
 
                 //RTC1d
                 it("should modify the realtime endpoint host if realtimeHost is assigned") {
-                    //let options = ARTClientOptions()
-                    // realtimeHost string, when set, will modify the realtime endpoint host used by this client library
-
-                    //Default: realtime.ably.io
-                    //let realtimeHost = options.realtimeHost
-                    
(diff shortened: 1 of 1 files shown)


ricardopereira:
@paddybyers Please merge.

ablySpec/RealtimeClient.swift
```diff
@@ -127,12 +121,12 @@ class RealtimeClient: QuickSpec {
                     
                     let oldRestHost = options.restHost
                     let oldRealtimeHost = options.realtimeHost
-                    
+
                     // Change REST and realtime endpoint hosts
-                    options.environment = "test"
+                    options.environment = "eu-central-1-a-sandbox"
```
paddybyers: I don't think we should do this because we don't want a hard-coded dependency on a specific datacenter. Assuming your test does not require the hosts to exist, I think `test` was better.

ricardopereira: @paddybyers Done.

ricardopereira:
@paddybyers Added RTC2 but has no implementation because there is no connection manager (too much code to refactor).

RTC3 is ready.

paddybyers:
Thanks

paddybyers:
> Added RTC2 but has no implementation because there is no connection manager (too much code to refactor).

We should talk about this with @mattheworiordan once he's back.

ricardopereira:
Ok, agree.

Issue #41, opened 2015-10-30
Realtime: timeouts implementations are outdated


ricardopereira:
I'm opening this issue because of:

```
(RTC7) The client library must use the configured timeouts specified in the ClientOptions, 
falling back to the client library defaults and defaults described in ClientOptions below
```

Currently, the `ClientOptions` doesn't have any of the `TO3l` specifications:
- **disconnectedRetryTimeout**: possible to implement with small refactoring
- **suspendedRetryTimeout**: possible to implement with small refactoring
- **httpOpenTimeout**: bigger refactoring
- **httpRequestTimeout**: bigger refactoring
- **httpMaxRetryCount**: bigger refactoring
- **httpMaxRetryDuration**: bigger refactoring

The Realtime client use some defaults (basic timeouts):
- **connectTimeout**: on connecting to the server
- **disconnectTimeout**: retry interval when Disconnected
- **suspendTimeout**: retry interval when Suspended

Different from what's defined on [DF1](http://docs.ably.io/client-lib-development-guide/features/#defaults).

`DF1a` and `DF1b` requires some refactoring.

ricardopereira:
With this I need to now what's important to test right now having in mind the current state of the lib.

@mattheworiordan /cc @paddybyers

ricardopereira:
Check changes:
https://github.com/ably/docs/commit/1f853e96b1a63fadf9c7c9f613d1a1d273a048cd

mattheworiordan:
@ricardopereira implementing configurable timeouts is probably lower priority for now.

I am not sure what you mean by `DF1a` and `DF1b` needing refactoring, to be clear the impact I assume is:

* `connectionStateTtl ` will simply map the existing max time in state for the disconnected state i.e. the iOS Realtime library should have already implemented some mechanism to move from the `disconnected` state to the `suspended` state after X time after being `disconnected`.  `connectionStateTtl` is simply this value.  Either way, this is not a public API attribute anyway
* `realtimeRequestTimeout` is simply a generic timeout value that is applied to any request that should auto-fail if a response does not come back in that time, for example `(RTN12b) If the CLOSED ProtocolMessage is not received within the default realtime request timeout, the transport will be disconnected and the connection will automatically move to the CLOSED state`

tcard:
This is done.

PR #42, opened 2015-10-30
RTC7

ricardopereira:
Test for Suspended timeout.
Waiting some answer on #41.

### DIFF
--- a/ably-ios/ARTDefault.h
+++ b/ably-ios/ARTDefault.h
@@ -14,8 +14,19 @@
 + (NSString*)restHost;
 + (NSString*)realtimeHost;
 + (int)TLSPort;
-+ (NSTimeInterval)connectTimeout;
-+ (NSTimeInterval)disconnectTimeout;
-+ (NSTimeInterval)suspendTimeout;
+
++ (NSTimeInterval)connectTimeout; //Realtime
+
+/**
+ When the client is in the DISCONNECTED state, once this TTL has passed, the client should change the state to the SUSPENDED state signifying that the state is now lost i.e. channels need to be reattached manually.
+ 
+ Note that this default is override by any connectionStateTtl specified in the ConnectionDetails of the CONNECTED ProtocolMessage.
+ */
++ (NSTimeInterval)connectionStateTtl;
+
+/**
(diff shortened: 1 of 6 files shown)


ably-ios/ARTDefault.h
```diff
@@ -14,8 +14,19 @@
...
-+ (NSTimeInterval)connectTimeout;
-+ (NSTimeInterval)disconnectTimeout;
-+ (NSTimeInterval)suspendTimeout;
+
++ (NSTimeInterval)connectTimeout; //Realtime
+
+/**
+ When the client is in the DISCONNECTED state, once this TTL has passed, the client should change the state to the SUSPENDED state signifying that the state is now lost i.e. channels need to be reattached manually.
+ 
+ Note that this default is override by any connectionStateTtl specified in the ConnectionDetails of the CONNECTED ProtocolMessage.
```
mattheworiordan: Yes, this is true, but not yet implemented in any client libraries so does not need to be implemented

ablySpec/RealtimeClient.swift
```diff
@@ -240,6 +240,48 @@ class RealtimeClient: QuickSpec {
...
+                            start = NSDate()
+                        default:
+                            break
+                        }
+                    }
+                }
+                client.close()
+
+                if let secs = endInterval {
+                    expect(secs).to(beLessThanOrEqualTo(UInt(options.suspendedRetryTimeout)))
```
mattheworiordan: I am not sure this test is actually doing what is required?  The `suspendedRetryTimeout` represents the frequency of connection reattempts whilst in the disconnected state.  How exactly is this test checking this?

ricardopereira: @mattheworiordan whilst in the disconnected state?! Did you mean suspended state?

ricardopereira: @mattheworiordan the test is checking if the retry connection is occurring on the right time. The suspended retry timeout has been changed to 6 seconds and the client should respect that.

mattheworiordan: Ah, yes, you are correct, thanks for clarifying

mattheworiordan:
Can this be merged?

ricardopereira:
@mattheworiordan Yes. Right now is just testing the `suspendTimeout`.

PR #43, opened 2015-10-30
Connection proxy

ricardopereira:


mattheworiordan:
Can we add some tests to ensure the properties are actually available in the API?

ricardopereira:
That is what I'm doing now.

mattheworiordan:
Any luck with the tests so that we can merge this PR?

ricardopereira:
@mattheworiordan This PR only has the `ARTConnection` proxy but I started the spec implementation on #44.

PR #47, opened 2015-11-07
ObjC tests update to use the latest API

ricardopereira:
It's compiling.


ricardopereira:
@mattheworiordan The changes about the authentication, channels and stats has complicated the task but it's ready to run.

I did an overview of the ObjC tests and it's not good. I have to analyse each test with a critical eye. It could take some time. Do I need to analyse the ObjC tests with very detail?

mattheworiordan:
Thanks @ricardopereira, looks good, I am surprised how much was needed to get this compiling :open_mouth: 

As discussed on Slack, please can you review all the tests with a critical eye so that we can work out which ones are usable, cover the requirements we have, and which ones need improving or rewriting.

ricardopereira:
@mattheworiordan I was surprised as well, believe me :smile: But it's done, hopefully. The majority of the tests are difficult to relate with the current specs. I will leave some notes on the spreadsheet.

mattheworiordan:
Do you think this is ready to be merged? Ideally we mark any known failing tests as failing (and skip them with a TODO) so that we have a passing suite.  If not, as we add new tests we'll never know if we're introducing new problems or not.

ricardopereira:
@mattheworiordan This one it's not ready. I'm fixing some core bugs because almost every ObjC test is failing.

mattheworiordan:
@ricardopereira closing this PR, can you open a PR in this repo next as you are now a collaborator

PR #51, opened 2015-11-13
Presence#history wasn't working (infinite recursion)

ricardopereira:
- (Conflicts) Separated base ARTPresence: ARTRealtimePresence and ARTRestPresence
 - ARTRestPresence has base class for ARTRealtimePresence

### DIFF
--- a/ably-ios/ARTChannel+Private.h
+++ b/ably-ios/ARTChannel+Private.h
@@ -13,10 +13,7 @@
 
 ART_ASSUME_NONNULL_BEGIN
 
-@interface ARTChannel() {
-@public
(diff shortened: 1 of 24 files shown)


mattheworiordan:
This looks good, however I am a bit confused by the naming now.  Previously it was something like `ARTObject`, so `ARTClient` stood for Ably Real-Time Client, however now it seems we are doubling up on Realtime i.e. `ARTRealtimePresence`.  I realise in Java the realtime class names are prefixed with Realtime, but it does feel a bit odd.  Out of interest, just so that I know, what do you think about the classes simply being `AblyRealtime*` and `AblyRest*` instead?  How difficult would this be to change i.e. are there refactoring tools built into Xcode?

PR #57, opened 2015-11-14
ObjC tests: ARTRealtimeInitTest

ricardopereira:
Tests passing:
 - [x] testInitWithOptions
 - [x] testInitWithHost
 - [x] testInitWithPort
 - [x] testInitWithKey
 - [x] testInitAutoConnectDefault
 - [x] testInitAutoConnectFalse

### DIFF
--- a/ably-iosTests/ARTRealtimeInitTest.m
+++ b/ably-iosTests/ARTRealtimeInitTest.m
@@ -99,9 +99,6 @@
         [_realtime.eventEmitter on:^(ARTRealtimeConnectionState state, ARTErrorInfo *errorInfo) {
             if (state == ARTRealtimeConnecting) {
                 XCTAssertEqual(_realtime.auth.options.key, options.key);
-                // FIXME: Temporary, because the environment is nil instead of `staging` because
-                //instanciating a new RealtimeClient with Key will create a "standard" ClientOptions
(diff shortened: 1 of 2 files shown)


ably-iosTests/ARTRealtimeInitTest.m
```diff
@@ -93,10 +94,14 @@ -(void)testInitWithPort {
...
-    [self getBaseOptions:^(ARTClientOptions * options) {
+    [self getBaseOptions:^(ARTClientOptions *options) {
         _realtime = [[ARTRealtime alloc] initWithKey:options.key];
         [_realtime.eventEmitter on:^(ARTRealtimeConnectionState state, ARTErrorInfo *errorInfo) {
-            if(state == ARTRealtimeConnected) {
+            if (state == ARTRealtimeConnecting) {
+                XCTAssertEqual(_realtime.auth.options.key, options.key);
+                // FIXME: Temporary, because the environment is nil instead of `staging` because
+                //instanciating a new RealtimeClient with Key will create a "standard" ClientOptions
+                XCTAssertEqual(((ARTClientOptions *)_realtime.auth.options).environment, nil);
```
mattheworiordan: I think it's OK to test this like that, we don't necessarily need to test that it connects to `sandbox` by default when we are only testing that a single key string works

ricardopereira: Ok then. I will remove the note.

PR #62, opened 2015-11-15
ObjC tests: ARTRestCapabilityTest

ricardopereira:
Tests passing:
 - [x] testPublishRestricted


ricardopereira:
Added `WARNING` when the response header has a `X-Ably-ErrorMessage`.

PR #65, opened 2015-11-16
ObjC tests: ARTRestTokenTest

ricardopereira:
Tests passing:
 - [x] testTokenSimple
 - [x] testInitWithBadToken
 - [x] testAuthURLForcesToken
 - [x] testTTLDefaultOneHour
 - [x] testInitWithBorrowedAuthCb

### DIFF
--- a/ably-ios/ARTAuthOptions.m
+++ b/ably-ios/ARTAuthOptions.m
@@ -113,7 +113,13 @@ NSString *const ARTAuthOptionsMethodDefault = @"GET";
 }
 
 - (BOOL)isBasicAuth {
-    return self.useTokenAuth == false && self.key != nil && self.clientId == nil;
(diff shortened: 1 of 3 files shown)


ricardopereira:
Reference:
> (RSC14b) Supports token authentication if one or more of token, tokenDetails, authCallback or authUrl are provided. If useTokenAuth is true or a clientId is provided when the library is instanced, then token authentication is always used.

mattheworiordan:
Would be good to have some test coverage for https://github.com/ably/ably-ios/pull/65/files#diff-4271d2d2cc6678caf542ebb41dc8c2f1R116, but that can be added in Swift perhaps. :+1:

ricardopereira:
@mattheworiordan Done: #68

PR #70, opened 2015-11-17
Swift tests: RSC9

ricardopereira:
Reference:
> (RSC9) Uses Auth to establish what authentication scheme to use, how to authenticate, and automatic issuing of tokens when necessary


mattheworiordan:
Looks good :+1:

PR #74, opened 2015-11-19
Swift tests: Review

ricardopereira:


ablySpec/RestClient.swift
```diff
@@ -198,6 +200,193 @@ class RestClient: QuickSpec {
...
+                            options.authUrl = NSURL(string: "http://test-auth.ably.io")
+
+                            let rest = ARTRest(options: options)
+                            rest.httpExecutor = mockExecutor
+
+                            // Delay for token expiration
+                            delay(tokenParams.ttl) {
+                                // 40140 - token expired and will not recover because authUrl is invalid
+                                publishTestMessage(rest) { error in
+                                    expect(mockExecutor.responses.first?.allHeaderFields["X-Ably-ErrorCode"] as? String).to(equal("40140"))
```
mattheworiordan: Ignore previous comment about checking 40140 as I can see this was introduced in a later commit.  However, the error object emitted should contain the error code and message, so I think it would make sense to check that the error object itself contains this information.  This is needed for a developer who potentially catches the error and needs to understand why there was a failure.

mattheworiordan:
@ricardopereira this is almost perfect, I will merge once you address / respond to vivisection studies conducted by the Japanese in wikipedia.orgUnit 731

ablySpec/RestClient.swift
ricardopereira: @mattheworiordan :disappointed: but it will always return the last error and the last one is `-1003: A server with the specified hostname could not be found` because it fails while trying to reach `http://test-auth.ably.io`.

ricardopereira: You want me to change the code now or later?

mattheworiordan: Ah, @ricardopereira I see. I will merge this.  Do we have a test where the token is not renewable where we check that an error with 40140 is emitted?

Issue #79, opened 2015-11-20
Travis: iOS 9 simulator


ricardopereira:
**Temporary solution:**
Try forcing Travis to use the iOS 8 simulator because the new version has ATP.

### App Transport Security from [Apple](https://developer.apple.com/library/prerelease/ios/releasenotes/General/WhatsNewIniOS/Articles/iOS9.html)

> App Transport Security (ATS) enforces best practices in the secure connections between an app and its back end. ATS prevents accidental disclosure, provides secure default behavior, and is easy to adopt; it is also on by default in iOS 9 and OS X v10.11. You should adopt ATS as soon as possible, regardless of whether you’re creating a new app or updating an existing one.

> If you’re developing a new app, you should use HTTPS exclusively. If you have an existing app, you should use HTTPS as much as you can right now, and create a plan for migrating the rest of your app as soon as possible. In addition, your communication through higher-level APIs needs to be encrypted using TLS version 1.2 with forward secrecy. If you try to make a connection that doesn't follow this requirement, an error is thrown. If your app needs to make a request to an insecure domain, you have to specify this domain in your app's Info.plist file.

Error:
```
2015-11-20 08:36:04.263 xctest[34491:1802313] NSURLSession/NSURLConnection HTTP load failed (kCFStreamErrorDomainSSL, -9807)
```

ricardopereira:
For more detail: http://ste.vn/2015/06/10/configuring-app-transport-security-ios-9-osx-10-11/

Issue #78, opened 2015-11-20
Travis: Integrate XCTool or Fastlane


ricardopereira:
The build and test output is very difficult to read.
[XCTool](https://github.com/facebook/xctool) or [Fastlane](https://github.com/fastlane/fastlane) could help.

PR #80, opened 2015-11-20
Travis: XCTool

ricardopereira:
Added XCTool: #78 
Added NSAppTransportSecurity: #79 (not sure if it is a workaround)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,8 @@ language: objective-c
 osx_image: xcode7.1
 
 before_install:
-  - gem install cocoapods -v '0.39.0'
+ - xcodebuild -showsdks
+ - gem install cocoapods -v '0.39.0'
 script:
  # Objectve-C tests
  - xctool test -workspace ably.xcworkspace -scheme ably -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,8 @@ osx_image: xcode7.1
 
 before_install:
   - gem install cocoapods -v '0.39.0'
-script: 
- - xcodebuild test -workspace ably.xcworkspace -scheme ablySpec -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
+script:
+ # Objectve-C tests
+ - xctool test -workspace ably.xcworkspace -scheme ably -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
+ # Swift tests
+ - xctool test -workspace ably.xcworkspace -scheme ablySpec -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
--- a/ably-ios/ARTAuth.h
+++ b/ably-ios/ARTAuth.h
@@ -24,7 +24,7 @@ ART_ASSUME_NONNULL_BEGIN
 
 @interface ARTAuth : NSObject
 
-@property (nonatomic, readonly, weak) ARTAuthOptions *options;
+@property (nonatomic, readonly, strong) ARTAuthOptions *options;
 @property (nonatomic, readonly, assign) ARTAuthMethod method;
 
 @property (nonatomic, weak) ARTLog *logger;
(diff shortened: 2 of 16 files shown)


ricardopereira:
Still occurring on Travis:
```
NSURLSession/NSURLConnection HTTP load failed (kCFStreamErrorDomainSSL, -9807)
```

ricardopereira:
New error:

```
2015-11-20 18:00:17.006 xctest[1275:4407] ERROR: POST https://sandbox-rest.ably.io:443/channels/channel/messages: error Error Domain=NSURLErrorDomain Code=-1202 "The certificate for this server is invalid. You might be connecting to a server that is pretending to be “sandbox-rest.ably.io” which could put your confidential information at risk." UserInfo={NSLocalizedDescription=The certificate for this server is invalid. You might be connecting to a server that is pretending to be “sandbox-rest.ably.io” which could put your confidential information at risk., NSLocalizedRecoverySuggestion=Would you like to connect to the server anyway?, _kCFStreamErrorDomainKey=3, NSUnderlyingError=0x7be3c2f0 {Error Domain=kCFErrorDomainCFNetwork Code=-1202 "(null)" UserInfo={_kCFStreamPropertySSLClientCertificateState=0, _kCFNetworkCFStreamSSLErrorOriginalValue=-9807, _kCFStreamErrorCodeKey=-9807, _kCFStreamErrorDomainKey=3, kCFStreamPropertySSLPeerTrust=<SecTrustRef: 0x7be422b0>, kCFStreamPropertySSLPeerCertificates=<CFArray 0x7be9b7d0 [0xa27098]>{type = immutable, count = 2, values = (
```

ricardopereira:
Fixed error `-9807`.

mattheworiordan:
@ricardopereira not sure what is going on with the SSL errors, go to https://cryptoreport.websecurity.symantec.com/checker/views/certCheck.jsp and enter the sandbox-rest.ably.io domain.  You will see it definitely has a valid cert

ricardopereira:
@mattheworiordan The `sandbox-rest.ably.io` has a valid cert but the client must have a cert as well. On iOS9 client certificates are needed to make SSL work properly and I updated the implementation to use a self-signed cert.

### App Transport Security from [Apple](https://developer.apple.com/library/prerelease/ios/releasenotes/General/WhatsNewIniOS/Articles/iOS9.html)

> App Transport Security (ATS) enforces best practices in the secure connections between an app and its back end. ATS prevents accidental disclosure, provides secure default behavior, and is easy to adopt; it is also on by default in iOS 9 and OS X v10.11. You should adopt ATS as soon as possible, regardless of whether you’re creating a new app or updating an existing one.

> If you’re developing a new app, you should use HTTPS exclusively. If you have an existing app, you should use HTTPS as much as you can right now, and create a plan for migrating the rest of your app as soon as possible. In addition, your communication through higher-level APIs needs to be encrypted using TLS version 1.2 with forward secrecy. **If you try to make a connection that doesn't follow this requirement, an error is thrown.** If your app needs to make a request to an insecure domain, you have to specify this domain in your app's Info.plist file.

mattheworiordan:
>  I updated the implementation to use a self-signed cert.

Why do you need a self-signed cert?  `sandbox-rest.ably.io` is signed by a root certificate that iOS should have the public keys for

ricardopereira:
@mattheworiordan Yes, you right. It doesn't use the public keys. I will debug the problem.

ricardopereira:
@mattheworiordan The iOS 9 has "App Transport Security" which encourages developers to use https instead of http. That is why the initial error `NSURLSession/NSURLConnection HTTP load failed (kCFStreamErrorDomainSSL, -9807)`.

So, I added an exception for `*.ably.io` host and I changed the code to always use a `NSURLSessionDelegate` object where it forces to get the server cert. I made the mistake to name it `NSURLSessionSelfSignedCertificateSync`. I will change it.

Some references:
[Authentication Challenges and TLS Chain Validation](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/URLLoadingSystem/Articles/AuthenticationChallenges.html)
[Overriding TLS Chain Validation Correctly](https://developer.apple.com/library/mac/documentation/NetworkingInternet/Conceptual/NetworkingTopics/Articles/OverridingSSLChainValidationCorrectly.html)
[NSURLProtectionSpace](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSURLProtectionSpace_Class/#//apple_ref/doc/constant_group/NSURLProtectionSpace_Authentication_Method)

mattheworiordan:
Ok, thanks for looking into this.

In your last commit f2e674c it's not immediately obvious how `*.ably.io` is trusted, can you explain?

mattheworiordan:
Mm, still not sure I understand, but I will trust your judgement and merge!

PR #85, opened 2015-11-25
Get Travis CI passing consistently 

ricardopereira:
Fixed:
 - Domain based on the testing environment.
 - if algorithm is nil then cipherEncoder should be ignored.
 - RTC5b.
 - RSN4: channel as a weak reference but the pointer does not mutate. The object is deallocated but it does not zero out the pointer.


mattheworiordan:
Looks good, would be nice (if available) to have tests marked as pending so that we see this in the test output.

ricardopereira:
Rebase from master.

ricardopereira:
It seems like, sometimes in a random test, it stops with an error: `Invalid connection: com.apple.coresymbolicationd`. I have a folder with that name on my HD: `/System/Library/Caches/com.apple.coresymbolicationd` Something related with cache.

ricardopereira:
@mattheworiordan Passing.

mattheworiordan:
> I change it to pending because it does not work on the test server. I need to investigate more.

Fine, however I would like to know why it isn't passing as well. I think it would make sense, where possible, to fix issues as and when we find them, as opposed to defer all the failing tests till the end of the process.  That way the pain is spread out, what do you think?

PS. If you enable a very verbose log level and send us the raw protocol and REST communication, we can confirm if the issue is on our side, and if not, help you to determine what may be wrong on your side

ricardopereira:
- Created a script `run-tests` to be sure we run the same commands on ci server and local machine
 - It is possible to choose which build tool to perform the tests: env BUILDTOOL
     Options: xctool (default), xcodebuild-pretty, xcodebuild-travis, xcodebuild
     Example: BUILDTOOL=xcodebuild ./Scripts/run-tests.sh
 - It is possible to run only one test class like: `CLASS=ablyTests:ARTRealtimePresenceHistoryTest ./Scripts/run-tests.sh`
 - Changed main scheme to run tests on the same build (the old way was runnning Swift tests and then ObjC tests. Now it runs both on the same execution)
 - Before testing, does a `clean` command

mattheworiordan:
@ricardopereira not quite sure what this PR is about anymore, it's titled stats.
Also, the tests are not passing on CI either.

ricardopereira:
Just for the record - some connections remain active after the test:

```
RealtimeClient__options__should_support_the_same_options_as_the_Rest_client_UsersricardopereiraRepositoriesWhitesmithablyiosablySpecRealtimeClientswift_34]' started.
2015-12-21 07:11:13.533 xctest[14418:4191633] DEBUG: (ARTWebSocketTransport.m:282) 0x7cfc9300 websocket did receive message {"action":8} <---- From another test
2015-12-21 07:11:13.533 xctest[14418:4191633] DEBUG: (ARTWebSocketTransport.m:220) 0x7cfc9300 websocket did disconnect (code 1000) 
2015-12-21 07:11:15.720 xctest[14418:4191584] DEBUG: (ARTRest.m:72) 0x7cef8d40 alloc HTTP
2015-12-21 07:11:15.720 xctest[14418:4191584] DEBUG: (ARTAuth.m:71) setting up auth method Token with key
2015-12-21 07:11:15.721 xctest[14418:4191584] DEBUG: (ARTRest.m:86) initialised 0x7c86a2c0
```

README.md
```diff
@@ -113,19 +117,23 @@ The following features are not implemented yet:
...
 
-## Support and feedback
+## Support, feedback and troubleshooting
+
+Please visit http://support.ably.io/ for access to our knowledgebase and to ask for any assistance.
+
+You can also view the [community reported Github issues](https://github.com/ably/ably-ios/issues).
 
-Please visit https://support.ably.io/ for access to our knowledgebase and to ask for any assistance.
+To see what has changed in recent versions of Bundler, see the [CHANGELOG](CHANGELOG.md).
```
tcard: Bundler?

ably-ios/ARTRest.m
```diff
@@ -92,6 +95,10 @@ - (instancetype)initWithKey:(NSString *) key {
     return [self initWithOptions:[[ARTClientOptions alloc] initWithKey:key]];
 }
 
+- (void)dealloc {
+    [self.logger debug:__FILE__ line:__LINE__ message:@"%p dealloc", self];
```
tcard: Missing `[super dealloc]`?

ably-ios/ARTRest.m
```diff
@@ -200,10 +207,20 @@ - (void)time:(void(^)(NSDate *time, NSError *error))callback {
     return nil;
 }
 
-- (void)stats:(ARTStatsQuery *)query callback:(void (^)(ARTPaginatedResult *, NSError *))callback {
-    NSParameterAssert(query.limit < 1000);
-    NSParameterAssert([query.start compare:query.end] != NSOrderedDescending);
-    
+- (BOOL)stats:(ARTStatsQuery *)query callback:(void (^)(ARTPaginatedResult *, NSError *))callback error:(NSError **)errorPtr {
```
tcard: Is this seriously how people usually do errors in Obj-C?

ably-ios/ARTRest.m
```diff
@@ -200,10 +207,20 @@ - (void)time:(void(^)(NSDate *time, NSError *error))callback {
...
+    if (query.limit > 1000 && errorPtr) {
```
tcard: We need to return `NO` even when errorPtr is nil, otherwise we will make an invalid request.

ably-ios/ARTRestChannel.m
```diff
@@ -51,9 +51,19 @@ - (ARTRestPresence *)presence {
     return _restPresence;
 }
 
-- (void)history:(ARTDataQuery *)query callback:(void(^)(ARTPaginatedResult *result, NSError *error))callback {
-    NSParameterAssert(query.limit < 1000);
-    NSParameterAssert([query.start compare:query.end] != NSOrderedDescending);
+- (BOOL)history:(ARTDataQuery *)query callback:(void(^)(ARTPaginatedResult *result, NSError *error))callback error:(NSError **)errorPtr {
+    if (query.limit > 1000 && errorPtr) {
```
tcard: Same as https://github.com/ably/ably-ios/pull/85/files#r48423985

ably-ios/ARTRestPresence.m
```diff
@@ -41,9 +41,19 @@ - (void)get:(void (^)(ARTPaginatedResult /* <ARTPresenceMessage *> */ *result, N
     [ARTPaginatedResult executePaginated:[self channel].rest withRequest:request andResponseProcessor:responseProcessor callback:callback];
 }
 
-- (void)history:(ARTDataQuery *)query callback:(void (^)(ARTPaginatedResult /* <ARTPresenceMessage *> */ *result, NSError *error))callback {
-    NSParameterAssert(query.limit < 1000);
-    NSParameterAssert([query.start compare:query.end] != NSOrderedDescending);
+- (BOOL)history:(ARTDataQuery *)query callback:(void (^)(ARTPaginatedResult /* <ARTPresenceMessage *> */ *result, NSError *error))callback error:(NSError **)errorPtr {
+    if (query.limit > 1000 && errorPtr) {
```
tcard: Same as https://github.com/ably/ably-ios/pull/85/files#r48423985

ably-ios/ARTRest.m
ricardopereira: dealloc is automatically called. The compiler do that work for you.

On 24 December 2015 at 18:53:51, Toni Cárdenas (notifications@github.com) wrote:

In ably-ios/ARTRest.m:

> @@ -92,6 +95,10 @@ - (instancetype)initWithKey:(NSString *) key {
>      return [self initWithOptions:[[ARTClientOptions alloc] initWithKey:key]];
>  }
>   
> +- (void)dealloc {
> +    [self.logger debug:__FILE__ line:__LINE__ message:@"%p dealloc", self];
Missing [super dealloc]?

ably-ios/ARTRest.m
ricardopereira: https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html#//apple_ref/doc/uid/TP40001806-CH204-BAJIIGCC

On 24 December 2015 at 18:59:41, Toni Cárdenas (notifications@github.com) wrote:

In ably-ios/ARTRest.m:

> @@ -200,10 +207,20 @@ - (void)time:(void(^)(NSDate *time, NSError *error))callback {
>      return nil;
>  }
>   
> -- (void)stats:(ARTStatsQuery *)query callback:(void (^)(ARTPaginatedResult *, NSError *))callback {
> -    NSParameterAssert(query.limit < 1000);
> -    NSParameterAssert([query.start compare:query.end] != NSOrderedDescending);
> -     
> +- (BOOL)stats:(ARTStatsQuery *)query callback:(void (^)(ARTPaginatedResult *, NSError *))callback error:(NSError **)errorPtr {
> +    if (query.limit > 1000 && errorPtr) {
We need to return NO even when errorPtr is nil, otherwise we will make an invalid request.

ricardopereira: Can you clarify?
If errorPtr is nil then the error is ignored by the caller.
On 24 December 2015 at 18:59:41, Toni Cárdenas (notifications@github.com) wrote:

In ably-ios/ARTRest.m:

> @@ -200,10 +207,20 @@ - (void)time:(void(^)(NSDate *time, NSError *error))callback {
>      return nil;
>  }
>   
> -- (void)stats:(ARTStatsQuery *)query callback:(void (^)(ARTPaginatedResult *, NSError *))callback {
> -    NSParameterAssert(query.limit < 1000);
> -    NSParameterAssert([query.start compare:query.end] != NSOrderedDescending);
> -     
> +- (BOOL)stats:(ARTStatsQuery *)query callback:(void (^)(ARTPaginatedResult *, NSError *))callback error:(NSError **)errorPtr {
> +    if (query.limit > 1000 && errorPtr) {
We need to return NO even when errorPtr is nil, otherwise we will make an invalid request.

ably-ios/ARTRest.m
ricardopereira: https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html#//apple_ref/doc/uid/TP40001806-CH204-BAJIIGCC

On 24 December 2015 at 18:55:49, Toni Cárdenas (notifications@github.com) wrote:

In ably-ios/ARTRest.m:

> @@ -200,10 +207,20 @@ - (void)time:(void(^)(NSDate *time, NSError *error))callback {
>      return nil;
>  }
>   
> -- (void)stats:(ARTStatsQuery *)query callback:(void (^)(ARTPaginatedResult *, NSError *))callback {
> -    NSParameterAssert(query.limit < 1000);
> -    NSParameterAssert([query.start compare:query.end] != NSOrderedDescending);
> -     
> +- (BOOL)stats:(ARTStatsQuery *)query callback:(void (^)(ARTPaginatedResult *, NSError *))callback error:(NSError **)errorPtr {
Is this seriously how people usually do errors in Obj-C?

ably-ios/ARTRest.m
tcard: Yes, but when called, because you're overriding the parent's implementation, which actually deallocates the object, if you don't pass the method up the inheritance chain, the deallocation won't happen and there'll be a memory leak.

See https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmPractical.html, the third example under `Don’t Use Accessor Methods in Initializer Methods and dealloc`.

At least this is what I understand from that, I've never worked with dealloc.

ably-ios/ARTRest.m
tcard: I get that if errorPtr is nil then the caller ignores the error, but I don't think that we should just continue when the preconditions are wrong. We should return early to avoid an inconsistent state that could even crash the program.

ricardopereira: Oh yes. You right. The condition should not check the errorPtr, only when
it is needed.

ricardopereira: I think that's pre-ARC docs. I am not on my laptop to check that with the
compiler but sure I will do it later.

tcard: I thought ARC just inserted calls to `release` when necessary. I don't see how that affects this. `release` will call `dealloc` when the reference count for the object reaches zero, and `NSObject` implements `dealloc` as calling `free` or whatever is needed to deallocate the object.

ricardopereira: So, it is not possible to call the super.dealloc:
screen shot 2015-12-28 at 08 18 35

I checked with the compiler and it call first the child's dealloc and then the parent dealloc, so it is fine.

tcard: Oh, OK then. Sorry for the noise. It's really confusing that the compiler makes a special case of this method, such that defining it for a class doesn't override the parent's version as would any other method.

tcard:
LGTM, FWIW :)

ably-ios/ARTRest.m
mattheworiordan: I can't comment on whether this is a suitable style for Obj-C, however typically in our APIs we raise exceptions directly when calling the Rest library as it is mostly a synchronous styled API, whereas with Realtime everything is more `async` and thus we pass an error to the callback unless the method being called is deterministically invalid i.e. you called the method with invalid arguments.

ricardopereira:
@mattheworiordan I forgot to mention but I finished the remaining work.

tcard:
Sorry to interrupt with this meta-issue, but in every team I've worked with before we tried to keep pull requests focused, atomic and as minimal as possible, to facilitate review, deployment, possible reverts, etc. But I keep seeing new commits in this one, e. g. a1272d9 changing the license, that really depart from this view. In those other teams I would've been asked to close this pull request and make smaller ones, with the commits adequately arranged and squashed for each one. I guess we can spare that to avoid even more distraction, but at least I think we should stop doing it from now on, unless in this team we want to work different to this for some reason.

It is especially hard for me to review this this way, because I'm new and I'm getting used to the codebase and the specs.

ricardopereira:
@tcard This PR is really a mess and it is my fault. It should not be that way. The license has changed on the master but then I did a merge where it was a mistake and now I revert it again. I believe you are struggling with this PR but it will never happen again. I can clarify you anything. I am happy to help.

@mattheworiordan I already reviewed the missing comments.

tcard:
OK, I've seen now the comments about merging master. That was really messy. E. g. we have b5695fc9b0aef316907cf35eddd0ddd635384a02 in this PR, then reverted by the merge master, then a1272d9f28a698c6cda06f0092516be12d035a27 doing exactly the same thing.

Can we rebase -i this at least to remove the master merge?

ricardopereira:
@tcard Yes, I tried to revert b78ac4f but there were necessary changes so I chose to recommit the missing parts.

tcard:
I've done the rebase -i together with @ricardopereira and now it's a bit less messy. However, when rebasing on top of master I had to handle conflicts again and I probably messed up something, so then I had to add 1df3e28. Sorry, it's not ideal, but at this point I think it isn't worth it to clean it up further.

tcard:
(By the way, the changes in the license and README got lost because they probably shouldn't have been there to begin with. They came from commits outside this pull request, like b5695fc9, but that also are not on master. They really should be merged to master separately.)

ricardopereira:
@tcard Seams everything fine. I messed up the _license_ & _readme_ with this 9731ecc 😓 I will create a PR with those changes. Thanks again. I learned one more thing about git.

mattheworiordan:
I think we should merge this so that this crazy PR can land and we can return to normality :)

:+1:

PR #89, opened 2015-11-26
RSA15a

ricardopereira:


ablySpec/Auth.swift
```diff
@@ -177,34 +177,66 @@ class Auth : QuickSpec {
...
+                        let expectedClientId = "client_string"
+                        let options = AblyTests.setupOptions(AblyTests.jsonRestOptions)
+                        options.clientId = expectedClientId
+
+                        let client = ARTRest(options: options)
+                        client.httpExecutor = mockExecutor
+
+                        waitUntil(timeout: 10) { done in
+                            // Token
+                            client.calculateAuthorization(ARTAuthMethod.Token) { token, error in
```
mattheworiordan: what is `calculateAuthorization`?

mattheworiordan:
Looks good, however I don't think it's testing the error condition when the provided `ClientOptions` `clientId` does not match, which is really what the spec item is about.

ablySpec/Auth.swift
ricardopereira: @mattheworiordan It is the REST method that checks the Auth method and completes the request with the Authorisation header: `Basic key` or `Bearer token`.

ricardopereira: The name could be better. Maybe: `calculateAuthorization` -> `buildAuthorisationHeader`

ricardopereira:
@mattheworiordan Could you clarify, please. Which error condition?

> (RSA15a) Any clientId provided in ClientOptions must match any non null clientId value in TokenDetails or connectionDetails of the CONNECTED ProtocolMessage, where applicable

ricardopereira:
@mattheworiordan Can you please merge this or review what is missing? I need the property that I added on the MockTransport.

ablySpec/Auth.swift
mattheworiordan: `prepareAuthorisationHeader`?

mattheworiordan:
> Could you clarify, please. Which error condition?

As a policy, we should always be testing the success and failure conditions.  In this example, you are only testing the success condition which passes i.e. `clientId` does match the `CONNECTED` or `TokenDetails` `clientId`.  Whilst this is needed, you have not covered off the other half of what this spec requires i.e. what happens when they are not compatible? When the `clientId` is not compatible, it should result in an error.  We need to have a test for this.

Please bear this in mind in all future tests, we are always aiming to cover both success and failures for any requirement to prove it's implemented.

ricardopereira:
Done

PR #90, opened 2015-11-26
RSA8a

ricardopereira:


mattheworiordan:
It is good, but I am not sure it really shows that a token request has been created, other than the fact a TokenDetails object is returned and assumed comes from a TokenRequest.  It would be nice to have some deeper test, but it's not that important either.

ricardopereira:
@mattheworiordan Yes I know, it is superficial but basically it calls the `Auth#createTokenRequest`. I could create some mock object for Auth to record all the steps of the authorisation process.

mattheworiordan:
No, I think lets proceed, it's not that important

PR #91, opened 2015-11-26
RSA8b

ricardopereira:


mattheworiordan:
Partial support, spec says `... which override defaults for Client Auth`.  There is no test that it overrides any defaults, so you should set those defaults such as clientId in the `ClientOptions`, and ensure that the options are overriden here.

ricardopereira:
@mattheworiordan Yes you right. But I do not see anything more than the `clientId` to override.

mattheworiordan:
> Yes you right. But I do not see anything more than the clientId to override

You can override any defaults, so if after calling `authorise` with a set of TokenParams (which it supports and uses for defaults for subsequent requests), you can test that not overriding them uses the defaults, and overriding them uses the new TokenParams

ricardopereira:
@mattheworiordan Done

mattheworiordan:
LGTM

PR #92, opened 2015-11-26
RSA9

ricardopereira:


ablySpec/Auth.swift
```diff
@@ -529,5 +529,148 @@ class Auth : QuickSpec {
...
+                }
+            }
+
+            // RSA9e
+            it("should TTL be optional and specified in milliseconds") {
+                let rest = ARTRest(options: AblyTests.commonAppSetup())
+
+                let tokenParams = ARTAuthTokenParams()
+                let expectedTtl = 6.0
+                tokenParams.ttl = expectedTtl
```
ricardopereira: @mattheworiordan Currently, `tokenParams.ttl` can not be nil. It is always initialised with the default value: 1 minute. Critical?

ablySpec/Auth.swift
```diff
@@ -529,5 +529,148 @@ class Auth : QuickSpec {
...
+                    expect(tokenRequest?.ttl).to(equal(expectedTtl))
+                })
+            }
+
+            // RSA9f
+            it("should provide capability has json text") {
+                let rest = ARTRest(options: AblyTests.commonAppSetup())
+
+                let tokenParams = ARTAuthTokenParams()
+                let expectedCapability = "{ - }" //Invalid
```
ricardopereira: @mattheworiordan It passes with invalid JSON. Should I fix this?

ablySpec/Auth.swift
mattheworiordan: 60 minutes is the default, not 1 minute please.  I understand perhaps it can't be omitted, but we need a sensible default if that is the case.

ablySpec/Auth.swift
mattheworiordan: I am not sure how it could, a token request with invalid JSON should fail server-side?

ablySpec/Auth.swift
ricardopereira: @mattheworiordan Sorry, it is 60 minutes (3600 seconds).

ablySpec/Auth.swift
ricardopereira: @mattheworiordan I can dig into it later or do you prefer that I check it now? Capability will have an object to manage it right?

ablySpec/Auth.swift
mattheworiordan: > Sorry, it is 60 minutes (3600 seconds).

And where is that?  It says `expectedTtl = 6.0`?

ablySpec/Auth.swift
mattheworiordan: If this is passing, then it indicates something is wrong with your implementation because that would fail.

```
$ curl -X POST "https://rest.ably.io/keys/xVLyHw.OpNddg/requestToken" \
>  -u "xVLyHw.OpNddg:tFG_5gBPYR9e_LtA" \
>  -H "Content-Type: application/json" \
>  --data '{ "keyName": "xVLyHw.OpNddg", "timestamp": 1449091700854, "capability": "{ - }" }'
{
	"error": {
		"statusCode": 400,
		"code": 40000,
		"message": "Malformed capability",
		"serverId": "frontend.2c91.production-eu-west-1-A.i-b1cf411b"
	}
}
```

So please look into why it is passing with an invalid capability, and make sure you add a test to ensure we don't get a regression please.  Once fixed, then we can look at merging this in

ably-ios/ARTAuthTokenParams.m
```diff
@@ -19,7 +20,7 @@ @implementation ARTAuthTokenParams
 
 - (instancetype)init {
     if (self = [super init]) {
-        _ttl = 60 * 60;
+        _ttl = [ARTDefault ttl] * 60; //seconds
```
mattheworiordan: The TTL is in miliseconds in the API, yet it appears the value here is in seconds meaning the intended `TTL` will be incorrect.  I would like to see a test that requests a token valid for 24 hours, then checks that the token does indeed expire the following day 24 hours later (or within a minute of that).

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+
+            // RSA9a
+            it("should create and sign a TokenRequest") {
+                let options = AblyTests.commonAppSetup()
+                let rest = ARTRest(options: options)
+                let expectedClientId = "client_string"
+                let tokenParams = ARTAuthTokenParams(clientId: expectedClientId)
+
+                rest.auth.createTokenRequest(tokenParams, options: options, callback: { tokenRequest, error in
+                    expect(tokenRequest?.clientId).to(equal(expectedClientId))
```
mattheworiordan: No check that it's actually a TokenRequest type, we should add that.  I appreciate it's a typed language, but then this code does not demonstrate that as the type is not specified.

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+                auth.createTokenRequest(nil, options: authOptions, callback: { tokenRequest, error in
+                    expect(tokenRequest?.keyName).to(equal("key"))
+                })
+            }
+
+            // RSA9c
+            it("should generate a unique 16+ character nonce if none is provided") {
+                let rest = ARTRest(options: AblyTests.commonAppSetup())
+
+                rest.auth.createTokenRequest(nil, options: nil, callback: { tokenRequest, error in
```
mattheworiordan: To test uniqueness please create at least least two token request objects and ensure the nonce is not the same

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+                        })
+                    }
+
+                    expect(serverTime).toNot(beNil(), description: "Server time is nil")
+
+                    let authOptions = ARTAuthOptions()
+                    authOptions.queryTime = true
+
+                    rest.auth.createTokenRequest(nil, options: authOptions, callback: { tokenRequest, error in
+                        expect(tokenRequest?.timestamp).to(beCloseTo(serverTime!, within: 6.0))
```
mattheworiordan: This is not actually testing that the server time has been used in any way as it is VERY likely the local time is within 6 seconds of the server time anyway.  You will need to think of a better way of checking that it's getting and using the server's time.

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+            it("should TTL be optional and specified in milliseconds") {
+                let rest = ARTRest(options: AblyTests.commonAppSetup())
+
+                rest.auth.createTokenRequest(nil, options: nil, callback: { tokenRequest, error in
+                    //In Seconds because TTL property is a NSTimeInterval but further it does the conversion to milliseconds
+                    expect(tokenRequest?.ttl).to(equal(ARTDefault.ttl() * 60.0))
+                })
+
+                let tokenParams = ARTAuthTokenParams()
+                expect(tokenParams.ttl).to(equal(ARTDefault.ttl() * 60.0))
```
mattheworiordan: I think you should change this so that you explicitly test that it's in milliseconds i.e. `60 * 60 * 1000`

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+                    expect(tokenRequest?.ttl).to(equal(expectedTtl))
+                })
+            }
+
+            // RSA9f
+            it("should provide capability has json text") {
+                let rest = ARTRest(options: AblyTests.commonAppSetup())
+
+                let tokenParams = ARTAuthTokenParams()
+                tokenParams.capability = "{ - }"
```
mattheworiordan: What is this? Why the dash?

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+            }
+
+            // RSA9g
+            it("should generate a valid HMAC") {
+                let rest = ARTRest(options: AblyTests.commonAppSetup())
+
+                let tokenParams = ARTAuthTokenParams(clientId: "client_string")
+
+                rest.auth.createTokenRequest(tokenParams, options: nil, callback: { tokenRequest, error in
+                    let signed = tokenParams.sign(rest.options.key!, withNonce: tokenRequest?.nonce ?? "")
```
mattheworiordan: Why the `?` in `tokenRequest?.nonce`?  We know it has a value.  Also, perhaps we should explicitly set the nonce for this test to ensure we are getting the intended result

mattheworiordan:
Lots of comments on this one, I don't think unfortunately that these tests do a good job at testing the spec functionality unfortunatley

ably-ios/ARTDefault.h
```diff
@@ -15,6 +15,11 @@
 + (NSString*)realtimeHost;
 + (int)TLSPort;
 
+/**
+ Default in minutes of requested time to live for the token.
+ */
++ (double)ttl;
```
tcard: Why is this not a NSTimeInterval?

ablySpec/Auth.swift
tcard: Elsewhere there are lots of checks like this:

```swift
if let e = error {
    XCTFail(e.description)
    done()
    return
}

guard let currentTokenDetails = tokenDetails else {
    XCTFail("expected tokenDetails not to be nil when error is nil")
    done()
    return
}
```

That applies here as well.

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+
+            // RSA9b
+            it("should support AuthOptions") {
+                let rest = ARTRest(options: AblyTests.commonAppSetup())
+                let auth: ARTAuth = rest.auth
+
+                let authOptions = ARTAuthOptions(key: "key:secret")
+
+                auth.createTokenRequest(nil, options: authOptions, callback: { tokenRequest, error in
+                    expect(tokenRequest?.keyName).to(equal("key"))
```
tcard: Same as https://github.com/ably/ably-ios/pull/92/files#r49773338

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+                })
```
tcard: What about the rest of the options? E. g. `force`?

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+            }
+
+            // RSA9c
+            it("should generate a unique 16+ character nonce if none is provided") {
+                let rest = ARTRest(options: AblyTests.commonAppSetup())
+
+                rest.auth.createTokenRequest(nil, options: nil, callback: { tokenRequest, error in
+                    expect(tokenRequest?.nonce.characters).to(haveCount(16))
```
tcard: Same as https://github.com/ably/ably-ios/pull/92/files#r49863069

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+            }
+
+            // RSA9d
+            context("should generate a timestamp") {
+
+                it("from current time if not provided") {
+                    let rest = ARTRest(options: AblyTests.commonAppSetup())
+
+                    rest.auth.createTokenRequest(nil, options: nil, callback: { tokenRequest, error in
+                        expect(tokenRequest?.timestamp).to(beCloseTo(NSDate(), within: 1.0))
```
tcard: Same as https://github.com/ably/ably-ios/pull/92/files#r49863069

ably-ios/ARTDefault.h
ricardopereira: I decided to express it in minutes. I could use `NSTimeInterval`.
This is maybe confusion https://github.com/ably/ably-ios/blob/b4bd0617bb96265e4e9f847bc397e383e33bbba8/ably-ios/ARTAuthTokenParams.m#L23.
What do you suggest?

tcard: NSTimeInterval should alleviate the confusion, shouldn't it? It makes the unit of time explicit.

ricardopereira: Yes, it will be less confusion.

ablySpec/Auth.swift
tcard: Or rather, use NSTimeInterval.

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+
+            // RSA9f
+            it("should provide capability has json text") {
+                let rest = ARTRest(options: AblyTests.commonAppSetup())
+
+                let tokenParams = ARTAuthTokenParams()
+                tokenParams.capability = "{ - }"
+
+                rest.auth.createTokenRequest(tokenParams, options: nil, callback: { tokenRequest, error in
+                    expect(error).toNot(beNil())
```
tcard: Same as https://github.com/ably/ably-ios/pull/92/files#r49863069, but inverted.

ablySpec/Auth.swift
tcard: I think this is testing the error case for a capability with invalid JSON, right?

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+                    expect(error?.description).to(contain("Capability"))
+                    expect(tokenRequest?.capability).to(beNil())
+                })
+
+                let expectedCapability = "{ \"cansubscribe:*\":[\"subscribe\"] }"
+                tokenParams.capability = expectedCapability
+
+                rest.auth.createTokenRequest(tokenParams, options: nil, callback: { tokenRequest, error in
+                    expect(error).to(beNil())
```
tcard: Same as https://github.com/ably/ably-ios/pull/92/files#r49863069

ablySpec/Auth.swift
```diff
@@ -599,5 +599,164 @@ class Auth : QuickSpec {
...
+                waitUntil(timeout: testTimeout) { done in
+                    rest.time({ date, error in
+                        serverTime = date
+                        done()
+                    })
+                }
+                expect(serverTime).toNot(beNil(), description: "Server time is nil")
+
+                rest.auth.createTokenRequest(tokenParams, options: authOptions, callback: { tokenRequest, error in
+                    expect(tokenRequest?.clientId).to(equal(expectedClientId))
```
tcard: Same as https://github.com/ably/ably-ios/pull/92/files#r49863069

ablySpec/Auth.swift
ricardopereira: Yes, with an invalid JSON

ablySpec/Auth.swift
ricardopereira: @mattheworiordan Sorry, can you clarify? I'm not using the local time.

ablySpec/Auth.swift
ricardopereira: Ok, I get it but I don't agree on hard coding the current default of TTL on the test. Shouldn't I use the `ARTDefaults.ttl`?

ablySpec/Auth.swift
mattheworiordan: Sorry, I was confused by the description `should provide capability has json text` which I thought meant it was testing the positive condition

ablySpec/Auth.swift
ricardopereira: @mattheworiordan I don't see a way to explicitly set the nonce with the current API. Any thoughts? Or perhaps I didn't get your idea.

ablySpec/Auth.swift
ricardopereira: `queryTime` is the only property used on the `createTokenRequest` method and it is already addressed [here](https://github.com/ably/ably-ios/blob/267c9256ff38fc180b6cd7902d679d597435d5b9/ablySpec/Auth.swift#L745).

ricardopereira:
Rebased with changes.

Lost comments:

[ably/ably-ios] New comment on commit b4bd061
> This is not actually testing that the server time has been used in any way as it is VERY likely the local time is within 6 seconds of the server time anyway. You will need to think of a better way of checking that it's getting and using the server's time.

@mattheworiordan Sorry, can you clarify? I'm not using the local time.

[ably/ably-ios] New comment on commit 6e5429d
> Why the ? in tokenRequest?.nonce? We know it has a value. Also, perhaps we should explicitly set the nonce for this test to ensure we are getting the intended result

@mattheworiordan I don't see a way to explicitly set the nonce with the current API. Any thoughts? Or perhaps I didn't get your idea.

mattheworiordan:
> Sorry, can you clarify? I'm not using the local time.

@ricardopereira the spec states `it("will retrieve the server time if queryTime is true")`, yet if you read the test it logically follows this path:
* query server time and store it
* set `AuthOptions#queryTime` attribute to true
* createTokenRequest and check that the time is close to the server time you retrieved earlier

Why this is flawed is that if you did the following it would pass to:
* query **LOCAL** time and store it
* set `AuthOptions#queryTime` attribute to true
* createTokenRequest and check that the time is close to the server time you retrieved earlier

So you are not actually testing that the server time is being used in any meaningful way.  You need to modify the test so that you are sure that the server time is being used as opposed to the local time when `queryTime` is set to true

>  I don't see a way to explicitly set the nonce with the current API. Any thoughts? Or perhaps I didn't get your idea.

You are right, the current documented API does not allow for this. I do support it, but unofficially in `TokenParams`.  Instead, perhaps just ensure that a check is in place to ensure that the nonce is unique for every token request.

ricardopereira:
@tcard @mattheworiordan Passing

mattheworiordan:
LGTM :+1:

tcard:
LGTM

PR #94, opened 2015-11-27
RSA10

ricardopereira:


ricardopereira:
([RSA10c](http://docs.ably.io/client-lib-development-guide/features/#RSA10c)) _Will not create a new token unless no previous token exists or current token has expired._

Not implemented: 
> Please note that a buffer of 15s for token expiry is recommended to avoid race conditions where the token is valid at the time of the request, but invalid when it reaches the server. Therefore, we recommend that a token is considered expired 15s before the time field `expires`

ricardopereira:
> (RSA10e) Adheres to the implementation of requestToken when issuing new tokens

I need to think about it. Right now I do not have any clue how to achieve this.

ricardopereira:
Not implemented:

> (RSA10g) Stores the AuthOptions and TokenParams arguments as defaults for subsequent authorisations

mattheworiordan:
>> (RSA10c) Will not create a new token unless no previous token exists or current token has expired.

> Not implemented:
> Please note that a buffer of 15s for token expiry is recommended to avoid race conditions where the token is valid at the time of the request, but invalid when it reaches the server. Therefore, we recommend that a token is considered expired 15s before the time field expires

@ricardopereira that is fine, but why?

>> (RSA10e) Adheres to the implementation of requestToken when issuing new tokens

> I need to think about it. Right now I do not have a clue how to achieve this.

Surely you just need to have a test that proves that requesting a token through an authorise call leads to a `requestToken` call with the arguments provided?  If you do that, then it adheres to the spec.

> Not implemented:
> (RSA10g) Stores the AuthOptions and TokenParams arguments as defaults for subsequent authorisations

Why not?  This is quite an important requirement.

ably-ios/ARTAuth.m
```diff
@@ -201,11 +201,12 @@ - (void)executeTokenRequest:(ARTAuthTokenRequest *)tokenRequest callback:(void (
     }];
 }
 
-- (void)authorise:(ARTAuthTokenParams *)tokenParams options:(ARTAuthOptions *)options force:(BOOL)force callback:(void (^)(ARTAuthTokenDetails *, NSError *))callback {
+- (void)authorise:(ARTAuthTokenParams *)tokenParams options:(ARTAuthOptions *)options callback:(void (^)(ARTAuthTokenDetails *, NSError *))callback {
     BOOL requestNewToken = NO;
+    BOOL useTheForce = options ? options.force : false;
```
mattheworiordan: useTheForce?  Sounds like something out of Star Wars.  Why not simply `forceTokenRenewal`?

ably-ios/ARTAuth.m
```diff
@@ -274,6 +275,11 @@ - (void)setProtocolClientId:(NSString *)clientId {
 
 - (void)setTokenDetails:(ARTAuthTokenDetails *)tokenDetails {
     _tokenDetails = tokenDetails;
+    // If a token has been setted and the current Auth method is Basic then for now on use Token Authentication
```
mattheworiordan: I don't understand this comment.  If a valid token is set, why not always use token auth from that point forwards?  Why is the current token method even relevant?

ablySpec/Auth.swift
```diff
@@ -529,5 +529,256 @@ class Auth : QuickSpec {
...
+                // Create a new token
+                waitUntil(timeout: testTimeout) { done in
+                    let rest = ARTRest(options: options)
+
+                    rest.auth.authorise(nil, options: nil, callback: { tokenDetails, error in
+                        expect(rest.auth.method).to(equal(ARTAuthMethod.Token))
+                        expect(tokenDetails).toNot(beNil())
+                        expect(tokenDetails?.token).toNot(beEmpty())
+
+                        if let token = tokenDetails?.token {
```
mattheworiordan: Why if, you have an assertion above that proves token is not empty?

ablySpec/Auth.swift
```diff
@@ -529,5 +529,256 @@ class Auth : QuickSpec {
...
+                            expect(error).to(beNil())
+                            done()
+                        })
+                    })
+                }
+
+                guard let currentToken = validToken else { return }
+
+                waitUntil(timeout: testTimeout) { done in
+                    // New client with Basic auth
```
mattheworiordan: Why not initialise the library with a token instead of with a common setup that includes an API key?  That way you can a) ensure the token is used, b) ensure the API key is not used to generate a new token, c) use the API of the client library instead of `setTokenDetails` which is an undocumented feature anyway

ablySpec/Auth.swift
```diff
@@ -529,5 +529,256 @@ class Auth : QuickSpec {
...
+                    let tokenParams = ARTAuthTokenParams()
+                    tokenParams.clientId = "client_from_params"
+
+                    let authOptions = ARTAuthOptions()
+                    authOptions.authCallback = { tokenParams, callback in
+                        rest.auth.createTokenRequest(ARTAuthTokenParams(clientId: expectedClientId), options: options, callback: callback)
+                    }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        rest.auth.authorise(tokenParams, options: authOptions) { tokenDetails, error in
```
mattheworiordan: If you are testing that the callback is indeed called as expected, then you should use two different clients.  One should be used to generate a token request that has an API key, and the other should be initialised with only a token request callback and no API key.  That way you can be sure that the callback system is working correctly.

ablySpec/Auth.swift
```diff
@@ -529,5 +529,256 @@ class Auth : QuickSpec {
...
+                    options.clientId = "client_string"
+                    let rest = ARTRest(options: options)
+
+                    let authOptions = ARTAuthOptions()
+                    authOptions.authUrl = NSURL(string: "http://auth.ably.io")!
+
+                    waitUntil(timeout: testTimeout) { done in
+                        rest.auth.authorise(nil, options: authOptions) { tokenDetails, error in
+                            expect(error).toNot(beNil())
+                            expect(tokenDetails).to(beNil())
```
mattheworiordan: Ok, makes sense.  Out of interest, how will we ever test that URL auth works? You may want to implement something like https://github.com/ably/ably-js/blob/334c260e192ae417f5b679013cdd6a62cf804c44/spec/realtime/auth.test.js#L53-L107 as well, one for a text token and another for a TokenDetails JSON object

mattheworiordan:
A few comments, but largely good progress, :+1:

ablySpec/Auth.swift
ricardopereira: @mattheworiordan Because without it I need to force the unbox with `!` and it could raise an exception. I can also use the nil coalescing opertor `??` something like `validToken = tokenDetails?.token ?? ""` where if the `tokenDetails` is nil then it assigns an empty string.

ably-ios/ARTAuth.m
ricardopereira: @mattheworiordan Just a way to guarantee the correct auth method is used when a `TokenDetails` but maybe it should be private.

ablySpec/Auth.swift
ricardopereira: @mattheworiordan I implemented this way because the spec is related just with `Auth#authorise` but I will use the API of the client library instead of `setTokenDetails`.

ablySpec/Auth.swift
mattheworiordan: I am still not following.  if `expect(tokenDetails).toNot(beNil())` fails, then everything after that line is invalid anyway as the test will fail.  So I am not clear why you need to use `tokenDetails?` as `tokenDetails` will always be non null.

ricardopereira: @mattheworiordan But the test will continue. The `expect(tokenDetails).toNot(beNil())` will fail but the execution continues to verify all the failing parts.

ablySpec/Auth.swift
mattheworiordan: >  I implemented this way because the spec is related just with `Auth#authorise`

I am not sure that is relevant to my comment to be honest, but that sounds good if you change to not use `setTokenDetails`

ablySpec/Auth.swift
ricardopereira: @mattheworiordan I can change that if you will.

mattheworiordan: No leave as is, it just feels a bit odd that you are writing code that asserts one thing, and then allows for something else later on.

ablySpec/Auth.swift
ricardopereira: @mattheworiordan To generate a token request I need this fix 9740b32. I will do a `cherry-pick`.

ablySpec/Auth.swift
ricardopereira: > b) ensure the API key is not used to generate a new token
Don't understand. It needs the API key to generate a new token, right?

ablySpec/Auth.swift
ricardopereira: The 9740b32 has some changes on `Auth.swift` 😐 damn

mattheworiordan: I'm not following this, are you saying it is possible or not?

ricardopereira: > should be initialised with only a token request callback and no API key

@mattheworiordan it gives an error: `Error Domain=io.ably.cocoa Code=80001 "API Key is missing" UserInfo={NSLocalizedDescription=API Key is missing}`

Am I missing something?
Code:
``` swift
let rest = ARTRest(options: AblyTests.commonAppSetup())

var tokenRequest: ARTAuthTokenRequest? = nil
rest.auth.createTokenRequest(nil, options: nil, callback: { currentTokenRequest, error in
    tokenRequest = currentTokenRequest
})
expect(tokenRequest).toEventuallyNot(beNil(), timeout: testTimeout)

let options = ARTClientOptions()
options.authCallback = { tokenParams, callback in
    return tokenRequest
}

ARTRest(options: options).auth.authorise(nil, options: nil) { tokenDetails, error in
    expect(error).to(beNil()) //<-- error
    expectTokenDetails(tokenDetails)
}
```

ricardopereira: The error is raised here: https://github.com/ably/ably-ios/blob/362a00b08657378188b7b12f6b41ca0cc0dc9e9d/ably-ios/ARTAuth.m#L137

ricardopereira: > should be initialised with only a token request callback and no API key

@mattheworiordan Without a key I can not include the base64 encoded key in the Authorisation header for the `requestToken`. I only can build the url: `"/keys/(tokenRequest.keyName)/requestToken"`

mattheworiordan: @ricardopereira if an error is being raised when calling authorise, then the `authCallback` is clearly broken.  Calling `authorise` should NEVER call `requestToken` if an `authCallback` option is provided because by providing an `authCallback` you are indicating to the library that the library should not issue tokens, but instead rely on the `authCallback` to provide it.  So if this is not working, you need to fix that problem.

ricardopereira:
@mattheworiordan Done

ricardopereira:
Tests are failing because of bug #114.
Should I fix?

mattheworiordan:
> Tests are failing because of bug #114. (test Auth.swift#L801)
> Should I fix?

Well up to you, it would be nice to not introduce PRs that change the build from green to red.  I think you should look at fixing the issue as part of this PR so that this PR passes, but do no more than necessary, and anything else should go in a separate PR to address the issue.  Alternatively, and probably better, perhaps fix https://github.com/ably/ably-ios/issues/114 in a new PR, and then rebase this one on top of that afterwards.

ably-ios/ARTAuth.h
```diff
@@ -30,7 +30,7 @@ ART_ASSUME_NONNULL_BEGIN
 @property (nonatomic, weak) ARTLog *logger;
 
 @property (art_nullable, readonly, getter=getClientId) NSString *clientId;
-@property (art_nullable, nonatomic, readonly, strong) ARTAuthTokenDetails *tokenDetails;
+@property (art_nullable, nonatomic, readwrite, strong) ARTAuthTokenDetails *tokenDetails;
```
mattheworiordan: Why is this `readwrite`?  I assume this means a consumer of this API could explicitly update the `tokenDetails` without using the correct API.  Is this intentional?  If so, why?

ablySpec/Auth.swift
```diff
@@ -599,5 +599,406 @@ class Auth : QuickSpec {
                 expect(createTokenRequestMethodWasCalled).to(beTrue())
             }
         }
+
+        // RSA9
+        describe("createRequestToken") {
```
mattheworiordan: This should be `createTokenRequest`

ablySpec/Auth.swift
```diff
@@ -599,5 +599,406 @@ class Auth : QuickSpec {
...
+            }
+
+        }
+
+        // RSA10
+        describe("authorise") {
+
+            // RSA10a
+            it("should create a token if needed and use it") {
+                let options = AblyTests.clientOptions(requestToken: true)
```
mattheworiordan: I am not familiar with this option `requestToken`, why does that exist?

ablySpec/Auth.swift
```diff
@@ -599,5 +599,406 @@ class Auth : QuickSpec {
...
+                expect(rest.auth.tokenDetails?.token).to(beNil())
+
+                // Created token
+                waitUntil(timeout: testTimeout) { done in
+                    rest.auth.authorise(tokenParams, options: nil) { tokenDetails, error in
+                        expect(error).to(beNil())
+                        expect(tokenDetails).toNot(beNil())
+                        expect(tokenDetails?.token).toNot(beEmpty())
+
+                        if let token = tokenDetails?.token {
```
mattheworiordan: I am not overly keen on this code in that we have what seems like a non-deterministic test.  Can we get read of this `if` condition and simply have one test process that succeeds every time?

ablySpec/Auth.swift
```diff
@@ -599,5 +599,406 @@ class Auth : QuickSpec {
...
+                        expect(error).to(beNil())
+                        expectTokenDetails(tokenDetails)
+                        expect(tokenDetails?.clientId).to(equal(options.clientId))
+                        done()
+                    }
+                }
+            }
+
+            // RSA10g
+            it("should store the AuthOptions and TokenParams arguments as defaults for subsequent authorisations") {
```
mattheworiordan: This test is good, but could we add additional attributes please such as `ttl` and `capability` as all other tests seem to only be testing one attribute `clientId`

mattheworiordan:
Few comments, but mostly looks good

ably-ios/ARTAuth.h
ricardopereira: It is a mistake. I removed the `setTokenDetails` and I change it to `readwrite` just to compile the code successfully and continue the remaining work of this spec.

ablySpec/Auth.swift
ricardopereira: Only for the test suite. It is just an easy way to do:
```
let options = ARTClientOptions()
options.token = getTestToken()
```

mattheworiordan: Ok, if this is not an option available outside of the test suite, then fine, thanks for clarifying

ricardopereira:
Needs #123.

ably-ios/ARTAuth.m
```diff
@@ -135,42 +135,61 @@ - (void)requestToken:(ARTAuthTokenParams *)tokenParams withOptions:(ARTAuthOptio
...
-                                      userInfo:@{ NSLocalizedDescriptionKey: NSLocalizedString(@"API Key is missing", nil) }]);
-    }
-
     if (mergedOptions.authUrl) {
         NSMutableURLRequest *request = [self buildRequest:mergedOptions withParams:currentTokenParams];
         
         [self.logger debug:__FILE__ line:__LINE__ message:@"using authUrl (%@ %@)", request.HTTPMethod, request.URL];
         
-        [_rest executeRequest:request withAuthOption:ARTAuthenticationUseBasic completion:^(NSHTTPURLResponse *response, NSData *data, NSError *error) {
+        [_rest executeRequest:request withAuthOption:ARTAuthenticationOff completion:^(NSHTTPURLResponse *response, NSData *data, NSError *error) {
```
mattheworiordan: Why `ARTAuthenticationOff`, what does that mean?

ablySpec/Auth.swift
```diff
@@ -599,5 +608,510 @@ class Auth : QuickSpec {
...
+                    let tokenParams = ARTAuthTokenParams()
+                    tokenParams.clientId = ExpectedTokenParams.clientId
+                    tokenParams.ttl = ExpectedTokenParams.ttl
+                    tokenParams.capability = ExpectedTokenParams.capability
+
+                    waitUntil(timeout: testTimeout) { done in
+                        rest.auth.authorise(tokenParams, options: nil) { tokenDetails, error in
+                            expect(error).to(beNil())
+                            expectTokenDetails(tokenDetails)
+                            expect(tokenDetails?.clientId).to(equal(ExpectedTokenParams.clientId))
```
mattheworiordan: We should remove this syntax with the trailing `?` as discussed

mattheworiordan:
Few small comments, please address and will merge.  Also, this won't merge automatically now unfottunately

ably-ios/ARTAuth.m
ricardopereira: `ARTAuthenticationOff` means the host does not require authentication (no `Authorization` http header)

ricardopereira:
Rebased with one change c670e8f.

mattheworiordan:
LGTM

tcard:
LGTM

PR #96, opened 2015-11-30
RSL4

ricardopereira:


ablySpec/RestChannel.swift
```diff
@@ -198,14 +198,40 @@ class RestChannel: QuickSpec {
...
+                        if message.payload.encoding.rangeOfString("cipher") == nil {
                             expect(message.content() as? NSObject).to(equal(fixtureMessage["data"].object as? NSObject))
                         }
                     }
                 }
             }
         }
+
+        // RSL4a
+        it("RSL4a") {
```
ricardopereira: @mattheworiordan What do you think? I need to add something to check the sent data.

ablySpec/RestChannel.swift
```diff
@@ -198,14 +212,203 @@ class RestChannel: QuickSpec {
...
+
+        describe("message encoding") {
+
+            struct TestCase {
+                let value: AnyObject?
+                let expected: JSON
+            }
+
+            let text = "John"
+            let integer = "5"
```
mattheworiordan: Not sure I see the value in this test given it's not an integer, but no point removing either

ablySpec/RestChannel.swift
```diff
@@ -198,14 +212,203 @@ class RestChannel: QuickSpec {
...
+                        expect { channel.publish(caseItem, callback: nil) }.to(raiseException())
+                        done()
+                    }
+                }
+            }
+
+            // RSL4b
+            it("encoding attribute should represent the encoding(s) applied in right to left") {
+                let encodingCases = [
+                    TestCase(value: text, expected: nil), //FIXME: Should be UTF-8 ?!
```
mattheworiordan: I believe this is in fact correct, see http://docs.ably.io/client-lib-development-guide/features/#RSL4d2 and https://github.com/ably/ably-js/pull/118#issuecomment-137590035

mattheworiordan: FYI, `utf-8` is only need when converting from binary data (such as encrypted data) so that the decoder knows what to convert it back into

ablySpec/RestChannel.swift
```diff
@@ -198,14 +212,203 @@ class RestChannel: QuickSpec {
...
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.publish(array, callback: { error in
+                            expect(error).to(beNil())
+
+                            if let request = mockExecutor.requests.last, let http = request.HTTPBody {
+                                // Array
+                                let json = JSON(data: http)
+                                // FIXME: "data" : "WyJKb2huIiwiTWFyeSJd"
+                                //expect(json["data"].object as? NSArray).to(equal(array))
+                                expect(json["encoding"].string).to(equal("json/base64")) //FIXME: only "json" ?!
```
mattheworiordan: This is wrong, so we should fix this and make this test pass before merging this PR

mattheworiordan:
Really good progress, however we need to get the tests passing now as opposed to moving on and leaving more TODO / FIXME items in the list.

ablySpec/RestChannel.swift
ricardopereira: #97

ricardopereira:
Needs #122.

mattheworiordan:
Thanks, looks very good, few small comments, but :+1:

ricardopereira:
Needs #122.

ricardopereira:
Rebased

mattheworiordan:
Ok, have the comments been addressed?  Hard to tell as old history is gone, if so, I think we should merge. @tcard do you agree?

ricardopereira:
Yes, all comments addressed.

mattheworiordan:
@tcard please merge if you are happy then

ably-ios/ARTPayload.m
```diff
@@ -317,6 +326,9 @@ - (ARTStatus *)decode:(ARTPayload *)payload output:(ARTPayload *__autoreleasing
 
 @end
 
+
+// MARK: Cipher
```
tcard: What is `MARK`?

ablySpec/RestChannel.swift
```diff
@@ -198,14 +222,228 @@ class RestChannel: QuickSpec {
...
+                ]
+
+                client.httpExecutor = mockExecutor
+
+                validCases.forEach { caseTest in
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.publish(caseTest.value, callback: { error in
+                            expect(error).to(beNil())
+
+                            if let request = mockExecutor.requests.last,
```
tcard: When there's no chance that an optional is nil, they should be unwrapped with `!` instead of if-let-ed. That way, if by any chance it _is_ nil (say we change and break MockHTTPExecutor by accident), this doesn't fail silently.

This would also fail silently if `request.HTTPBody` is nil, but in that case a guard-let would be a better solution, as we've talked about before, because that's actually tested code.

ably-ios/ARTPayload.m
ricardopereira: It helps with the `Show Document Items`:
screen_shot_2016-01-14_at_12_35_34

tcard: Oh, nice.

ablySpec/RestChannel.swift
```diff
@@ -198,14 +222,228 @@ class RestChannel: QuickSpec {
...
+                ]
+
+                client.httpExecutor = mockExecutor
+
+                encodingCases.forEach { caseItem in
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.publish(caseItem.value, callback: { error in
+                            expect(error).to(beNil())
+
+                            if let request = mockExecutor.requests.last,
```
tcard: Same as https://github.com/ably/ably-ios/pull/96/files#r49720773

ablySpec/RestChannel.swift
```diff
@@ -198,14 +222,228 @@ class RestChannel: QuickSpec {
...
+            }
+
+            context("json") {
+                // RSL4d1
+                it("binary payload should be encoded as Base64 and represented as a JSON string") {
+                    client.httpExecutor = mockExecutor
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.publish(binaryData, callback: { error in
+                            expect(error).to(beNil())
+
```
tcard: Same as https://github.com/ably/ably-ios/pull/96/files#r49720773

ablySpec/RestChannel.swift
```diff
@@ -198,14 +222,228 @@ class RestChannel: QuickSpec {
...
+                                done()
+                            })
+                        }
+                    }
+
+                    var totalReceived = 0
+                    try! channel.history(nil) { result, error in
+                        expect(error).to(beNil())
+                        expect(result).toNot(beNil())
+                        expect(result?.hasNext).to(beFalse())
```
tcard: We agreed in the end to replace this with guard-lets, right?

ricardopereira: I am doing it sense we agreed to use guard but I did not change the existing code. Shall we?

tcard: I'm changing it as I review the tests, but because this is not yet on master it would be helpful if you do it before merging.

ricardopereira: Ok, I will

mattheworiordan:
Aside from @tcard's feedback, looks good

tcard:
LGTM

PR #123, opened 2016-01-08
Auth.authorise: store TokenParams and AuthOptions

ricardopereira:


ably-ios/ARTAuth.m
```diff
@@ -84,8 +86,26 @@ - (ARTAuthOptions *)mergeOptions:(ARTAuthOptions *)customOptions {
     return customOptions ? [self.options mergeWith:customOptions] : self.options;
 }
 
+- (void)storeOptions:(ARTAuthOptions *)customOptions {
```
tcard: If ARTAuthOptions some day adds new properties, I think chances are this method will be left outdated. Maybe add this functionality to ARTAuthOptions, as described [here](https://developer.apple.com/library/ios/documentation/General/Conceptual/DevPedia-CocoaCore/ObjectCopying.html)?

Why do we need to copy the options anyway?

ricardopereira: But it should not store all the elements and `copyWithZone:` should copy all of them as good practice.

ricardopereira: [RSA10g](http://docs.ably.io/client-lib-development-guide/features/#RSA10g)

tcard: Oh, fair enough.

tcard:
LGTM

mattheworiordan:
LGTM

Issue #111, opened 2015-12-30
Realtime: add `dispose` method


ricardopereira:
Method to explicitly dispose all resources

 - [ ] Update tests

ricardopereira:
Fixed fb3c2e0

PR #128, opened 2016-01-14
Realtime: `dispose` method

ricardopereira:
#111

### DIFF
--- a/ably-ios/ARTRealtime.h
+++ b/ably-ios/ARTRealtime.h
@@ -56,6 +56,9 @@ Instance the Ably library with the given options.
 - (BOOL)connect;
 - (BOOL)isActive;
 
+/// Explicitly dispose all resources
+- (void)dispose;
+
 - (ARTRealtimeConnectionState)state;
 - (NSString *)connectionId;
 - (NSString *)connectionKey;
--- a/ably-ios/ARTRealtime.m
+++ b/ably-ios/ARTRealtime.m
@@ -52,8 +52,13 @@
 @property (readwrite, assign, nonatomic) int64_t connectionSerial;
 @property (readwrite, assign, nonatomic) int64_t msgSerial;
 
-@property (readwrite, strong, nonatomic) NSMutableArray *queuedMessages;
-@property (readonly, strong, nonatomic) NSMutableArray *pendingMessages;
+/// List of queued messages on a connection in the disconnected or connecting states
+@property (readwrite, strong, nonatomic) __GENERIC(NSMutableArray, ARTQueuedMessage*) *queuedMessages;
+
+/// List of pending messages waiting for ACK/NACK action to confirm the success receipt and acceptance
+@property (readonly, strong, nonatomic) __GENERIC(NSMutableArray, ARTQueuedMessage*) *pendingMessages;
+
+/// To verify skipped messages, earlier messages... comparing with `msgSerial` response from ACK/NACK action
 @property (readwrite, assign, nonatomic) int64_t pendingMessageStartSerial;
 
 @property (nonatomic, copy) ARTRealtimePingCb pingCb;
@@ -145,6 +150,8 @@
 }
 
 - (void)dispose {
+    [self.pendingMessages removeAllObjects];
+    self.pendingMessageStartSerial = 0;
     [self removeAllChannels];
     [self.eventEmitter removeEvents];
 }
--- a/ably-ios/ARTRealtime.m
+++ b/ably-ios/ARTRealtime.m
@@ -144,6 +144,11 @@
     return self;
 }
 
+- (void)dispose {
+    [self removeAllChannels];
+    [self.eventEmitter removeEvents];
+}
+
 - (id<ARTRealtimeTransport>)getTransport {
     return _transport;
 }


tcard:
I recall some comments about this, but some context about why do we want this would be helpful.

Also, I couldn't find this method on other client libs. What's the story there?

mattheworiordan:
> Also, I couldn't find this method on other client libs. What's the story there?

Do you mean `dispose`?  If so, I believe this was introduced because @ricardopereira identified memory issues in our test suite where each invocation of the library was never being freed up, so an explicit `dispose` to free up the memory was introduced.  If you feel this is not valid for an iOS lib, please advise.

tcard:
Why would iOS be different that any other library? If there is a memory leak, I guess it's because we keep a live reference to the ARTRealtime instance somewhere we shouldn't, right?

mattheworiordan:
> If there is a memory leak, I guess it's because we keep a live reference to the ARTRealtime instance somewhere we shouldn't, right?

I would assume so, although I thought Objective-C had quirky GC support.  You will have to make your recommendation as I don't know the specifics of memory management with Objective-C

tcard:
The quirks of Obj-C/Swift memory management shouldn't affect this case, as far as I can see. Leaks can be caused by exactly the same reasons as in any GCed language (Ruby, Java, etc.), plus cycles, but I don't see a problem of cycles (and if there is, it should be fixed with weak references).

@ricardopereira, could you provide the full story for this method?

ricardopereira:
Some of the tests were failing with _Questionable API usage errors_ when the Travis server ran all the tests. I digged into it and I found out that some callbacks (e.g.: state event subscriptions) were called after the `RealtimeClient` were disposed. I fixed some retain cycles that reduced some of the errors but I still had some problems. I checked memory issues with the `Instruments` tool and I noticed that some `RealtimeClients` took some time to freed up but in the end all the instances were getting cleanup. So, I decided to explicitly clean the channels and events on every test invocation just for a precaution.

Another example where I decided to clean the channels and events just for convenience:

``` swift
context("test") {
    options.autoConnect = false

    it("one test") {
        client.connect()
        defer {
            // Without dispose
            //client.dispose()
            client.close()
        }

            client.eventEmitter.on { state, error in
                print("1. received state \(state)")
                done()
            }
        }
    }
    
    it("second test") {
        client.connect()
        defer {
            // Without dispose
            //client.dispose()
            client.close()
        }
        
            client.eventEmitter.on { state, error in
                print("2. received state \(state)")
                done()
            }
        }
    }
}
```

**Result:**
```
Connection__test__one_test_UsersricardopereiraRepositoriesWhitesmithablyiosablySpecRealtimeClientconnectionswift_401]' started.
1. received state ARTRealtimeConnectionState
1. received state ARTRealtimeConnectionState
1. received state ARTRealtimeConnectionState
Test Case '-[ablySpec.RealtimeClientConnection Connection__test__one_test_UsersricardopereiraRepositoriesWhitesmithablyiosablySpecRealtimeClientconnectionswift_401]' passed (0.019 seconds).
Test Case '-[ablySpec.RealtimeClientConnection Connection__test__second_test_UsersricardopereiraRepositoriesWhitesmithablyiosablySpecRealtimeClientconnectionswift_415]' started.
1. received state ARTRealtimeConnectionState
2. received state ARTRealtimeConnectionState
1. received state ARTRealtimeConnectionState
2. received state ARTRealtimeConnectionState
1. received state ARTRealtimeConnectionState
2. received state ARTRealtimeConnectionState
Test Case '-[ablySpec.RealtimeClientConnection Connection__test__second_test_UsersricardopereiraRepositoriesWhitesmithablyiosablySpecRealtimeClientconnectionswift_415]' passed (0.014 seconds).
```

tcard:
If this method is used as a test convenience, I'd say it should be kept in the tests, don't you think? Either as an extension or as just a function.

There are some other methods (like removeAllChannels) that are not part of the documented API and mainly bloat the API, in my opinion. I think it's really important to keep the interface clean and minimal.

mattheworiordan:
> There are some other methods (like removeAllChannels) that are not part of the documented API and mainly bloat the API, in my opinion

Agreed.  In the case of Ruby there is a convention to document some methods as private APIs, even though they are not private methods, but even in those cases I have kept this to a minimum as the less exposed methods that are not part of the API the better.

ricardopereira:
Well, I never have in mind to keep the API interface at the same level from other platforms but it makes sense. I added `removeAllChannels` because I thought it would be helpful but it is not my call to decide those changes. Lesson learned.

ricardopereira:
So, I will include `dispose` only for the test suite.

@mattheworiordan Just to clarify one thing: if the user closes the connection, should the library clean the `pendingMessages`? AFAIK, `queuedMessages` should remain for the next connection, right?

mattheworiordan:
> Just to clarify one thing: if the user closes the connection, should the library clean the pendingMessages? AFAIK, queuedMessages should remain for the next connection, right?

Well if you look at http://docs.ably.io/client-lib-development-guide/features/#RTN18b and http://docs.ably.io/client-lib-development-guide/features/#RTN18c it states that in suspended or failed states the messages should be marked as failed.  As the `CLOSED` state is not transient, the messages should also be treated as failed as the client library is no longer trying to send them.

PR #136, opened 2016-01-20
Use guard-let instead of if-let, ?-chaining and force-unwrapping.

tcard:
This is more idiomatic, plus it forces the code to handle the else
case, plus avoids an additional level of indentation inside the
if-let.


mattheworiordan:
Nice, :+1:

tcard:
Swift tests succeed, Obj-C tests aren't run (#143), but this PR can't possibly break those. So tests should be considered passing.

ricardopereira:
LGTM

PR #137, opened 2016-01-20
Replace ?-chains with force-unwraps.

tcard:
There is some abuse of ?-chaining in the test code. If we expect a
thing to be there and if not being that the case would mean that
test code (e. g. a helper) is wrong, force-unwrap should be used
instead of silently ignoring this condition with a ?-chain so that
the error is caught from the test framework. The test framework
should only tell us about the code being tested, not about the
code that tests the code being tested.

### DIFF
--- a/ablySpec/Auth.swift
+++ b/ablySpec/Auth.swift
@@ -53,10 +53,10 @@ class Auth : QuickSpec {
                 }
 
                 let key64 = NSString(string: "\(client.options.key!)")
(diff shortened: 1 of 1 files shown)


tcard:
Tests seem to be failing due to what #139 is supposed to fix.

mattheworiordan:
Agreed, :+1:

ricardopereira:
LGTM

PR #138, opened 2016-01-20
Don't use if-let to unwrap when it is sure that it will succeed.

tcard:
Force-unwrap with `!` instead. Otherwise, if the if-let fails to
unwrap the optional, it will be silently ignored. `!` crashes the
program at least, so that we know that we have something wrong
in the test code.

### DIFF
--- a/ablySpec/RestClient.swift
+++ b/ablySpec/RestClient.swift
@@ -227,11 +227,11 @@ class RestClient: QuickSpec {
                     }
                 }
 
-                if let requestUrlA = mockExecutor.requests.first?.URL,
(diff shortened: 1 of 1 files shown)


tcard:
Tests seem to be failing due to what #139 is supposed to fix.

mattheworiordan:
:+1:

tcard:
Now tests fail because of #142, but succeed in Swift otherwise. Obj-C suite isn't run, but can't be broken by this PR. So tests can be considered to be passing.

ricardopereira:
LGTM

PR #139, opened 2016-01-20
waitUntil instead of toEventually in RTN9b test.

tcard:
Tests were intermitently failing because of timing. This uses the
same strategy as the RTN8b test above.

This should be merged with some priority so that other PR tests
can pass more consistently.


mattheworiordan:
LGTM

tcard:
Got LGTM from @ricardopereira on chat.

PR #141, opened 2016-01-20
Properly annotate ARTConnection.id, ARTConnection.key as optional.

tcard:
RTN8a and RTN9a require them to be nullable.


mattheworiordan:
LGTM

tcard:
Tests fail because of #142, but succeed otherwise, also in Obj-C. So tests can be considered to be passing.

ricardopereira:
LGTM

PR #147, opened 2016-01-20
RTL4: Fix test that was doing nothing.

tcard:
hasAttached was never being written. Also, now expectation is
fullfilled when the channel is closed so that the no-op
`[channel attach]` calls can have an observable effect if they're
wrong, ie. if they transition again to ATTACHED.

### DIFF
--- a/ably-iosTests/ARTRealtimeAttachTest.m
+++ b/ably-iosTests/ARTRealtimeAttachTest.m
@@ -51,19 +51,24 @@
 
                 __block bool hasAttached = false;
                 [channel subscribeToStateChanges:^(ARTRealtimeChannelState state, ARTStatus *reason) {
-                    XCTAssertEqual(ARTStateOk, reason.state);
                     if(state == ARTRealtimeChannelAttaching) {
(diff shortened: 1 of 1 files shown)


ricardopereira:
LGTM

ricardopereira:
```
/Users/travis/build/ably/ably-ios/ably-iosTests/ARTRealtimeAttachTest.m:63:38: no visible @interface for 'ARTRealtimeChannel' declares the selector 'close'
```

tcard:
Sorry, I forgot to run the Obj-C suite on my machine. PTAL

mattheworiordan:
Ok, bit of a messy test but it's legacy so looks good in that at least it's fixed

tcard:
Pretty much all the Obj-C tests are like this. Each of them is huge, and test several things. IMHO it's not worth fixing most of them, they should be superseded by proper unit tests in the Swift suite.

mattheworiordan:
Agreed, not worth fixing, we have bigger fish to fry!

PR #151, opened 2016-01-20
Uncomment RSN4 test that now works.

tcard:


mattheworiordan:
:+1:

ricardopereira:
The test is marked has pending: `pending("releaseChannel")`. I will change that.

mattheworiordan:
Ok, didn't see that in the commit as it was off the page, thanks

ricardopereira:
The test is passing.

I made a test:
```
class Item {
    var name: String = ""
    init(name: String) {
        self.name = name
    }
}

class Collection {
    var dictionary = NSMutableDictionary()
    func add(name: String) -> Item {
        dictionary[name] = Item(name: name)
        return dictionary[name] as! Item
    }
    func release(item: Item) {
        dictionary.removeObjectForKey(item.name)
    }
}

let items = Collection()
weak var item1: Item! = items.add("asjflkajsdf")

print(item1.name)
items.release(item1)
print(item1.name) //<-- EXC_BAD_INSTRUCTION 💥
```

And `item1 == nil` is true.

The `ARTChannelCollection` also uses `removeObjectForKey`, so it is something retaining the channel.

mattheworiordan:
Ok, leave for now and we'll fix later. Let's see how the write tests approach works over the next few days

ricardopereira:
Ok :+1:

tcard:
Oh, come on... Maybe we can just write this test in Obj-C for now. Anyway, if it's pending it shouldn't be also commented, don't you think?

ricardopereira:
Yes. First I commented it and then I found out about `pending` because it isn't referenced on the API docs.

tcard:
So should I merge this anyway?

PR #154, opened 2016-01-21
RTN7c

ricardopereira:
Need fix.

### DIFF
--- a/ablySpec/RealtimeClient.connection.swift
+++ b/ablySpec/RealtimeClient.connection.swift
@@ -656,16 +656,20 @@ class RealtimeClientConnection: QuickSpec {
                         }
 
                         let channel = client.channel("channel")
+                        let transport = client.transport as! TestProxyTransport
+                        transport.actionsIgnored += [.Ack, .Nack]
 
                         waitUntil(timeout: testTimeout) { done in
                             channel.subscribeToStateChanges { state, status in
                                 if state == .Attached {
                                     channel.publish("message", cb: { status in
                                         expect(status.state).to(equal(ARTState.Error))
-                                        expect(status.errorInfo!.code).to(equal(80003))
                                         done()
                                     })
-                                    client.close()
+                                    // Wait until the message is pushed to Ably first
+                                    delay(1.0) {
+                                        transport.simulateIncomingNormalClose()
+                                    }
                                 }
                             }
                             channel.attach()
@@ -680,16 +684,20 @@ class RealtimeClientConnection: QuickSpec {
                         }
 
                         let channel = client.channel("channel")
+                        let transport = client.transport as! TestProxyTransport
+                        transport.actionsIgnored += [.Ack, .Nack]
 
                         waitUntil(timeout: testTimeout) { done in
                             channel.subscribeToStateChanges { state, status in
                                 if state == .Attached {
                                     channel.publish("message", cb: { status in
                                         expect(status.state).to(equal(ARTState.Error))
-                                        expect(status.errorInfo!.code).to(equal(80000))
                                         done()
                                     })
-                                    client.fail()
+                                    // Wait until the message is pushed to Ably first
+                                    delay(1.0) {
+                                        transport.simulateIncomingError()
+                                    }
                                 }
                             }
                             channel.attach()
@@ -697,19 +705,30 @@ class RealtimeClientConnection: QuickSpec {
                     }
 
                     it("lost connection state") {
-                        let client = ARTRealtimeExtended(options: AblyTests.commonAppSetup())
+                        let options = AblyTests.commonAppSetup()
+                        options.autoConnect = false
+                        let client = ARTRealtimeExtended(options: options)
+                        client.setTransportClass(TestProxyTransport.self)
+                        client.connect()
                         defer { client.close() }
 
                         let channel = client.channel("channel")
 
-                        waitUntil(timeout: testTimeout) { done in
+                        let transport = client.transport as! TestProxyTransport
+                        transport.actionsIgnored += [.Ack, .Nack]
+
+                        waitUntil(timeout: testTimeout + options.disconnectedRetryTimeout) { done in
                             channel.subscribeToStateChanges { state, status in
                                 if state == .Attached {
                                     channel.publish("message", cb: { status in
                                         expect(status.state).to(equal(ARTState.Error))
                                         done()
                                     })
-                                    client.lostConnection()
+                                    // Wait until the message is pushed to Ably first
+                                    delay(1.0) {
+                                        client.simulateLostConnection()
+                                        expect(client.connection().state).toEventually(equal(ARTRealtimeConnectionState.Connecting), timeout: options.disconnectedRetryTimeout)
+                                    }
                                 }
                             }
                             channel.attach()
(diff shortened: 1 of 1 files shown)


ricardopereira:
Tests are failing because of:
 - published a message (the connection gets closed) and the callback isn't called.
 - published a message (the connection gets a failure) and the callback isn't called.

tcard:
Test looks good. I guess it's an internal problem.

ricardopereira:
I think that too.

mattheworiordan:
This needs rebasing unfortunately, can you do that before I review & merge?

ricardopereira:
Rebased with no changes.

> Tests are failing because of:
 - published a message (the connection gets closed) and the callback isn't called.
 - published a message (the connection gets a failure) and the callback isn't called.

Should I open an issue?

ablySpec/RealtimeClient.connection.swift
```diff
@@ -648,6 +648,78 @@ class RealtimeClientConnection: QuickSpec {
                         expect(nacks[0].msgSerial).to(equal(6))
                         expect(nacks[0].count).to(equal(1))
                     }
+                }
+
+                // RTN7c
+                context("should deliver as failed the remaining pending messages if") {
```
mattheworiordan: I think the wording here should be "should trigger the failure callback for the remaining pending messages if"

ablySpec/RealtimeClient.connection.swift
```diff
@@ -648,6 +648,78 @@ class RealtimeClientConnection: QuickSpec {
...
+
+                        waitUntil(timeout: testTimeout) { done in
+                            channel.subscribeToStateChanges { state, status in
+                                if state == .Attached {
+                                    channel.publish("message", cb: { status in
+                                        expect(status.state).to(equal(ARTState.Error))
+                                        expect(status.errorInfo!.code).to(equal(80000))
+                                        done()
+                                    })
+                                    client.fail()
```
mattheworiordan: Does fail need to be a public method?  It would be nice, if possible, to not expose a method like that

ricardopereira: No, it is an extension only available with `TestUtilities` (test suite).

ablySpec/RealtimeClient.connection.swift
```diff
@@ -648,6 +648,78 @@ class RealtimeClientConnection: QuickSpec {
...
+
+                        waitUntil(timeout: testTimeout) { done in
+                            channel.subscribeToStateChanges { state, status in
+                                if state == .Attached {
+                                    channel.publish("message", cb: { status in
+                                        expect(status.state).to(equal(ARTState.Error))
+                                        expect(status.errorInfo!.code).to(equal(80003))
+                                        done()
+                                    })
+                                    client.close()
```
mattheworiordan: We need to wait until the message has actually been sent along the transport.  As it stands, the message is unlikely to have been sent to Ably by the time the connection has closed, so instead you should publish, wait until the message has been sent, then simulate an incoming `CLOSED` messages and close the transport before the `ACK` arrives.  Calling an explicit close is not correct as the CLOSED response should, in most cases, come back after the `ACK`

ablySpec/RealtimeClient.connection.swift
mattheworiordan: As above too, we need to ensure the message is pushed to Ably first, then you can simulate an incoming ERROR message on the transport that would cause the lib to move to the FAILED state

ablySpec/RealtimeClient.connection.swift
```diff
@@ -648,6 +648,78 @@ class RealtimeClientConnection: QuickSpec {
...
+                        let channel = client.channel("channel")
+
+                        waitUntil(timeout: testTimeout) { done in
+                            channel.subscribeToStateChanges { state, status in
+                                if state == .Attached {
+                                    channel.publish("message", cb: { status in
+                                        expect(status.state).to(equal(ARTState.Error))
+                                        done()
+                                    })
+                                    client.lostConnection()
```
mattheworiordan: As above, please ensure the message has been sent first, and then perhaps also check that the connection has been reconnected before checking that the message callback with an error has been called.

ablySpec/RealtimeClient.connection.swift
```diff
@@ -648,6 +648,78 @@ class RealtimeClientConnection: QuickSpec {
...
+                        let client = ARTRealtimeExtended(options: AblyTests.commonAppSetup())
+                        defer { client.close() }
+
+                        let channel = client.channel("channel")
+
+                        waitUntil(timeout: testTimeout) { done in
+                            channel.subscribeToStateChanges { state, status in
+                                if state == .Attached {
+                                    channel.publish("message", cb: { status in
+                                        expect(status.state).to(equal(ARTState.Error))
```
mattheworiordan: @tcard are you happy with this style of callback i.e. we don't have an error attribute that could be null but instead have a status object with the state and I assume the error info?

ablySpec/RealtimeClient.connection.swift
mattheworiordan: > No, it is an extension only available with TestUtilities (test suite).

OK

ablySpec/RealtimeClient.connection.swift
tcard: I think it's basically the same, but having two arguments that need to be at least typed down explicitly encourages the user to acknowledge the error, so it probably leads to more robust code.

Ideally we would use an enum, so that you need to pattern-match it to extract either the OK value or the error, but there's no such thing in Objective-C, only in Swift.

mattheworiordan: > I think it's basically the same, but having two arguments that need to be at least typed down explicitly encourages the user to acknowledge the error, so it probably leads to more robust code.

Ok, what would you suggest then?

ablySpec/RealtimeClient.connection.swift
ricardopereira: @mattheworiordan The message is sent before it reaches the `close` call. It isn't an async call. Even so, I will check that.

ricardopereira: It sends the action 15 before calling the `close()`, so it waits until the message has been sent. I will change the code to simulate an incoming CLOSED.

ablySpec/RealtimeClient.connection.swift
ricardopereira: > As above too, we need to ensure the message is pushed to Ably first, then you can simulate an incoming ERROR message on the transport that would cause the lib to move to the FAILED state

That's the current behaviour.

ablySpec/RealtimeClient.connection.swift
tcard: Maybe `errorInfo` could be a separate argument? But anyway, if it breaks the API I don't think it's so important in this case.

ablySpec/RealtimeClient.connection.swift
mattheworiordan: > It isn't an async call. Even so, I will check that.

Exactly, because it is async unfortunately there are no guarantees about what has happened i.e. it could be in the queue or it could be sent and waiting an `N?ACK`

ablySpec/RealtimeClient.connection.swift
mattheworiordan: Nope, not in this case, it was more of a question for you when reviewing the API @tcard

ricardopereira:
Rebased with changes.

mattheworiordan:
LGTM

tcard:
LGTM

PR #157, opened 2016-01-21
Make ARTPaginatedResult generic.

tcard:
At first I thought this would help to write test code, as Xcode would help with autocomplete. Turns out that Swift ignores generics, except for Foundation collections. Anyway, it's done, and Objective-C clients can benefit from them.


ricardopereira:
Really great! :+1:

mattheworiordan:
:+1:

PR #158, opened 2016-01-21
RSC1: Add token string constructor.

tcard:


ablySpec/RestClient.swift
```diff
@@ -45,6 +45,32 @@ class RestClient: QuickSpec {
...
+                                return
+                            }
+
+                            guard let tokenDetails = tokenDetails else {
+                                XCTFail("expected tokenDetails not to be nil when error is nil")
+                                done()
+                                return
+                            }
+
+                            let client = ARTRest(token: tokenDetails.token)
```
ricardopereira: If you just need a token you can use `getTestToken` from test suite.

mattheworiordan: Should we not be initialising with just a token i.e. `ARTRest(tokenDetails.token)` as well?

tcard: @ricardopereira Good point. Will do.

Sorry @mattheworiordan , I don't know what you mean. What you say is how it looks like now (except for the named parameter).

mattheworiordan: In other languages we have an overloaded constructor so that you can pass in just a token or API key string, or options.  If this is not possible in iOS then it's absolutely fine, but if it is possible it's a nice touch.

tcard: It can be done, but it's strongly anti-idiomatic. All constructors that take arguments are like `initWithSomething:something`. So we have `initWithOptions`, `initWithToken`, `initWithKey`.

tcard: (Note that Swift automatically removes the "With", but still requires a named argument.)

ricardopereira: The conversion removes the `initWith`: 

> Objective-C initializers begin with init, or initWith: if the initializer takes one or more arguments. When an Objective-C initializer is imported by Swift, the init prefix becomes an init keyword to indicate that the method is a Swift initializer. If the initializer takes an argument, the With is removed and the rest of the selector is divided up into named parameters accordingly.
https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithObjective-CAPIs.html#//apple_ref/doc/uid/TP40014216-CH4-ID35

ricardopereira: Only the designated initialiser can be used like `ARTRest(value)`. Right now, I think the default initialiser is `initWithLogger:andOptions`.

tcard: > If you just need a token you can use getTestToken from test suite.

Done at 3fd9a731ec4a1490c7c11ecbc9b22195f60ac1de

mattheworiordan:
LGTM

ricardopereira:
LGTM

PR #159, opened 2016-01-22
RTN10b

ricardopereira:
Not passing.

### DIFF
--- a/ablySpec/RealtimeClient.connection.swift
+++ b/ablySpec/RealtimeClient.connection.swift
@@ -793,6 +793,26 @@ class RealtimeClientConnection: QuickSpec {
                     }
                 }
 
+                // RTN10b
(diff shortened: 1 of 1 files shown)


mattheworiordan:
Ok, looks good.

ricardopereira:
Rebased with changes.

Connection serial isn't set to -1 when a new connection is opened: `expected to equal <-1>, got <0>`.
Should I open an issue?

mattheworiordan:
> Should I open an issue?

We should discuss at our stand up today what we do with pull requests with tests that fail.  I am concerned that if we merge them in we are only compounding the problem of our unreliable test suite.  One to discuss later

tcard:
LGTM

We agreed on marking those as pending, right?

ricardopereira:
Yes, after two _LGTM_.

PR #160, opened 2016-01-22
RTN10c

ricardopereira:


ricardopereira:
Not passing but I think I implemented it well.

mattheworiordan:
LGTM

ablySpec/RealtimeClient.connection.swift
```diff
@@ -713,6 +713,44 @@ class RealtimeClientConnection: QuickSpec {
...
+                it("should have last known connection serial from restored connection") {
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                    defer { client.close() }
+                    let channel = client.channel("test")
+
+                    var lastSerial: Int64 = 0
+                    for _ in 1...5 {
+                        channel.publish("message", cb: { status in
+                            expect(status.state).to(equal(ARTState.Ok))
+                            lastSerial = client.connection().serial
```
tcard: Correct me if I'm wrong, but there's no guarantee that the last callback called will be that of the last published, since it could take the server more time to respond to any of them. (I've checked the client code and that seems to be indeed what's happening.)

tcard: Sorry, scratch that, I though that connection.serial was updated before publishing, but no, it is set when the N?ACK message is received.

tcard:
The test looks good to me.

ablySpec/RealtimeClient.connection.swift
```diff
@@ -713,6 +713,44 @@ class RealtimeClientConnection: QuickSpec {
...
+                        })
+                    }
+                    expect(lastSerial).toEventually(equal(4), timeout: testTimeout)
+
+                    let options = AblyTests.clientOptions(requestToken: true)
```
mattheworiordan: Why are you using token auth here?  Not sure how that applies to this test when you have not used token auth for the first `client`

mattheworiordan:
Mostly good, one small comment re: token auth

ablySpec/RealtimeClient.connection.swift
ricardopereira: No reason in particular.

ricardopereira:
Rebased with no changes.

> Not passing but I think I implemented it well. The connectionSerial from the recovered client is 0.

Should I open an issue and merge?

ablySpec/RealtimeClient.connection.swift
mattheworiordan: I would remove as it's just more noise

ricardopereira: Done

PR #162, opened 2016-01-22
RTN12b

ricardopereira:
Not passing. The `realtimeRequestTimeout` is not used, so... feature not implemented.

### DIFF
--- a/ablySpec/RealtimeClient.connection.swift
+++ b/ablySpec/RealtimeClient.connection.swift
@@ -887,6 +887,50 @@ class RealtimeClientConnection: QuickSpec {
                     expect(states[1]).to(equal(ARTRealtimeConnectionState.Closed))
                 }
 
+                // RTN12b
+                it("should transition to CLOSED action when the close process timeouts") {
+                    let options = AblyTests.commonAppSetup()
+                    options.autoConnect = false
+                    let client = ARTRealtime(options: options)
+                    client.setTransportClass(TestProxyTransport.self)
+                    client.connect()
+
+                    let transport = client.transport as! TestProxyTransport
+                    transport.actionsIgnored += [.Closed]
+
+                    var states: [ARTRealtimeConnectionState] = []
+                    var start: NSDate?
+                    var end: NSDate?
+
+                    client.eventEmitter.on { state, errorInfo in
+                        switch state {
+                        case .Connected:
+                            client.close()
+                        case .Closing:
+                            start = NSDate()
+                            states += [state]
+                        case .Closed:
+                            end = NSDate()
+                            states += [state]
+                        case .Disconnected:
+                            states += [state]
+                        default:
+                            break
+                        }
+                    }
+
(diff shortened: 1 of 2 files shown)


ablySpec/RealtimeClient.connection.swift
```diff
@@ -713,6 +713,56 @@ class RealtimeClientConnection: QuickSpec {
...
+                        switch state {
+                        case .Connected:
+                            client.close()
+                        case .Closing:
+                            start = NSDate()
+                            states += [state]
+                        case .Closed:
+                            end = NSDate()
+                            states += [state]
+                        case .Disconnected:
```
tcard: Why should this happen?

ablySpec/RealtimeClient.connection.swift
```diff
@@ -713,6 +713,56 @@ class RealtimeClientConnection: QuickSpec {
                 }
 
             }
+
+            // RTN12
+            context("close") {
+
+                // RTN12b
+                it("should receive CLOSED action within the default request timeout") {
```
tcard: As I read RTN12b, what should be tested is that the transport is disconnected and the connection transitions to 'closed' when the close process timeouts. The test looks good but this sentence should describe that.

ablySpec/RealtimeClient.connection.swift
```diff
@@ -713,6 +713,56 @@ class RealtimeClientConnection: QuickSpec {
...
+                    expect(start).toEventuallyNot(beNil(), timeout: testTimeout)
+                    expect(end).toEventuallyNot(beNil(), timeout: ARTDefault.realtimeRequestTimeout())
+
+                    if states.count != 3 {
+                        fail("Invalid number of connection states. Expected CLOSING, DISCONNECTED and CLOSE states")
+                        return
+                    }
+
+                    expect(states[0]).to(equal(ARTRealtimeConnectionState.Closing))
+                    expect(states[1]).to(equal(ARTRealtimeConnectionState.Disconnected))
```
mattheworiordan: There is nothing in the spec that says the connection should move to `Disconnected`, this is wrong.  The connection, when closing, will only ever move to closed directly.

ablySpec/RealtimeClient.connection.swift
ricardopereira: Ok, changed to `should transition to CLOSED action when the close process timeouts`.

ablySpec/RealtimeClient.connection.swift
ricardopereira: It should not happen but I want to catch it and check if `states` doesn't have a `.Disconnected` event.

ablySpec/RealtimeClient.connection.swift
ricardopereira: Fixed.

ricardopereira:
Rebased with changes.

mattheworiordan:
LGTM

tcard:
LGTM

PR #163, opened 2016-01-22
RTN12c

ricardopereira:
Passing

### DIFF
--- a/ablySpec/RealtimeClient.connection.swift
+++ b/ablySpec/RealtimeClient.connection.swift
@@ -926,6 +926,47 @@ class RealtimeClientConnection: QuickSpec {
                     expect(states[1]).to(equal(ARTRealtimeConnectionState.Closed))
                 }
 
+                // RTN12c
+                it("transitions to the CLOSING state and then to the CLOSED state if the transport is abruptly closed") {
+                    let options = AblyTests.commonAppSetup()
+                    options.autoConnect = false
+                    let client = ARTRealtime(options: options)
+                    client.setTransportClass(TestProxyTransport.self)
+                    client.connect()
+
+                    let transport = client.transport as! TestProxyTransport
+                    var states: [ARTRealtimeConnectionState] = []
+
+                    waitUntil(timeout: testTimeout) { done in
+                        client.eventEmitter.on { state, errorInfo in
+                            switch state {
+                            case .Connected:
+                                states += [state]
+                                client.close()
+                            case .Closing:
+                                states += [state]
+                                transport.simulateIncomingAbruptlyClose()
+                            case .Closed:
+                                states += [state]
+                                done()
+                            case .Disconnected, .Suspended, .Failed:
+                                states += [state]
+                            default:
+                                break
+                            }
+                        }
+                    }
+
+                    if states.count != 3 {
+                        fail("Invalid number of connection states. Expected CONNECTED, CLOSING and CLOSE states (got \(states.map{ $0.rawValue }))")
+                        return
+                    }
+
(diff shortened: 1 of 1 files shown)


ablySpec/TestUtilities.swift
```diff
@@ -414,6 +414,13 @@ class TestProxyTransport: ARTWebSocketTransport {
         super.receive(msg)
     }
 
+    func closeAbruptly() {
+        // Inject a ProtocolMessage.CLOSE
```
tcard: I don't think this is a valid mock, because if the real transport does indeed close abruptly it won't make up a ProtocolMessage.CLOSE. I guess the connection should be notified of the close somehow (with a callback, I guess) and then check the current state, so that if it is CLOSING it transitions to CLOSED.

ablySpec/RealtimeClient.connection.swift
```diff
@@ -713,6 +713,49 @@ class RealtimeClientConnection: QuickSpec {
                 }
 
             }
+
+            // RTN12
+            context("close") {
+
+                // RTN12c
+                it("close move to CLOSED state if transport is abruptly closed") {
```
mattheworiordan: This description is not clear, it should be `transitions to the CLOSING state and then to the CLOSED state if the transport is abruptly closed`

ablySpec/RealtimeClient.connection.swift
```diff
@@ -713,6 +713,49 @@ class RealtimeClientConnection: QuickSpec {
...
+
+                    let transport = client.transport as! TestProxyTransport
+                    var states: [ARTRealtimeConnectionState] = []
+
+                    waitUntil(timeout: testTimeout) { done in
+                        client.eventEmitter.on { state, errorInfo in
+                            switch state {
+                            case .Connected:
+                                states += [state]
+                                transport.closeAbruptly()
```
mattheworiordan: Why are you waiting until it is connected for an abrupt close?  The spec says after calling `close`, if the transport is abruptly terminated then the connection simply moves from the `closing` state it would be in straight to the `closed` state

ablySpec/RealtimeClient.connection.swift
```diff
@@ -713,6 +713,49 @@ class RealtimeClientConnection: QuickSpec {
...
+                        }
+                    }
+
+                    if states.count != 2 {
+                        fail("Invalid number of connection states. Expected CONNECTED and CLOSE states")
+                        return
+                    }
+
+                    expect(states[0]).to(equal(ARTRealtimeConnectionState.Connected))
+                    expect(states[1]).to(equal(ARTRealtimeConnectionState.Closed))
```
mattheworiordan: This is not the sequence we expect, it is `CONNECTED -> CLOSING -> CLOSED`

ablySpec/TestUtilities.swift
mattheworiordan: Agreed, this is completely the wrong behaviour.  We want the transport `on(closed)` event equivalent to be triggered by the test suite so that the client library thinks the transport was abruptly terminated.

ricardopereira:
Rebased with changes.

ricardopereira:
https://github.com/ably/ably-ios/blob/b755c200717e81188651c8bdb19bf71f31b9d7d6/ably-ios/ARTWebSocketTransport.m#L251 I think the `ARTWsAbnormalClose, ...` should also call the `delegate.realtimeTransportClosed` instead of `delegate.realtimeTransportDisconnected`.

mattheworiordan:
LGTM

tcard:
LGTM

ricardopereira:
It's important to merge #169 first, remove 435317c, rebase with master and check if it's ok.

Issue #143, opened 2016-01-20
Travis CI is stalling in some builds


tcard:
Travis builds sometimes fail with something like this:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

This happens most often after running the Swift suite and before starting the Obj-C suite. But I've also seen it happen in the middle of the Swift suite, with a message like "main thread has stalled".

This happens inconsistently.

ricardopereira:
I will try to join both tests into one target to have only one execution. Right know it executes the Swift target and then the Objective-C target and in some way it gets stuck. It should resolve this issue.

I will also do some changes on the project because I want to use the bridging header to import the library code to the Swift test target like Apple suggest: [Importing Objective-C into Swift](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/MixandMatch.html). CocoaPods in a Swift project and with the `use_frameworks!` setting use also a bridging header so I think it is a way to guarantee we use it in the same way on our `ablySpec`(Swift one).

ricardopereira:
Could be related: https://openradar.appspot.com/24222858.

ricardopereira:
I tried to join both test targets into one but  it isn't possible because there are too different. I will change the Travis file to use `scan --verbose` to see if we get anything from the log.

ricardopereira:
I found out my test suite was hanging when I was checking #171. I run the test suite with `scan` then only with `xcodebuild` and it hanged as well. I run it from Xcode and the test stopped because of an exception. So, I assume that there is something raising an exception on the CI as well. Something to look further.

ricardopereira:
There is a bug in `xcodebuild` which has been fixed in the latest release of XCode, 7.2.1 (which was released an hour ago). From the release notes

> Fixed in Xcode 7.2.1:
• Command line tool ‘xcodebuild test’ will no longer time out waiting for Simulator.app to launch

mattheworiordan:
What is the status on this?

tcard:
No news. There are periods in which this happens less often. I still don't know how we could debug this, we would need to make Travis builds more explicit somehow, haven't looked into it though.

mattheworiordan:
@tcard @ricardopereira should we not raise an issue with Travis themselves to get their help?  I can't believe that their system doesn't work given how ubiquitous it is in the open source community.  Can one of you do that?

ricardopereira:
[...]

PR #166, opened 2016-01-26
Fix compiler warnings.

tcard:
Mostly about using NSInteger in formatted strings, which the compiler dislikes.


mattheworiordan:
Looks good, although Travis build stalled?

tcard:
@mattheworiordan The CI build is broken right now (#143). It just stalls sometimes. We talked about this at the standup (I thought @ricardopereira said he would write it down). He has tried out this weekend Circe CI and something else and it still stalls there, so there must be something wrong with our build script or whatever. We agreed that fixing this is non-trivial and low priority right now, as we are only three people looking at PRs and we can just trust each other that the test suite passes in our machines for now. What do you think?

mattheworiordan:
> It just stalls sometimes.

I do recall that discussion, however I understood we had a potential solution.  I did not realise this issue was still present.

I agree it's not the highest priority, but it is an issue.  Do we have an issue for it?

ricardopereira:
Yes, we have. The issue is #143.

ricardopereira:
LGTM

PR #167, opened 2016-01-26
RTN14b

ricardopereira:
The test doesn't pass.

### DIFF
--- a/ablySpec/RealtimeClient.connection.swift
+++ b/ablySpec/RealtimeClient.connection.swift
@@ -1080,6 +1080,104 @@ class RealtimeClientConnection: QuickSpec {
                 }
             }
 
+            // RTN14b
+            context("connection request fails") {
+
+                it("should not emit error with a renewable token") {
+                    let options = AblyTests.commonAppSetup()
+                    options.autoConnect = false
+                    let tokenTtl = 1.0
+                    options.token = getTestToken(key: options.key, ttl: tokenTtl)
+
+                    let client = ARTRealtime(options: options)
+                    client.setTransportClass(TestProxyTransport.self)
+                    defer { client.close() }
+
+                    // Let the token expire
+                    waitUntil(timeout: testTimeout) { done in
+                        delay(tokenTtl) {
+                            done()
+                        }
+                    }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        client.eventEmitter.on { state, errorInfo in
+                            switch state {
+                            case .Connected:
+                                expect(errorInfo).to(beNil())
+                                // New token
+                                expect(client.auth().tokenDetails!.token).toNot(equal(options.token))
+                                done()
+                            case .Failed, .Disconnected, .Suspended:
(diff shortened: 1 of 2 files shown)


ricardopereira:
The test is not passing. The lib only gives the `40142 - token expired` and it doesn't retry to request a new token. Need fix.

ricardopereira:
Rebased with changes.

mattheworiordan:
I still don't understand why one of those tests passes and the other fails when both seem to not have any mechanism to renew the token.  The tests look wrong to me

ricardopereira:
Rebased with changes.

ablySpec/RealtimeClient.connection.swift
```diff
@@ -1080,6 +1080,106 @@ class RealtimeClientConnection: QuickSpec {
...
+
+                    waitUntil(timeout: testTimeout) { done in
+                        client.eventEmitter.on { state, errorInfo in
+                            switch state {
+                            case .Connected:
+                                expect(errorInfo).to(beNil())
+                                // New token
+                                expect(client.auth().tokenDetails!.token).toNot(equal(options.token))
+                                done()
+                            case .Failed:
```
mattheworiordan: Mind adding the Disconnected state in here too as it should not reach Disconnected, Suspended or Failed states?

ablySpec/RealtimeClient.connection.swift
```diff
@@ -1080,6 +1080,106 @@ class RealtimeClientConnection: QuickSpec {
...
+
+                    if failures.count != 1 {
+                        fail("Should have only one connection request fail")
+                        return
+                    }
+
+                    expect(failures[0].error!.code).to(equal(40142)) //Token expired
+                }
+
+                it("should fail again due a token request error") {
```
mattheworiordan: This should say "it should transition to the Failed state because the token is invalid and not renewable"

ablySpec/RealtimeClient.connection.swift
```diff
@@ -1080,6 +1080,106 @@ class RealtimeClientConnection: QuickSpec {
...
+                    client.setTransportClass(TestProxyTransport.self)
+                    defer { client.close() }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        client.eventEmitter.on { state, errorInfo in
+                            switch state {
+                            case .Connected:
+                                fail("Should not be connected")
+                                done()
+                            case .Failed:
```
mattheworiordan: Again please check that it never reaches the `Disconnected` or `Suspended` states either please

mattheworiordan:
Few small comments, please address those and from my side I am happy for you to merge then

ablySpec/RealtimeClient.connection.swift
ricardopereira: Done

ricardopereira:
All comments addressed.
@tcard PTAL

tcard:
LGTM

PR #170, opened 2016-01-28
Separate ARTProtocolMessage private parts.

tcard:
To adjust the public API to the spec.

### DIFF
--- a/ably-ios/ARTProtocolMessage+Private.h
+++ b/ably-ios/ARTProtocolMessage+Private.h
@@ -1,0 +1,23 @@
+//
+//  ARTProtocolMessage+Private.h
+//  ably-ios
+//
+//  Created by Toni Cárdenas on 26/01/2016.
+//  Copyright (c) 2014 Ably. All rights reserved.
+//
+
+ART_ASSUME_NONNULL_BEGIN
+
+@interface ARTProtocolMessage ()
+
+@property (art_nullable, readwrite, strong, nonatomic) NSString *clientId;
+@property (readwrite, assign, nonatomic) BOOL hasConnectionSerial;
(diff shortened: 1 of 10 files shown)


ricardopereira:
LGTM

ably-ios/ARTProtocolMessage+Private.h
```diff
@@ -0,0 +1,23 @@
...
+//
+//  Created by Toni Cárdenas on 26/01/2016.
+//  Copyright (c) 2014 Ably. All rights reserved.
+//
+
+ART_ASSUME_NONNULL_BEGIN
+
+@interface ARTProtocolMessage ()
+
+@property (art_nullable, readwrite, strong, nonatomic) NSString *clientId;
```
mattheworiordan: Don't think a ProtocolMessage needs to have a clientId, http://docs.ably.io/client-lib-development-guide/features/#types

mattheworiordan:
One comment, but otherwise good

ably-ios/ARTProtocolMessage+Private.h
tcard: It is implemented kind of backwards. I can see the point, which is to store the ConnectionDetails in the ProtocolMessage itself, and just have the former point to the latter, but I think the other way around (ie. having a ProtocolMessage.connectionKey getter that points to ConnectionDetails.connectionKey if exists) . The API isn't affected, though (this is a private header).

Done at 7bef28a81c3be38f4c3f7dfd84d49e57cf88eac8.

tcard:
PTAL

mattheworiordan:
:+1:

ricardopereira:
LGTM and tests are passing :+1:

PR #172, opened 2016-01-29
Mark "RestClient initializer should accept a token" as pending.

tcard:
It's not working. Already marked in the sheet.


ricardopereira:
One test [passed](https://travis-ci.org/ably/ably-ios/builds/105595233). Should be something with the staging environment?

tcard:
Merging because this is just standard procedure.

Issue #140, opened 2016-01-20
Issues noticed while reviewing tests marked as done


tcard:
I've kept a list of notes about different. Some of them are just questions, some of them are are things to be done. I've put a checkbox along with each one to keep track of what's been addressed.

- [RSC1](http://docs.ably.io/client-lib-development-guide/features/#RSC1)
	- [x] Says "key, token", but iOS only takes key, and it's that way in the docs, but Ruby takes both and discriminates with regexp. (Update: it should be done per the spec.)
- [RSC7a](http://docs.ably.io/client-lib-development-guide/features/#RSC7a)
	- [x] Why is it grey?
- [RSC10](http://docs.ably.io/client-lib-development-guide/features/#RSC10)
	- [x] "Using the Auth class a token should be automatically issued if possible, otherwise it should result in an error" The "otherwise" case isn't tested.
- [RSC14b](http://docs.ably.io/client-lib-development-guide/features/#RSC14b)
	- [x] Is it the same as [RSA4](http://docs.ably.io/client-lib-development-guide/features/#RSA4)? It is linked as such. If so, maybe it should be removed.
- [RSL1d](http://docs.ably.io/client-lib-development-guide/features/#RSL1d)
	- [x] Untested, but it says "yes"
- [RSL2b1](http://docs.ably.io/client-lib-development-guide/features/#RSL2b1)
	- [x] "start must be equal to or less than end and is unaffected by the request direction" That seems to be untested.	
- [RSL5b](http://docs.ably.io/client-lib-development-guide/features/#RSL5b)
	- [x] The test code doesn't seem to be encrypting anything.
- [RTC1c](http://docs.ably.io/client-lib-development-guide/features/#RTC1c)
	- [x] I'm not sure the test is testing this at all, but I don't know how it could be tested.
- [RTC1d](http://docs.ably.io/client-lib-development-guide/features/#RTC1d)
	- [x] There's just a comment in the test. realtimeHost is read-only... why?
- [RTC2](http://docs.ably.io/client-lib-development-guide/features/#RTC2)
	- [x] Says 'yes', but code says TODO.
- [RTN4e](http://docs.ably.io/client-lib-development-guide/features/#RTN4e)
	- [x] Says 'yes', but code says a TODO and has a comment that has nothing to do with this. (I've removed it.)
- [RTS3a](http://docs.ably.io/client-lib-development-guide/features/#RTS3a)
[...]

PR #178, opened 2016-02-01
RTS2

ricardopereira:


ricardopereira:
@tcard Maybe it is best to change the `ARTRealtime.channels` function to a property to have a nice Swift integration: `client.channels()["test"]` -> `client.channels["test"]`.

mattheworiordan:
FYI, the documented API is `client.channels.get('name')` not `client.channel('name')`.  The latter is a nice to have and supported in some libraries.

ricardopereira:
Ok, it is possible and easy to add an extension to NSDictionary with categories (`get`).

tcard:
I'd rather just stick to the API on this one. This is one of the issues listed in #140, I'll fix it soon.

tcard:
To be clear, I mean that I'd rather use a custom collection class than to add stuff to NSDictionary. Just for the sake of simplicity and keeping the API minimal. E. g. if we add a `get` method now we'd have `get` and `objectForKey` to do the same thing.

ricardopereira:
Rebased with no changes.

ricardopereira:
Can I merge this?

tcard:
Yep, sorry.

PR #179, opened 2016-02-01
Rewrite ARTEventEmitter per the spec and adjust Connection.

### LINKED ISSUE #133: Realtime: connection state changes is inconsistent with the API
**Problem:**
`client.eventEmitter.on` should be `client.on`.

[EventEmitter mixin / interface](http://docs.ably.io/client-lib-development-guide/features/#RTE3)

tcard:
Previously, ARTConnection managed state subscriptions and
ARTEventEmitter was just a shell. Now the heavy-lifting of
passing events around is done by ARTEventEmitter, which is
generic and can be reused by others, and ARTConnection just uses
it to emit events and expose its listening interface.

This also introduces ARTConnectionStateChange, used in connection
state events' callbacks, as the spec says.

Something that went away is that starting listening to events
dispatched the current event to the callback. Now it doesn't. It
starts listening to _changes_ from that point on. I think this
makes more sense and is consistent with what you expect from an
event emitter.

Fixes #133.

### DIFF
--- a/ably-ios/ARTEventEmitter+Private.h
+++ b/ably-ios/ARTEventEmitter+Private.h
@@ -1,0 +1,36 @@
+//
+//  ARTEventEmitter+Private.h
+//  ably
+//
+//  Created by Toni Cárdenas on 29/1/16.
+//  Copyright © 2016 Ably. All rights reserved.
+//
+
+#include "ARTEventEmitter.h"
+#include "CompatibilityMacros.h"
+
+ART_ASSUME_NONNULL_BEGIN
+
+@interface __GENERIC(ARTEventEmitterEntry, ItemType) : NSObject
+
+@property (readwrite, strong, nonatomic) __GENERIC(ARTEventListener, ItemType) *listener;
+@property (readwrite, nonatomic) BOOL once;
+
+- (instancetype)initWithListener:(__GENERIC(ARTEventListener, ItemType) *)listener once:(BOOL)once;
+// ARTEventEmitterEntry are compared only using their listener.
+- (BOOL)isEqual:(id)object;
+
+@end
+
+@interface __GENERIC(ARTEventEmitter, EventType, ItemType) ()
+
+// The reason we use arrays as sets is that blocks are not hashable.
+
+@property (readwrite, nonatomic) __GENERIC(NSMutableDictionary, EventType, __GENERIC(NSMutableArray, __GENERIC(ARTEventEmitterEntry, ItemType) *) *) *listeners;
+// totalListeners listen to all events, except those in exceptions.
+@property (readwrite, nonatomic) __GENERIC(NSMutableArray, __GENERIC(ARTEventEmitterEntry, ItemType) *) *totalListeners;
+@property (readwrite, nonatomic) __GENERIC(NSMutableDictionary, EventType, __GENERIC(NSMutableArray, __GENERIC(ARTEventListener, ItemType) *) *) *ignoring;
+
+@end
+
+ART_ASSUME_NONNULL_END
--- a/ably-ios/ARTEventEmitter+Private.h
+++ b/ably-ios/ARTEventEmitter+Private.h
@@ -17,19 +17,13 @@ ART_ASSUME_NONNULL_BEGIN
 @property (readwrite, nonatomic) BOOL once;
 
 - (instancetype)initWithListener:(__GENERIC(ARTEventListener, ItemType) *)listener once:(BOOL)once;
-// ARTEventEmitterEntry are compared only using their listener.
-- (BOOL)isEqual:(id)object;
 
 @end
 
 @interface __GENERIC(ARTEventEmitter, EventType, ItemType) ()
 
-// The reason we use arrays as sets is that blocks are not hashable.
-
 @property (readwrite, nonatomic) __GENERIC(NSMutableDictionary, EventType, __GENERIC(NSMutableArray, __GENERIC(ARTEventEmitterEntry, ItemType) *) *) *listeners;
-// totalListeners listen to all events, except those in exceptions.
 @property (readwrite, nonatomic) __GENERIC(NSMutableArray, __GENERIC(ARTEventEmitterEntry, ItemType) *) *totalListeners;
-@property (readwrite, nonatomic) __GENERIC(NSMutableDictionary, EventType, __GENERIC(NSMutableArray, __GENERIC(ARTEventListener, ItemType) *) *) *ignoring;
 
 @end
 
--- a/ably-ios/ARTEventEmitter+Private.h
+++ b/ably-ios/ARTEventEmitter+Private.h
@@ -23,7 +23,7 @@ ART_ASSUME_NONNULL_BEGIN
 @interface __GENERIC(ARTEventEmitter, EventType, ItemType) ()
 
 @property (readwrite, nonatomic) __GENERIC(NSMutableDictionary, EventType, __GENERIC(NSMutableArray, __GENERIC(ARTEventEmitterEntry, ItemType) *) *) *listeners;
-@property (readwrite, nonatomic) __GENERIC(NSMutableArray, __GENERIC(ARTEventEmitterEntry, ItemType) *) *totalListeners;
+@property (readwrite, nonatomic) __GENERIC(NSMutableArray, __GENERIC(ARTEventEmitterEntry, ItemType) *) *anyListeners;
 
 @end
 
--- a/ably-ios/ARTEventEmitter.h
+++ b/ably-ios/ARTEventEmitter.h
@@ -24,17 +24,17 @@ ART_ASSUME_NONNULL_BEGIN
 - (__GENERIC(ARTEventListener, ItemType) *)on:(EventType)event call:(void (^)(ItemType))cb;
 - (void)on:(EventType)event callListener:(__GENERIC(ARTEventListener, ItemType) *)listener;
 
-- (__GENERIC(ARTEventListener, ItemType) *)onAll:(void (^)(ItemType))cb;
-- (void)onAllCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener;
+- (__GENERIC(ARTEventListener, ItemType) *)on:(void (^)(ItemType))cb;
+- (void)onCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener;
 
 - (__GENERIC(ARTEventListener, ItemType) *)once:(EventType)event call:(void (^)(ItemType))cb;
 - (void)once:(EventType)event callListener:(__GENERIC(ARTEventListener, ItemType) *)listener;
 
-- (__GENERIC(ARTEventListener, ItemType) *)onceAll:(void (^)(ItemType))cb;
-- (void)onceAllCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener;
+- (__GENERIC(ARTEventListener, ItemType) *)once:(void (^)(ItemType))cb;
+- (void)onceCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener;
 
 - (void)off:(EventType)event listener:(__GENERIC(ARTEventListener, ItemType) *)listener;
-- (void)offAll:(__GENERIC(ARTEventListener, ItemType) *)listener;
+- (void)off:(__GENERIC(ARTEventListener, ItemType) *)listener;
 
 - (void)emit:(EventType)event with:(ItemType)data;
 
@@ -47,17 +47,17 @@ ART_ASSUME_NONNULL_BEGIN
 #define ART_EMBED_INTERFACE_EVENT_EMITTER(EventType, ItemType) - (__GENERIC(ARTEventListener, ItemType) *)on:(EventType)event call:(void (^)(ItemType))cb;\
 - (void)on:(EventType)event callListener:(__GENERIC(ARTEventListener, ItemType) *)listener;\
 \
-- (__GENERIC(ARTEventListener, ItemType) *)onAll:(void (^)(ItemType))cb;\
-- (void)onAllCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener;\
+- (__GENERIC(ARTEventListener, ItemType) *)on:(void (^)(ItemType))cb;\
+- (void)onCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener;\
 \
 - (__GENERIC(ARTEventListener, ItemType) *)once:(EventType)event call:(void (^)(ItemType))cb;\
 - (void)once:(EventType)event callListener:(__GENERIC(ARTEventListener, ItemType) *)listener;\
 \
-- (__GENERIC(ARTEventListener, ItemType) *)onceAll:(void (^)(ItemType))cb;\
-- (void)onceAllCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener;\
+- (__GENERIC(ARTEventListener, ItemType) *)once:(void (^)(ItemType))cb;\
+- (void)onceCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener;\
 \
 - (void)off:(EventType)event listener:(__GENERIC(ARTEventListener, ItemType) *)listener;\
-- (void)offAll:(__GENERIC(ARTEventListener, ItemType) *)listener;
+- (void)off:(__GENERIC(ARTEventListener, ItemType) *)listener;
 
 // This macro adds methods to a class implementation that just bridge calls to an internal
 // instance variable, which must be called _eventEmitter, of type ARTEventEmitter *.
@@ -71,12 +71,12 @@ return [_eventEmitter on:event call:cb];\
 [_eventEmitter on:event callListener:listener];\
 }\
 \
-- (__GENERIC(ARTEventListener, ItemType) *)onAll:(void (^)(ItemType))cb {\
-return [_eventEmitter onAll:cb];\
+- (__GENERIC(ARTEventListener, ItemType) *)on:(void (^)(ItemType))cb {\
+return [_eventEmitter on:cb];\
 }\
 \
-- (void)onAllCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener {\
-[_eventEmitter onAllCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener];\
+- (void)onCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener {\
+[_eventEmitter onCallListener:(__GENERIC(ARTEventListener, ItemType) *)listener];\
 }\
 \
 - (__GENERIC(ARTEventListener, ItemType) *)once:(EventType)event call:(void (^)(ItemType))cb {\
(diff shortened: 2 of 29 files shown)


tcard:
I didn't run the Objective-C test suite. Sorry. Working on that now.

tcard:
Passing:

captura de pantalla 2016-02-01 a les 19 51 04

ably-ios/ARTEventEmitter.h
```diff
@@ -9,14 +9,102 @@
...
+
+- (void)emit:(EventType)event with:(ItemType)data;
 
 @end
+
+// This macro adds methods to a class header file that mimic the API of an event emitter.
+// This way you can automatically "implement the EventEmitter pattern" for a class
+// as the spec say. It's supposed to be used together with ART_EMBED_IMPLEMENTATION_EVENT_EMITTER
+// in the implementation of the class.
+#define ART_EMBED_INTERFACE_EVENT_EMITTER(EventType, ItemType) - (__GENERIC(ARTEventListener, ItemType) *)on:(EventType)event call:(void (^)(ItemType))cb;\
```
ricardopereira: Why did you not use a protocol?

ricardopereira: Well, I'm thinking about it and maybe it is not possible using objc protocols with generics?

ably-iosTests/ARTRealtimeResumeTest.m
```diff
@@ -86,7 +86,9 @@ - (void)testSimple {
             }
         }];
 
-        [_realtime.eventEmitter on:^(ARTRealtimeConnectionState state, ARTErrorInfo *errorInfo) {
+        [_realtime onAll:^(ARTConnectionStateChange *stateChange) {
+            ARTRealtimeConnectionState state = stateChange.current;
+            ARTErrorInfo *errorInfo = stateChange.reason;
```
ricardopereira: I think `errorInfo` is not used.

ably-iosTests/ARTRealtimeResumeTest.m
```diff
@@ -154,7 +156,9 @@ -(void) testSimpleDisconnected {
             }
         }];
         
-        [_realtime.eventEmitter on:^(ARTRealtimeConnectionState state, ARTErrorInfo *errorInfo) {
+        [_realtime onAll:^(ARTConnectionStateChange *stateChange) {
+            ARTRealtimeConnectionState state = stateChange.current;
```
ricardopereira: Same here. `state ` and `errorInfo` aren't used.

mattheworiordan:
> starts listening to changes from that point on. I think this
makes more sense and is consistent with what you expect from an
event emitter.

@tcard yes, that is correct. FYI, we have an unspecified equivalent of how this used to work though, see https://github.com/ably/ably-js/blob/master/common/lib/util/eventemitter.js#L160-L176 and https://github.com/ably/ably-ruby/blob/master/lib/ably/modules/state_emitter.rb#L61-L103. Obviously this is out of spec so not needed

ably-ios/ARTEventEmitter.h
```diff
@@ -9,14 +9,102 @@
...
+
+@interface __GENERIC(ARTEventEmitter, EventType, ItemType) : NSObject
+
+- (__GENERIC(ARTEventListener, ItemType) *)on:(EventType)event call:(void (^)(ItemType))cb;
+- (void)on:(EventType)event callListener:(__GENERIC(ARTEventListener, ItemType) *)listener;
 
-- (instancetype)initWithRealtime:(ARTRealtime *)realtime;
-- (id<ARTSubscription>)on:(ARTRealtimeConnectionStateCb)cb;
-- (void)removeEvents;
+- (__GENERIC(ARTEventListener, ItemType) *)onAll:(void (^)(ItemType))cb;
```
mattheworiordan: Based on our previous discussions about method overloading, I thought we had preferred to use overloading in the end.  If so, why do we have ` onAll` as opposed to `on(void (^)(ItemType))cb`?

ably-ios/ARTEventEmitter.m
```diff
@@ -6,39 +6,266 @@
...
+                NSMutableArray *ign = [self.ignoring objectForKey:event];
+                if (ign && [ign member:entry.listener]) {
+                    // If you call 'onAll', then 'once(A)' and then 'off(A)',
+                    // 'onAll' will add to totalListeners, 'once(A)' will add
+                    // to listeners and 'off(A)' will add to ignoring. We've found
+                    // here one of those cases; just keep going. (If 'on' or 'once'
+                    // is called again, it will remove A from ignoring and replace this
+                    // entry, so it's fine to leave it here.)
+                    [toRemoveFromListenersForEvent addObjectReplacing:entry];
+                    continue;
```
mattheworiordan: I am not entirely clear why you need to keep track of ignored listeners. The way the Ruby lib works is simply:

`on(event, listener)` adds the listener to a list indexed by event
`on(listener)` adds the listener to a list without an event
`once(event, listener)` adds the listener to a list indexed by an event but with a flag to indicate it should be deleted after use

If a user used the same listener var and did this:
```
on(listener)
on(listener)
once(event, listener)
on(event, listener)
off(listener)
```

It would result in all listeners that match listener being removed.  I am not following why we need to keep track of ignored listeners.

ably-ios/ARTTypes.h
```diff
@@ -120,4 +118,16 @@ typedef void (^ARTTokenCallback)(ARTAuthTokenDetails *__art_nullable tokenDetail
...
 
+@interface ARTConnectionStateChange : NSObject
+
+- (instancetype)initWithCurrent:(ARTRealtimeConnectionState)current
+                       previous:(ARTRealtimeConnectionState)previous
+                         reason:(ARTErrorInfo *__art_nullable)reason;
+
+@property (readonly, nonatomic) ARTRealtimeConnectionState current;
+@property (readonly, nonatomic) ARTRealtimeConnectionState previous;
+@property (readonly, nonatomic, art_nullable) ARTErrorInfo *reason;
```
mattheworiordan: you're missing retryIn

ably-iosTests/ARTRealtimeChannelTest.m
```diff
@@ -39,13 +39,13 @@ - (void)tearDown {
...
         [_realtime removeAllChannels];
-        [_realtime.eventEmitter removeEvents];
+        [_realtime resetEventEmitter];
         [_realtime close];
     }
     _realtime = nil;
     if (_realtime2) {
         [_realtime2 removeAllChannels];
-        [_realtime2.eventEmitter removeEvents];
+        [_realtime2 resetEventEmitter];
```
mattheworiordan: The API for this is simply `EventEmitter.off()`, not sure why we have resetEventEmtter

ablySpec/Utilities.swift
```diff
@@ -0,0 +1,137 @@
...
+                    eventEmitter.emit("foo", with: 111)
+                    
+                    expect(receivedAll).to(equal(789))
+                    
+                    eventEmitter.emit("bar", with: 222)
+                    
+                    expect(receivedAll).to(equal(222))
+                }
+
+                it("should receive only once after calling onAll and then once") {
```
mattheworiordan: This test is wrong as far as I understand.  If you register a listener twice, then it will be called twice. There is no de-duping of events by listener.  Try this Node.js example yourself:

```javascript
const EventEmitter = require('events');
const util = require('util');

function MyEmitter() {
  EventEmitter.call(this);
}
util.inherits(MyEmitter, EventEmitter);

const myEmitter = new MyEmitter();
let listener = function() {
  console.log('an event occurred!');
};
myEmitter.on('event', listener);
myEmitter.on('event', listener);
myEmitter.once('event', listener);

console.log("> Emitting event first time");
myEmitter.emit('event');

console.log("> Emitting event second time");
myEmitter.emit('event');

/* 
 
> Emitting event first time
> myEmitter.emit('event');
an event occurred!
an event occurred!
an event occurred!

> Emitting event second time
> myEmitter.emit('event');
an event occurred!
an event occurred!

*/
```

mattheworiordan: Weirdly though, following on from the above example, I did not expect this:

```javascript


myEmitter.removeListener('event', listener);
console.log("> Emitting event third time after remove listener");
myEmitter.emit('event');

/* result from Node.js 
> myEmitter.removeListener('event', listener);
> Emitting event third time after remove listener
> myEmitter.emit('event');
an event occurred!

*/
```

I assumed that calling an `off` handler would match all events, whereas it seems only one is matched.  The Node.js spec says `removeListener will remove, at most, one instance of a listener from the listener array`.

@paddybyers this is worrying as we have not specified the intended behaviour on our 0.8 spec for the EventEmitter. In Ruby I certainly remove all, not sure about the other languages. What do you think we should do, what does Java do?

paddybyers: It's unusual to register the same function multiple times for the same event, so it's not too surprising that it's under-specified/under-documented/buggy.

Generally I would prefer that when you call `off()`, or `removeListener` you specify criteria - listener and/or event - and all registered handlers matching the given criteria are removed. This makes the most consistent spec when looking at all combinations of event or listener being unspecified.

I realise that this stops you from removing one of the two identical registrations but I think if you wanted to do that properly you'd have a different API - ie one where you get back a handle or id when you register, and then you use that id to de-register. And if you want that complexity, then do that yourself in your app.

mattheworiordan: > Generally I would prefer that when you call off(), or removeListener you specify criteria - listener and/or event - and all registered handlers matching the given criteria are removed.

@paddybyers Ok, I agree and what is implemented in Ruby. I will add a quick test to ably-js to make sure that functionality is followed there, and add an issue to ably-java to ensure it matches that spec.  I will also quickly update the 0.8 spec to reflect this.

ably-ios/ARTEventEmitter.h
tcard: Can protocols implement methods and have properties? I think they were just interfaces that implementors had to, well, implement themselves.

ably-ios/ARTEventEmitter.h
tcard: Overloading is problematic when there is a different number of arguments. Plus, I think we agreed on doing the idiomatic thing... See https://github.com/ably/ably-ios/pull/158#discussion_r50473673

ably-iosTests/ARTRealtimeChannelTest.m
tcard: `off` is for listeners to unsubscribe themselves, isn't it? This is a (private) method that removes all listeners from the emitter's side. It's just a helper for tests, to clean up and avoid calling callbacks when closing the channels once the test is done.

ably-ios/ARTEventEmitter.m
tcard: If I do this:

```
on(listenerA)
off(eventA, listenerA)
```

I want `listener` to receive all events _but_ `event`. That's why I understand from this:

> (RTE5) EventEmitter#off deregisters a listener for the specified event if given, else it is deregisters that listener for all registrations (global and event-specific)

ablySpec/Utilities.swift
tcard: Hmm, OK, that changes things. It's simpler to implement also. I prefer the behavior I implemented, as I think set-like on/off operations are more robust and more coherent with the API than list-like behavior (as you just noticed yourself), but well.

tcard: By the way, I'm not sure I understood in the end.

> Generally I would prefer that when you call off(), or removeListener you specify criteria - listener and/or event - and all registered handlers matching the given criteria are removed.

So I should keep the implemented behavior, ie. `off` removing all instances of the given criteria, which opposite to Node.js' behavior, right? But then I don't understand your next paragraph, "I realise that this stops you from removing one of the two identical registrations". Didn't you say `off` should remove "all registered handlers"?

ably-iosTests/ARTRealtimeResumeTest.m
tcard: Fixed at 0e5cf5f.

ably-iosTests/ARTRealtimeResumeTest.m
tcard: Fixed at 0e5cf5f.

ably-ios/ARTEventEmitter.h
ricardopereira: Yes, you can have properties. I guess you can have the protocol for the interface and have the macro for the implementation.

ablySpec/Utilities.swift
tcard: I've just seen https://github.com/ably/ably-js/pull/227, which I guess answers my question.

ably-ios/ARTTypes.h
tcard: Fixed at e354f9e. (It's only been defined, not used.)

ably-ios/ARTEventEmitter.h
mattheworiordan: I thought we agreed on something quite different, see https://github.com/ably/ably-ios/issues/121#issuecomment-169501841.  I thought #158 was an exception because it was an initializer / constructor?

tcard: Hmm, you're right, I didn't consider that Objective-C allows overloading with different number of arguments. My previous comment makes no sense. Will change.

ably-ios/ARTEventEmitter.h
tcard: In any case, a protocol suggests that classes can (should?) bring their own implementation for the protocol, when in fact that's not the case here. In this case, they should just provide access to the `ARTEventEmitter` implementation.

I'd prefer not to overcomplicate the thing with a protocol besides the class.

ably-ios/ARTEventEmitter.h
tcard: Fixed at 8aa38f6.

ably-ios/ARTEventEmitter.m
mattheworiordan: I see why that's nice, but I think it's too complicated from a developer's perspective to keep track of what their ignore list is.  Lets discuss today with @paddybyers though as we need to agree on a simple spec so that you don't waste time on this, sorry the spec was not clear enough in the first place.

ably-iosTests/ARTRealtimeChannelTest.m
mattheworiordan: Ok, sure, but I would have thought `off` would achieve the same thing as it should remove all listeners.

ably-ios/ARTEventEmitter.m
tcard: We agreed on call that my implementation was wrong; `off` will just be a no-op if called with a specific event and a total listener.

tcard:
Rebased, plus two changes: 5cba6ca and 43f6833.

@ricardopereira @mattheworiordan PTAL

ably-ios/ARTEventEmitter.h
```diff
@@ -9,14 +9,74 @@
...
+// as the spec say. It's supposed to be used together with ART_EMBED_IMPLEMENTATION_EVENT_EMITTER
+// in the implementation of the class.
+#define ART_EMBED_INTERFACE_EVENT_EMITTER(EventType, ItemType) - (__GENERIC(ARTEventListener, ItemType) *)on:(EventType)event call:(void (^)(ItemType __art_nullable))cb;\
+- (__GENERIC(ARTEventListener, ItemType) *)on:(void (^)(ItemType __art_nullable))cb;\
+\
+- (__GENERIC(ARTEventListener, ItemType) *)once:(EventType)event call:(void (^)(ItemType __art_nullable))cb;\
+- (__GENERIC(ARTEventListener, ItemType) *)once:(void (^)(ItemType __art_nullable))cb;\
+\
+- (void)off:(EventType)event listener:(__GENERIC(ARTEventListener, ItemType) *)listener;\
+- (void)off:(__GENERIC(ARTEventListener, ItemType) *)listener;
```
mattheworiordan: Is there not also meant to be another overloaded `off` without any params?

ably-ios/ARTEventEmitter+Private.h
```diff
@@ -0,0 +1,30 @@
...
+@property (readwrite, nonatomic) BOOL once;
+
+- (instancetype)initWithListener:(__GENERIC(ARTEventListener, ItemType) *)listener once:(BOOL)once;
+
+@end
+
+@interface __GENERIC(ARTEventEmitter, EventType, ItemType) ()
+
+@property (readwrite, nonatomic) __GENERIC(NSMutableDictionary, EventType, __GENERIC(NSMutableArray, __GENERIC(ARTEventEmitterEntry, ItemType) *) *) *listeners;
+@property (readwrite, nonatomic) __GENERIC(NSMutableArray, __GENERIC(ARTEventEmitterEntry, ItemType) *) *totalListeners;
```
mattheworiordan: Not mad about the name `totalListeners` as that makes it sound like it's a list compiled from all the listeners, whereas this this Array contains a list of listeners for any event.  Could it not be called something like `anyListeners` or `allEventListeners`?

mattheworiordan:
LGTM, only thing missing from what I can see is an `off` without any args to remove all listeners

ably-ios/ARTEventEmitter.h
tcard: > (RTE5) EventEmitter#off deregisters a listener for the specified event if given, else it is deregisters that listener for all registrations (global and event-specific)

I understand from that that there is at least a listener parameter.

ably-ios/ARTEventEmitter+Private.h
tcard: Fixed at 9be57ca.

ably-ios/ARTEventEmitter.h
tcard: Fixed at e9f6153997cd248376bd0db611c8e77fad733e08.

ably-ios/ARTEventEmitter.m
```diff
@@ -6,39 +6,201 @@
...
+        for (ARTEventEmitterEntry *entry in self.anyListeners) {
+            if (entry.once) {
+                [toRemoveFromTotalListeners addObject:entry];
+            }
+            [toCall addObject:entry];
+        }
+    }
+    @finally {
+        for (ARTEventEmitterEntry *entry in toRemoveFromListenersForEvent) {
+            [listenersForEvent removeObject:entry];
```
ricardopereira: It only removes the occurrence in that array (`listenersForEvent`).
I think it should be `self.listeners`.

mattheworiordan:
LGTM :+1:

ably-ios/ARTEventEmitter.m
tcard: Hey, good catch. Although I'm not sure what you mean. This should be calling `[removeObject:entry fromArrayWithKey:event inDictionary:self.listeners]`, because that handles the case in which `self.listeners[event]` is left empty and needs to be deleted to avoid entry leaks. I fixed that at 859e474.

ricardopereira: I didn't thought about the special case when I wrote about it but looks great. :+1:

Issue #121, opened 2016-01-06
Channel: consistent `publish` methods between other platform clients


ricardopereira:
**Current publish methods**
``` objc
- (void)publish:(art_nullable id)payload callback:(art_nullable ARTErrorCallback)callback;
- (void)publish:(art_nullable id)payload name:(art_nullable NSString *)name callback:(art_nullable ARTErrorCallback)callback;
- (void)publishMessage:(ARTMessage *)message callback:(art_nullable ARTErrorCallback)callback;
- (void)publishMessages:(__GENERIC(NSArray, ARTMessage *) *)messages callback:(art_nullable ARTErrorCallback)callback;
```

**Should be something like**
``` objc
- (void)publish:(art_nullable NSString *)name data:(art_nullable id)data callback:(art_nullable ARTPublishCallback)callback;
- (void)publish:(__GENERIC(NSArray, ARTMessage *) *)messages callback:(art_nullable ARTPublishCallback)callback;
```

I am assuming renaming `ARTErrorCallback` to `ARTPublishCallback`. `ARTErrorCallback` feels like it is a callback when something fails and it is not.

mattheworiordan:
Agreed, renamed to `ARTPublishCallback`
I also would prefer to be consistent and use the method `publish`.  However, whilst I did not suggest this approach, I thought that Objective-C recommended not overloading methods and used this approach, is this not the case? @tcard thoughts?

ricardopereira:
@mattheworiordan Well, it is a good practice but with Swift it might be better.

The result with the change:

**Objective-C**
``` objc
[channel publish:@"event" data:nil callback:nil];
[channel publish:@[message1, message2] callback:nil];
```

**Swift**
``` swift
channel.publish("event", data: nil, callback: nil)
channel.publish([message1, message2], callback: nil)
```

tcard:
I'm not sure what is our "policy" about preferring Ably conventions vs. language conventions. I think we favor the former, right? For example, I've seen that the Go library takes callbacks for async operations, which is really not Go-y. That favors uniformity on documentation I guess, so you can apply doc resources written for one language to one trivially, but it makes it uglier and maybe more difficult to actually use the libraries. So I think I prefer the latter, but what's really important is to be consistent, so if we're doing the former elsewhere, I'm OK with this change.

tcard:
OK, I've checked and I was wrong about Go taking callbacks. It's also wrong in the front page example, where I saw it:

[...]

PR #180, opened 2016-02-02
RTL1

ricardopereira:


ably-ios/ARTPresenceMap.h
```diff
@@ -7,22 +7,24 @@
...
+
 /// Used to maintain a list of members present on a channel
-@interface ARTPresenceMap : NSObject {
-    
-}
+@interface ARTPresenceMap : NSObject
+
+/// List of members.
+/// The key is the clientId and the value is the latest relevant ARTPresenceMessage for that clientId.
+@property (readonly, atomic, getter=getMembers) __GENERIC(NSDictionary, NSString *, ARTPresenceMessage *) *members;
```
mattheworiordan: There is no public members property in the API, only `get`, see http://docs.ably.io/client-lib-development-guide/features/#RSP3

ablySpec/RealtimeClientChannel.swift
```diff
@@ -13,6 +13,70 @@ class RealtimeClientChannel: QuickSpec {
...
+                expect(channel1.presenceMap.members).to(haveCount(1))
+                expect(channel2.presenceMap.members).to(haveCount(1))
+
+                // Check if receives incoming messages
+                channel2.subscribeToName("Client 1") { message, errorInfo in
+                    expect(message.payload.payload as? String).to(equal("message"))
+                }
+
+                waitUntil(timeout: testTimeout) { done in
+                    channel1.publish("message", cb: { status in
```
mattheworiordan: I realise this may be an API issue so happy for you to simply add an issue, but calling `publish` with just one data argument would set the message `name` to "message", not the payload as the signature is `publish(name, data)`

ablySpec/RealtimeClientChannel.swift
```diff
@@ -13,6 +13,70 @@ class RealtimeClientChannel: QuickSpec {
...
+
+                expect(channel2.state).toEventually(equal(ARTRealtimeChannelState.Attached), timeout: testTimeout)
+
+                expect(channel2.presence().isSyncComplete()).toEventually(beTrue(), timeout: testTimeout)
+
+                expect(channel1.presenceMap.members).to(haveCount(1))
+                expect(channel2.presenceMap.members).to(haveCount(1))
+
+                // Check if receives incoming messages
+                channel2.subscribeToName("Client 1") { message, errorInfo in
```
mattheworiordan: Another API issue, but why do we have errorInfo here. When you subscribe to a message, you only ever get messages, never errors.  Also, `subscribeToName` is simply `subscribe`, please raise an issue if not addressed in your changes @tcard

ablySpec/RealtimeClientChannel.swift
ricardopereira: Addressed here #121.

ablySpec/RealtimeClientChannel.swift
ricardopereira: Addressed here 183.

ably-ios/ARTPresenceMap.h
ricardopereira: But this is related with the PresenceMap, not with the Presence himself.

> (RTP2) A PresenceMap should be used to maintain a list of members present on a channel.

tcard: The presence API is pretty broken anyway already. I see that you're only adding generics, which I think is OK. I'll get the API right.

tcard:
LGTM

ably-ios/ARTPresenceMap.h
mattheworiordan: Ok, thanks @tcard, @ricardopereira ignore my comments tehne

mattheworiordan:
:+1:

PR #182, opened 2016-02-02
RTL3a

ricardopereira:
❌ Not passing.

### DIFF
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -81,7 +81,7 @@ class RealtimeClientChannel: QuickSpec {
             context("connection state") {
 
                 // RTL3a
-                context("changes to FAILED") {
+                pending("changes to FAILED") {
 
                     it("ATTACHING channel should transition to FAILED") {
                         let options = AblyTests.commonAppSetup()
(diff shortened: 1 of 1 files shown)


ablySpec/RealtimeClientChannel.swift
```diff
@@ -13,6 +13,63 @@ class RealtimeClientChannel: QuickSpec {
...
+
+                        let channel = client.channel("test")
+                        channel.attach()
+                        expect(channel.state).toEventually(equal(ARTRealtimeChannelState.Attached), timeout: testTimeout)
+
+                        waitUntil(timeout: testTimeout) { done in
+                            channel.subscribeToStateChanges { state, status in
+                                if state == .Failed {
+                                    expect(status.state).to(equal(ARTState.Error))
+                                    expect(status.errorInfo!.code).to(equal(90000))
```
mattheworiordan: Might be nice in both of these tests to simply double check that the channel state itself as failed as opposed to simply emitting the failed state.  I appreciate this is probably covered elsewhere, but the test explicitly states that the channel should transition to failed, and as such, it's nice to check that it both emits a failed state and has a failed state.

mattheworiordan:
LGTM other than one comment

ablySpec/RealtimeClientChannel.swift

tcard: > Done

Hmm, did you forget to push maybe? It's still the same.

tcard:
LGTM

ablySpec/RealtimeClientChannel.swift
ricardopereira: I edit the commit: https://github.com/ably/ably-ios/commit/8352fd528335672cc2a122805016e94e6518e380#diff-74dac661e35645777f94b2a135363313R68

PR #186, opened 2016-02-03
RTL4a

ricardopereira:


ablySpec/RealtimeClientChannel.swift
```diff
@@ -13,6 +13,32 @@ class RealtimeClientChannel: QuickSpec {
...
+
+                // RTL4a
+                it("if already ATTACHED or ATTACHING nothing is done") {
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                    defer { client.close() }
+
+                    var errorInfo: ARTErrorInfo?
+                    let channel = client.channel("test")
+
+                    errorInfo = channel.attach()
```
ricardopereira: Seems inconsistent with other clients. If so, I can open an issue.

mattheworiordan: Yes, this is inconsistent

ablySpec/RealtimeClientChannel.swift
```diff
@@ -13,6 +13,32 @@ class RealtimeClientChannel: QuickSpec {
...
+                    expect(errorInfo).to(beNil())
+                    expect(channel.state).to(equal(ARTRealtimeChannelState.Attaching))
+
+                    errorInfo = channel.attach()
+                    expect(errorInfo!.message).to(contain("Already attached"))
```
mattheworiordan: It should not indicate an error if calling attach, it should simply proceed. This is the wrong behaviour.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -13,6 +13,32 @@ class RealtimeClientChannel: QuickSpec {
...
+                    expect(channel.state).toEventually(equal(ARTRealtimeChannelState.Attached), timeout: testTimeout)
+
+                    errorInfo = channel.attach()
+                    expect(errorInfo!.message).to(contain("Already attached"))
```
mattheworiordan: As above, it should not be indicating an error at all

ablySpec/RealtimeClientChannel.swift
tcard: @ricardopereira There's no need to open issues for API inconsistencies, as I'm reviewing everything. Do it if you feel like it, but I'm on it anyway!

ablySpec/RealtimeClientChannel.swift
ricardopereira: I can fix the test after the API changes.

ricardopereira:
Rebased with no changes.

ablySpec/RealtimeClientChannel.swift
tcard: I think this goes beyond the API change. This should _not_ return an error, even with this API.

ricardopereira: The `attach` is inconsistent with the current API. @tcard Can you address this problem after the `attach` changes?

tcard: @ricardopereira Well, yes, I can, but this can be fixed right now, by changing this expectation to `expect(errorInfo).to(beNil())`. Then, the API change only would need to change, well, the API, to make it a callback.

ricardopereira:
Rebased with changes.

PR #192, opened 2016-02-03
RTL4f

ricardopereira:
❌ Not passing.

### DIFF
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -315,6 +315,25 @@ class RealtimeClientChannel: QuickSpec {
                     expect(channel.state).toEventually(equal(ARTRealtimeChannelState.Failed), timeout: testTimeout)
                 }
 
+                // RTL4f
+                it("should transition the channel state to FAILED if ATTACHED ProtocolMessage is not received") {
+                    let options = AblyTests.commonAppSetup()
+                    options.autoConnect = false
+                    let client = ARTRealtime(options: options)
+                    client.setTransportClass(TestProxyTransport.self)
+                    client.connect()
+                    defer { client.close() }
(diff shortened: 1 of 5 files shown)


tcard:
LGTM

ablySpec/RealtimeClientChannel.swift
```diff
@@ -181,6 +181,25 @@ class RealtimeClientChannel: QuickSpec {
...
+                    client.setTransportClass(TestProxyTransport.self)
+                    client.connect()
+                    defer { client.close() }
+
+                    expect(client.connection().state).toEventually(equal(ARTRealtimeConnectionState.Connected), timeout: testTimeout)
+                    let transport = client.transport as! TestProxyTransport
+                    transport.actionsIgnored += [.Attached]
+
+                    let channel = client.channel("test")
+                    channel.attach()
```
mattheworiordan: To make this test more deterministic, I would prefer that you instanced the library with a lower [`realtimeRequestTimeout`](http://docs.ably.io/client-lib-development-guide/features/#DF1b) such as 3s, and then following `attach`, you should expect the state to change to failed in exactly 3s.

ricardopereira: @mattheworiordan `realtimeRequestTimeout` is currently readonly and I can't hack that method because it is a static method. The user should have access to change that option or not?

mattheworiordan: See http://docs.ably.io/client-lib-development-guide/features/#DF1b, it's not a documented API, but we've made this configurable in our client libraries and left it undocumented for exactly this reason

ricardopereira:
Rebased with changes.

mattheworiordan:
LGTM, :+1:

PR #171, opened 2016-01-28
Rename payload -> data, refactor message data encoding.

tcard:
Well, this took a bit of time to get right. It started just
wanting to comply with the spec, and access message data just by
doing `message.data` instead of `message.payload.payload`. So
I wanted to remove just the amount of code necessary to do that,
but ended up opening a can of worms and didn't stop until it
kind of made sense.

It seems like the whole ARTPayload business was designed with a
use case which didn't really fit the spec at all, so it was then
hacked here and there to kind of work in some (not all) cases
by bypassing the design. It was also really overengineered, with
a whole class hierarchy thingy just for a few predefined cases.
I simplified the whole thing a lot and made it fit the spec.

I could've just set up a custom `message.data` getter method that
extracted the data from the ARTPayload instance to make it comply
with the spec, but well, this would've hit us down the road
anyway.

PR #193, opened 2016-02-04
RTL6a

ricardopereira:


ricardopereira:
I'm getting errors after the #171:

```
2016-02-04 18:50:15.161 xctest[85089:13705808] ERROR: ARTDataDecoded failed to decode data as 'json': (__NSCFDictionary){
    value = 1;
}
2016-02-04 18:50:15.162 xctest[85089:13705808] ERROR: ARTRealtimeChannel: error decoding data, status: 2038569904
```

@tcard Any thoughts?

ricardopereira:
I can debug it later.

tcard:
Seems like it's trying to decode a piece of JSON that is already decoded. The `data` of the incoming message should be a string, not a dictionary

ricardopereira:
The test is publishing a dictionary. The existing `RSL4d4` test is also failing.
@tcard Should I open an issue?

tcard:
@ricardopereira Please.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -186,6 +187,56 @@ class RealtimeClientChannel: QuickSpec {
...
+                    let data = ["value":1]
+                    var restEncodedMessage: ARTMessage?
+                    var realtimeEncodedMessage: ARTMessage?
+
+                    let rest = ARTRest(options: AblyTests.commonAppSetup())
+                    let restChannel = rest.channels.get("test")
+
+                    let restBlock: @convention(block) (AspectInfo, ARTMessage) -> Void = { info, message in
+                        let invocation = info.originalInvocation()
+                        invocation.invoke()
```
tcard: I'm not sure how this aspects thing works, but if `message` here is the original, unencoded message that is passed to `encodeMessageIfNeeded`, take into account that `ARTMessage.encode` returns a _copy_ of the ARTMessage instance, and thus this `message` won't be mutated by `invocation.invoke()`, and you're assigning the unencoded version to `restEncodedMessage`.

Ignore this if this `message` is what's returned by `encodeMessageIfNeeded`. (But then I don't understand what `invocation.invoke()` is invoking.)

ricardopereira: `invocation.invoke()` calls the original implementation, so the `encodeWithEncoder` will write to the current `ARTMessage` instance (look https://github.com/ably/ably-ios/blob/bff8f901b67e9ab12b38fa477470ed773ae0a902/ably-ios/ARTChannel.m#L54). Even so, I saw the documentation and I think it is possible to retrieve the original return value of the mehotd directly with the [NSInvocation.getReturnValue](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSInvocation_Class/#//apple_ref/occ/instm/NSInvocation/getReturnValue:). I think both approaches are correct. What you think?

tcard: That link shows how `encodeWithEncoder` takes an `output` argument, where it writes the _new_ message, which is then a different instance than the `message` in this block. For this to be correct you need to use the returned message.

ricardopereira: Yes, you right, It is doing a copy of the `message`. I will try to use the `NSInvocation.getReturnValue`. Ty.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -186,6 +187,56 @@ class RealtimeClientChannel: QuickSpec {
...
+                    let _ = try? realtimeHook("encodeMessageIfNeeded:", withOptions: .PositionAfter, usingBlock: unsafeBitCast(realtimeBlock, ARTRealtimeChannel.self))
+
+                    waitUntil(timeout: testTimeout) { done in
+                        realtimeChannel.publish(data, cb: { status in
+                            expect(status.state).to(equal(ARTState.Ok))
+                            done()
+                        })
+                    }
+
+                    expect(restEncodedMessage!.data as? NSObject).to(equal(realtimeEncodedMessage!.data as? NSObject))
```
mattheworiordan: Is it worth checking that the encoding value matches as well, and that encoding is not null i.e. the messages are indeed encoded.

ablySpec/RealtimeClientChannel.swift
ricardopereira: Fixed 357c5db.

ablySpec/RealtimeClientChannel.swift
ricardopereira: Done 357c5db.

mattheworiordan:
Ok, looks good to me

tcard:
LGTM

PR #194, opened 2016-02-04
RTL6i

ricardopereira:
[Review on Reviewable](https://reviewable.io/reviews/ably/ably-ios/194)

### DIFF
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -552,6 +552,52 @@ class RealtimeClientChannel: QuickSpec {
 
                 }
 
+                // RTL6i
+                context("expect either") {
+
+                    it("an array of Message objects") {
+                        let options = AblyTests.commonAppSetup()
+                        options.autoConnect = false
+                        let client = ARTRealtime(options: options)
+                        client.setTransportClass(TestProxyTransport.self)
+                        client.connect()
+                        defer { client.close() }
+                        let channel = client.channels.get("test")
+                        typealias JSONObject = NSDictionary
+
+                        var result = [JSONObject]()
+                        channel.subscribe { message in
+                            result.append(message.data as! JSONObject)
+                        }
+
+                        let messages = [ARTMessage(data: ["key":1], name: nil), ARTMessage(data: ["key":2], name: nil)]
+                        channel.publish(messages)
+
+                        let transport = client.transport as! TestProxyTransport
+
+                        expect(transport.protocolMessagesSent.filter{ $0.action == .Message }).toEventually(haveCount(1), timeout: testTimeout)
+                        expect(result).toEventually(equal(messages.map{ $0.data as! JSONObject }), timeout: testTimeout)
+                    }
+
+                    it("a name string and data payload") {
+                        let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                        defer { client.close() }
+                        let channel = client.channels.get("test")
+
+                        let expectedResult = "string_data"
+                        var result: String?
+
+                        channel.subscribe("event") { message in
+                            result = message.data as? String
+                        }
+
+                        channel.publish("event", data: expectedResult, cb: nil)
+
+                        expect(result).toEventually(equal(expectedResult), timeout: testTimeout)
+                    }
+
+                }
+
             }
 
             // RTL7
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -596,6 +596,105 @@ class RealtimeClientChannel: QuickSpec {
                         expect(result).toEventually(equal(expectedResult), timeout: testTimeout)
                     }
 
+                    it("allows name to be null") {
+                        let options = AblyTests.commonAppSetup()
+                        options.autoConnect = false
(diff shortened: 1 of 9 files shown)


tcard:
LGTM

ablySpec/RealtimeClientChannel.swift
```diff
@@ -290,6 +290,46 @@ class RealtimeClientChannel: QuickSpec {
...
+                        let messages = [ARTMessage(data: ["value":1], name: nil), ARTMessage(data: ["value":2], name: nil)]
+                        var result: [String]?
+
+                        channel.subscribe { message, errorInfo in
+                            result = message.data as? [String]
+                        }
+
+                        channel.publish(messages, callback: nil)
+
+                        expect(result).toEventually(equal(messages), timeout: testTimeout)
```
mattheworiordan: I am not quite sure this is doing what is required. You need to check that both messages are received (which you may be doing, but I'm not quite sure how `messages` will ever equal `result` given you are simply assigning each message's data to the `result` variable), but you also need to check that a single `ProtocolMessage` is being used to send the messages to Ably which you are not doing.

ricardopereira:
Rebased with changes.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -526,6 +526,52 @@ class RealtimeClientChannel: QuickSpec {
...
+                        let channel = client.channels.get("test")
+
+                        let expectedResult = "string_data"
+                        var result: String?
+
+                        channel.subscribeToName("event") { message, errorInfo in
+                            result = message.data as? String
+                        }
+
+                        channel.publish(expectedResult, name: "event", callback: nil)
```
mattheworiordan: Is the signature `publish(name, data)` not supported?

ricardopereira: I think #204 fixes that.

mattheworiordan: @tcard does it?

tcard: Yes: https://github.com/ably/ably-ios/pull/204/files#diff-04c6e90faac2675aa89e2176d2eec7d8R187

ricardopereira:
Rebased with changes.

tcard:
LGTM

ablySpec/RealtimeClientChannel.swift
```diff
@@ -552,6 +552,52 @@ class RealtimeClientChannel: QuickSpec {
...
+                        client.setTransportClass(TestProxyTransport.self)
+                        client.connect()
+                        defer { client.close() }
+                        let channel = client.channels.get("test")
+
+                        let messages = [ARTMessage(data: ["value":1], name: nil), ARTMessage(data: ["value":2], name: nil)]
+                        var result: [ARTMessage]?
+
+                        channel.subscribe { message in
+                            result = message.data as? [ARTMessage]
```
mattheworiordan: I think I've asked this before but don't think I got an answer, and unfortunately I just don't understand this. How can `expect(result).toEventually(equal(messages))` when result is being assigned with a `=` operator here i.e. it can only ever have one value?

tcard: I second that, I don't think this works, it should be an append of every `message`. Good catch.

ricardopereira: Yes, it is wrong. I missed it, sorry.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -552,6 +552,52 @@ class RealtimeClientChannel: QuickSpec {
...
+                        let client = ARTRealtime(options: options)
+                        client.setTransportClass(TestProxyTransport.self)
+                        client.connect()
+                        defer { client.close() }
+                        let channel = client.channels.get("test")
+                        typealias JSONObject = NSDictionary
+
+                        var result = [JSONObject]()
+                        channel.subscribe { message in
+                            result.append(message.data as! JSONObject)
```
ricardopereira: @mattheworiordan @tcard Done

tcard:
LGTM

tcard:
By the way, you're missing:

> (RTL6i3) Allows name and or data to be null. If any of the values are null, then key is not sent to Ably i.e. a payload with a null value for data would be sent as follows { "name": "click" }

It can be done in another pull request, just wanted to point it out since the title is "RTL6i".

ricardopereira:
> By the way, you're missing:

> > (RTL6i3) Allows name and or data to be null. If any of the values are null, then key is not sent to Ably i.e. a payload with a null value for data would be sent as follows { "name": "click" }

@tcard Done

mattheworiordan:
LGTM

tcard:
LGTM

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ably/ably-ios/194#-:-KApC-LFcqygGhAie3Az:676834233)*

tcard:
Reviewed 2 of 3 files at r1, 7 of 7 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ably/ably-ios/194)*

PR #198, opened 2016-02-08
Fix RTL3b

ricardopereira:


tcard:
LGTM

We should make sure that we rebase before merging. I've also merged #187 prematurely. (Fixed in #199.)

PR #199, opened 2016-02-08
Fix RTL4b: use client.channels.get.

tcard:


tcard:
I'm merging this and #198 as they're blocking others and I suspect they are uncontroversial.

PR #202, opened 2016-02-08
Minor changes to adjust to API spec.

tcard:
I think those aren't relevant enough to grant each its own pull request.

### DIFF
--- a/ably-ios/ARTPresenceMessage.m
+++ b/ably-ios/ARTPresenceMessage.m
@@ -34,4 +34,8 @@
     return description;
 }
 
+- (NSString *)memberKey {
+    return [NSString stringWithFormat:@"%@:%@", self.connectionId, self.clientId];
+}
+
 @end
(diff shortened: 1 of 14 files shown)


ably-ios/ARTPresenceMessage.m
```diff
@@ -34,4 +34,8 @@ - (NSString *)description {
     return description;
 }
 
+- (NSString *)memberKey {
+    return [NSString stringWithFormat:@"%@:%@", self.connectionId, self.clientId];
```
ricardopereira: Should it return an empty string if the `connectionId` or `clientId` are nil?
It will return `(null):(null)` if both properties are nil.

tcard: In `PresenceMessage` both `connectionId` and `clientId` are defined as non-nullable, so that's something we don't have to worry about.

mattheworiordan:
LGTM

ably-ios/ARTPresenceMessage.m
ricardopereira: Yes, sorry. I thought both were nullable.

ricardopereira:
LGTM

PR #203, opened 2016-02-09
TG4: PaginatedResult.next calls back with nil if there's no next.

tcard:
This behavior is consistent with [the spec](http://docs.ably.io/client-lib-development-guide/features/#TG4).


ricardopereira:
LGTM

mattheworiordan:
Looks good, but I'd like to see a test to prevent a regression.  @ricardopereira could you add a test?

ricardopereira:
Yes, sure.

PR #204, opened 2016-02-09
General adjustment to API spec.

tcard:
This pull request is a rather big chain of fixes to adjust to the API spec.

Because most commits rely on some changes from previous commits in the chain, it's not possible to separate them in independent pull requests, so rather than have a series of pull requests each waiting on the previous one to be merged I'm just pushing them all together. It's probably easier to review each commit than the whole thing at once.

Sorry I didn't do this more gradually, but I often edited old commits as I discovered things to fix in an old commit's concern.


README.md
```diff
@@ -8,114 +8,480 @@ An iOS client library for [ably.io](https://www.ably.io), the realtime messaging
 
 Visit https://www.ably.io/documentation for a complete API reference and more examples.
 
-## CocoaPod Installation
-add pod ably to your Podfile. 
+## Installation
 
-## Manual Installation 
+You can install Ably for iOS either thorugh CocoaPods or manually.
```
ricardopereira: Typo: _thorugh_

mattheworiordan:
@tcard wow, this is a MEGA PR :)
I am quite sad to see how far from the API the client library was, but am impressed at how much you've fixed in this PR. A big :+1:

README.md
```diff
@@ -8,114 +8,480 @@ An iOS client library for [ably.io](https://www.ably.io), the realtime messaging
...
+        // members is the array of members present
+    }
+}
 ```
 
-### Querying the History
+**Objective-C**
+
+```objective-c
+[ARTTestUtil testRealtime:options callback:^(ARTRealtime *client) {
```
ricardopereira: Should this example use `ARTTestUtil`?

ably-ios/ARTClientOptions.h
```diff
@@ -14,24 +14,32 @@ ART_ASSUME_NONNULL_BEGIN
 
 @interface ARTClientOptions : ARTAuthOptions
 
-@property (readonly, getter=getRestHost) NSString *restHost;
-@property (readonly, getter=getRealtimeHost) NSString *realtimeHost;
+@property (readwrite, strong, nonatomic, getter=getRestHost) NSString *restHost;
```
ricardopereira: I understand the reason why you changed the property to readwrite but I noticed the property has no setter.

tcard: Does it need one? I assumed this synthesizes a `_restHost` variable with a setter.

README.md
tcard: Fixed at 493e0bf.

README.md
tcard: Oops, no. I copypasted more than appropriate from the test. Fixed at 493e0bf.

ably-ios/ARTClientOptions.h
ricardopereira: No, scratch that. I'm drunk.

"_5-line code review vs 5000-line code review_" ;)
she_didn_t_know_what_to_do_

ably-ios/ARTRealtimeChannel.m
```diff
@@ -424,23 +404,22 @@ - (ARTErrorInfo *)attach {
     [self.realtime send:attachMessage cb:nil];
```
ricardopereira: I think you need to pass the `cb`, right?

ably-ios/ARTRealtimeChannel.m
```diff
@@ -450,7 +429,10 @@ - (ARTErrorInfo *)detach {
     [self.realtime send:detachMessage cb:nil];
```
ricardopereira: Same as above.

ably-ios/ARTRealtimePresence.m
```diff
@@ -48,33 +72,45 @@ - (void)enterClient:(NSString *)clientId data:(id)data cb:(ARTStatusCallback)cb
...
-- (void)updateClient:(NSString *)clientId data:(id)data cb:(ARTStatusCallback)cb {
+- (void)updateClient:(NSString *)clientId data:(id)data {
+    [self updateClient:clientId data:data cb:nil];
+}
+
+- (void)updateClient:(NSString *)clientId data:(id)data cb:(void (^)(ARTErrorInfo * _Nullable))cb {
     ARTPresenceMessage *msg = [[ARTPresenceMessage alloc] init];
     msg.action = ARTPresenceUpdate;
     msg.clientId = clientId;
-    if(!msg.clientId) {
```
ricardopereira: Why did you removed this?

ably-ios/ARTRealtimePresence.m
```diff
@@ -85,11 +121,7 @@ - (void) leaveClient:(NSString *) clientId data:(id) data cb:(ARTStatusCallback)
     msg.action = ARTPresenceLeave;
     msg.data = data;
     msg.clientId = clientId;
-    msg.connectionId = [self channel].realtime.connectionId;
-    if(!msg.clientId) {
```
ricardopereira: Same as above.

ricardopereira: Ok, forget that. The `publishPresence` do the check, right?

ably-ios/ARTRestPresence.m
```diff
@@ -47,7 +47,7 @@ - (void)get:(void (^)(__GENERIC(ARTPaginatedResult, ARTPresenceMessage *) *resul
     [ARTPaginatedResult executePaginated:[self channel].rest withRequest:request andResponseProcessor:responseProcessor callback:callback];
 }
 
-- (BOOL)history:(ARTDataQuery *)query callback:(void (^)(__GENERIC(ARTPaginatedResult, ARTPresenceMessage *) *result, NSError *error))callback error:(NSError **)errorPtr {
+- (BOOL)history:(ARTDataQuery *)query error:(NSError **)errorPtr callback:(void (^)(__GENERIC(ARTPaginatedResult, ARTPresenceMessage *) *result, NSError *error))callback {
```
ricardopereira: Does this become a method that throw when imported by Swift?

ricardopereira: Yes, it's fine.

> If the **last non-block parameter** of an Objective-C method is of type NSError **, Swift replaces it with the throws keyword, to indicate that the method can throw an error. [Using Swift with Cocoa and Objective-C](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html#//apple_ref/doc/uid/TP40014216-CH7-ID6)

ably-ios/ARTTokenParams.m
```diff
@@ -134,19 +143,18 @@ - (NSDictionary *)toDictionaryWithUnion:(NSArray *)items {
...
 
-- (ARTTokenRequest *)sign:(NSString *)key withNonce:(NSString *)randomNonce {
+- (ARTTokenRequest *)sign:(NSString *)key withNonce:(NSString *)nonce {
     NSArray *keyComponents = decomposeKey(key);
     NSString *keyName = keyComponents[0];
     NSString *keySecret = keyComponents[1];
-    NSString *nonce = randomNonce;
     NSString *clientId = self.clientId ? self.clientId : @"";
-
+
```
ricardopereira: Sorry. _Whitespace_

ably.xcworkspace/contents.xcworkspacedata
```diff
@@ -2,6 +2,9 @@
 <Workspace
    version = "1.0">
    <FileRef
+      location = "group:ablySpec/ReadmeExamples.swift">
```
ricardopereira: I always _gitignore_ this file. Never notice it was here.
Can you remove it and add it to the `.gitignore`?
This file is generated by the CocoaPods.

ricardopereira: BTW, use this https://github.com/github/gitignore/blob/master/Objective-C.gitignore

ricardopereira:
👏🏻👏🏻👏🏻 Great work.
I contributed in some of the API inconsistency and I'm sorry.
Luckily, I'm learning a lot with you guys.

ably.xcworkspace/contents.xcworkspacedata
tcard: This has not much to do with this pull request. Can it be done in a separate one?

ricardopereira: Yes, sure.

ably-ios/ARTRealtimePresence.m
tcard: `updateClient` takes a non-nullable clientId as argument. The compiler will catch this. In any case, we can't be adding null checks everywhere.

ably-ios/ARTRealtimeChannel.m
tcard: Right, good catch. Fixed at fb040caec2965cae8310a9ba33c97a11a401b508.

ably-ios/ARTRealtimeChannel.m
tcard: Fixed at fb040caec2965cae8310a9ba33c97a11a401b508.

ably-ios/ARTRealtimePresence.m
tcard: Same as above: `leaveClient` takes a non-nullable clientId as argument. The compiler will catch this. In any case, we can't be adding null checks everywhere.

ably-ios/ARTRealtimePresence.m
ricardopereira: I was just wondering because a dedicated state `ARTStateNoClientId` seems to be a special case but I agree with you :+1:

tcard: Actually, you were right here. This method is not only public, it's also called from `update`, which passes the channels' `clientId` which _can_ be nil. Joke's on me for fiddling with internals without need... Fixed at 500d0f874e86759c7c6fa6c0f37821497b8b6e0d.

tcard:
OK, I finally got all tests passing in my machine. Let's see what the CI says, but it should be mergeable now after your LGTM. @ricardopereira @mattheworiordan

tcard:
There are some tests that pass when ran individually but fail when ran in bulk. Although this is introducing this bug, I think we can merge this since it is blocking all other work and we can deal with that afterwards.

mattheworiordan:
> There are some tests that pass when ran individually but fail when ran in bulk. Although this is introducing this bug, I think we can merge this since it is blocking all other work and we can deal with that afterward

Agreed, failures like this are test suite issues, not necessarily issues with the library.  Perhaps merge and add an issue

ricardopereira:
@tcard I removed this block of code: https://github.com/ably/ably-ios/blob/5d85a840cd31d1c4d687b7bd848a86628695d18b/ablySpec/RealtimeClientConnection.swift#L469-L473. Tests are passing now. Maybe it is something wrong whit `dispose`.

Why didn't you use Quick? https://github.com/ably/ably-ios/blob/5d85a840cd31d1c4d687b7bd848a86628695d18b/ablySpec/ReadmeExamples.swift#L15
If you want to run only one test, it will always run all the examples first and then the test that you selected.

This is not good for debugging: https://github.com/ably/ably-ios/blob/5d85a840cd31d1c4d687b7bd848a86628695d18b/ably-ios/ARTConnection.m#L79
. E.g.: I can't step into `- (__GENERIC(ARTEventListener, ItemType) *)on:(void (^)(ItemType __art_nullable))cb`.

tcard:
@ricardopereira Nice! Thanks.

> Why didn't you use Quick? https://github.com/ably/ably-ios/blob/5d85a840cd31d1c4d687b7bd848a86628695d18b/ablySpec/ReadmeExamples.swift#L15
> If you want to run only one test, it will always run all the examples first and then the test that you selected.

Well, that's a problem about how badly Quick works; we should be able to run tests individually. I didn't need all the spec boilerplate (describe/context/it) for these tests, just to run some snippets. There's no use case to describe here.

> This is not good for debugging: https://github.com/ably/ably-ios/blob/5d85a840cd31d1c4d687b7bd848a86628695d18b/ably-ios/ARTConnection.m#L79
> . E.g.: I can't step into - (__GENERIC(ARTEventListener, ItemType) *)on:(void (^)(ItemType __art_nullable))cb.

Hmm, why not? You should be able to. I've tried and you go directly to `ARTEventEmitter on`, you can't look into `ARTConnection on` but that's just a wrapper that calls `ARTEventEmitter on`.

tcard:
Rebased without changes.

tcard:
Sorry, scratch that, there were some tests added to master that now don't compile. Fixing.

ricardopereira:
> Well, that's a problem about how badly Quick works; we should be able to run tests individually. I didn't need all the spec boilerplate (describe/context/it) for these tests, just to run some snippets. There's no use case to describe here.

@tcard You right. Quick generates each test dynamically. That's why it sucks when you try to execute one test with the Xcode shortcut.

> Hmm, why not? You should be able to. I've tried and you go directly to ARTEventEmitter on, you can't look into ARTConnection on but that's just a wrapper that calls ARTEventEmitter on.

Ok, if you say so. I didn't pay to much attention to it but it felt like it wasn't stepping into the ARTEventEmitter.

> Sorry, scratch that, there were some tests added to master that now don't compile. Fixing.

Can you try clean all the warnings too?

tcard:
> Can you try clean all the warnings too?

Done that as well. PTAL

ricardopereira:
Great, thank you.
LGTM

PR #205, opened 2016-02-10
RTL6g

ricardopereira:
❌ Not passing.

Can't publish a message of type `ARTMessage`: `channel.publish(message, ...)`
```
2016-02-10 15:31:37.259 xctest[4649:209394] ERROR: ARTChannel: error encoding data, status: 2029696992
/Users/ricardopereira/Repositories/Whitesmith/ably-ios/ablySpec/RealtimeClientChannel.swift:560: error: -[ablySpec.RealtimeClientChannel Channel__publish__Identified_clients_with_clientId__When_publishing_a_Message_with_clientId_set_to_null__should_be_unnecessary_to_set_clientId_of_the_Message_before_publishing_UsersricardopereiraRepositoriesWhitesmithablyiosablySpecRealtimeClientChannelswift_536] : failed: caught "NSInvalidArgumentException", "message encoding failed"
(
	0   CoreFoundation                      0x00b5ea94 __exceptionPreprocess + 180
	1   libobjc.A.dylib                     0x0061fe02 objc_exception_throw + 50
	2   ably                                0x0a5a4817 -[ARTChannel encodeMessageIfNeeded:] + 487
	3   ably                                0x0a5a613b __41-[ARTRealtimeChannel publishMessages:cb:]_block_invoke + 91
	4   ably                                0x0a5ea357 -[NSArray(ARTFunctional) artMap:] + 455
	5   ably                                0x0a5a6012 -[ARTRealtimeChannel publishMessages:cb:] + 370
	6   ably                                0x0a5a5e2c -[ARTRealtimeChannel publish:withName:cb:] + 268
	7   ably                                0x0a5a5cde -[ARTRealtimeChannel publish:cb:] + 318
```

It is possible in JavaScript:
```
channel.publish({ data: 'boom!', name: 'event' });
```

### DIFF
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -557,7 +557,7 @@ class RealtimeClientChannel: QuickSpec {
                             resultClientId = message.clientId
                         }
 
-                        let message = ARTMessage(data: "message", name: nil)
+                        let message = ARTMessage(name: nil, data: "message")
                         message.clientId = "client_string"
 
                         channel.publish([message]) { errorInfo in
@@ -587,7 +587,7 @@ class RealtimeClientChannel: QuickSpec {
                             result.append(message.data as! JSONObject)
                         }
 
-                        let messages = [ARTMessage(data: ["key":1], name: nil), ARTMessage(data: ["key":2], name: nil)]
+                        let messages = [ARTMessage(name: nil, data: ["key":1]), ARTMessage(name: nil, data: ["key":2])]
                         channel.publish(messages)
 
                         let transport = client.transport as! TestProxyTransport
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -742,6 +742,32 @@ class RealtimeClientChannel: QuickSpec {
                             expect(messageSent.messages![0].clientId).to(beNil())
                         }
 
+                        // RTL6g1b
+                        it("should have clientId value as null for the Message when received") {
+                            let options = AblyTests.commonAppSetup()
+                            options.autoConnect = false
+                            let client = ARTRealtime(options: options)
+                            client.setTransportClass(TestProxyTransport.self)
+                            client.connect()
+                            defer { client.close() }
+
+                            let channel = client.channels.get("test")
+
+                            waitUntil(timeout: testTimeout) { done in
+                                channel.subscribe { message in
+                                    expect(message.clientId).to(beNil())
+                                    done()
+                                }
+                                channel.publish(nil, data: "message")
+                            }
+
+                            let transport = client.transport as! TestProxyTransport
+                            expect(transport.protocolMessagesReceived.filter { $0.action == .Message }).toEventually(haveCount(1), timeout: testTimeout)
+
+                            let messageReceived = transport.protocolMessagesReceived.filter({ $0.action == .Message })[0]
+                            expect(messageReceived.messages![0].clientId).to(beNil())
+                        }
+
                     }
 
                 }
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -720,49 +720,32 @@ class RealtimeClientChannel: QuickSpec {
                     // RTL6g1
                     context("When publishing a Message with clientId set to null") {
 
-                        // RTL6g1a
-                        pending("should be unnecessary to set clientId of the Message before publishing") {
+                        // RTL6g1a & RTL6g1b
+                        pending("should be unnecessary to set clientId of the Message before publishing and have clientId value as null for the Message when received") {
                             let options = AblyTests.commonAppSetup()
+                            options.clientId = "client_string"
                             options.autoConnect = false
                             let client = ARTRealtime(options: options)
                             client.setTransportClass(TestProxyTransport.self)
                             client.connect()
                             defer { client.close() }
-
                             let channel = client.channels.get("test")
 
-                            channel.publish(nil, data: "message") { errorInfo in
-                                expect(errorInfo).to(beNil())
-                            }
-
-                            let transport = client.transport as! TestProxyTransport
-                            expect(transport.protocolMessagesSent.filter { $0.action == .Message }).toEventually(haveCount(1), timeout: testTimeout)
-
-                            let messageSent = transport.protocolMessagesSent.filter({ $0.action == .Message })[0]
-                            expect(messageSent.messages![0].clientId).to(beNil())
-                        }
-
-                        // RTL6g1b
-                        pending("should have clientId value as null for the Message when received") {
-                            let options = AblyTests.commonAppSetup()
-                            options.autoConnect = false
-                            let client = ARTRealtime(options: options)
-                            client.setTransportClass(TestProxyTransport.self)
-                            client.connect()
-                            defer { client.close() }
-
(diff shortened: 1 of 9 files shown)


ablySpec/RealtimeClientChannel.swift
```diff
@@ -526,6 +526,78 @@ class RealtimeClientChannel: QuickSpec {
...
+                context("Identified clients with clientId") {
+
+                    // RTL6g1
+                    context("When publishing a Message with clientId set to null") {
+
+                        // RTL6g1a
+                        pending("should be unnecessary to set clientId of the Message before publishing") {
+                            let options = AblyTests.commonAppSetup()
+                            options.autoConnect = false
+                            options.clientId = "client_string"
```
mattheworiordan: This test says "When publishing a Message with clientId set to null" yet you are setting the clientId?

ablySpec/RealtimeClientChannel.swift
```diff
@@ -526,6 +526,78 @@ class RealtimeClientChannel: QuickSpec {
...
+                            expect(messageSent.messages![0].clientId).to(beNil())
+
+                            expect(resultClientId).toEventually(equal(options.clientId), timeout: testTimeout)
+                        }
+
+                        // RTL6g1b
+                        pending("should have clientId value as null for the Message when received") {
+                            let options = AblyTests.commonAppSetup()
+                            options.autoConnect = false
+                            options.clientId = "client_string"
```
mattheworiordan: As above, this test fails because it shoudl...

ablySpec/RealtimeClientChannel.swift
ricardopereira: @mattheworiordan Ok, I see. A little confusing because the `RTL6g` says _Identified clients with a clientId _... I thought the client needs  a `clientId` but the message is published without a `clientId` (_clientId attribute set to null_) like I did:

```
let message = ARTMessage(data: "message", name: nil)
expect(message.clientId).to(beNil())
```

I will fix the test.

mattheworiordan: OK

ablySpec/RealtimeClientChannel.swift
```diff
@@ -526,6 +526,60 @@ class RealtimeClientChannel: QuickSpec {
...
+
+                            channel.publish("message", cb: { status in
+                                expect(status.state).to(equal(ARTState.Ok))
+                            })
+
+                            let transport = client.transport as! TestProxyTransport
+                            expect(transport.protocolMessagesSent.filter { $0.action == .Message }).toEventually(haveCount(1), timeout: testTimeout)
+
+                            let messageSent = transport.protocolMessagesSent.filter({ $0.action == .Message })[0]
+                            expect(messageSent.messages![0].clientId).to(beNil())
```
ricardopereira: Failing with: `failed - expected to be nil, got <>`

ablySpec/RealtimeClientChannel.swift
```diff
@@ -526,6 +526,60 @@ class RealtimeClientChannel: QuickSpec {
...
+
+                            channel.publish("message", cb: { status in
+                                expect(status.state).to(equal(ARTState.Ok))
+                            })
+
+                            let transport = client.transport as! TestProxyTransport
+                            expect(transport.protocolMessagesReceived.filter { $0.action == .Message }).toEventually(haveCount(1), timeout: testTimeout)
+
+                            let messageReceived = transport.protocolMessagesReceived.filter({ $0.action == .Message })[0]
+                            expect(messageReceived.messages![0].clientId).to(beNil())
```
ricardopereira: As above.

ricardopereira:
Rebased with changes.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -552,6 +552,60 @@ class RealtimeClientChannel: QuickSpec {
...
+                context("Identified clients with clientId") {
+
+                    // RTL6g1
+                    context("When publishing a Message with clientId set to null") {
+
+                        // RTL6g1a
+                        pending("should be unnecessary to set clientId of the Message before publishing") {
+                            let options = AblyTests.commonAppSetup()
+                            options.autoConnect = false
+                            let client = ARTRealtime(options: options)
```
tcard: The spec says "Identified clients with a clientId (as a result of either an explicitly configured clientId in ClientOptions, or implicitly through Token Auth)", but this client doesn't have a clientId, does it?

ablySpec/RealtimeClientChannel.swift
mattheworiordan: I would also add a normal `subscribe` check to ensure that a message received using the normal documented API will deliver a message without a clientId to the listener

ablySpec/RealtimeClientChannel.swift
ricardopereira: > The spec says "Identified clients with a clientId (as a result of either an explicitly configured clientId in ClientOptions, or implicitly through Token Auth)", but this client doesn't have a clientId, does it?

@tcard I removed the `clientId` ^

ablySpec/RealtimeClientChannel.swift

ablySpec/RealtimeClientChannel.swift

mattheworiordan: This is still wrong as far as I can see. This is supposed to be testing publishing a message without an explicit `clientId`, yet the client must be identified i.e. with a `clientId`.  This client is not identified

tcard: @mattheworiordan @ricardopereira I've seen the discussion that @ricardopereira linked, and isn't that just what was done before? The client was being identified through ClientOptions.clientId.

mattheworiordan: I'm not following you unfortunately. This test needs an identified client to publish a message without specifying an explicit `clientId` for the message.  Not much more to it than that.

ablySpec/RealtimeClientChannel.swift
mattheworiordan: This looks good, however RTL6g1b and RTL6g1a could really be tested in one test block instead as this is pretty straightforward and both appear to largely be doing the same thing

ablySpec/RealtimeClientChannel.swift
tcard: Yes, but unless I'm reading the code very wrong, isn't that what @ricardopereira did initially and you asked him to change?

From the code snippet commented [here](8 days ago):

```swift
+                        pending("should be unnecessary to set clientId of the Message before publishing") {
+                            let options = AblyTests.commonAppSetup()
+                            options.autoConnect = false
+                            options.clientId = "client_string"
```

That's setting the clientId in the connection options, not in the message.

ricardopereira: Yes, I had that before. The client with a valid `clientId` and the message published with no `clientId`.

mattheworiordan: If that's the case then I apologise, I misread it the first time. Bit confused why no one pulled me up on my mistake though!  

I've tried to review the rest of the code beneath the `options.clientId` at the time of my comment and Github is not allowing me to see that, so it's possible the `message` clientId was being set?

Anyway, setting a clientId is valid in the options, and setting one in the message is not valid, in the context of this test.

ricardopereira:
Done

tcard:
LGTM

ricardopereira:
@mattheworiordan PTAL

PR #208, opened 2016-02-10
RTL7c

ricardopereira:
[Review on Reviewable](https://reviewable.io/reviews/ably/ably-ios/208)

### DIFF
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -630,6 +630,8 @@ class RealtimeClientChannel: QuickSpec {
 
                     waitUntil(timeout: testTimeout) { done in
                         channel.subscribe { message in
+                            // FIXME: error handling
+                            //https://github.com/ably/ably-ios/pull/208#discussion_r53043622
                             done()
                         }
                     }
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -607,6 +607,34 @@ class RealtimeClientChannel: QuickSpec {
                     expect(Test.counter).toEventually(equal(2), timeout: testTimeout)
                 }
 
+                // RTL7c
+                it("should implicitly attach the channel") {
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                    defer { client.close() }
+
+                    let channel = client.channels.get("test")
+
+                    channel.subscribe { _ in }
+
+                    expect(channel.state).toEventually(equal(ARTRealtimeChannelState.Attached), timeout: testTimeout)
+                }
+
+                // RTL7c
+                it("should result in an error if channel is in the FAILED state") {
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                    defer { client.close() }
+
+                    let channel = client.channels.get("test")
+                    channel.onError(AblyTests.newErrorProtocolMessage())
+                    expect(channel.state).to(equal(ARTRealtimeChannelState.Failed))
+
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.subscribe { message in
+                            done()
+                        }
+                    }
+                }
+
                 // RTL7f
                 it("should exist ensuring published messages are not echoed back to the subscriber when echoMessages is false") {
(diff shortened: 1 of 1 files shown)


tcard:
LGTM

ablySpec/RealtimeClientChannel.swift
```diff
@@ -550,6 +550,40 @@ class RealtimeClientChannel: QuickSpec {
...
+                pending("should result in an error if channel is in the FAILED state") {
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                    defer { client.close() }
+
+                    let channel = client.channels.get("test")
+                    channel.onError(AblyTests.newErrorProtocolMessage())
+                    expect(channel.state).to(equal(ARTRealtimeChannelState.Failed))
+
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.subscribe { message, errorInfo in
```
mattheworiordan: Ok, but there is no `errorInfo` argument in the callback.  We need to discuss how best to do this in iOS, @tcard any thoughts?

tcard: After #204, the "However, if the channel is in the FAILED state, the subscribe request result in an error" part is not done. The error story is not so consistent yet in this library, sometimes we throw exceptions for things like this, sometimes we move NSErrors around...

I think (and it's not an original thought) that exceptions should be used for programmer errors (e. g. bad arguments passed to a function, state violations, etc.), because those just shouldn't happen ever and if they do the code should be fixed, while "external world" errors (like a channel being FAILED) are best handled with normal "error values", like those `ErrorInfo`s, or returning nil.

In this case, I think returning a nil listener would be best; it's easy to check if you are indeed interested in whether the subscription succeeded, and it doesn't have any effect otherwise. An exception would be truly unfortunate because you would have to either always handle it or always check the channel state before subscribing, because the channel could always be in a FAILED state.

mattheworiordan: Your interpretation of whether to raise an exception or pass an error in a listener is 100% correct.  Paddy, Simon and I [debated this for some time](https://github.com/ably/ably-js/commit/1b151df5558339d8b2fd80bbadd2375d9a138b4a), and if the developer at the time of writing their code is doing the right thing, then we should not be raising an exception simply because the state of the library may have changed outside of their control.

However, returning a `nil` listener to me is also wrong.  Subscribing to a channel that is failed is not necessarily wrong, it just means the listener will not be called until the channel is attached again.  So I don't think we should do anything if someone registers a listener for a failed channel because later, the channel could feasibly become attached if explicitly reattached.

I wonder if perhaps http://docs.ably.io/client-lib-development-guide/features/#RTL7c is simply wrong and whilst we implicitly attach, if we cannot we simply don't, or we emit an error on the channel at best.

@paddybyers @SimonWoolf WDYT?

ricardopereira:
Rebased with changes.

tcard:
LGTM

ablySpec/RealtimeClientChannel.swift
```diff
@@ -607,6 +607,34 @@ class RealtimeClientChannel: QuickSpec {
...
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                    defer { client.close() }
+
+                    let channel = client.channels.get("test")
+                    channel.onError(AblyTests.newErrorProtocolMessage())
+                    expect(channel.state).to(equal(ARTRealtimeChannelState.Failed))
+
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.subscribe { message in
+                            done()
```
mattheworiordan: How is this resuling in an error?

tcard: I've LGTM'd this because `subscribe` currently doesn't support testing this (#220). It is marked as pending. It's true that maybe an exception here would make sure that we fix this.

mattheworiordan: Well I would at least prefer a comment to state that this is not complete because the callback is not yet supported else one could easily merge this in mistakenly by simply switching the test from `pending` to `it`

ricardopereira: Done d640299

mattheworiordan:
LGTM

PR #211, opened 2016-02-12
RTL7e

ricardopereira:
✅

Picked: 4b6829d.

### DIFF
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -1150,7 +1150,7 @@ class RealtimeClientChannel: QuickSpec {
                 context("message cannot be decoded or decrypted") {
 
                     // RTL7e
-                    it("should deliver with encoding attribute set indicating the residual encoding and error should be emitted") {
+                    pending("should deliver with encoding attribute set indicating the residual encoding and error should be emitted") {
                         let options = AblyTests.commonAppSetup()
                         options.autoConnect = false
                         let client = ARTRealtime(options: options)
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -1147,6 +1147,55 @@ class RealtimeClientChannel: QuickSpec {
 
                 }
 
+                context("message cannot be decoded or decrypted") {
+
+                    // RTL7e
+                    it("should deliver with encoding attribute set indicating the residual encoding and error should be emitted") {
+                        let options = AblyTests.commonAppSetup()
+                        options.autoConnect = false
+                        let client = ARTRealtime(options: options)
+                        client.setTransportClass(TestProxyTransport.self)
+                        client.connect()
+                        defer { client.close() }
+
+                        let channelOptions = ARTChannelOptions(cipher: ["key":ARTCrypto.generateRandomKey()])
+                        let channel = client.channels.get("test", options: channelOptions)
+
+                        let expectedMessage = ["key":1]
+                        let expectedData = try! NSJSONSerialization.dataWithJSONObject(expectedMessage, options: NSJSONWritingOptions(rawValue: 0))
+
+                        let transport = client.transport as! TestProxyTransport
+
+                        transport.beforeProcessingReceivedMessage = { protocolMessage in
+                            if protocolMessage.action == .Message {
+                                let messageReceived = protocolMessage.messages![0]
+                                // Replacement: `json/cipher+aes-128-cbc/base64` to `invalid/cipher+aes-128-cbc/base64`
+                                let newEncoding = "invalid" + messageReceived.encoding!.substringFromIndex("json".endIndex)
+                                messageReceived.encoding = newEncoding
+                            }
+                        }
+
+                        waitUntil(timeout: testTimeout) { done in
+                            let logTime = NSDate()
+
+                            channel.subscribe { message in
+                                // Last decoding failed: NSData -> JSON object, so...
+                                expect(message.data as? NSData).to(equal(expectedData))
+                                expect(message.encoding).to(equal("invalid"))
+
+                                let logs = querySyslog(forLogsAfter: logTime)
+                                let line = logs.reduce("") { $0 + "; " + $1 } //Reduce in one line
+                                expect(line).to(contain("ERROR: ARTDataDecoded failed to decode data as 'invalid'"))
+
+                                done()
+                            }
+                            
+                            channel.publish(nil, data: expectedMessage)
+                        }
+                    }
+                    
+                }
+
                 // RTL7f
                 it("should exist ensuring published messages are not echoed back to the subscriber when echoMessages is false") {
                     let options = AblyTests.commonAppSetup()
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -1185,7 +1185,9 @@ class RealtimeClientChannel: QuickSpec {
 
                                 let logs = querySyslog(forLogsAfter: logTime)
                                 let line = logs.reduce("") { $0 + "; " + $1 } //Reduce in one line
-                                expect(line).to(contain("ERROR: ARTDataDecoded failed to decode data as 'invalid'"))
+                                expect(line).to(contain("ERROR: Failed to decode data as 'invalid' encoding is unknown"))
+
+                                expect(channel.errorReason!.message).to(contain("Failed to decode data as 'invalid' encoding is unknown"))
 
                                 done()
                             }
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -1183,11 +1183,11 @@ class RealtimeClientChannel: QuickSpec {
                                 expect(message.data as? NSData).to(equal(expectedData))
                                 expect(message.encoding).to(equal("invalid"))
 
-                                let logs = querySyslog(forLogsAfter: logTime)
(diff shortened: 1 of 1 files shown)


ricardopereira:
Rebased with changes.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -607,6 +607,55 @@ class RealtimeClientChannel: QuickSpec {
...
+                            channel.subscribe { message in
+                                // Last decoding failed: NSData -> JSON object, so...
+                                expect(message.data as? NSData).to(equal(expectedData))
+                                expect(message.encoding).to(equal("invalid"))
+
+                                let logs = querySyslog(forLogsAfter: logTime)
+                                let line = logs.reduce("") { $0 + "; " + $1 } //Reduce in one line
+                                expect(line).to(contain("ERROR: ARTDataDecoded failed to decode data as 'invalid'"))
+
+                                done()
```
tcard: Same as https://github.com/ably/ably-ios/pull/209#discussion_r53037608:

> Not tested: "an ErrorInfo error object is emitted as an error on the Channel". (I guess this means that the error is set in RealtimeChannel.errorReason.)

Although now I think "an error should be emitted on the channel" means that it should be emitted to the `channel.on` listeners. @mattheworiordan is this right? Should it also be set in `RealtimeChannel.errorReason`?

ablySpec/RealtimeClientChannel.swift
```diff
@@ -607,6 +607,55 @@ class RealtimeClientChannel: QuickSpec {
...
+                    // RTL7e
+                    it("should deliver with encoding attribute set indicating the residual encoding and error should be emitted") {
+                        let options = AblyTests.commonAppSetup()
+                        options.autoConnect = false
+                        let client = ARTRealtime(options: options)
+                        client.setTransportClass(TestProxyTransport.self)
+                        client.connect()
+                        defer { client.close() }
+
+                        let channelOptions = ARTChannelOptions(encrypted: ARTCrypto.defaultParams())
```
mattheworiordan: This an incorrect definition for channel options for two reasons as far as I can see:

* A `key` value must always be provided as part of the `cipherParams`
* `encrypted` is a boolean, and `chiperParams`: should be the `cipherParams`

See http://docs.ably.io/client-lib-development-guide/features/#TB2 and second example at http://docs.ably.io/realtime/channels-messages/#channels

ablySpec/RealtimeClientChannel.swift
```diff
@@ -607,6 +607,55 @@ class RealtimeClientChannel: QuickSpec {
...
+                            let logTime = NSDate()
+
+                            channel.subscribe { message in
+                                // Last decoding failed: NSData -> JSON object, so...
+                                expect(message.data as? NSData).to(equal(expectedData))
+                                expect(message.encoding).to(equal("invalid"))
+
+                                let logs = querySyslog(forLogsAfter: logTime)
+                                let line = logs.reduce("") { $0 + "; " + $1 } //Reduce in one line
+                                expect(line).to(contain("ERROR: ARTDataDecoded failed to decode data as 'invalid'"))
```
mattheworiordan: Might be nice to fix this error message whilst you are at it as it does not make sense.  How about `ERROR: ARTMessage Failed to decode data as 'invalid' encoding is unknown`

ablySpec/RealtimeClientChannel.swift
mattheworiordan: Yes, ideally we want to have a generic `on('error')` handler, although this is not fully supported in all libs unfortunately. Yes, `Channel.errorReason` should be set as well.

ablySpec/RealtimeClientChannel.swift
ricardopereira: I think #217 will change this log error.

ablySpec/RealtimeClientChannel.swift
ricardopereira: @tcard Did you changed this when you enhance the API?

tcard: No, I didn't. This is about constructors, which aren't specified in the API. We have `[ARTChannelOptions unencrypted]`, which makes an unencrypted `ChannelOptions`, and `[[ARTChannelOptions alloc] initEncrypted:params]`, which makes . The properties, `encrypted` and `cipherParams`, are as in the spec.

We probably should just change it to a single `initWithEncrypted:(bool) cipherParams:(ARTCipherParams *)` constructor. Can you make that change here, @ricardopereira? I will otherwise.

@mattheworiordan `ARTCrypto.defaultParams()` generates a random key. It's pretty stupid, yes, but it's for internal use, as in tests. `ARTCrypto` is not supposed to be part of the external API.

ricardopereira: I think JS is doing the same: https://github.com/ably/ably-js/blob/be64c9aa91d8b4b07ab20b8888f0543045ce0f71/spec/realtime/crypto.test.js#L185.

ablySpec/RealtimeClientChannel.swift
tcard: I've merged #217, so we'll know if you rebase this.

ricardopereira: @tcard Rebased and now it's failing with an uncaught exception: `Unexpected exception raised: ARTRealtimeChannel: error decoding data: Error Domain=io.ably.cocoa Code=0 "decoding failed" UserInfo={NSLocalizedFailureReason=ARTErrorInfo with code 0, message: unknown encoding: 'bad_encoding_type', NSLocalizedDescription=decoding failed}`

https://github.com/ably/ably-ios/blob/66b38d811a09846c46c2c2cdffea3f7e3bb4e449/ably-ios/ARTRealtimeChannel.m#L329

But the spec says "_If there is an error decoding a message, the message is still delivered_" [RTL7d](http://docs.ably.io/client-lib-development-guide/features/#RTL7d)

ablySpec/RealtimeClientChannel.swift
mattheworiordan: Nope, JS is not generating a random key.

@tcard Crypto is meant to be part of the public API, see http://docs.ably.io/client-lib-development-guide/features/#TZ3.  

The pseudo code we'd need for this to be consistent would be:
```
cipherParams = ARTCrypto.defaultParams()
cipherParams.key = randomString()
channelOptions = ARTChannelOptions(encrypted: true, cipherParams: cipherParams)
```

So this needs to be fixed as far as I can tell i.e. ARTCrypto, if public, should not generate a random key ever

ablySpec/RealtimeClientChannel.swift
mattheworiordan: Did you fix the error message?

tcard: Mark as pending then, will fix. (Also the error message.)

tcard: (So, please change the expected error message also.)

ablySpec/RealtimeClientChannel.swift
tcard: @mattheworiordan Sorry, you're right, Crypto is public. But TZ3b says that the key parameter is optional, and TZ3a says that if the key isn't passed it is generated randomly. Thus the only thing that should be changed here is the ARTChannelOptions constructor. (I'll submit a PR with that.)

By the way, and this is kind of off-topic here, but does it make sense how we're using the IV? The specified API doesn't even let you pass a IV to `getCipherParams`, it just generates one randomly. But AFAICT you won't be able to decrypt those messages unless you have the same IV on the receiving end. Besides, we're CBC-chaining _all_ the messages in a channel, so the receiving end will need to get _all_ the messages published in that channel, and if it misses any of them because it subscribes later or there's a suspension or something.

Not saying that it's wrong, and I don't immediately see a better way to do it, just want to confirm that I'm getting it right.

ricardopereira:
Done

ablySpec/RealtimeClientChannel.swift
tcard: See #231.

mattheworiordan: > and TZ3a says that if the key isn't passed it is generated randomly.

Shit, you're right. I now recall why `TZ3a` does that as the spec was largely based on what was implemented in Java, although I don't really like the fact that it does that. We should leave in the spec, but it's a weird default behaviour if you ask me.

In regards to IV, it feels a bit like an extension of the `key` issue we have with `defaultParams()`, why are we auto-generating an `IV` randomly that both sides will need to synchronise on.  The `IV` should either be shared between clients before communicating, or both sides should have a means to generate an `IV` and keep it in sync.  The latter option is clearly more complicated, but equally far more robust in the long term.

I have not documented Crypto yet, and was intentionally leaving this to later as I didn't think it was high priority, but I expect I will struggle a bit if we're not all entirely clear on this.

@paddybyers is this worth a quick chat to discuss?

tcard:
This should be rebased on top of #231 once it's merged.

mattheworiordan:
I am assuming this is somewhat blocked by https://github.com/ably/docs/pull/89#discussion_r53929144 so we can review once that is finalised and merged in?

ricardopereira:
Correct.

tcard:
@ricardopereira #231 is merged, you can rebase this now.

ricardopereira:
Rebased with changes: e1cc273.

mattheworiordan:
LGTM

tcard:
LGTM

PR #219, opened 2016-02-16
RTL8

ricardopereira:
No description

### DIFF
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -828,6 +828,33 @@ class RealtimeClientChannel: QuickSpec {
 
             }
 
+            // RTL8
+            context("unsubscribe") {
+
+                // RTL8a
+                it("with no arguments unsubscribes the provided listener to all messages if subscribed") {
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                    defer { client.close() }
+
+                    let channel = client.channels.get("test")
+
+                    waitUntil(timeout: testTimeout) { done in
+                        let listener = channel.subscribe { message in
+                            fail("Listener shouldn't exist")
+                            done()
+                        }
+
+                        channel.unsubscribe(listener)
+
+                        channel.publish(nil, data: "message") { errorInfo in
+                            expect(errorInfo).to(beNil())
+                            done()
+                        }
+                    }
+                }
+
+            }
+
             // RTL10
             context("history") {
 
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -853,6 +853,28 @@ class RealtimeClientChannel: QuickSpec {
                     }
                 }
 
+                // RTL8b
+                it("with a single name argument unsubscribes the provided listener if previously subscribed with a name-specific subscription") {
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                    defer { client.close() }
+
+                    let channel = client.channels.get("test")
+
+                    waitUntil(timeout: testTimeout) { done in
+                        let eventAListener = channel.subscribe("eventA") { message in
+                            fail("Listener shouldn't exist")
+                            done()
+                        }
+
+                        channel.unsubscribe("eventA", listener: eventAListener)
+
+                        channel.publish("eventA", data: "message") { errorInfo in
+                            expect(errorInfo).to(beNil())
+                            done()
+                        }
+                    }
+                }
+
             }
 
             // RTL10


ablySpec/RealtimeClientChannel.swift
```diff
@@ -609,6 +609,88 @@ class RealtimeClientChannel: QuickSpec {
...
+
+                    channel2.subscribe { message in
+                        expect(message.data as? String).to(equal("message"))
+                        Test.counter += 1
+                    }
+
+                    channel1.publish(nil, data: "message")
+                    channel1.publish(nil, data: "message")
+
+                    expect(Test.counter).toEventually(equal(3), timeout: testTimeout)
```
mattheworiordan: I really don't like the readability of the test.  Why not simply set up one channel, subscribe, publish and check that message is received. Then unsubscribe and check that no message is received?  This is overly complex and I don't really follow why one channel unsubscribes whilst the other one doesn't.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -609,6 +609,88 @@ class RealtimeClientChannel: QuickSpec {
...
+                        channel1.unsubscribe("eventA", listener: eventAListener)
+                    }
+
+                    channel1.publish("eventA", data: "message")
+                    channel1.publish("eventB", data: "message")
+                    channel2.publish("eventA", data: "message")
+                    channel2.publish("eventB", data: "message")
+
+                    expect(Counter.eventA).toEventually(equal(3), timeout: testTimeout)
+                    expect(Counter.eventB).toEventually(equal(2), timeout: testTimeout)
```
mattheworiordan: As above, why complicate this with two channels.  Why not simply do what I described above with one channel, publish eventA and eventB, receive both via subscribers, unsubscribe eventA subscriber, publish both, check that only eventB is received.

ablySpec/RealtimeClientChannel.swift
ricardopereira: Yes, you right. I will simplify.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -609,6 +609,62 @@ class RealtimeClientChannel: QuickSpec {
...
+                        channel.publish(nil, data: "message")
+                    }
+
+                    channel.unsubscribe()
+
+                    channel.publish(nil, data: "message")
+
+                    // If `unsubscribe` fails then the test suite will raise "Done closure's was called multiple times."
+                    waitUntil(timeout: testTimeout) { done in
+                        delay(5.0) { done() }
```
tcard: Could you please make this delay much smaller and move it to the callback of the second `channel.publish`? The test suite is already way too slow as it is...

ablySpec/RealtimeClientChannel.swift
```diff
@@ -609,6 +609,62 @@ class RealtimeClientChannel: QuickSpec {
...
+                        channel.publish("eventA", data: "message")
+                    }
+
+                    channel.unsubscribe(eventAListener!)
+
+                    channel.publish("eventA", data: "message")
+
+                    // If `unsubscribe` fails then the test suite will raise "Done closure's was called multiple times."
+                    waitUntil(timeout: testTimeout) { done in
+                        delay(5.0) { done() }
```
tcard: Same as above.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -609,6 +609,62 @@ class RealtimeClientChannel: QuickSpec {
...
+
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.subscribe { message in
+                            expect(message.data as? String).to(equal("message"))
+                            done()
+                        }
+                        channel.publish(nil, data: "message")
+                    }
+
+                    channel.unsubscribe()
```
tcard: This is bad wording of the spec, but "with no arguments" means "with no arguments apart from the listener", because it also talks about "the provided listener". This is unsubscribing _all_ listeners.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -609,6 +609,62 @@ class RealtimeClientChannel: QuickSpec {
...
+                    var eventAListener: ARTEventListener?
+                    waitUntil(timeout: testTimeout) { done in
+                        eventAListener = channel.subscribe("eventA") { message in
+                            expect(message.data as? String).to(equal("message"))
+                            done()
+                        }
+                        channel.publish("eventA", data: "message")
+                    }
+
+                    channel.unsubscribe(eventAListener!)
```
tcard: Missing the name argument that the spec talks about.

ablySpec/RealtimeClientChannel.swift
ricardopereira: :+1:

ablySpec/RealtimeClientChannel.swift
ricardopereira: :+1:

tcard:
LGTM

ablySpec/RealtimeClientChannel.swift
```diff
@@ -609,6 +609,65 @@ class RealtimeClientChannel: QuickSpec {
...
+                        }
+                        channel.publish(nil, data: "message")
+                    }
+
+                    channel.unsubscribe(listener!)
+
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.publish(nil, data: "message") { errorInfo in
+                            expect(errorInfo).to(beNil())
+                            // If `unsubscribe` fails then the test suite will raise "Done closure's was called multiple times."
```
mattheworiordan: This test is unnecessarily complex because of this.  Why are we testing publish on line 628 at all. Surely we just add a listener, and the unsubscribe the listener.  We have other tests to test that the listener registration works, so if we don't test something that this test is not about, you can instead raise an error in the listener block because we never expect it to be called.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -609,6 +609,65 @@ class RealtimeClientChannel: QuickSpec {
...
+                        }
+                        channel.publish("eventA", data: "message")
+                    }
+
+                    channel.unsubscribe("eventA", listener: eventAListener!)
+
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.publish("eventA", data: "message") { errorInfo in
+                            expect(errorInfo).to(beNil())
+                            // If `unsubscribe` fails then the test suite will raise "Done closure's was called multiple times."
```
mattheworiordan: As above, but also, can we please register two listeners, one for eventA and one for eventB. Then remove only eventA, and ensure that eventB still fires.  This is sensible to ensure that a unsubscribe for one listener has no side effects on other listeners.

PR #229, opened 2016-02-17
RTL10a

ricardopereira:
[Review on Reviewable](https://reviewable.io/reviews/ably/ably-ios/229)


ricardopereira:
Did 31db95f because of this strange conversion:
screen shot 2016-02-16 at 15 27 57

 - Changed `historyWithError` with just `history`.

tcard:
I think a more definitive test would be to use Aspects to check that calling Realtime's history calls Rest's history. It's hard to completely test "Supports all the same params" otherwise; you would have to check what happens if you remove some, if they do the same, etc.

ricardopereira:
> I think a more definitive test would be to use Aspects to check that calling Realtime's history calls Rest's history. It's hard to completely test "Supports all the same params" otherwise; you would have to check what happens if you remove some, if they do the same, etc.

@tcard Done

tcard:
LGTM

mattheworiordan:
:+1:

PR #230, opened 2016-02-17
Fix RTN10b.

tcard:
The test was checking the connection serial before the connection
was established. The code behave correctly, but before CONNECTED
serial was 0 just because it's the default int value, then set to
-1 at CONNECTED. I've changed serial's initial value to -1, to
make it not important whether you check it before of after the
connection is established.

[Review on Reviewable](https://reviewable.io/reviews/ably/ably-ios/230)


ricardopereira:
LGTM

mattheworiordan:
:+1:

PR #231, opened 2016-02-19
Adjust Crypto, CipherParams and ChannelOptions to spec.

tcard:
https://github.com/ably/ably-ios/pull/211#discussion_r53042652

I've kept the IV param in Crypto.getDefaultParams for the reasons mentioned in the discussion.

### DIFF
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -10,6 +10,8 @@ import Quick
 import Nimble
 import Aspects
 
+import Ably.Private
+
 class RealtimeClientChannel: QuickSpec {
     override func spec() {
         describe("Channel") {
@@ -1085,8 +1087,8 @@ class RealtimeClientChannel: QuickSpec {
                             let (keyData, ivData, messages) = AblyTests.loadCryptoTestData(cryptoTest)
                             let testMessage = messages[0]
 
-                            let cipherParams = ARTCrypto.defaultParamsWithKey(keyData, iv: ivData)
-                            let channelOptions = ARTChannelOptions(encrypted: cipherParams)
+                            let cipherParams = ARTCipherParams(algorithm: "aes", key: keyData, keyLength: UInt(keyData.length), iv: ivData)
+                            let channelOptions = ARTChannelOptions(encrypted: true, cipherParams: cipherParams)
                             let channel = client.channels.get("test", options: channelOptions)
 
                             let transport = client.transport as! TestProxyTransport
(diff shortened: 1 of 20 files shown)


tcard:
After a conversation with @paddybyers and @mattheworiordan, it seems like the iOS library is doing this pretty wrong. The IV is supposed to be randomly generated _in every message_ (it's currently being set when instantiating the channel only), and it's communicated as the first block in every encrypted message data. Java already does this, could take a look there.

This PR can be merged for now, but the IV param will be gone in a followup.

mattheworiordan:
Sure, looks like it can be merged.  Equally, not difficult to add another commit onto this PR to address `iv` in the correct way i.e. randomly generated for each encoding

tcard:
@mattheworiordan Done. Turns out it was actually doing that already, so I've only added a test in the end. I also checked with another library to confirm that the implementations are compatible.

ricardopereira:
LGTM

ablySpec/RealtimeClientChannel.swift
```diff
@@ -1297,6 +1299,66 @@ class RealtimeClientChannel: QuickSpec {
 
             }
 
+            context("crypto") {
+                it("if configured for encryption, channels encrypt and decrypt messages' data") {
```
mattheworiordan: I like this test, however I am concerned that at no point are we actually checking that the message is encrypted i.e. if you removed the cipher options completely this test would still pass.  A simple way to make sure this test is actually checking for encrypted messages is to add another subscriber that has encryption disabled and check that the messages are received with an encrypted payload i.e. `data` should be different and `encoding` should indicate it's an encrypted payload

tcard: Right now that wouldn't work, because that behavior is not working properly (see #211). Can I just intercept the raw `ProtocolMessage` and check that the data is encoded and the encoding is set there?

mattheworiordan: Ok sure intercept is fine

tcard: Done at 88ea27c.

tcard:
API is about to change, discussion at https://github.com/ably/docs/pull/89#discussion_r53929144. Once that's settled I make the changes here.

tcard:
@mattheworiordan @ricardopereira PTAL

mattheworiordan:
@tcard this looks good, but can you confirm that the `iv` is randomly generated and used for every message that is encoded?

tcard:
> @tcard this looks good, but can you confirm that the iv is randomly generated and used for every message that is encoded?

@mattheworiordan Fixed at 2d6e9b7. I realized along the way that the IV was being chained between messages when encrypting, I thought I checked this already but apparently I only did so for decryption, which was right. So I had to fix that.

Now CipherParams only has an IV for testing. If set, `encrypt` will use that, and if not, it will generate a random one.

ricardopereira:
LGTM

mattheworiordan:
LGTM

PR #232, opened 2016-02-19
Fix attach callbacks, and add to RealtimeChannel.subscribe.

### LINKED ISSUE #220: subscribe should optionally take an additional error callback
> IDL is `subscribe(String, (Message) ->) => io // RTL7b` which means you can register a listener for messages that come in, but there is also `=> io` which is a callback that will result in an error if the implicit attach does not work, see https://github.com/ably/docs/pull/87/files#diff-11900b395df266bc2bbfc1d11bdfc7a0R407

tcard:
Had to fix channel attach first. You can more easily review first one commit and then the other.

Fixes #220.

### DIFF
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -239,7 +239,7 @@ class RealtimeClientChannel: QuickSpec {
 
                     channel.attach { errorInfo in
                         expect(errorInfo).to(beNil())
-                        expect(channel.state).to(equal(ARTRealtimeChannelState.Attaching))
+                        expect(channel.state).to(equal(ARTRealtimeChannelState.Attached))
                     }
 
                     expect(channel.state).toEventually(equal(ARTRealtimeChannelState.Attached), timeout: testTimeout)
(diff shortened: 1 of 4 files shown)


ablySpec/RealtimeClientChannel.swift
```diff
@@ -769,6 +876,31 @@ class RealtimeClientChannel: QuickSpec {
...
+                    defer { client.close() }
+
+                    let channel = client.channels.get("test")
+
+                    let publishedMessage = ARTMessage(name: "foo", data: "bar")
+
+                    waitUntil(timeout: testTimeout) { done in
+                        expect(channel.state).to(equal(ARTRealtimeChannelState.Initialised))
+
+                        channel.subscribeWithAttachCallback({ errorInfo in
```
mattheworiordan: Does this method need to be overloaded given we've "preferred" to stick with the overloaded IDL?

mattheworiordan:
This PR looks good, but should we not also have a scenario for calling attach when failed, or calling attach when channel becomes failed, or calling detach when failed, or calling detach when becomes failed to ensure the callback is called with an error in those cases?

ablySpec/RealtimeClientChannel.swift
tcard: There's already a `subscribe:cb:` method that takes a message name in its first parameter. Objective-C doesn't let you overload only by type, you need to either have different number of params, different names for them, or just take any type (`id`, the equivalent to Java's `Object`) and downcast it manually, which we don't do anywhere else and is strongly anti-idiomatic.

tcard:
@mattheworiordan That's already done for `attach`, see https://github.com/ably/ably-ios/blob/subscribe-cb/ablySpec/RealtimeClientChannel.swift#L257 onwards. For `detach` it's still pending, I guess because it's not in the prioritization list.

mattheworiordan:
> @mattheworiordan That's already done for attach, see https://github.com/ably/ably-ios/blob/subscribe-cb/ablySpec/RealtimeClientChannel.swift#L257 

Nope, that is different.  A channel can be FAILED or become FAILED because of a permissions issue, whereas that test is for a connection state issue blocking the attach

> For detach it's still pending, I guess because it's not in the prioritization list.

Ok, well perhaps we should defer to later.  I thought it would be easy enough to add given you simply need to copy & paste the current test and cause the channel to fail, using permissions is the easiest way

ablySpec/RealtimeClientChannel.swift
ricardopereira: Yes, @tcard is right. Swift conversion could be better with this. 😞 It should imported by Swift like the initialisers where if the method takes an argument, the `With` should be removed and the rest of the selector is divided up into named parameters accordingly.

tcard:
@ricardopereira Could you take a look at this? It's blocking a couple of other changes.

ricardopereira:
LGTM

ablySpec/RealtimeClientChannel.swift
mattheworiordan: OK

PR #233, opened 2016-02-19
RTL10b

ricardopereira:


mattheworiordan:
I don't think mocking is the right approach here.

A robust integration test approach is simply (assume each step is synchronous)

* Create client1 and client2, client2 remains disconnected `autoConnect: false`
* client1 attaches to channel `foo` and once attached publishes 20 messages with id 0 -> 19
* client2 attaches
* client1 publishes 20 messages 20->39
* check client2 has received messages 20-39
* client2 gets history with `untilAttach` and should expect exactly messages `19->0` in that order

ricardopereira:
@mattheworiordan PTAL

ablySpec/RealtimeClientChannel.swift
```diff
@@ -980,6 +980,124 @@ class RealtimeClientChannel: QuickSpec {
...
+                    struct CaseTest {
+                        let untilAttach: Bool
+                        let expected: Int
+                    }
+
+                    let cases = [CaseTest(untilAttach: true, expected: 1),
+                                 CaseTest(untilAttach: false, expected: 2)]
+
+                    for caseItem in cases {
+                        it("where value is \(caseItem.untilAttach)") {
```
tcard: This should be describing what the test is checking, not just the context.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -980,6 +980,124 @@ class RealtimeClientChannel: QuickSpec {
...
+                            waitUntil(timeout: testTimeout) { done in
+                                channel.publish(nil, data: "postMessage") { errorInfo in
+                                    expect(errorInfo).to(beNil())
+                                    done()
+                                }
+                            }
+
+                            waitUntil(timeout: testTimeout) { done in
+                                try! channel.history { result, _ in
+                                    expect(result!.items.count).to(equal(caseItem.expected))
```
tcard: This is basically testing the same as the `it` below, only at a smaller scale. To simplify, maybe we can remove it, and just test the `from_serial` in the URL query?

tcard:
Untested: "If the untilAttach param is specified when the channel is not attached, it results in an error".

ablySpec/RealtimeClientChannel.swift
```diff
@@ -980,6 +980,124 @@ class RealtimeClientChannel: QuickSpec {
...
+                                expect(errorInfo).to(beNil())
+                                done()
+                            }
+                        }
+
+                        let query = ARTRealtimeHistoryQuery()
+                        query.untilAttach = true
+
+                        waitUntil(timeout: testTimeout) { done in
+                            try! channel2.history { result, errorInfo in
```
tcard: Should use the query.

ablySpec/RealtimeClientChannel.swift
```diff
@@ -980,6 +980,124 @@ class RealtimeClientChannel: QuickSpec {
...
+
+                            waitUntil(timeout: testTimeout) { done in
+                                channel.publish(nil, data: "postMessage") { errorInfo in
+                                    expect(errorInfo).to(beNil())
+                                    done()
+                                }
+                            }
+
+                            waitUntil(timeout: testTimeout) { done in
+                                try! channel.history { result, _ in
```
tcard: Should use the query.

ablySpec/RealtimeClientChannel.swift
ricardopereira: Pretty basic 😓Thanks

ablySpec/RealtimeClientChannel.swift
ricardopereira: Changed f343107.

ricardopereira:
> Untested: "If the untilAttach param is specified when the channel is not attached, it results in an error".

@tcard Done 630f61a.

tcard:
LGTM

PR #236, opened 2016-02-22
Force full channel release.

tcard:
Just removing the channel from the channels collection isn't
enough; the channel might have references elsewehere, including
reference cycles if the listener/subscriber callbacks close over
it.

### DIFF
--- a/ablySpec/RealtimeClientChannel.swift
+++ b/ablySpec/RealtimeClientChannel.swift
@@ -1297,6 +1297,40 @@ class RealtimeClientChannel: QuickSpec {
 
             }
 
+            context("channels") {
+                // RTS4
+                context("release") {
+                    it("should release a channel") {
+                        let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                        defer { client.close() }
+
+                        let channel = client.channels.get("test")
+                        channel.subscribe { _ in
+                            fail("shouldn't happen")
+                        }
+                        channel.presence.subscribe { _ in
+                            fail("shouldn't happen")
+                        }
+                        waitUntil(timeout: testTimeout) { done in
+                            client.channels.release("test") { errorInfo in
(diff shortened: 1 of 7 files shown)


ablySpec/RealtimeClientChannel.swift
```diff
@@ -1114,6 +1114,40 @@ class RealtimeClientChannel: QuickSpec {
...
+                                expect(channel.state).to(equal(ARTRealtimeChannelState.Detached))
+                                done()
+                            }
+                        }
+
+                        let sameChannel = client.channels.get("test")
+                        waitUntil(timeout: testTimeout) { done in
+                            sameChannel.subscribe { _ in
+                                sameChannel.presence.enterClient("foo", data: nil) { _ in
+                                    delay(0.0) { done() } // Delay to make sure EventEmitter finish its cycle.
```
ricardopereira: Delay of 0 seconds

ricardopereira:
LGTM

ablySpec/RealtimeClientChannel.swift
tcard: Yes, the delay is just to move the block to an "immediate event" in the event loop, so that we make sure that the presence subscription above gets called before `done()` is called in this line in case `release` doesn't work, ie. doesn't unsubscribe it.

Maybe there's a less confusing way of doing this...

ricardopereira: It's clear for me that the presence subscription should get called before `done()` is called but the delay of `0` is kind of strange. It's holding a couple of milliseconds. Is it enough?

tcard: It doesn't have to _actually_ delay, it just has to be done after the `EventEmitter.emit` cycle finishes. It doesn't matter if it happens immediately after.

PR #241, opened 2016-02-22
RTL2c

ricardopereira:


tcard:
The spec also says that the error should be emitted. I guess you can just add an `on` that checks that the error emitted is the same that `errorReason`.

ricardopereira:
> The spec also says that the error should be emitted. I guess you can just add an on that checks that the error emitted is the same that errorReason.

Done 14d53e7.
Thanks.

tcard:
LGTM

PR #242, opened 2016-02-23
History request without a query shouldn't throw errors

ricardopereira:
I notice this problem from the README examples.


tcard:
LGTM

PR #228, opened 2016-02-17
Fix RTN7c: notify publish callbacks of broken connection.

tcard:
[Review on Reviewable](https://reviewable.io/reviews/ably/ably-ios/228)

PR #244, opened 2016-02-23
RTN15a

ricardopereira:


ricardopereira:
Well, I realised right now that what I did here is more about (RTN15d)[http://docs.ably.io/client-lib-development-guide/features/#RTN15d].

ablySpec/RealtimeClientConnection.swift
```diff
@@ -1302,6 +1302,53 @@ class RealtimeClientConnection: QuickSpec {
...
+
+                    channel2.publish(nil, data: "message") { errorInfo in
+                        expect(errorInfo).to(beNil())
+                    }
+
+                    waitUntil(timeout: options.disconnectedRetryTimeout) { done in
+                        client1.connection.on { stateChange in
+                            switch stateChange!.current {
+                            case .Connecting:
+                                channel1.unsubscribe(listener1)
```
tcard: `listener1` should still work, shouldn't it? Maybe you can instead of failing inside unconditionally, add a condition that it has reached the reconnection. That way we test that previous subscriptions still work.

ricardopereira: Done 551d534.

ricardopereira:
> Connection state is only maintained for a brief period, up to a minute, so if a client is disconnected for a longer period connection state cannot be resumed
@tcard What do you think, should I test this? I think it's a server behaviour.

tcard:
@ricardopereira I don't see how we could test that meaningfully. Maybe @mattheworiordan does.

ricardopereira:

ricardopereira:

ablySpec/RealtimeClientConnection.swift
```diff
@@ -1302,6 +1302,63 @@ class RealtimeClientConnection: QuickSpec {
...
+                                reconnected = true
+                            default:
+                                break
+                            }
+                        }
+                    }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        // Check the queued message
+                        channel1.subscribe { message in
```
mattheworiordan: Why are you resubscribing, you have a listener already defined above on line 1322. Also, subscribing after you're connected is not going to work because it's quite possible the message has already been received at this point.

ablySpec/RealtimeClientConnection.swift
```diff
@@ -1302,6 +1302,63 @@ class RealtimeClientConnection: QuickSpec {
...
+                    let client1 = ARTRealtime(options: options)
+                    defer { client1.close() }
+                    let channel1 = client1.channels.get("test")
+
+                    let client2 = ARTRealtime(options: options)
+                    defer { client2.close() }
+                    let channel2 = client2.channels.get("test")
+
+                    var reconnected = false
+                    channel1.subscribe { message in
```
mattheworiordan: It would be great to add one small step into here so that we can be sure the serial numbers are being respected.  Publish two messages whilst both clients are connected, then disconnect, publish a message on the connected client, and then resume the connection, and ensure the message arrives on the existing listener

mattheworiordan:
>>  Connection state is only maintained for a brief period, up to a minute, so if a client is disconnected for a longer period connection state cannot be resumed

> @tcard What do you think, should I test this? I think it's a server behaviour.

You're right, you can't test that behaviour.  We can however test that a connection resume fails i.e. this would happen after connection state is lost.  We can test this by changing the connection key explicitly so that a connection cannot be resumed

ablySpec/RealtimeClientConnection.swift
ricardopereira: I will address this into #249.

ricardopereira:
Rebased with master with changes: b274f12.

ablySpec/RealtimeClientConnection.swift
```diff
@@ -1307,6 +1307,66 @@ class RealtimeClientConnection: QuickSpec {
 
             }
 
+            // RTN15
+            context("connection failures once CONNECTED") {
+
+                // RTN15a
+                it("should attempt to reconnect immediately and try to restore the connection") {
```
mattheworiordan: If this test is testing the failure condition, this description should reflect that.

ricardopereira: Done 6c55525.

ablySpec/RealtimeClientConnection.swift
```diff
@@ -782,7 +782,7 @@ class RealtimeClientConnection: QuickSpec {
                             delay(1.0) { done() }
                         }
 
-                        client.simulateLostConnection()
+                        client.simulateLostConnectionAndState()
```
mattheworiordan: Ah I see this is fixed in this PR so ignore my previous comment @ricardopereira, just working my way through the backlog...

mattheworiordan:
LGTM

ricardopereira:
@tcard PTAL

tcard:
LGTM

PR #245, opened 2016-02-23
Connection EventEmitter: use enum instead of NSNumber as event type.

tcard:
Internally we still need to wrap the raw enum values in NSNumber,
but that shouldn't be exposed.


ricardopereira:
LGTM

PR #246, opened 2016-02-23
RTL10b: Implement untilAttach for RealtimeChannel.history.

### LINKED ISSUE #234: Channel#history: `query.untilAttach` feature
[RTL10b](http://docs.ably.io/client-lib-development-guide/features/#RTL10b)

Java: https://github.com/ably/ably-java/pull/102

tcard:
Fixes #234.

### DIFF
--- a/ably-ios/ARTDataQuery+Private.h
+++ b/ably-ios/ARTDataQuery+Private.h
@@ -7,12 +7,19 @@
 //
 
 #import "ARTDataQuery.h"
+#import "ARTRealtimeChannel.h"
 
 ART_ASSUME_NONNULL_BEGIN
 
 @interface ARTDataQuery(Private)
 
 - (NSMutableArray /* <NSURLQueryItem *> */ *)asQueryItems;
+
+@end
+
+@interface ARTRealtimeHistoryQuery ()
+
+@property (weak, readwrite) ARTRealtimeChannel *realtimeChannel;
 
 @end
 
--- a/ably-ios/ARTDataQuery.m
+++ b/ably-ios/ARTDataQuery.m
@@ -7,6 +7,7 @@
 //
 
 #import "ARTDataQuery+Private.h"
+#import "ARTRealtimeChannel+Private.h"
 
 @implementation ARTDataQuery
 
@@ -48,5 +49,17 @@ static NSString *queryDirectionToString(ARTQueryDirection direction) {
 @end
 
 @implementation ARTRealtimeHistoryQuery
+
+- (NSMutableArray *)asQueryItems {
+    NSMutableArray *items = [super asQueryItems];
+    if (self.untilAttach) {
+        NSAssert(self.realtimeChannel, @"ARTRealtimeHistoryQuery used from outside ARTRealitmeChannel.history");
+        if (self.realtimeChannel.state != ARTRealtimeChannelAttached) {
+            @throw [NSError errorWithDomain:ARTAblyErrorDomain code:ARTRealtimeHistoryErrorNotAttached userInfo:nil];
+        }
+        [items addObject:[NSURLQueryItem queryItemWithName:@"fromSerial" value:self.realtimeChannel.attachSerial]];
+    }
+    return items;
+}
 
 @end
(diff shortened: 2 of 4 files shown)


ably-ios/ARTDataQuery+Private.h
```diff
@@ -16,4 +17,10 @@ ART_ASSUME_NONNULL_BEGIN
 
 @end
 
+@interface ARTRealtimeHistoryQuery ()
+
+@property (strong, readwrite) ARTRealtimeChannel *realtimeChannel;
```
ricardopereira: Maybe it's better to be `weak`.

ricardopereira:
Beside a small comment, LGTM

ably-ios/ARTDataQuery+Private.h
tcard: Hmm, why? realtimeChannel doesn't hold references to the ARTRealtimeHistoryQuery. There's no cycle here as far as I can see.

ricardopereira: The `query` can retain the channel if you create the `ARTRealtimeHistoryQuery` in other part of the main code and the client tries to release the channel.

tcard: Right, but as soon as the query is gone (ie. as soon as the call site returns) the channel will lose its only reference left and all will be good. How can that leak?

ricardopereira: The user can hold the query anyway and with `weak` you will have the guarantee.

ably-ios/ARTDataQuery.m
```diff
@@ -49,4 +50,16 @@ - (NSMutableArray *)asQueryItems {
...
 @implementation ARTRealtimeHistoryQuery
 
+- (NSMutableArray *)asQueryItems {
+    NSMutableArray *items = [super asQueryItems];
+    if (self.untilAttach) {
+        NSAssert(self.realtimeChannel, @"ARTRealtimeHistoryQuery used from outside ARTRealitmeChannel.history");
+        if (self.realtimeChannel.state != ARTRealtimeChannelAttached) {
+            @throw [NSError errorWithDomain:ARTAblyErrorDomain code:ARTRealtimeHistoryErrorNotAttached userInfo:nil];
+        }
+        [items addObject:[NSURLQueryItem queryItemWithName:@"from_serial" value:self.realtimeChannel.attachSerial]];
```
mattheworiordan: I believe this should be `fromSerial` as we changed all of our query params to use camelCase

mattheworiordan:
Looks good other than small comment

ably-ios/ARTDataQuery.m
tcard: Why is it working then? It is `from_serial` in the spec.

mattheworiordan: Because we support the old format as well for backwards compatibility :)

ably-ios/ARTDataQuery+Private.h
tcard: I really can't see and scenario in which this changes anything in practice, because the only way this property will be set is by `ARTRealtimeChannel.history` itself, and if you're calling that method obviously you _have_ the channel, and even if you drop the channel and hold to the query you can't do anything with the query but reuse it, which will change its `realtimeChannel`. But anyway, done at 547e39c.

ably-ios/ARTDataQuery.m
tcard: Oh, OK. Fixed at 7532c73.

ricardopereira: Oh, right. @tcard Please enhance https://github.com/ably/ably-ios/blob/9c7bc2c5122556bfe9ac66c20d762c67eafb833a/ablySpec/RealtimeClientChannel.swift#L1309

tcard: @ricardopereira That's done at 7532c73.

ricardopereira: Thank you

PR #247, opened 2016-02-23
TokenParams.timestamp: use current at getter, not at constructor.

### LINKED ISSUE #129: AuthTokenParams should not contain a default timestamp
See https://github.com/ably/ably-ios/blob/master/ably-ios/ARTTokenParams.m#L30

The problem with this is that if a use a default token params object in my code, and then later use to authorise or request a token, the timestamp may be out of date.  Instead, it is better that it is left `null`, and if `null` when authenticating the local current time should be used

tcard:
Fixes #129.


ricardopereira:
LGTM

mattheworiordan:
LGTM

tcard:
@mattheworiordan @ricardopereira I've had to change it because it needs to be sticky, ie. the timestamp shouldn't change once set. Otherwise, you can't check the MAC.

mattheworiordan:
LGTM

PR #249, opened 2016-02-23
RTN15d

ricardopereira:


ricardopereira:

ricardopereira:

ablySpec/RealtimeClientConnection.swift
```diff
@@ -1302,6 +1302,54 @@ class RealtimeClientConnection: QuickSpec {
 
             }
 
+            // RTN15
+            context("connection failures once CONNECTED") {
+
+                // RTN15d
+                pending("should recover from disconnection and messages delivered once the connection is resumed") {
```
mattheworiordan: Can we change this to "should recover from disconnection and publish queued messages once the connection is resumed"?

mattheworiordan: Actually, I may have misread RTN15d and this suggested change is wrong. However, is this test then not the same as RTN15a, I'm not sure why there are two different tests?

ricardopereira: You right, I noticed [that](https://github.com/ably/ably-ios/pull/244#issuecomment-187715273). I will change the #244.

ricardopereira:
Rebased with master with changes: 7b8e8f5.

mattheworiordan:
LGTM

tcard:
LGTM

PR #250, opened 2016-02-24
RTN15g

ricardopereira:


ricardopereira:

ricardopereira:

ablySpec/RealtimeClientConnection.swift
```diff
@@ -1302,6 +1302,35 @@ class RealtimeClientConnection: QuickSpec {
...
+
+                    channel.on(.Failed) { errorInfo in
+                        expect(errorInfo!.message).to(contain("Unable to recover connection"))
+                        expect(errorInfo!.code).to(equal(80008))
+                        expect(channel.errorReason).to(beIdenticalTo(errorInfo))
+                    }
+
+                    client.simulateResumeFailure()
+
+                    expect(channel.state).toEventually(equal(ARTRealtimeChannelState.Detached), timeout: testTimeout)
```
mattheworiordan: Shouldn't this be waiting for a `Failed` state and not a `Detached` state?

ricardopereira: Isn't suppose to on detach the channel?

ricardopereira:
Rebased with master and removed `pending`.

ablySpec/RealtimeClientConnection.swift
```diff
@@ -1307,6 +1307,35 @@ class RealtimeClientConnection: QuickSpec {
...
+                    options.disconnectedRetryTimeout = 1.0
+
+                    let client = ARTRealtime(options: options)
+                    defer { client.close() }
+                    let channel = client.channels.get("test")
+
+                    channel.attach()
+                    expect(channel.state).toEventually(equal(ARTRealtimeChannelState.Attached), timeout: testTimeout)
+
+                    channel.on(.Failed) { errorInfo in
```
mattheworiordan: Why are you checking `on(Failed)`, this should not happen when a connection resume has failed, the channel goes to a detached state.

tcard: More generally, every test that uses callbacks should make sure that the callback is being called. Please add that as well.

ricardopereira: Right, `on(Failed)` does nothing in the test.

ricardopereira: Done ea4a06e.

ablySpec/RealtimeClientConnection.swift
```diff
@@ -1307,6 +1307,33 @@ class RealtimeClientConnection: QuickSpec {
...
+                    let channel = client.channels.get("test")
+
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.attach() { errorInfo in
+                            expect(errorInfo).to(beNil())
+                            done()
+                        }
+                    }
+
+                    client.simulateLostConnection()
```
mattheworiordan: As mentioned previously, can we change the name of this method because it's misleading.  Lost connection is not the same as lost connection state.  Please update to something like `simulateLostConnectionAndState`

tcard:
@ricardopereira #231 is merged, you can rebase this now.

mattheworiordan:
@ricardopereira are you going to rebase this one?

ricardopereira:
Yes.

PR #251, opened 2016-02-24
RTN15e

ricardopereira:


ricardopereira:

ricardopereira:

ablySpec/RealtimeClientConnection.swift
```diff
@@ -1302,6 +1302,44 @@ class RealtimeClientConnection: QuickSpec {
...
+                        let client = ARTRealtime(options: options)
+                        client.setTransportClass(TestProxyTransport.self)
+                        client.connect()
+                        defer { client.close() }
+                        let channel = client.channels.get("test")
+
+                        channel.attach()
+                        expect(channel.state).toEventually(equal(ARTRealtimeChannelState.Attached), timeout: testTimeout)
+
+                        client.onDisconnected()
```
mattheworiordan: Why don't you add a `once(.Connecting)` step that explicitly resets the key to another value. That way you can be sure the iOS code is actually updating the `Connection#key`?

ricardopereira:
Rebased with master with changes: cc419e7.

ablySpec/RealtimeClientConnection.swift
ricardopereira: Done cc419e7.

mattheworiordan:
LGTM

tcard:
LGTM

PR #253, opened 2016-02-24
Fix RTL7d: error on decoding message.

tcard:


ablySpec/RealtimeClientChannel.swift
```diff
@@ -1068,7 +1068,7 @@ class RealtimeClientChannel: QuickSpec {
                 }
 
                 // RTL7d
-                pending("should deliver the message even if there is an error while decoding") {
+                context("TMP should deliver the message even if there is an error while decoding") {
```
ricardopereira: TMP?

ricardopereira:
LGTM

ably-ios/ARTRealtimeChannel.m
```diff
@@ -371,8 +371,10 @@ - (void)onMessage:(ARTProtocolMessage *)message {
             NSError *error = nil;
             msg = [msg decodeWithEncoder:dataEncoder error:&error];
             if (error != nil) {
-                [self.logger error:@"ARTRealtimeChannel: error decoding data: %@", error];
-                [NSException raise:NSInvalidArgumentException format:@"ARTRealtimeChannel: error decoding data: %@", error];
+                ARTErrorInfo *errorInfo = [ARTErrorInfo wrap:(ARTErrorInfo *)error.userInfo[NSLocalizedFailureReasonErrorKey] prepend:@"Failed to decode data: "];
+                [self.logger error:@"%@", errorInfo.message];
+                _errorReason = errorInfo;
+                [self emit:self.state with:errorInfo];
```
mattheworiordan: Does this mean on every error we are emitting the the current state again?  If you look at http://docs.ably.io/client-lib-development-guide/features/#RTL2c, it states we emit an error, but not emit the same state. It's a bit of a fudge because error is not a state, yet we emit an error from the eventemitter, see https://github.com/ably/ably-ruby/blob/master/spec/acceptance/realtime/channel_spec.rb#L174-L178 for an example. I am not sure how we have an equivalent for iOS, perhaps one to discuss?

ablySpec/RealtimeClientChannel.swift
tcard: Sorry, leftover. Fixed at f7c5ab5.

ably-ios/ARTRealtimeChannel.m
tcard: I see, so apart from the channel states we have an "error" event that is emitted through `RealtimeChannel.on`, right?

I can see two alternatives: create a new enum `ChannelEvent`, with all the variants in `ChannelState` plus an "error" variant; or a more convoluted class `ChannelState` with two properties: `newState: ChannelState?` and `error: ErrorInfo?`, which are only populated when appropriate.

mattheworiordan: I definitely prefer option 1...

tcard: Fixed at 24f9136. Changed in the IDL in ably/docs#92.

tcard:
@ricardopereira @mattheworiordan PTAL

mattheworiordan:
LGTM @tcard however it was somewhat dependent on https://github.com/ably/docs/pull/92#issuecomment-188851260 hence why I was waiting for @paddybyers' feedback.  Just to be clear, by changing the Enum type, will all on handlers both in Swift & iOS still work i.e. were there never previous on(event) handlers that used the ChannelState?  Can you clarify how this is handled now that it's been changed?

mattheworiordan:
From @paddybyers on my previous questions:

> the issue I suppose is whether or not different classes of events (state changes, errors, whatever) have different event arguments/payloads. Is it just a happy convenience that there’s no difference in this case?

tcard:
@mattheworiordan You're right it's dependent on https://github.com/ably/docs/pull/92#issuecomment-188851260, sorry. @ricardopereira You can cherry-pick 9954bc4 if it's blocking other things.

> will all on handlers both in Swift & iOS still work i.e. were there never previous on(event) handlers that used the ChannelState?

Correct.

> Can you clarify how this is handled now that it's been changed?

What do you mean? The only change is that

paddybyers:
> What do you mean? The only change is that RealtimeChannel's EventEmitter now has ChannelEvent as event type. When there's a state change, it emits the ChannelEvent that corresponds to the new state.

The question is what happens if different events have different parameters. I know it doesn't apply in this case, but suppose `attached`, `detached` etc have a `ChannelStateChange` event payload, but an `ERROR` has an `ErrorInfo` payload. Does that information get communicated in the IDL? If so, does it mean that the space of event names isn't a single space of names, but a set of grouped names corresponding to different payload types?

tcard:
About @paddybyers ' concerns, I agree that we're conflating several kind of events, and that could be troublesome if not for this happy convenience of them bringing the same kind of payload. Thus I don't think it's a problem, we can keep it simple like this.

paddybyers:
> Thus I don't think it's a problem, we can keep it simple like this.

I agree; I'm fine with it so long as we understand this won't necessarily generalise, so it might be revisited at a later time.

tcard:
> but suppose attached, detached etc have a ChannelStateChange event payload, but an ERROR has an ErrorInfo payload. Does that information get communicated in the IDL?

If we wanted to do that, we would need either separate event emitters for separate kind of payloads or a more convoluted EventEmitter that does something like this:

```rust
class StateChangeEvent<StateType>: // This can be reused for connection state.
    previous: <StateType>
    current: <StateType>
    reason: ErrorInfo?

// This enum could be implemented as a parent class that is downcasted in a Java-like language.
enum ChannelEvent:
    StateChange(StateChangeEvent<ChannelState>)
    Error(ErrorInfo)

class RealtimeChannel:
    on(callback: (ChannelEvent) ->) // Registers listener for all kinds of events
    on(ChannelState, callback: (StateChangeEvent<ChannelState> ->) // Registers listener only for a particular ChannelState
    onError(callback: ErrorInfo) ->) // Registers listener only for errors not related to state changes.

    // Realtime would have to have separate EventEmitters internally for each case, as the generic EventEmitter can't deal with this association of event type's values -> payload type generically.
```

So this is all the complexity we avoid by just having a `ErrorReason?` payload that is "overloaded" to mean both "error, if any, that justifies the state change" or "some error happened for some reason, not involving a state change, although you will have to check that for yourself".

tcard:
> I agree; I'm fine with it so long as we understand this won't necessarily generalise, so it might be revisited at a later time.

Agreed. I take this as that you approve of ably/docs#92 then?

paddybyers:
> I take this as that you approve of ably/docs#92 then?

I'm fine. @mattheworiordan ?

mattheworiordan:
> Agreed. I take this as that you approve of ably/docs#92 then?

Yes

tcard:
Failing because master is failing. I'm merging, #266 should then fix part of the problem.

ricardopereira:
I will rebase #266.

PR #255, opened 2016-02-24
Fix RSC1: Rest constructor should accept a token.

tcard:


ablySpec/RestClient.swift
```diff
@@ -44,7 +46,10 @@ class RestClient: QuickSpec {
                     expect(publishTask.error?.code).toEventually(equal(40005), timeout:testTimeout)
                 }
 
-                pending("should accept a token") {
+                it("should accept a token") {
+                    ARTClientOptions.setDefaultEnvironment("sandbox")
+                    defer { ARTClientOptions.setDefaultEnvironment("sandbox") }
```
ricardopereira: Assigning again with `sandbox`?

mattheworiordan: Yup, not sure I follow what this is doing

mattheworiordan:
Other than one comment, LGTM

ablySpec/RestClient.swift
tcard: Oh, silly me. Of course, this is meant to restore the default `nil` value. Fixed at b30271d.

ricardopereira:
:+1:

PR #260, opened 2016-02-25
Fix RTN14b test.

tcard:


tcard:
I just realized this is missing: "If the token creation fails, the next connection attempt fails again due to a token error [...] the connection will transition to the FAILED state and the Connection#errorReason will be set". Will do soon.

ricardopereira:
LGTM

ricardopereira:
@mattheworiordan PTAL

mattheworiordan:
NIce, LGTM

PR #265, opened 2016-03-01
Xcode 7.2 (use xcodebuild)

ricardopereira:

PR #267, opened 2016-03-01
Bitrise CI

ricardopereira:
[Bitrise.io](http://bitrise.io) is also free and has a great support service.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -1,0 +1,27 @@
+---
+format_version: 1.1.0
+default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
+app:
+  envs:
+  - BITRISE_PROJECT_PATH: ably.xcworkspace
+    opts:
+      is_expand: false
+  - BITRISE_SCHEME: ably
+    opts:
+      is_expand: false
+trigger_map:
+- pattern: "*"
+  is_pull_request_allowed: true
+  workflow: primary
+workflows:
+  primary:
+    steps:
+    - activate-ssh-key@3.1.0:
+        title: Activate App SSH key
+        inputs:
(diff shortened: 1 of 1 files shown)


ricardopereira:
I will try this: http://devcenter.bitrise.io/docs/step-hangs-times-out-after-a-period-without-any-logs-on-bitrise

tcard:
I don't see how any of those can apply to our case. Anyway, can't you try this on a separate branch?

ricardopereira:
Yes, I can but you're right. I was wondering if the scheme was `shared` and it is. Any other doesn't apply.

ricardopereira:
Have you faced any stalled build after #265?

tcard:
@ricardopereira None so far.

ricardopereira:
Me too.
Many builds are failing because of this:
```
-[RestClientChannels RestClient__channels__releaseChannel__should_release_a_channel()]
```

mattheworiordan:
I don't care which CI we use, however I would be surprised if Travis is at fault here.  If Bitrise works though, then we should move across.

ricardopereira:
@mattheworiordan No, Bitrise is failing too.
Before this PR I updated the Travis server with the latest stable version of Xcode and removed `scan`. Currently, the test suite is executed with the standard xcodebuild command.

ricardopereira:
So, currently the test suite seems to be fine. We run a couple of tests and none of them hanged.
I read that `xcpretty` (used by `scan`) seems to be buggy with the latest Xcode versions but like @tcard said the test suite was hanging before the `scan` was introduced so maybe the new Xcode version solved it. From the release notes:
> Fixed in Xcode 7.2.1:
• Command line tool ‘xcodebuild test’ will no longer time out waiting for Simulator.app to launch

I will close this and get back to the priority list.

PR #222, opened 2016-02-16
Remove 'pending' from RSN4 test.

tcard:
It is working now.

[Review on Reviewable](https://reviewable.io/reviews/ably/ably-ios/222)

PR #268, opened 2016-03-01
Fix RSN4: releaseChannel

ricardopereira:


tcard:
I don't understand why this wasn't working, it definitely was working before (#222 shows green). Anyway, LGTM.

ricardopereira:
Before this change, the channel, sometimes or most of the times, was released when the test goes off. The `release` implementation was always correct.

PR #273, opened 2016-03-03
Use ErrorInfo instead of internal ARTStatus in ping callback.

tcard:
Plus, remove a couple of deprecated typedefs.


ricardopereira:
LGTM

PR #274, opened 2016-03-03
Rename all cb: to callback:

tcard:
We were using both, inconsistently. `callback:` is longer, but
clearer, plus that's considered a good thing in iOS.


ricardopereira:
LGTM

PR #275, opened 2016-03-03
Remove block typedefs, use block syntax explicitly.

tcard:
Typedefs for block types are pretty annoying, because when checking how you should call some method that takes a callback you then need also to check the signature of the callback type. Now it's all inlined and explicit, if more verbose.


ricardopereira:
Awesome :+1:

mattheworiordan:
Ok, looks good. I suppose there is an argument that sharing types could be useful, but I expect in reality, this is not the case an inlining it like this helps developers.  Nice one @tcard

ricardopereira:
I think it's a known bug from Xcode but for now it's better so. It's quite annoying to select one method from code completion and then you need to check the list of arguments for a block. Great one.

ricardopereira:
Need a rebase after #276

PR #276, opened 2016-03-04
Travis enhancements

ricardopereira:


ricardopereira:
I did a little optimisation. If the first test target fails then it will stop the test suite immediately.
@mattheworiordan @tcard PTAL. I want to merge this.

mattheworiordan:
>  I want to merge this.

And I'm watching [the build](https://travis-ci.org/ably/ably-ios/builds/113623972) :)

ricardopereira:
Unfortunately, the log will only update in the end of the execution 😞 but at least the test suite will be more stable... I hope so 😁

ricardopereira:
The `AblySpec` target failed and the `AblyTests` didn't start :+1:
@mattheworiordan Can I run the test again and see if it passes?

tcard:
I'm not sure this is a good idea. I want to know all the tests I'm breaking so that I can fix all of them.

ricardopereira:
But it takes time and you will know all the failing tests from the current test target.

ricardopereira:
>> I'm not sure this is a good idea. I want to know all the tests I'm breaking so that I can fix all of them.

> @tcard But it takes time and you will know all the failing tests from the current test target.

@mattheworiordan What is your opinion about this?

mattheworiordan:
>> But it takes time and you will know all the failing tests from the current test target.

> What is your opinion about this?

I am not sure I understand the problem.  Are you saying we don't get the list of broken tests in the output?  Generally, I prefer for a build not to abort and run all tests so that when you get the feedback, you can go through all broken tests, fix them and rebuild.  If we abort a test early, won't it then prevent this?  Why was 81eca60  needed?

ricardopereira:
@mattheworiordan We have two steps: `AblySpec` (Swift) and `AblyTests` (ObjC). I did 81eca60 in case of one of the steps fails, the test suite stops immediately and is marked as errored. It will show all the broken tests about the current step.

ricardopereira:
Sorry, I didn't mention why I did it. Well, I thought it would be nice to abort the test suite immediately if one of the steps fails to win time but maybe it makes more sense to have a clear idea on what it's failing.

ricardopereira:
Ok, I will undo 81eca60.

ricardopereira:
I decided to do a new commit to maintain the old ones in the history. Could be helpful for something else.

ricardopereira:
@mattheworiordan @tcard Can I merge?

tcard:
Yes, LGTM.

Issue #142, opened 2016-01-20
RestClientStats tests are failing inconsistently.


tcard:
From time to time, a RestClientStats test fails with an off-by-one error.

``````
[16:18:08]: ▸ ablySpec.RestClientStats
[16:18:08]: ▸ RestClient__stats__result__should_match_hour_level_inbound_and_outbound_fixture_data__forwards__UserstravisbuildablyablyiosablySpecRestClientstatsswift_160, failed - expected to equal <1>, got <2>
[16:18:08]: ▸ /Users/travis/build/ably/ably-ios/ablySpec/RestClient.stats.swift:181
[16:18:08]: ▸ ```
[16:18:08]: ▸ 
[16:18:08]: ▸                         expect(result.items.count).to(equal(1))
[16:18:08]: ▸                         expect(totalInbound).to(equal(50 + 60 + 70))
[16:18:08]: ▸   ```
[16:18:08]: ▸      Executed 190 tests, with 1 failure (0 unexpected) in 508.631 (509.033) seconds
[16:18:08]: ▸ 
``````

tcard:
Another failure that has nothing to do with the new code (#144):

	[17:08:50]: ▸ ablySpec.RestChannel
	[17:08:50]: ▸ presence__get__should_return_presence_fixture_data_UserstravisbuildablyablyiosablySpecRestChannelswift_206, failed - expected to eventually equal <6>, got <0>
	[17:08:50]: ▸ /Users/travis/build/ably/ably-ios/ablySpec/RestChannel.swift:216
	[17:08:50]: ▸ ```
	[17:08:50]: ▸ 
	[17:08:50]: ▸                     expect(presenceMessages.count).toEventually(equal(presenceFixtures.count), timeout: testTimeout)
	[17:08:50]: ▸                     for message in presenceMessages {
	[17:08:50]: ▸   ```
	[17:08:50]: ▸ 	 Executed 185 tests, with 1 failure (0 unexpected) in 551.620 (551.911) seconds

This looks like a timeout in Presence.get. testTimeout is 10 s, so it's taking longer than that, which is not good. Maybe it's a problem of Travis itself? The alternative is that something is wrong in realtime.

tcard:
@ricardopereira Why did you close this? Last time I checked (when trying out #173) it was still happening.

ricardopereira:
I thought it was fixed.

tcard:
#278 fixes this.

PR #278, opened 2016-03-07
Remove pending from RestClientStats

ricardopereira:
#142

### DIFF
--- a/Spec/RestClientStats.swift
+++ b/Spec/RestClientStats.swift
@@ -81,10 +81,7 @@ class RestClientStats: QuickSpec {
     override func spec() {
         describe("RestClient") {
             // RSC6
-            pending("stats") {
-                // Temporarily disabled until #https://github.com/ably/ably-ios/issues/142
-                // is figured out.
-
+            context("stats") {
                 // RSC6a
                 context("result") {
                     let calendar = NSCalendar(identifier: NSCalendarIdentifierGregorian)!


mattheworiordan:
LGTM, but not passing?

ricardopereira:
Yes, both failed with: `Channel__subscribe__should_exist_ensuring_published_messages_are_not_echoed_back_to_the_subscriber_when_echoMessages_is_false, failed - Waited more than 10.0 seconds`

Will check.

tcard:
I don't think this will just work now. Remember that this happened intermittently, with an off-by-one error, which suggests something related to timing, like some events being recorded in the stats in the next interval than expected. Nothing has changed that would prevent that, AFAIK.

ricardopereira:
I think it's more than that. The problem is affecting all the test suite. All the current PRs aren't passing with this similar problem: tests failing intermitently. I'm catching all of those fails to see any relation between them: #283. The CI is quite useless with this problem. It's really frustrating.

tcard:
That's a different issue @ricardopereira, I'm talking specifically about the stats tests.

mattheworiordan:
FYI @ricardopereira, sandbox was a bit unstable in the last 24 hours or so so could have been contributing to this.

ricardopereira:
Lets see if we get fails from `stats` and then I analyse what could be wrong.

PR #282, opened 2016-03-07
RTP1

ricardopereira:


Spec/RealtimeClientPresence.swift
```diff
@@ -9,11 +9,83 @@
...
+                    let transport = client.transport as! TestProxyTransport
+                    let attached = transport.protocolMessagesReceived.filter({ $0.action == .Attached })[0]
+
+                    expect(attached.flags & 0x1).to(equal(0))
+                    expect(attached.isSyncEnabled()).to(beFalse())
+                    expect(channel.presence.syncComplete).to(beFalse())
+                    expect(channel.presenceMap.isSyncComplete()).to(beFalse())
+                }
+
+                it("when members are present") {
```
tcard: This tests when _one_ member is present. Spec says "subsequent ProtocolMessage SYNC messages will be delivered", which requires a test that expects more then one SYNC messages to arrive and be processed properly.

mattheworiordan: I would agree, please add 250 present members, and expect 3 SYNC pages of members.

ricardopereira: Done d6a47ec.

tcard:
LGTM

PR #286, opened 2016-03-08
Fix RTL1

ricardopereira:


mattheworiordan:
Seems good to me, but not sure about the error, will have to leave that to @tcard

Spec/RealtimeClientChannel.swift
```diff
@@ -70,11 +67,11 @@ class RealtimeClientChannel: QuickSpec {
                     }
```
tcard: I think line 49 above (can't comment on it) also needs a `toEventually`, or a wait until for the `channel2.attach()` callback.

tcard: Also, the `subscribe` on line 52 should be moved above the `channel2.attach()`, to make sure it happens before it is attached.

ricardopereira: Done 60b1ef1. Thanks.

tcard:
LGTM

PR #289, opened 2016-03-10
Fix uncaught HTTP exception

ricardopereira:
- should always enqueue the callback block

### DIFF
--- a/Source/ARTHttp.m
+++ b/Source/ARTHttp.m
@@ -216,8 +216,7 @@
     request.HTTPBody = artRequest.body;
     [self.logger debug:@"ARTHttp: makeRequest %@", [request allHTTPHeaderFields]];
 
-    CFRunLoopRef rl = CFRunLoopGetCurrent();
-    CFRetain(rl);
+    __block CFRunLoopRef rl = CFRunLoopGetCurrent();
 
     [self.urlSession get:request completion:^(NSHTTPURLResponse *response, NSData *data, NSError *error) {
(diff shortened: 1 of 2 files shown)


mattheworiordan:
Not sure I understand why this change is needed, @tcard over to you

ricardopereira:
@mattheworiordan I fixed it because it was raising an exception when a request was returning 500 ([see here](https://github.com/ably/ably-ios/pull/289/files#diff-6decbd6ac2aba5822e5b35c79f08d353L85)). It was calling the callback ([here](https://github.com/ably/ably-ios/pull/289/files#diff-6decbd6ac2aba5822e5b35c79f08d353L88)) without enqueuing the block on the current loop.

BTW, the sandbox seems to be unstable.

mattheworiordan:
> BTW, the sandbox seems to be unstable.

Yes, we are oddly getting an unstable sandbox quite frequently.  I will ask Paddy to fix it now, sorry

ricardopereira:
And with this bug (all night on this one 😅) I found out that the `[xcpretty](https://github.com/supermarin/xcpretty)` ignores abnormal failures, consequently the log hangs and Travis kills the process. I reproduce it successfully: https://github.com/ricardopereira/CrashTestSuiteExample.

mattheworiordan:
> And with this bug (all night on this one 😅) I found out that the xcpretty ignores abnormal failures, 

Well done finding that one!

tcard:
LGTM

PR #295, opened 2016-03-12
Fix RTL5b

ricardopereira:


Tests/ARTRealtimeChannelTest.m
```diff
@@ -320,33 +320,6 @@ - (void)testAttachFails {
     [self waitForExpectationsWithTimeout:[ARTTestUtil timeout] handler:nil];
 }
 
-- (void)testDetachFails {
```
ricardopereira: I removed it because the same test already exists in Swift (RTL5b)

tcard:
LGTM

Issue #271, opened 2016-03-03
RealtimePresence.get shouldn't wrap RestPresence.get.


tcard:
It should return a `NSArray<PresenceMessage *> *` from the PresenceMap instead.

mattheworiordan:
@tcard as this is a known API incompatibility, I've added to the spreadsheet

mattheworiordan:
I meant @ricardopereira :)

PR #296, opened 2016-03-12
RTP11a

ricardopereira:


ricardopereira:
Oh. I forgot about #271 😖
I need to fix it first.

Spec/RealtimeClientPresence.swift
```diff
@@ -98,6 +98,86 @@ class RealtimeClientPresence: QuickSpec {
...
+                    ARTRealtimePresenceQuery.testSuite_injectIntoClassMethod("init") { // Default initialiser
+                        presenceQueryWasCreated = true
+                    }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.presence.get { membersPage, error in
+                            expect(error).to(beNil())
+
+                            let membersPage = membersPage!
+                            expect(membersPage.items).to(haveCount(100))
```
mattheworiordan: This is wrong unfortunately. The realtime API for Presence#get does not return a page result set.  Check the IDL, the REST API returns `PaginatedResult<PresenceMessage>` whereas the Realtime API returns `[PresenceMessage]`.  There is no need for paging as all the members are retained in memory anyway

ricardopereira: Yes, I'm aware. I'm currently dealing with #299 and then I will fix this.

mattheworiordan: BTW. I would not throw this away, this test is pretty much what is needed for RSP3 so perhaps just repurpose it and do both at the same time?

ricardopereira: Yes, sure.

Spec/RealtimeClientPresence.swift
```diff
@@ -195,40 +195,15 @@ class RealtimeClientPresence: QuickSpec {
...
-                                expect(lastItem.clientId).toNot(beNil())
-                                expect(lastItem.data as? NSObject).to(equal(expectedData))
-
-                                expect(nextPage.hasNext).to(beFalse())
-                                expect(nextPage.isLast).to(beTrue())
-                                done()
-                            }
+                            expect(members).to(haveCount(150))
+                            expect(members!.first).to(beAnInstanceOf(ARTPresenceMessage))
+                            expect(members!.first!.clientId).to(contain("user"))
```
mattheworiordan: Not sure if we have this anywhere else, but I would quickly build an array of clientId values and ensure when `uniqued` there are 150 of them.

mattheworiordan:
Looks good, are you able to reuse the existing code to do a REST get test as well?  You could even just peg a REST test onto the end of this test to check that 150 members across 2 pages is returned, keeps the test suite time down... WDYT?

ricardopereira:
> Looks good, are you able to reuse the existing code to do a REST get test as well? You could even just peg a REST test onto the end of this test to check that 150 members across 2 pages is returned, keeps the test suite time down... WDYT?

@mattheworiordan I added the test on another PR: #308.

Spec/RealtimeClientPresence.swift
ricardopereira: @mattheworiordan You mean something like this: https://github.com/ably/ably-ios/blob/dd01cf80ce12eb9077b0a198a4c0df61db38e30a/Spec/TestUtilities.swift#L161-L168. Do you want to check every `memberItem.clientId` and `memberItem.data` from `Presence.get`?

ricardopereira: Done d10fbfb.

tcard:
LGTM

PR #303, opened 2016-03-16
Test suite: avoid calling the `fulfill` after the wait context has ended

ricardopereira:
I declared each expectation variable as `__weak` to avoid API violations. The expectation will be released when the timeout expires. So if the task takes longer than the timeout, the expectation variable will be nil when the task completion handler is called. Thus the `fulfill` method will be called on nil, doing nothing.

### DIFF
--- a/Tests/ARTFallbackTest.m
+++ b/Tests/ARTFallbackTest.m
@@ -14,12 +14,11 @@
 
 @implementation ARTFallbackTest
 
--(void)testAllHostsIncludedOnce {
-    
-    ARTFallback * f = [[ARTFallback alloc] init];
-    NSArray * defaultHosts = [ARTDefault fallbackHosts];
-    NSSet * defaultSet = [NSSet setWithArray:defaultHosts];
-    NSMutableArray * hostsRandomised = [NSMutableArray array];
+- (void)testAllHostsIncludedOnce {
+    ARTFallback *f = [[ARTFallback alloc] init];
+    NSArray *defaultHosts = [ARTDefault fallbackHosts];
+    NSSet *defaultSet = [NSSet setWithArray:defaultHosts];
+    NSMutableArray *hostsRandomised = [NSMutableArray array];
     for(int i=0;i < [defaultHosts count]; i++) {
         [hostsRandomised addObject:[f popFallbackHost]];
     }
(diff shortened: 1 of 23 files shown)


tcard:
Doesn't assigning a freshly allocated object to a weak variable just deallocate the object right away?

From a quick test:

```swift
 10> func test() { 
 11.     weak var x = NSString(string: "not nil") 
 12.     print(x) 
 13. }    
 14> test()
nil
```

ricardopereira:
Yes, that's my point but it will wait because of the `waitForExpectationsWithTimeout`. So the `expectation` variable is available until the `fullfill` or the `timeout`.

ricardopereira:
Can I merge?

tcard:
I mean that, unless I'm mistaken, the `expectation` variable will never be not nil. I think the snippet I pasted above proves that; `x` is immediately nil after assigning it.

ricardopereira:
Yes, correct. In your example the instance has no strong reference but when you create an expectation  `__weak XCTestExpectation *expectation = [self expectationWithDescription:@"testCase"];` it will have a strong reference. Although not clearly documented, the expectation will be released when the timeout expires. You can check the reference count with `[expectation retainCount]`:
```
2016-03-17 08:25:34.155 xctest[59800:9879713] ------- testSubscribeToName
(lldb) po [exp retainCount]
2
```
Thus there are two references to it. I don't know if one of them is a strong but I tested it and it's working as expected:

```objective-c
- (void)testSubscribeToName {
    __weak XCTestExpectation *expectation = [self expectationWithDescription:@"testSubscribeToName"];
    NSString *channelName = @"channel";
    [ARTTestUtil testRealtime:^(ARTRealtime *realtime) {
        NSLog(@"------- %@", expectation);
        _realtime = realtime;
        ARTRealtimeChannel *channel = [realtime.channels get:channelName];
        [expectation fulfill];
    }];
    [self waitForExpectationsWithTimeout:[ARTTestUtil timeout] handler:nil];
    NSLog(@"------- %@", expectation);
}
```

```
2016-03-17 08:25:34.155 xctest[59800:9879713] ------- testSubscribeToName
2016-03-17 08:25:34.156 xctest[59800:9879713] ------- testSubscribeToName
```

tcard:
Ah, OK, of course, `self` should be retaining a reference to the expectation so that `waitForExpectationsWithTimeout` works. Sorry, just wanted to understand what's going on. LGTM then!

PR #311, opened 2016-03-18
RTP2

ricardopereira:


Spec/RealtimeClientPresence.swift
```diff
@@ -77,6 +77,59 @@ class RealtimeClientPresence: QuickSpec {
...
+
+                channel.presence.subscribe { presence in
+                    expect(presence.action).to(equal(ARTPresenceAction.Leave))
+                    expect(presence.clientId).to(equal("user50"))
+                }
+
+                channel.presenceMap.testSuite_getArgumentFrom("put:", atIndex: 0) { arg0 in
+                    let member = arg0 as! ARTPresenceMessage
+                    if member.clientId == "user50" {
+                        print(member.timestamp)
```
tcard: Leftover? Should this whole "paragraph" testing anything?

Spec/RealtimeClientPresence.swift
```diff
@@ -77,6 +77,59 @@ class RealtimeClientPresence: QuickSpec {
...
+                    }
+                }
+
+                waitUntil(timeout: testTimeout) { done in
+                    channel.attach() { _ in
+                        let transport = client.transport as! TestProxyTransport
+                        transport.beforeProcessingReceivedMessage = { protocolMessage in
+                            // A leave event for a member can arrive before that member is later registered as present as part of the initial SYNC operation.
```
tcard: I don't get how is this being enforced in the test. How do you know that the leave event for `user50` is arriving to `client` before its enter event? The only thing I see here is that `clientSecondary` is leaving `user50`, but that event can reach `client` _after_ the SYNC is complete, as far as I can tell.

mattheworiordan: Agreed, I think the only viable way to test this is to send a mock protocol message to the library mid sync before `user50` has been received simulating that message being received.

Spec/RealtimeClientPresence.swift
ricardopereira: Ups, yes. Thanks.

Spec/RealtimeClientPresence.swift
ricardopereira: PTAL 9d80ec2.

tcard:
LGTM

tcard:
In fact, no LGTM, see failing test. `leavePresenceTimestamp` isn't being set before you compare to it.

tcard:
PTAL

ricardopereira:
LGTM

PR #317, opened 2016-03-20
RTP16b

ricardopereira:


tcard:
Should also be tested the "and ClientOptions#queueMessages has not been explicitly set to false" part, ie. that the behavior changes.

ricardopereira:
> Should also be tested the "and ClientOptions#queueMessages has not been explicitly set to false" part, ie. that the behavior changes.

Done e993184. It doesn't return the error. Thanks.

tcard:
LGTM

PR #318, opened 2016-03-20
RTP16c

ricardopereira:


Spec/RealtimeClientPresence.swift
```diff
@@ -136,6 +136,59 @@ class RealtimeClientPresence: QuickSpec {
                 }
             }
 
+            // RTP16
+            context("Connection state conditions") {
+
+                // RTP16c
+                pending("should result in an error if the connection state is INITIALIZED") {
```
tcard: I'm probably missing something, but where in this test is the connection INITIALIZED? Won't it be CONNECTING right away?

Also, I'm not really sure this applies to INITIALIZED since it's not an state _change_. @mattheworiordan could you confirm this?

ricardopereira: Yes, you right. The connection is in the `CONNECTING` state. I verified the ruby test (keep in mind that I'm not comfortable with ruby) and it seems to check the `INITIALIZED` state: https://github.com/ably/ably-ruby/blob/6e8e1ea7949026f19fa4c1405ccb5d8b96c3aeb9/spec/acceptance/realtime/presence_spec.rb#L58-L64

ricardopereira: Done aef47db.

tcard:
LGTM

PR #321, opened 2016-03-21
RTP6c

ricardopereira:


mattheworiordan:
Looks good, think we have to modify the interface to allow the error callback to be provided

ricardopereira:
Done #352.

PR #337, opened 2016-03-21
RTP10b

ricardopereira:


mattheworiordan:
LGTM

tcard:
PTAL. The test wasn't actually working because we were simulating an ERROR protocol message that had nothing to do with the LEAVE operation, so it didn't affect it at all; the ACK arrived just fine. What we need is to replace the arriving ACK with a NACK to trigger the error in the callback.

ricardopereira:
Cool, thanks.

Spec/RealtimeClientPresence.swift
```diff
@@ -868,7 +868,7 @@ class RealtimeClientPresence: QuickSpec {
                     defer { client.close() }
                     let channel = client.channels.get("test")
 
-                    channel.attach()
+                                        channel.attach()
```
mattheworiordan: Odd spacing here?

ricardopereira: Done 3200787.

PR #341, opened 2016-03-21
RTP12b

ricardopereira:


tcard:
PTAL

Source/ARTRealtimePresence.m
```diff
@@ -74,7 +75,16 @@ - (void)history:(void (^)(__GENERIC(ARTPaginatedResult, ARTPresenceMessage *) *,
 }
 
 - (BOOL)history:(ARTRealtimeHistoryQuery *)query callback:(void (^)(__GENERIC(ARTPaginatedResult, ARTPresenceMessage *) *, ARTErrorInfo *))callback error:(NSError **)errorPtr {
-    return [super history:query callback:callback error:errorPtr];
+    query.realtimeChannel = self.channel;
+    @try {
+        return [super history:query callback:callback error:errorPtr];
```
ricardopereira: Why did you added a try catch? Is it necessary?

tcard: Yes, [this](https://github.com/ably/ably-ios/blob/master/Source/ARTDataQuery.m#L58) could happen.

ricardopereira: Thanks for clarifying :+1:

PR #344, opened 2016-03-21
RTN19b

ricardopereira:


mattheworiordan:
LGTM, although could we do the same for detaching?

tcard:
PTAL

ricardopereira:
LGTM

PR #347, opened 2016-03-22
RSA8d

ricardopereira:


Spec/Auth.swift
```diff
@@ -713,6 +713,67 @@ class Auth : QuickSpec {
...
+            context("When authCallback option is set, it will invoke the callback") {
+
+                pending("with a token string") {
+                    let options = AblyTests.clientOptions()
+                    let expectedTokenParams = ARTTokenParams()
+
+                    options.authCallback = { tokenParams, completion in
+                        expect(tokenParams).to(beIdenticalTo(expectedTokenParams))
+                        // TODO: call the completion with a token string
+                        //completion("token_string", nil)
```
mattheworiordan: Why is this commented out?

Spec/Auth.swift
```diff
@@ -713,6 +713,67 @@ class Auth : QuickSpec {
...
+                }
+
+                pending("with a TokenRequest") {
+                    let options = AblyTests.clientOptions()
+                    let expectedTokenParams = ARTTokenParams()
+
+                    options.authCallback = { tokenParams, completion in
+                        expect(tokenParams).to(beIdenticalTo(expectedTokenParams))
+                        // TODO: call the completion with a TokenRequest
+                        //completion(ARTTokenRequest(), nil)
```
mattheworiordan: Why commented out?

mattheworiordan:
Few comments, but otherwise LGTM

Spec/Auth.swift
ricardopereira: Because it doesn't compile.

Spec/Auth.swift
ricardopereira: Because it doesn't compile. The completion only receives a `TokenDetails` but I saw on other libs that it is also possible to call it with a Token string.

tcard:
PTAL

ricardopereira:
LGTM

PR #349, opened 2016-03-22
RSA8f2

ricardopereira:


Spec/Auth.swift
```diff
@@ -713,6 +713,27 @@ class Auth : QuickSpec {
                 }
 
             }
+
+            // RSA8f2
+            it("ensure that the message is rejected") {
+                let token = getTestToken()
+                let options = ARTClientOptions(token: token)
```
mattheworiordan: Has this token got a null value clientId?

Spec/Auth.swift
```diff
@@ -713,6 +713,27 @@ class Auth : QuickSpec {
...
+                options.environment = "sandbox"
+                options.clientId = "john"
```
mattheworiordan: Like RSA8f3 & RSA8f4, this is wrong.

Spec/Auth.swift
ricardopereira: Done 70ab463.

Spec/Auth.swift
ricardopereira: Done 70ab463.

tcard:
LGTM

PR #350, opened 2016-03-22
RSA8f3

ricardopereira:


Spec/Auth.swift
```diff
@@ -715,6 +715,32 @@ class Auth : QuickSpec {
             }
         }
 
+        // RSA8f3
+        pending("ensure the message published with a wildcard '*' does not have a clientId") {
+            let token = getTestToken()
+            let options = ARTClientOptions(token: token)
+            options.environment = "sandbox"
+            options.clientId = "john"
```
mattheworiordan: See PR for RSA8f4, this is wrong, you need a token with a wildcard clientId and clientId is not set in the options.

Spec/Auth.swift
```diff
@@ -715,6 +715,32 @@ class Auth : QuickSpec {
...
+            let channel = rest.channels.get("test")
+
+            waitUntil(timeout: testTimeout) { done in
+                channel.publish([ARTMessage(name: nil, data: "no client")]) { error in
+                    expect(error).to(beNil())
+                    switch extractBodyAsJSON(mockExecutor.requests.first) {
+                    case .Failure(let error):
+                        fail(error)
+                    case .Success(let httpBody):
+                        expect(httpBody.unbox["clientId"]).to(beNil())
```
mattheworiordan: As with RSA8f4, we need to check this also via a history API request

Spec/Auth.swift
ricardopereira: Done 5092393.

Spec/Auth.swift
ricardopereira: Done 5092393.

Spec/Auth.swift
```diff
@@ -845,6 +845,37 @@ class Auth : QuickSpec {
...
+                    }
+                    channel.history { page, error in
+                        expect(error).to(beNil())
+                        expect(page!.items).to(haveCount(1))
+                        expect((page!.items[0] as! ARTMessage).clientId).to(beNil())
+                        done()
+                    }
+                }
+            }
+            expect(rest.auth.clientId).to(beNil())
```
tcard: RSA8f3 says: "Check that Auth#clientId is a string with value '*'"

ricardopereira: @tcard Then this is wrong https://github.com/ably/ably-ios/blob/master/Source/ARTAuth.m#L346? I saw [RSA12](http://docs.ably.io/client-lib-development-guide/features/#RSA12) and it should be nil.  The nil _value indicates that the client has not yet been able to confirm its identity_. I think RSA8f3 spec is wrong. WDYT?

tcard: The code is wrong, yes. I don't really think there's a conflict between RSA8f3 and RSA12. RSA12a has as precondition "The clientId attribute of a TokenRequest or TokenDetails used for authentication is null, or ConnectionDetails#clientId is null following a connection to Ably", which doesn't apply here.

ricardopereira: Done cf241cd. (not passing)

mattheworiordan:
@ricardopereira this was merged without https://github.com/ably/ably-ios/commit/cf241cd380f02d44a8cc5c50cef76099c0c8d20f?  I think that was a mistake

ricardopereira:
You're right. Thanks.
#402 created.

PR #351, opened 2016-03-22
RSA8f4

ricardopereira:


Spec/Auth.swift
```diff
@@ -713,6 +713,33 @@ class Auth : QuickSpec {
                 }
 
             }
+
+            // RSA8f4
+            pending("ensure the message published with a wildcard '*' has the provided clientId") {
+                let token = getTestToken()
+                let options = ARTClientOptions(token: token)
+                options.environment = "sandbox"
+                options.clientId = "john"
```
mattheworiordan: Why are you setting the `clientId`?

Spec/Auth.swift
```diff
@@ -713,6 +713,33 @@ class Auth : QuickSpec {
                 }
 
             }
+
+            // RSA8f4
+            pending("ensure the message published with a wildcard '*' has the provided clientId") {
+                let token = getTestToken()
+                let options = ARTClientOptions(token: token)
+                options.environment = "sandbox"
```
mattheworiordan: Is this necessary, can this not be a default in test?

Spec/Auth.swift
```diff
@@ -713,6 +713,33 @@ class Auth : QuickSpec {
                 }
 
             }
+
+            // RSA8f4
+            pending("ensure the message published with a wildcard '*' has the provided clientId") {
+                let token = getTestToken()
+                let options = ARTClientOptions(token: token)
```
mattheworiordan: The token is supposed to have a wildcard clientId yet that is not configured?

Spec/Auth.swift
```diff
@@ -713,6 +713,33 @@ class Auth : QuickSpec {
...
+                let channel = rest.channels.get("test")
+
+                waitUntil(timeout: testTimeout) { done in
+                    channel.publish([ARTMessage(name: nil, data: "no client", clientId: "john")]) { error in
+                        expect(error).to(beNil())
+                        switch extractBodyAsJSON(mockExecutor.requests.first) {
+                        case .Failure(let error):
+                            fail(error)
+                        case .Success(let httpBody):
+                            expect(httpBody.unbox["clientId"] as? String).to(equal("john"))
```
mattheworiordan: I think you should just check that the message has the correct clientId when retrieved from history

ricardopereira: Done 8f1da6a.

tcard:
LGTM

Issue #126, opened 2016-01-13
Core: `ARTErrorInfo` should be related with `NSError`


ricardopereira:
**Problem:**
To many inconsistencies on how to encapsulate an error in the client.

**e.g.** two different callback types where one uses `NSError` and another uses `ARTErrorInfo`:
``` swift
public typealias ARTStatsCallback = (ARTPaginatedResult?, NSError?) -> Void

public typealias ARTRealtimeChannelMessageCb = (ARTMessage, ARTErrorInfo?) -> Void
```

**Two solutions:**
 1. `ARTErrorInfo` should have `NSError` has base class
 2. Remove `ARTErrorInfo` and just use `NSError`

[`NSError`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSError_Class/) is the error object from Cocoa.
It has:
 - `code`: Int
 - `domain`: String
 - `userInfo`: Dictionary (user info dictionary)

IMO
we can do _2._ because the `NSError` has a `userInfo` property where we can add more data if we need.

mattheworiordan:
I am not sure, I prefer ARTErrorInfo inheriting from NSError, @tcard I need your view on this please

tcard:
I guess ARTErrorInfo could inherit from NSError.

But the real problem I see here is that, according to the docs, the REST library [uses anonymous objects with a defined structure as errors](http://localhost:4000/rest/statistics/#stats) while the Realtime API uses a named type [ErrorInfo](http://localhost:4000/realtime/types/#error-info) with just the same structure. I don't see why we shouldn't have the REST library use ErrorInfo as well. I see e. g. ably-java and ably-js does just this, but it is not specified in the docs.

In that case, ARTStatsCallback would be `(ARTPaginatedResult?, ARTErrorInfo?) -> Void`.

tcard:
To be clear, ARTErrorInfo could inherit NSError and use its userInfo property under the hood to store the message, but we'd get no gain for this except implementing Obj-C's standard error interface, which our users could find helpful in their own code. But it's a secondary thing.

mattheworiordan:
Agreed we should be consistent between REST and Realtime in regards to Error Types.  I like the idea of our own custom error type, we've erred towards that generally and mostly inherited from another compatible standard error type.

PR #353, opened 2016-03-24
RSA15c

ricardopereira:


mattheworiordan:
LGTM

tcard:
PTAL

Spec/Auth.swift
```diff
@@ -693,17 +692,17 @@ class Auth : QuickSpec {
 
                 it("overriding defaults") {
                     // Custom values
-                    let expectedClientId = "token_client"
                     let expectedTtl = 4800.0
                     let expectedCapability = "{\"canpublish:*\":[\"publish\"]}"
 
-                    let tokenParams = ARTTokenParams(clientId: expectedClientId)
+                    let tokenParams = ARTTokenParams(clientId: currentCliend)
```
ricardopereira: I don't understand the `currentCliend`. Where does it come from?

ricardopereira:
Beside a small comment, LGTM.

ricardopereira:
2ff8945 will close #126. 👍

Spec/Auth.swift
tcard: It's the client ID passed to the `rest` instance used below. RSA15c says that the client ID can't change in a TokenRequest or TokenDetails from the one passed in the ClientOptions, so we can't expect it to be different.

ricardopereira: Oh, I see now. Sorry. Can you change it to `currentClient`?

tcard: Done at 52b87bb.

ricardopereira: Thanks.

PR #355, opened 2016-03-24
RSA12b

ricardopereira:


Spec/Auth.swift
```diff
@@ -441,13 +441,32 @@ class Auth : QuickSpec {
...
+                        options.autoConnect = false
+                        options.defaultTokenParams = ARTTokenParams(clientId: "tester")
+                        let realtime = ARTRealtime(options: options)
+                        defer { realtime.close() }
+                        expect(realtime.auth.clientId).to(beNil())
+
+                        waitUntil(timeout: testTimeout) { done in
+                            realtime.connection.once(.Connecting) { stateChange in
+                                expect(stateChange!.reason).to(beNil())
+                                expect(realtime.auth.clientId).to(beNil())
```
mattheworiordan: I am not sure this is right because the clientId is set as a token param.  What you should instead do is create a token string that is assigned to a clientId, and use that for auth, so the client lib then has no means to determine whether it is assigned a `clientId` until the Connected protocol message is sent back

ricardopereira: Done f379f16.

PR #356, opened 2016-03-24
RSA7b3

ricardopereira:


mattheworiordan:
LGTM

tcard:
LGTM

Spec/Auth.swift
```diff
@@ -491,7 +491,27 @@ class Auth : QuickSpec {
                     }
                     
                     // RSA7b3
-                    // TODO: Realtime.connection
+                    it("should CONNECTED ProtocolMessages contain a clientId") {
+                        let options = AblyTests.commonAppSetup()
+                        options.clientId = "john"
```
mattheworiordan: This is not a great test because it appears the `clientId ` is set to `john` before the ProcotolMessage is received.  You should have a null clientId but use a token with an implicit `clientId` and then check that the client library later sets the `clientId` based solely on the `connectionDetails.clientId`

ricardopereira: Done #403.

PR #361, opened 2016-03-24
RSL1g3

ricardopereira:


mattheworiordan:
Bit like my comments on the other, is the client library responsible for the rejection?

ricardopereira:
> Bit like my comments on the other, is the client library responsible for the rejection?

Done ca84933.

ricardopereira:
@tcard PTAL

tcard:
LGTM

PR #362, opened 2016-03-24
RSL1g4

ricardopereira:


mattheworiordan:
The only issue I can see is that there is no way to be sure that the test is not rejecting the message because the server is rejecting it, or as intended by the spec, the client lib is rejecting it.  Can you somehow ensure that the message is never sent to the server i.e. rejected before that happens?

ricardopereira:
@mattheworiordan @tcard PTAL.

Spec/RestClientChannel.swift
```diff
@@ -155,6 +155,54 @@ class RestClientChannel: QuickSpec {
...
+                    client.httpExecutor = mockExecutor
+                    let channel = client.channels.get("test")
+
+                    mockExecutor.beforeRequest = { request in
+                        if request.URL!.absoluteString.containsString("/channels/test/messages") {
+                            switch extractBodyAsMessages(request) {
+                            case .Failure(let error):
+                                fail(error)
+                            case .Success(let rawMessages):
+                                expect(rawMessages.unbox.first!["clientId"] as? String).to(equal("tester"))
```
tcard: I'm not sure I get this. The spec says that the library should catch the error, so the request this block is intercepting shouldn't happen, isn't that right?

tcard: To clarify, the library should catch the error _before_ sending the request, thus failing early and preventing it from happening.

ricardopereira: Yes, you right. The request shouldn't happen. Fixed 436b265.

ricardopereira:
@tcard PTAL

tcard:
LGTM, but isn't this doing exactly the same as #361 ?

ricardopereira:
Quite but RSL1g3 is checking if the connection and channel remains active.

PR #368, opened 2016-03-30
RSP4b1

ricardopereira:


Spec/RestClientPresence.swift
```diff
@@ -77,6 +77,83 @@ class RestClientPresence: QuickSpec {
...
+                    it("start must be equal to or less than end and is unaffected by the request direction") {
+                        let client = ARTRest(options: AblyTests.commonAppSetup())
+                        let channel = client.channels.get("test")
+
+                        let query = ARTDataQuery()
+                        query.direction = .Backwards
+                        query.end = NSDate()
+                        query.start = NSDate()
+
+                        expect { try channel.presence.history(query) { _, _ in } }.to(throwError { (error: ErrorType) in
```
mattheworiordan: Why are we expecting an error?  As far as I can tell, start should be equal to or less than end, well it's hard to tell given the way you've set that up without explicitly setting start as later then end

ricardopereira: Done 917182b.

tcard:
LGTM

PR #372, opened 2016-04-01
RTP8a, RTL7f: attach first to avoid race condition.

tcard:
Before we were subscribing the presence listener and entering in
parallel connections, it was possible that the latter finished
before the former attached, thus missing the event.


tcard:
Note that all "push" Travis builds (including #311 right now) can fail because of this.

ricardopereira:
LGTM

Issue #345, opened 2016-03-22
Add support for Swift 2.2


ricardopereira:
_Xcode 7.3 includes Swift 2.2 and SDKs for iOS 9.3..._

ricardopereira:
This is important right now. The Xcode 7.3 is already on the AppStore. It won't be possible to use it if someone want to try out the latest client version. Swift syntax has changed but we can use different versions on the code. Swift 2.2 introduces a new compiler directive that makes cross-version compatibility. It is a compile time feature:

```
#if swift(>=3.0)
print("Running Swift 3.0 or later")
#else
print("Running Swift 2.2 or earlier")
#endif
```

tcard:
Sorry @ricardopereira, are you saying that the library right now doesn't work with Swift 2.2?

ricardopereira:
Yes because we are using the SwiftWebSocket. I think we only need to update the dependency because CocoaPods will only install the dependency and select the ObjC code of the client. We can update the `Spec` to Swift 2.2 later.

ricardopereira:
I'm downloading the Xcode 7.3 and I will test it.

PR #373, opened 2016-04-05
Swift 2.2

ricardopereira:
#345


tcard:
LGTM

ricardopereira:
I can create a new version on CocoaPods but maybe I should do it after the current PRs are merged.
@tcard WDYT?

tcard:
@ricardopereira I talked with @mattheworiordan and we should do the 0.8 release ASAP, but I want to get at least #347 in as it has a significant API change. The rest could be included in 0.8.1 eventually.

PR #376, opened 2016-04-08
Rearrange types relationships to avoid ambiguities.

tcard:
Previously, we had some situations in which, for convenience, we were
inheriting rather than composing (RealtimeChannel inheriting from
RestChannel). This was causing problems in places where some parent
class implemented `method:(A)a` while a child class was implementing
`method:(B)b`. Swift 2.2 doesn't let you do this, so it's necessary for
that, but it's also a good idea for Swift <2.2 if only for code
organization and soundness.

Also added some changes in the test Swift code for forwards 
compatibility.


ricardopereira:
Great 👌🏻 LGTM

mattheworiordan:
LGTM.  Can you confirm this does not impact on our documentation?

PR #377, opened 2016-04-08
Remove Ably.podspec.

tcard:
We can just rename AblyRealtime.podspec to Ably.podspec when the
time comes.


ricardopereira:
LGTM

mattheworiordan:
Is that time now now @tcard, I believe we can post to Ably now?  If so, we need to deprecate AblyRealtime pod

PR #378, opened 2016-04-08
Adapt to Swift 2.2

tcard:
WIP to check via Travis if it's backwards compatible. If it isn't we need to figure out how/if we're going to maintain two versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode7.2
+osx_image: xcode7.3
 env:
   global:
   - LC_CTYPE=en_US.UTF-8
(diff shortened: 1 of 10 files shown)


ricardopereira:
You need to change the travis file to use Xcode 7.3.

tcard:
Please don't merge anything before merging this. I'm making a `0.8.1-swift2.1` release with the current master branch (which is _finally_ passing the tests). Then a new branch, `swift2.1`, from the current master as well. Then we can merge this PR, and make a `0.8.1` release that is supposed to be compatible with Swift 2.2 onwards. From then on, every PR merged to `master` will have to be merged to `swift2.1` as well. (We'll use squashed commits and cherry-picking to facilitate this.)

Does the plan sound good?

PR #381, opened 2016-04-11
RSC15b

ricardopereira:


tcard:
LGTM

Spec/RestClient.swift
```diff
@@ -315,124 +315,176 @@ class RestClient: QuickSpec {
...
-                                expect(rest.auth.tokenDetails!.token).toNot(equal(currentTokenDetails.token))
-                                done()
-                            }
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.publish(nil, data: "nil") { _ in
+                            done()
                         }
                     }
+
+                    expect(testHTTPExecutor.requests).to(haveCount(2))
```
mattheworiordan: Why do we expect only two requests?  It should try all the fallback hosts surely, well at least up until `httpMaxRetryCount`

PR #385, opened 2016-04-13
RTN17b

ricardopereira:


tcard:
PTAL.

I had to switch from SwiftWebSocket to PocketSocket to be able to inspect which kind of error the websocket connection is failing with. SwiftWebSocket uses a Swift enum for this, which can't be used from Objective-C.

Originally I tried to add some Swift to our Objective-C codebase to do the bridging, but Xcode gave me a extremely hard time with this, so I gave up. Forking SwiftWebSocket to add Objective-C support is another possibility, but that would've required fighting against Swift's way of doing things from inside a Swift project, so I decided to just use a Objective-C library.

ricardopereira:
LGTM but it's not passing.

tcard:
@ricardopereira Not yet. The switch to PocketSocket unveiled a few things that are wrong that were working basically by chance. We need some of the changes in https://github.com/ably/docs/pull/112 for this, will do that first.

mattheworiordan:
What's the status on this?

tcard:
@mattheworiordan I'm about to try to reproduce why the Travis build stalls with this.

tcard:
Sitrep: this might be the worst case of Works On My Machine™ I've encountered so far. The whole test suite works on my machine. I've already implemented #387, #388, #389 and #390 on top of this and they also work. I've implemented MessagePack encoding on top of this: also works. So the only thing missing now is pretty much to have this stop crashing on Travis.

Since there doesn't seem to be a way of looking into stdout from Travis, I've tried to get some insight into what's going on by capturing every single line of logging, sending it in a HTTP request to a server, and writing them there. But I can't spot anything wrong from the logs; it just seems to stall at some random point.

I'll try to add some more external monitoring. Any idea would be appreciated.

mattheworiordan:
Is this since PocketRocket change?

ricardopereira:
@tcard Change the `.travis.xml` and run the test suite with standard `xcodebuild` and see what happens on the raw log.

ricardopereira:
Something like: `xcodebuild -workspace './Ably.xcworkspace' -scheme 'Ably' -destination 'platform=iOS Simulator,OS=latest,name=iPhone 5s' build test`

tcard:
@mattheworiordan Yes, although it was still happening with SocketRocket. I need one or the other to implement this spec line, as explained above.

Now it works. There's only one test failing, unrelated to this PR (I'll fix it separately). I don't know what changed, but it's passed consistently a few times now. Maybe PocketSocket was broken, and since the switch to SocketRocket it was just a coincidence with some other condition? Anyway, it's so annoying that it's so hard to get better diagnostics.

Please confirm I can merge this.

mattheworiordan:
> Please confirm I can merge this.

👍

PR #392, opened 2016-04-14
RTN16b

ricardopereira:


tcard:
LGTM

tcard:
@ricardopereira Note how `testAttachedChannelFails` fails from time to time with a timeout. I believe that's because it's the first tests that uses the test app, so it's the one issuing the slow POST /apps request. I still think we should increase the timeout at least for the old test suite.

ricardopereira:
@tcard Ok, I agree but we should only increase the timeout for the [app request](https://github.com/ably/ably-ios/blob/c2904d8c836b68d681fc05e27366d7d64b8a93c4/Tests/ARTTestUtil.m#L308) and leave the `testTimeout` used in our tests.

Spec/RealtimeClientConnection.swift
```diff
@@ -1735,6 +1735,24 @@ class RealtimeClientConnection: QuickSpec {
...
+            context("Connection recovery") {
+
+                // RTN16b
+                it("Connection#recoveryKey should be composed with the connection key and latest serial received") {
+                    let options = AblyTests.commonAppSetup()
+                    let client = ARTRealtime(options: options)
+                    defer { client.close() }
+                    waitUntil(timeout: testTimeout) { done in
+                        client.connection.once(.Connected) { _ in
+                            expect(client.connection.recoveryKey).to(equal("\(client.connection.key!):\(client.connection.serial)"))
```
mattheworiordan: Could we send one message after this to make sure the serial increases?

Spec/RealtimeClientConnection.swift
```diff
@@ -1894,6 +1894,25 @@ class RealtimeClientConnection: QuickSpec {
...
+                    defer { client.close() }
+                    let channel = client.channels.get("test")
+                    waitUntil(timeout: testTimeout) { done in
+                        client.connection.once(.Connected) { _ in
+                            expect(client.connection.serial).to(equal(-1))
+                            expect(client.connection.recoveryKey).to(equal("\(client.connection.key!):\(client.connection.serial)"))
+                        }
+                        channel.publish(nil, data: "message") { error in
+                            expect(error).to(beNil())
+                            expect(client.connection.serial).to(equal(0))
```
ricardopereira: Done. (I rebase it. It was not intended. Sorry)

Spec/RealtimeClientConnection.swift
ricardopereira: https://github.com/ably/ably-ios/pull/392#discussion_r61413476

PR #394, opened 2016-04-14
RTN16d

ricardopereira:


tcard:
LGTM

Spec/RealtimeClientConnection.swift
```diff
@@ -1735,6 +1735,39 @@ class RealtimeClientConnection: QuickSpec {
...
+                    let clientRecover = AblyTests.newRealtime(options)
+                    defer { clientRecover.close() }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        clientRecover.connection.once(.Connected) { _ in
+                            let transport = clientRecover.transport as! TestProxyTransport
+                            let firstConnectionDetails = transport.protocolMessagesReceived.filter{ $0.action == .Connected }.first!.connectionDetails
+                            expect(firstConnectionDetails!.connectionKey).toNot(beNil())
+                            expect(clientRecover.connection.id).to(equal(expectedConnectionId))
+                            expect(clientRecover.connection.key).to(equal(firstConnectionDetails!.connectionKey))
```
mattheworiordan: I am not sure this is right, the connection key can change when the connection is recovered, the only guarantee is that the connection ID remains the same

mattheworiordan:
Please review as I am not sure this is correct @tcard @ricardopereira

Spec/RealtimeClientConnection.swift
tcard: This is right. It's checking that the connection key is the _new_ one received in the CONNECTED message from the server (`firstConnectionDetails`), not that it's keeping the old connection key. (Maybe the name `firstConnectionDetails` is a bit misleading?)

ricardopereira: Yes, it should be something like `recoveredConnectionDetails` :/

PR #397, opened 2016-04-15
Increase test timeouts.

tcard:
Sadly, 10 seconds seems to be not enough given our current sandbox
performance. See some passing tests almost reaching the 10 s timeout:

```
[17:38:25]: ▸ ✓ RestClient__should_use_Auth_to_manage_authentication (9.815 seconds)
[17:38:34]: ▸ ✓ RestClient__should_request_another_token_after_current_one_is_no_longer_valid (9.421 seconds)
[17:38:39]: ▸ ✓ RestClient__initializer__should_accept_an_API_key (4.559 seconds)
```

Travis can be slower than that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,4 @@ script:
   # Use `travis_wait` when a long running command or compile step regularly takes longer than 10 minutes without producing any output.
   # It writes a short line to the build log every minute for 20 minutes, extending the amount of time your command has to finish.
   # Prefix `travis_wait` with a greater number to extend the wait time.
-  - travis_wait 30 scan --scheme "Ably" --open_report false
+  - travis_wait 35 scan --scheme "Ably" --open_report false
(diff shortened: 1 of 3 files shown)


mattheworiordan:
This is worrying as it should not happen even in sandbox.  However, if this increases stability of the iOS tests then 👍

ricardopereira:
I run those tests on my machine and I've got similar results:

```
'-[RestClient__should_use_Auth_to_manage_authentication]' started.
'-[RestClient__should_use_Auth_to_manage_authentication]' passed (4.191 seconds).
'-[RestClient__should_request_another_token_after_current_one_is_no_longer_valid]' started.
'-[RestClient__should_request_another_token_after_current_one_is_no_longer_valid]' passed (8.902 seconds).
'-[RestClient__initializer__should_accept_an_API_key]' started.
'-[RestClient__initializer__should_accept_an_API_key]' passed (4.552 seconds).
```

I think there's nothing wrong with Travis and the timeout of 10 seconds is fine because we use it for asynchronous expectations (doesn't mean the test has to end in 10 seconds).

PR #402, opened 2016-04-18
Fix RSA8f3

ricardopereira:


tcard:
PTAL. I'm a bit unsure about this change as I've removed this explicit check from `getClientId`:

```objective-c
// Check wildcard
if ([self.tokenDetails.clientId isEqual:@"*"])
    // Any client
    return nil;
```

The commit that introduced that change (dd2baf01) mentions RSA15, but I don't see anything in [RSA15](http://docs.ably.io/client-lib-development-guide/features/#RSA15) to justify that behavior, and the whole Auth test suite is passing after removing it.

ricardopereira:
Yes, you're right. I think it's good.

PR #406, opened 2016-04-20
RTL6c

ricardopereira:


mattheworiordan:
LGTM but will be revisited following @tcard's updates to the spe

tcard:
PTAL

tcard:
(By the way, that's up to date with ably/docs#112.)

Spec/RealtimeClientChannel.swift
```diff
@@ -1110,6 +1110,208 @@ class RealtimeClientChannel: QuickSpec {
...
+                            channel.publish(nil, data: "message") { error in
+                                expect(error).to(beNil())
+                                done()
+                            }
+                            expect((client.transport as! TestProxyTransport).protocolMessagesSent.filter({ $0.action == .Message })).to(haveCount(1))
+                        }
+                    }
+
+                    // RTL6c2
+                    context("the message should be queued and delivered as soon as the connection state returns to CONNECTED if the connection is") {
```
mattheworiordan: The tests in this context test `the message should be queued`, but don't test `and delivered as soon as the connection state returns to CONNECTED`.  Should we?

mattheworiordan:
One comment but otherwise great 👍

Spec/RealtimeClientChannel.swift
tcard: It is tested. All of them are calling a local `publish` function (line 1150) which publishes and waits for the callback. Or do you want to also subscribe to the channel and check if it's received? I don't think that's necessary, as other tests already prove that the publish callback is called upon successful delivery.

mattheworiordan: Yes,  you are right, leave as is

PR #410, opened 2016-04-20
RTL12

ricardopereira:


Spec/RealtimeClientChannel.swift
```diff
@@ -1636,6 +1636,32 @@ class RealtimeClientChannel: QuickSpec {
                     }
                 }
 
+                // RTL12
+                pending("attached channel may receive an additional ATTACHED ProtocolMessage") {
```
mattheworiordan: I would like to see an additional channel attach message WITHOUT an error because that is what the spec is largely about, it should have no effect on the lib if it receives more than one, so long as it is without an error

ricardopereira: Done 55feb77.

tcard:
PTAL

ricardopereira:
👍

PR #415, opened 2016-04-21
RSC12

ricardopereira:


Spec/RestClient.swift
```diff
@@ -200,6 +200,14 @@ class RestClient: QuickSpec {
 
                 expect(authOptions == options).to(beTrue())
             }
+
+            // RSC12
+            it("REST endpoint host should be configurable in the Client constructor with the option restHost") {
+                let options = ARTClientOptions()
+                expect(options.restHost).to(equal("rest.ably.io"))
+                options.restHost = "rest.ably.test"
+                expect(options.restHost).to(equal("rest.ably.test"))
```
mattheworiordan: This is not testing that the client uses the host but simply tests that the options is editable. You need to instance a library with a custom host and check that it's used

ricardopereira: Done 110d97f.

tcard:
LGTM

PR #416, opened 2016-04-21
RSC13

ricardopereira:


mattheworiordan:
Looks good, but I'm not sure if that will work when the tests are no longer pending due to assumptions on timings

tcard:
PTAL

ricardopereira:
👍

PR #420, opened 2016-04-21
Test suites: reuse test app.

tcard:
For this, we isolate tests by prepending a prefix to channel names in
different tests (ie. that use different calls to 
TestUtilties.setupOptions.)

This should make #397 unnecessary.

### DIFF
--- a/Spec/TestUtilities.swift
+++ b/Spec/TestUtilities.swift
@@ -78,35 +78,46 @@ class AblyTests {
         }
     }
 
-    class func setupOptions(options: ARTClientOptions, debug: Bool = false) -> ARTClientOptions {
-        let request = NSMutableURLRequest(URL: NSURL(string: "https://\(options.restHost):\(options.tlsPort)/apps")!)
-        request.HTTPMethod = "POST"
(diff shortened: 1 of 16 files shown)


Spec/TestUtilities.swift
```diff
@@ -78,35 +78,46 @@ class AblyTests {
...
+                XCTFail(error.localizedDescription)
+                return options
+            }
+
+            testApplication = JSON(data: responseData!)
             
             if debug {
                 options.logLevel = .Verbose
-                print(response)
+                print(testApplication!)
```
ricardopereira: Leftover

ricardopereira:
LGTM 😃 perfect... tests in 10min!

Spec/TestUtilities.swift
ricardopereira: Ups, nevermind. Didn't saw the `debug`.

tcard: Well, it is printing exactly the same as it was doing before. Note the `if debug {`.

mattheworiordan:
Now that is what I like, a 3x saving in time!

PR #430, opened 2016-04-26
RTP15c

ricardopereira:


Spec/RealtimeClientPresence.swift
```diff
@@ -1283,6 +1283,32 @@ class RealtimeClientPresence: QuickSpec {
...
+                        channel.presence.updateClient("john", data:"mobile") { error in
+                            expect(error).to(beNil())
+                            done()
+                        }
+                    }
+                }
+
+                waitUntil(timeout: testTimeout) { done in
+                    channel.presence.get() { members, error in
+                        expect(members!.first!.data as? String).to(equal("mobile"))
```
mattheworiordan: Why are you expecting `mobile` first, ordering of the member set is not guaranteed.  Also, the tests says that you you should make sure it has no side effects, so you should check that both members exist

ricardopereira: But the test has only one member. I will change it to use two clientId's to be more complete.

ricardopereira: Done 08a0a3e.

tcard:
LGTM

PR #433, opened 2016-04-26
RSA7a4

ricardopereira:


Spec/Auth.swift
```diff
@@ -440,6 +440,27 @@ class Auth : QuickSpec {
...
+                it("ClientOptions#clientId takes precendence when a clientId value is provided in both ClientOptions#clientId and ClientOptions#defaultTokenParams") {
+                    let options = AblyTests.commonAppSetup()
+                    options.clientId = "john"
+                    options.defaultTokenParams = ARTTokenParams(clientId: "tester")
+                    let client = ARTRest(options: options)
+                    let channel = client.channels.get("test")
+
+                    expect(client.auth.clientId).to(equal("john"))
+                    waitUntil(timeout: testTimeout) { done in
+                        channel.publish(nil, data: "message") { error in
```
tcard: This doesn't prove that it's using the right client ID for Auth operations. I think you should at least remove the key from `options` and set the `authCallback`, and check that the client ID passed to the `authCallback` is the right one.

ricardopereira: Done e8f618e.

tcard:
LGTM

PR #434, opened 2016-04-26
RSA7b4

ricardopereira:


Spec/Auth.swift
```diff
@@ -558,6 +558,18 @@ class Auth : QuickSpec {
                         }
                     }
 
+                    // RSA7b4
+                    it("client does not have an identity when a wildcard string '*' is present") {
+                        let realtime = ARTRealtime(options: AblyTests.commonAppSetup())
```
mattheworiordan: This is not clear, we need to demonstrate `When a wildcard string '*' is present in the TokenRequest, TokenDetails, or ProtocolMessage#connectionDetails object` in the test.  So whilst it may in fact be in the `connectionDetails`, I would prefer to be more explicit here and use a token with a wildcard

ricardopereira: Done 1b98664.

Spec/Auth.swift
```diff
@@ -558,6 +558,20 @@ class Auth : QuickSpec {
...
 
+                    // RSA7b4
+                    it("client does not have an identity when a wildcard string '*' is present") {
+                        let options = AblyTests.clientOptions()
+                        options.token = getTestToken(clientId: "*")
+                        let realtime = ARTRealtime(options: options)
+                        defer { realtime.close() }
+                        waitUntil(timeout: testTimeout) { done in
+                            realtime.connection.on(.Connected) { _ in
+                                expect(realtime.auth.clientId).to(equal("*"))
```
tcard: Can you check that the `ProtocolMessage.connectionDetails.clientId` has a `*`?

tcard: Well, forget that, it's the only way `realtime.auth.clientId` can be `*` here, really.

tcard:
LGTM

PR #436, opened 2016-04-26
RTN13b

ricardopereira:


Spec/RealtimeClientConnection.swift
```diff
@@ -1169,6 +1169,32 @@ class RealtimeClientConnection: QuickSpec {
...
+                        client.connection.on(.Connected) { _ in
+                            client.onError(AblyTests.newErrorProtocolMessage())
+                            done()
+                        }
+                    }
+                    waitUntil(timeout: testTimeout) { done in
+                        client.ping() { error in
+                            expect(error).toNot(beNil())
+                            let transport = client.transport as! TestProxyTransport
+                            expect(transport.protocolMessagesSent.filter{ $0.action == .Heartbeat }).to(haveCount(1))
```
mattheworiordan: Why is this one when the spec says it should immediately indicate an error, if the connection is `CLOSED` why are we sending a message?

ricardopereira: Done 1218765.

tcard:
PTAL. RTN13b is expanded by ably/docs#112, and there was a test for it introduced in a parallel PR. I've removed your test (sorry for the confusion) and changed the behavior of `ping` to pass the error to the callback instead of raising an exception.

ricardopereira:
👍

PR #438, opened 2016-04-26
RTN15f

ricardopereira:


Spec/RealtimeClientConnection.swift
```diff
@@ -1660,6 +1660,34 @@ class RealtimeClientConnection: QuickSpec {
...
+                it("ACK and NACK responses for published messages can only ever be received on the transport connection on which those messages were sent") {
+                    let options = AblyTests.commonAppSetup()
+                    options.disconnectedRetryTimeout = 0.5
+                    let client = AblyTests.newRealtime(options)
+                    defer { client.close() }
+                    let channel = client.channels.get("test")
+                    var resumed = false
+                    waitUntil(timeout: testTimeout) { done in
+                        client.connection.once(.Connected) { _ in
+                            channel.publish(nil, data: "message") { _ in
```
mattheworiordan: I am not sure I follow hot his method is called twice, can you explain?

ricardopereira: Is called twice? Can you clarify? The `publish` only receives the `ACK` (callback is called) after the connection has been resumed.

ricardopereira: Maybe I'm wrong. @tcard do you have any suggestion for this one?

tcard: I don't understand @mattheworiordan 's point either. This looks good to me. The message is sent in one transport, which is immediately dropped. The fact that publish's callback is called _after_ the connection is resumed proves that the library resent the message in the new transport.

mattheworiordan: Sorry, I misunderstood the code.  However, as the spec states that the message is resent on the new transport, can we also include that check so that this test code explains what is happening i.e. it's not that on a new connection the ACK is sent, it's actually an ACK that is received from the message being resent.  Do you agree?

ricardopereira: Done 5c3487a.

tcard:
LGTM (by the way).

PR #439, opened 2016-04-27
RTN20a

ricardopereira:


tcard:
As-is, I don't think this test is providing anything. I'd rather leave this functionality out for now, and just have it marked in the spec spreadsheet instead of having this test lingering around.

tcard:
Actually, scratch that, I doesn't seem hard to just do it now.

PR #441, opened 2016-04-27
Conform to spec changes at ably/docs#112.

tcard:
This is required by #385. Note that ably/docs#112 is not merged yet.

Split in several commits for ease of consumption.

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -139,7 +139,26 @@
 }
 
 - (void)close {
-    [self transition:ARTRealtimeClosing];
+    switch (self.connection.state) {
+    case ARTRealtimeInitialized:
+    case ARTRealtimeClosing:
+    case ARTRealtimeClosed:
+    case ARTRealtimeFailed:
+        return;
+    case ARTRealtimeConnecting: {
+        [_connection once:^(ARTConnectionStateChange *change) {
+            [self close];
+        }];
+        return;
+    }
+    case ARTRealtimeDisconnected:
+    case ARTRealtimeSuspended:
+        [self transition:ARTRealtimeClosed];
+        break;
+    case ARTRealtimeConnected:
+        [self transition:ARTRealtimeClosing];
+        break;
+    }
 }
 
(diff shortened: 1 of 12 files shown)


ricardopereira:
LGTM, now timeouts are way simple! Great one.

Source/ARTRealtime.m
```diff
@@ -158,20 +139,54 @@ - (void)connect {
...
+        return;
+    case ARTRealtimeConnecting:
+    case ARTRealtimeDisconnected: {
+        [_connection once:^(ARTConnectionStateChange *change) {
+            [self ping:cb];
+        }];
+        return;
+    }
+    case ARTRealtimeConnected:
+        [_pingEventEmitter timed:[_pingEventEmitter once:cb] deadline:10.0 onTimeout:^{
```
ricardopereira: 10s is the default timeout?

Source/ARTRealtime.m
```diff
@@ -238,15 +247,15 @@ - (void)transition:(ARTRealtimeConnectionState)state withErrorInfo:(ARTErrorInfo
...
 
             break;
-        case ARTRealtimeConnected:
-            [self cancelSuspendTimer];
-            break;
-        case ARTRealtimeClosing:
-            [self startCloseTimer];
+        }
+        case ARTRealtimeClosing: {
+            [self unlessStateChangesBefore:10.0 do:^{
```
ricardopereira: Same here.

Source/ARTRealtime.m
tcard: It was hardcoded like that before. Fixed at 0aa8cb6.

Source/ARTRealtime.m
tcard: Fixed at 0aa8cb6.

mattheworiordan:
FYI, I've reviewed https://github.com/ably/docs/pull/112 and had some comments. it is possible that we may need to revisit this after :(

tcard:
Of course, I counted on that @mattheworiordan, but some parts of this were blocking other things and the refactor was nice anyway. Changing things should be easy now.

PR #442, opened 2016-04-27
RTN15h

ricardopereira:


mattheworiordan:
LGTM

tcard:
I don't think this test is right. It should be like [RTN14b](https://github.com/ably/ably-ios/blob/bc22daea6906a675d54ab9a84f4c861e13056e55/Spec/RealtimeClientConnection.swift#L1379), whose spec is exactly the same except it's an ERROR ProtocolMessage instead of DISCONNECTED. I'll adapt it.

tcard:
PTAL

ricardopereira:
👍

PR #443, opened 2016-05-02
Several fixes in test suites.

tcard:
I've removed a couple of old failing tests that have more exhaustive,
more maintainable counterparts in the new suite.


mattheworiordan:
Thanks, LGTM.  Good to see some old Objective-C tests being removed as well

PR #444, opened 2016-05-02
Fix RTN20.

tcard:
I've tested the ARTOSReachability "live" manually (by switching on/off
the real Internet connection on my machine) and it works as expected,
but it needs #385 to properly handle transport errors.

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -278,6 +278,7 @@
             break;
         }
         case ARTRealtimeClosing: {
+            [_reachability off];
             [self unlessStateChangesBefore:[ARTDefault realtimeRequestTimeout] do:^{
                 [self transition:ARTRealtimeClosed];
             }];
@@ -285,6 +286,7 @@
             break;
         }
         case ARTRealtimeClosed:
+            [_reachability off];
             [self.transport close];
             self.transport.delegate = nil;
             _connection.key = nil;


Spec/RealtimeClientConnection.swift
```diff
@@ -359,20 +359,24 @@ class RealtimeClientConnection: QuickSpec {
...
-                            case .Failed:
-                                errorInfo = reason
-                                done()
-                            default:
-                                break
+                            print("CHANGE \(stateChange!.current)")
+                            if stateChange!.current == .Failed {
+                                print("FAILED")
                             }
+                            // let stateChange = stateChange!
```
mattheworiordan: Is this meant to be commented out?

Source/ARTOSReachability.m
```diff
@@ -0,0 +1,87 @@
...
+static void ARTOSReachability_Callback(SCNetworkReachabilityRef target, SCNetworkReachabilityFlags flags, void* info) {
+    [(ARTOSReachability *)ARTOSReachability_instances[[NSValue valueWithPointer:target]] internalCallback:flags & kSCNetworkReachabilityFlagsReachable];
+}
+
+- (void)listenForHost:(NSString *)host callback:(void (^)(BOOL))callback {
+    [self off];
+    _host = host;
+    _callback = callback;
+
+    _reachabilityRef = SCNetworkReachabilityCreateWithName(NULL, [host UTF8String]);
```
mattheworiordan: Out of interest, is this the only API available for connectivity checks?  In other client libraries the OS internet connection checks are not host specific.

mattheworiordan:
LGTM

Source/ARTOSReachability.m
tcard: It can be also done just for the connection. I thought why not make it more precise.

Spec/RealtimeClientConnection.swift
tcard: No, sorry, leftover. Fixed at 65bfc55.

Source/ARTOSReachability.m
mattheworiordan: Ok, sure, although of course it does add more complexity, but entirely up to you...

ricardopereira:
@mattheworiordan I rebased it with master.
`RTN20` tests are passing and they look good.
Can I force push? /cc @tcard

mattheworiordan:
Yes, please force push.  Are any other tests failing?

ricardopereira:
@mattheworiordan Yes, there is at least one test failing with something related this PR:

```
Test Case '-[AblySpec.Auth Token__token_auth_and_clientId__Incompatible_client__with_Realtime__it_should_change_the_connection_state_to_FAILED_and_emit_an_error]' started.
INFO: R:0x7fc6daa1ce80 ARTRealtime closed
INFO: Reachability: stopped listening for host (null)
ERROR: R:0x7fc6daa377c0 WS:0x7fc6daa38f80 ARTWebSocketTransport: token auth failed with ARTErrorInfo with code 401, message: incompatible credentials
EXC_BAD_ACCESS
```

mattheworiordan:
@ricardopereira well from the title of that test, it should raise an incompatible credentials error I believe, so it sounds like that is a test that genuinely needs to be fixed then.

ricardopereira:
@mattheworiordan The test is fine. I made a fix and it is running now.
I'm running all the test suite on my machine.

Source/ARTOSReachability.m
```diff
@@ -0,0 +1,93 @@
...
+__weak NSMutableDictionary *ARTOSReachability_instances;
+
+static void ARTOSReachability_Callback(SCNetworkReachabilityRef target, SCNetworkReachabilityFlags flags, void* info) {
+    [(ARTOSReachability *)ARTOSReachability_instances[[NSValue valueWithPointer:target]] internalCallback:flags & kSCNetworkReachabilityFlagsReachable];
+}
+
+- (void)listenForHost:(NSString *)host callback:(void (^)(BOOL))callback {
+    [self off];
+
+    if (!host) {
```
tcard: Why would `host` be nil here? If it's because the user provides a nil host, this validation should happen at a higher level, where the user provides it, and properly return an error to the user (probably as an exception). Internal code can just assume then that `host` is correct.

ricardopereira: @tcard It makes sense what you said but I think this check is important to help out when you're debugging. The `host` was nil because the `transport` was nil. Now, why was the `transport` nil? I don't know. I need to check the code.

ricardopereira: Just for the sake of clarification: the `transport` is nil because the client transitions to `FAILED`, see  https://github.com/ably/ably-ios/blob/4f1bd1098c54663ce20cf3dfb7b4595f4fbe4b40/Source/ARTRealtime.m#L302.

ricardopereira: @tcard I'm sure you faced something similar when you run the test suite on your machine.

The force cast of a `transport` like `realtime.transport as! TestProxyTransport` sometimes raises a bad access because the `realtime.transport` is created only when the connection starts CONNECTING, so we need to guarantee that the current connection has a valid transport instance before force casting it. The test suite will stall when a `EXC_BAD_ACCESS` occurs.

Then we have as well situations like this:
> Just for the sake of clarification: the `transport` is nil because the client transitions to `FAILED`, see  https://github.com/ably/ably-ios/blob/4f1bd1098c54663ce20cf3dfb7b4595f4fbe4b40/Source/ARTRealtime.m#L302.

What is your opinion on creating the `transport` once and release it only when the `RealtimeClient` deinitialise?

tcard: Well, I don't know how much that would affect things, but I suspect we would be fixing the wrong problem. If we're trying to use the transport when there is no transport, forcing a transport to be there even when there's no connection will probably just move the failure elsewhere.

Instead, things requiring a transport should make sure that there's indeed a transport present I think.

Why does the client transition to FAILED?

ricardopereira: @tcard The client transitions to FAILED because it is an incompatible client. That is what the spec is about. It checks if the error is well formed.

So, based on your answer and not thinking on this particular case, I think the best is to replace:

```
let transport = client.transport as! TestProxyTransport
```

to something like

```
guard let transport = client.transport as? TestProxyTransport else {
    fail("RealtimeTransport is invalid"); return
}
```
What do you think?

tcard: AFAICS that doesn't fix the underlying issue, just improves error reporting (which is good, but that's the kind of changes [I was talking about here](https://github.com/ably/ably-ios/pull/467#issuecomment-242024512) arguing that they're not critical right now.)

If we have a test in which the transport is unexpectedly nil, why not fix it so that the transport is not nil?

ricardopereira: @tcard Yes, you're right. The error reporting is another discussion.

The transport is unexpectedly nil because when the client connects, the `_transport` is nil but the instance is created and starts the connection, then the `_transport` emits the Failed state because of the incompatible client, the `ARTRealtimeFailed` is handled and the `_transport` is [cleared](https://github.com/ably/ably-ios/blob/4f1bd1098c54663ce20cf3dfb7b4595f4fbe4b40/Source/ARTRealtime.m#L302), then comes the `_reachability listenForHost:[_transport host]`  and `_transport` is nil.

ricardopereira: @tcard Now, when is the right moment to start listening for a host?

IMO, if the connections fails immediately then it doesn't makes sense to start listening for a host. So, if there is no transport at the moment when the client is setting up the OS reachability, then ignore it.

tcard: > then comes the _reachability listenForHost:[_transport host] and _transport is nil.

I guess you mean that `[_transport host]` is nil? `_transport` cannot be nil at this point since it's instantiated [right above](https://github.com/ably/ably-ios/pull/444/files/5c2436e3427bb653c71df7f655acf969c6be60fb#diff-d9df428f90fcfee8426c1e46435f952aR250).

Now, why is `[_transport host]` nil? That surely looks like a inconsistent state of `_transport`.

ricardopereira: @tcard No, I mean `_transport` is nil. The [connection](https://github.com/ably/ably-ios/pull/444/files/5c2436e3427bb653c71df7f655acf969c6be60fb#diff-d9df428f90fcfee8426c1e46435f952aR252) will emit a Failed state and set `_transport` to nil.

ricardopereira:
Legacy tests:

screen shot 2016-08-25 at 08 48 58

ricardopereira:
Specs:

```
Test Suite 'AblySpec.xctest' failed at 2016-08-25 10:57:51.589.
	 Executed 520 tests, with 6 failures (0 unexpected) in 417.708 (418.077) seconds
Test Suite 'All tests' failed at 2016-08-25 10:57:51.591.
	 Executed 520 tests, with 6 failures (0 unexpected) in 417.708 (418.080) seconds
```

Failed:
```
/Users/ricardopereira/Repositories/Whitesmith/ably-ios/Spec/RealtimeClientPresence.swift:907: error: -[AblySpec.RealtimeClientPresence Presence__enter__should_result_in_an_error_if_the_client_does_not_have_required_presence_permission] : expected to contain <Channel denied access based on given capability>, got <Invalid clientId for credentials>
Test Case '-[AblySpec.RealtimeClientPresence Presence__enter__should_result_in_an_error_if_the_client_does_not_have_required_presence_permission]' failed (0.894 seconds).

/Users/ricardopereira/Repositories/Whitesmith/ably-ios/Spec/RealtimeClientPresence.swift:980: error: -[AblySpec.RealtimeClientPresence Presence__update__should_result_in_an_error_if_the_client_does_not_have_required_presence_permission] : expected to contain <Channel denied access based on given capability>, got <Invalid clientId for credentials>
Test Case '-[AblySpec.RealtimeClientPresence Presence__update__should_result_in_an_error_if_the_client_does_not_have_required_presence_permission]' failed (0.859 seconds).

/Users/ricardopereira/Repositories/Whitesmith/ably-ios/Spec/RealtimeClientPresence.swift:1221: error: -[AblySpec.RealtimeClientPresence Presence__leave__should_result_in_an_error_if_the_client_does_not_have_required_presence_permission] : expected to contain <Channel denied access based on given capability>, got <Invalid clientId for credentials>
Test Case '-[AblySpec.RealtimeClientPresence Presence__leave__should_result_in_an_error_if_the_client_does_not_have_required_presence_permission]' failed (0.918 seconds).

/Users/ricardopereira/Repositories/Whitesmith/ably-ios/Spec/RealtimeClientPresence.swift:1885: error: -[AblySpec.RealtimeClientPresence Presence__history__should_return_a_PaginatedResult_page] : expected to all pass a condition, but failed first at element <<ARTPresenceMessage: 0x7fb114258ea0> {id: 
...
/Users/ricardopereira/Repositories/Whitesmith/ably-ios/Spec/RealtimeClientPresence.swift:1900: error: -[AblySpec.RealtimeClientPresence Presence__history__should_return_a_PaginatedResult_page] : expected to all pass a condition, but failed first at element <<ARTPresenceMessage: 0x7fb11417f610> {id: 
...
Test Case '-[AblySpec.RealtimeClientPresence Presence__history__should_return_a_PaginatedResult_page]' failed (2.681 seconds).

/Users/ricardopereira/Repositories/Whitesmith/ably-ios/Spec/RestClientPresence.swift:501: error: -[AblySpec.RestClientPresence Presence__presence_messages_retrieved_are_decoded_in_the_same_way_that_messages_are_decoded] : expected to equal <{test = 1;}>, got <nil> (use beNil() to match nils)
2016-08-25 10:56:47.359 xctest[77206:4299336] INFO: Reachability: stopped listening for host sandbox-realtime.ably.io
Test Case '-[AblySpec.RestClientPresence Presence__presence_messages_retrieved_are_decoded_in_the_same_way_that_messages_are_decoded]' failed (1.507 seconds).
```

ricardopereira:
@tcard @mattheworiordan Ready to be reviewed.

tcard:
LGTM

PR #447, opened 2016-05-05
Add MessagePack support.

tcard:
And make it the default.


tcard:
@mattheworiordan Please confirm I should merge this and #446, I should really make a new release.

mattheworiordan:
Looks good.  Shame we don't have more permutations of tests with / without the binary protocol, but I think this should prove it works sufficiently for now.  I was not entirely clear when reviewing, but do you have any tests using MsgPack over REST as well?

tcard:
@mattheworiordan All the tests that weren't JSON-specific before are now using MessagePack, because it's the new default. So yes, most REST tests use MessagePack.

PR #455, opened 2016-07-13
Ensure graceful handling of DETACH and DISCONNECT.

### LINKED ISSUE #454: Important: Ensure DETACHED or DISCONNECTED with error is non-fatal
See a bug discovered today at https://github.com/ably/ably-js/issues/312 for ably-js.

This is quite problematic because if we explicitly have to DETACH a channel (due to a timeout for example), then the client now will move the channel into the failed state.  This means most customers will never re-attach that channel automatically, see https://www.ably.io/documentation/realtime/connection#handling-failures.

With some urgency, we need to add test coverage to ensure that:

* A server initiated `DETACH` with or without an error moves the channel to the `DETACHED` state and not a failed state
* A server initiated `DISCONNECT` with or without an error moves the connection to the `DISCONNECTED` state and not a failed state.

tcard:
Fixes #454.

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -712,6 +712,7 @@
             // Event
             [self onConnected:message];
             break;
+        case ARTProtocolMessageDisconnect:
         case ARTProtocolMessageDisconnected:
             [self onDisconnected:message];
             break;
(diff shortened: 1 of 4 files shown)


mattheworiordan:
Nice @tcard thanks!

PR #456, opened 2016-07-26
RSC7b - Added 'X-Ably-Lib' header

eltiren:
Completed RSC7b - Added 'X-Ably-Lib' header


Source/ARTDefault.m
```diff
@@ -34,6 +35,10 @@ + (NSString *)version {
     return ARTDefault_version;
 }
 
++ (NSString *)libVersion {
+    return [@[ARTDefault_platform, ARTDefault_version] componentsJoinedByString:@"-"];
```
mattheworiordan: Thanks, but I don't think `ARTDefault_version` is what we are after, I believe that represents the API version, not the client library version. Perhaps introduce a new var and fix the old one for clarity.

PR #457, opened 2016-07-28
Rtn2g

eltiren:
Completed with notes from previous pullrequest

### DIFF
--- a/Source/ARTDefault.m
+++ b/Source/ARTDefault.m
@@ -13,6 +13,7 @@
 NSString *const ARTDefault_restHost = @"rest.ably.io";
 NSString *const ARTDefault_realtimeHost = @"realtime.ably.io";
 NSString *const ARTDefault_version = @"0.8";
+NSString *const ARTDefault_platform = @"ios";
 
 
 static int _realtimeRequestTimeout = 10.0;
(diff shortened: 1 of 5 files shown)


Spec/RealtimeClient.swift
```diff
@@ -36,6 +36,7 @@ class RealtimeClient: QuickSpec {
                         expect(error).to(beNil())
                         let transport = client.transport as! TestProxyTransport
                         expect(transport.lastUrl!.query).to(haveParam("v", withValue: "0.8"))
+                        expect(transport.lastUrl!.query).to(haveParam("lib", withValue: "ios-0.8"))
```
mattheworiordan: Why a test for realtime transport URL and not for the header over REST?

Source/ARTDefault.m
```diff
@@ -13,6 +13,7 @@ @implementation ARTDefault
 NSString *const ARTDefault_restHost = @"rest.ably.io";
 NSString *const ARTDefault_realtimeHost = @"realtime.ably.io";
 NSString *const ARTDefault_version = @"0.8";
+NSString *const ARTDefault_library_version = @"0.8";
```
mattheworiordan: This is not the version, see https://github.com/ably/ably-ios/releases

PR #459, opened 2016-08-03
Encode JSON message data as string.

tcard:
Previously, we were just inlining it:

    "messages": [{"data": {"foo": "bar"}}]

What the JS library expects, and really what [RSL4d3](http://docs.ably.io/client-lib-development-guide/features/#RSL4d3) says:

    "messages": [{"data": "{\"foo\": \"bar\"}"}]

So now we do that.

### DIFF
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -3066,30 +3066,41 @@ class RealtimeClientConnection: QuickSpec {
                 let msgpackOptions = AblyTests.commonAppSetup()
                 msgpackOptions.useBinaryProtocol = true
 
-                it("should send messages through JSON and retrieve equal messages through MsgPack") {
+                it("should send messages through raw JSON POST and retrieve equal messages through MsgPack and JSON") {
                     let restPublishClient = ARTRest(options: jsonOptions)
-                    let realtimeSubscribeClient = AblyTests.newRealtime(msgpackOptions)
-                    defer { realtimeSubscribeClient.close() }
+                    let realtimeSubscribeClientMsgPack = AblyTests.newRealtime(msgpackOptions)
+                    let realtimeSubscribeClientJSON = AblyTests.newRealtime(jsonOptions)
+                    defer {
+                        realtimeSubscribeClientMsgPack.close()
+                        realtimeSubscribeClientJSON.close()
+                    }
 
-                    let realtimeSubscribeChannel = realtimeSubscribeClient.channels.get("test-subscribe")
-                    realtimeSubscribeChannel.attach()
+                    let realtimeSubscribeChannelMsgPack = realtimeSubscribeClientMsgPack.channels.get("test-subscribe")
+                    let realtimeSubscribeChannelJSON = realtimeSubscribeClientJSON.channels.get(realtimeSubscribeChannelMsgPack.name)
 
-                    expect(realtimeSubscribeChannel.state).toEventually(equal(ARTRealtimeChannelState.Attached), timeout: testTimeout)
-                    if realtimeSubscribeChannel.state != .Attached {
-                        return
+                    waitUntil(timeout: testTimeout) { done in
+                        let partlyDone = AblyTests.splitDone(2, done: done)
+                        realtimeSubscribeChannelMsgPack.attach { _ in partlyDone() }
+                        realtimeSubscribeChannelJSON.attach { _ in partlyDone() }
                     }
 
                     for (_, fixtureMessage) in fixtures["messages"] {
-                        var receivedMessage: ARTMessage?
-
                         waitUntil(timeout: testTimeout) { done in
-                            realtimeSubscribeChannel.subscribe { message in
-                                realtimeSubscribeChannel.unsubscribe()
-                                receivedMessage = message
-                                done()
+                            let partlyDone = AblyTests.splitDone(2, done: done)
+
+                            realtimeSubscribeChannelMsgPack.subscribe { message in
+                                realtimeSubscribeChannelMsgPack.unsubscribe()
+                                expectDataToMatch(message, fixtureMessage)
+                                partlyDone()
                             }
 
-                            let request = NSMutableURLRequest(URL: NSURL(string: "/channels/\(realtimeSubscribeChannel.name)/messages")!)
+                            realtimeSubscribeChannelJSON.subscribe { message in
+                                realtimeSubscribeChannelJSON.unsubscribe()
+                                expectDataToMatch(message, fixtureMessage)
+                                partlyDone()
+                            }
+
+                            let request = NSMutableURLRequest(URL: NSURL(string: "/channels/\(realtimeSubscribeChannelMsgPack.name)/messages")!)
                             request.HTTPMethod = "POST"
                             request.HTTPBody = try! fixtureMessage.rawData()
                             request.allHTTPHeaderFields = [
@@ -3102,20 +3113,16 @@ class RealtimeClientConnection: QuickSpec {
                                 }
                             })
                         }
-
-                        guard let message = receivedMessage else {
-                            continue
-                        }
-
-                        expectDataToMatch(message, fixtureMessage)
                     }
                 }
 
-                it("should send messages through MsgPack and retrieve equal messages through JSON") {
-                    let restPublishClient = ARTRest(options: msgpackOptions)
+                it("should send messages through MsgPack and JSON and retrieve equal messages through raw JSON GET") {
+                    let restPublishClientMsgPack = ARTRest(options: msgpackOptions)
+                    let restPublishClientJSON = ARTRest(options: jsonOptions)
                     let restRetrieveClient = ARTRest(options: jsonOptions)
 
-                    let restPublishChannel = restPublishClient.channels.get("test-publish")
+                    let restPublishChannelMsgPack = restPublishClientMsgPack.channels.get("test-publish")
+                    let restPublishChannelJSON = restPublishClientJSON.channels.get(restPublishChannelMsgPack.name)
 
                     for (_, fixtureMessage) in fixtures["messages"] {
                         var data: AnyObject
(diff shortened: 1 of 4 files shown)


tcard:
I've added a test at fff273298b76e96cf9eca3775bd0a218f89c1168 for what we've talked about at ably/wiki#22. If it looks good I can do it in Java as well.

Please review @mattheworiordan @paddybyers @SimonWoolf.

mattheworiordan:
This looks good, but got a few comments:

* It seems from https://github.com/ably/ably-common/blob/messages-encoding-fixtures/test-resources/messages-encoding.json#L6-L7 that you are checking that a RAW post of the payload and encoding translates to a realtime publish of that decoded body, but you are not checking the body is ever as expected.  For example, if the realtime library incorrectly decoded an object, and then incorrect encoded the object, we'd never know as the first RAW post check and subsequent history check will pass, but the intervening payload may be incorrect.  I think we need to add a "decodedData" object to messages which we check matches exactly what we get following the RAW post
* I don't think we should call the expectedType `map` and `array` as that is platform specific. Instead, we have typically described those types as JSON Objects and JSON Arrays.  I think we should stick to that for consistency and to avoid confusion.

mattheworiordan:
BTW. Once done, I will add a corresponding test to Ruby and then a new spec item to the 0.9 spec

tcard:
> For example, if the realtime library incorrectly decoded an object, and then incorrect encoded the object, we'd never know as the first RAW post check and subsequent history check will pass, but the intervening payload may be incorrect.

It would be very unlikely that an incorrectly decoded object is then encoded correctly just as the fixture expects, wouldn't it? Plus, we really can't specify this in a platform-independent manner. Other tests should check this for each platform (they do already if they test [RSL4](http://docs.ably.io/client-lib-development-guide/features/#RSL4) properly); this one is focused on interoperability.

> I don't think we should call the expectedType map and array as that is platform specific. 

Renamed "map" to "object" at e7f68e1. (I just picked "map" randomly really, they're called dictionaries in Swift.)

mattheworiordan:
> It would be very unlikely that an incorrectly decoded object is then encoded correctly just as the fixture expects, wouldn't it? Plus, we really can't specify this in a platform-independent manner. Other tests should check this for each platform (they do already if they test RSL4 properly); this one is focused on interoperability.

But it's not testing interoperability because of exactly that reason.  Every platform supports JSON-like objects.  If a developer is unable to rely on our platform to provide consistent JSON-like objects then we have a much bigger problem.  We state clearly we support string, binary and JSON, so those are three types we need to test.  I am sorry, but I disagree here and feel we must test that the object received is what we expected else we're open to yet more problems.

> Renamed "map" to "object" at e7f68e1. (I just picked "map" randomly really, they're called dictionaries in Swift.)

Still not sure I would agree with that.  We don't support Objects (that is very generic), we support JSON Objects and JSON Arrays, see http://docs.ably.io/client-lib-development-guide/features/#RSL4c3 and http://docs.ably.io/client-lib-development-guide/features/#RSL4d3

paddybyers:
> But it's not testing interoperability because of exactly that reason. Every platform supports JSON-like objects. If a developer is unable to rely on our platform to provide consistent JSON-like objects then we have a much bigger problem. We state clearly we support string, binary and JSON, so those are three types we need to test. I am sorry, but I disagree here and feel we must test that the object received is what we expected else we're open to yet more problems.

Yes but the issue is how we represent the expected value in the fixture data when the value is a JSON object or array, or binary, without things getting a bit circular.

I think we do have to include an `expectedValue` as a javascript literal value (for string, object and array cases), expecting that any platform capable of parsing the fixture data also parses these values, and allows a test to be written that does a deep comparison of the decoded value and the parsed fixture `expectedValue`.

In the case of binary we just have to include a serialised representation in the fixture data (lets say hex instead of base64 because it minimises ambiguity), and the tests know that if the `expectedType` is `binary`, then it does a hex decode of the `expectedValue` before comparing.

> Still not sure I would agree with that. We don't support Objects (that is very generic), we support JSON Objects and JSON Arrays

Yes, but since the JSON spec defines what those values are it seems to make sense to use the names of the corresponding types from the JSON BNF (ie `object` and `array`, as in http://rfc7159.net/rfc7159#rfc.section.3.

paddybyers:
See https://github.com/ably/ably-common/tree/expected-value-proposal

tcard:
PTAL, implemented following @paddybyers ' proposal, and prepended "json" to "array" and "object".

mattheworiordan:
Thanks lgtm :+1:

tcard:
PTAL

Spec/RealtimeClientConnection.swift
```diff
@@ -3028,26 +2989,165 @@ class RealtimeClientConnection: QuickSpec {
...
                     }
+                }
 
-                    waitUntil(timeout: testTimeout) { done in
-                        channel.publish([message]) { err in
-                            if let err = err {
-                                fail("\(err)")
+                // https://github.com/ably/wiki/issues/22
+                it("should encode and decode fixture messages as expected") {
+                    let options = AblyTests.commonAppSetup()
```
mattheworiordan: Should this not explicitly use `JSON` i.e. `binaryProtocol = false` and should the test description be updated to reflect that these tests are about JSON based interoperability?

mattheworiordan:
One comment otherwise LGTM 👍

Spec/RealtimeClientConnection.swift
tcard: Hmm, I don't see why. Whether this uses JSON or MsgPack doesn't matter, that's not the layer we're testing. We have fully independent two layers:

* Message data encoding JSON-like to string and back again.
* Whole protocol message encoding to JSON or MsgPack and back again.

This test is about the former. When the encoded message data hits the latter, it is already a string in all cases.

mattheworiordan: > Message data encoding JSON-like to string and back again.

Well you are never technically doing that because every time you use JSON on one end MsgPack on the other.

tcard: I really don't see your point here. How am I never doing that? I am pretty literally doing that, concretely [here](https://github.com/ably/ably-ios/blob/json-encode-string/Source/ARTDataEncoder.m#L65). _Then_ I'm turning that _string_ (doesn't matter if it's JSON or not) into either MsgPack or a JSON blob or any other protocol-level encoding we might use in the future. Then I'm turning that into maybe a gzipped blob over HTTP and then encrypting that with TLS and then Wi-Fi signals or whatever, all those encodings that also don't affect the first one at all. Am I wrong here?

mattheworiordan: The point here is that tests don't know how the internals of a library work and should not care.  The point of our interoperability tests was to check that a message when decoded/encoded via JSON or MsgPack are consistent with the types we expect to see. We agreed the only reliable way to get data into the Ably system was via a raw REST request to Ably, and then to check that it was as expected, was also via a raw REST history request to Ably and then to parse the JSON (or inspect the type if binary/string) and compare with what we see the library interprets that over JSON and MsgPack.  So if we want to test how our Ably library encoded/decodes objects with a JSON transport it should use that, and if we want to see how our Ably library does that MsgPack it should use a MsgPack transport.  This test uses MsgPack, so the JSON encoding / decoding of a message injected or received via a raw request is not being done. That is why in Ruby I have a set of tests [using the JSON transport](https://github.com/ably/ably-ruby/pull/94/files#diff-5a92463d6ab8ad3fc8cf8aff5cfde6f8R746) and then using the [MsgPack transport](https://github.com/ably/ably-ruby/pull/94/files#diff-5a92463d6ab8ad3fc8cf8aff5cfde6f8R798) but all the time using a [raw JSON request to inject or check the fixture in Ably](https://github.com/ably/ably-ruby/pull/94/files#diff-5a92463d6ab8ad3fc8cf8aff5cfde6f8R823) irrespective of the transport being tested for interoperability.

tcard: > The point of our interoperability tests was to check that a message when decoded/encoded via JSON or MsgPack are consistent with the types we expect to see.

That's the point of the second interoperability test. The point of this one is to check it the message _data_ (not the message itself) when decoded/encoded as string, JSON or base64. We need both precisely because they test two completely independent things.

>  We agreed the only reliable way to get data into the Ably system was via a raw REST request to Ably, and then to check that it was as expected, was also via a raw REST history request to Ably and then to parse the JSON

That part I agree with, and if that's what you meant then OK. But since I'm doing a direct raw POST with a JSON payload ([here](https://github.com/ably/ably-ios/pull/459/files/3176c628c244243510171e5852d62780c41df921..1ee851cb4020e66c1a893573bd687aca9076f0a2#diff-8c3eccfb7635c28b9fadd0e4fcb6e6b0R3019)) it doesn't matter what I set `useBinaryProtocol` to. Same for the raw GET for history. It still doesn't affect how the message gets transmitted via realtime.

> This test uses MsgPack, so the JSON encoding / decoding of a message injected or received via a raw request is not being done.

Yes it is, because what we're encoding/decoding is the JSON _inside_ that message, which comes as a string. [These lines](https://github.com/ably/ably-ios/blob/master/Source/ARTDataEncoder.m#L142-L149) that decode the JSON from the message data are being executed both if the message came as JSON or as MsgPack. At that point, it's just an `ARTMessage`.

mattheworiordan: Perhaps I am misunderstanding still, but in these tests, at what point do we publish with a raw JSON message (i.e. getting the fixture into Ably in a guaranteed portable way) and get that message over a JSON transport and confirm the `message.data` is as expected then?

mattheworiordan: As above, and where do we publish over JSON using the library, and then confirm with a raw HTTP request that the JSON in Ably is as expected?

tcard: > at what point do we publish with a raw JSON message

[Here](https://github.com/ably/ably-ios/pull/459/files/3176c628c244243510171e5852d62780c41df921..1ee851cb4020e66c1a893573bd687aca9076f0a2#diff-8c3eccfb7635c28b9fadd0e4fcb6e6b0R3019). That's a `POST /channels/.../messages`. `fixtureMessage` is a `JSON` object (from a library for easy JSON manipulation) and `rawData()` converts it into a JSON string. That's quite like cutting and pasting from the `messages-encoding.json` directly into the POST request body.

> and get that message over a JSON transport

[Here](https://github.com/ably/ably-ios/pull/459/files/3176c628c244243510171e5852d62780c41df921..1ee851cb4020e66c1a893573bd687aca9076f0a2#diff-8c3eccfb7635c28b9fadd0e4fcb6e6b0R3054). `data` there is the HTTP response body, which the `JSON` library parses.

> and confirm the message.data is as expected

[Here](https://github.com/ably/ably-ios/pull/459/files/3176c628c244243510171e5852d62780c41df921..1ee851cb4020e66c1a893573bd687aca9076f0a2#diff-8c3eccfb7635c28b9fadd0e4fcb6e6b0R3055). Both `persistedMessage` and `fixtureMessage` are `JSON` objects that don't go through the `ARTRealtime` instance at any point. The `ARTRealtime` instance is used to test that the iOS library encodes and decodes the data as expected (ie. as `expectedValue` or `expectedHexValue` tell us), be it JSON, string or base64.

> As above, and where do we publish over JSON using the library, and then confirm with a raw HTTP request that the JSON in Ably is as expected?

That's in the second test, [here](https://github.com/ably/ably-ios/pull/459/files/3176c628c244243510171e5852d62780c41df921..1ee851cb4020e66c1a893573bd687aca9076f0a2#diff-8c3eccfb7635c28b9fadd0e4fcb6e6b0R3069).

mattheworiordan: >>  and get that message over a JSON transport
> Here. data there is the HTTP response body, which the JSON library parses.

But that's using `client.rest.executeRequest` to get that data.  So it's not passing through either a Realtime subscribe API or a history API request.  The goal was end-to-end API test, yet that is not using an API that handles encoding & decoding implicitly i.e. publish/subscribe/history.

>> As above, and where do we publish over JSON using the library, and then confirm with a raw HTTP request that the JSON in Ably is as expected?
That's in the second test, here.

tcard:
@mattheworiordan PTAL. If it looks good I'll do the same for Java.

mattheworiordan:
LGTM.  Once both are done, can we bump versions so that I can update that customer who had the issue?

PR #476, opened 2016-08-29
RSL2b

ricardopereira:


tcard:
I'm not really sure what this test is testing. We're not testing: "start must be equal to or less than end and is unaffected by the request direction", "if omitted the direction defaults to the REST API default (backwards)", "limit supports up to 1,000 items", "if omitted the direction defaults to the REST API default (100)".

ricardopereira:
I tried to cover all the parameters (`start`, `end`, `direction` and `limit`). I saw that some of the spec was already implemented in the legacy tests but maybe I should clarify this implementation.

ricardopereira:
@tcard PTAL

mattheworiordan:
👍

tcard:
LGTM.

@ricardopereira Any idea about what happens with CI now? There's a ton of "account blocked" errors, wasn't 5663598b61d69809ce1a40a29169712ecdff83f2 supposed to fix that?

PR #483, opened 2016-08-31
RSC7a

ricardopereira:


Spec/RestClient.swift
```diff
@@ -832,6 +832,24 @@ class RestClient: QuickSpec {
...
+                let client = ARTRest(options: options)
+                client.httpExecutor = testHTTPExecutor
+                waitUntil(timeout: testTimeout) { done in
+                    client.channels.get("test").publish(nil, data: "message") { error in
+                        expect(error).to(beNil())
+                        guard let headerAblyVersion = testHTTPExecutor.requests.first?.allHTTPHeaderFields?["X-Ably-Version"] else {
+                            fail("X-Ably-Version header not found"); done()
+                            return
+                        }
+                        expect(headerAblyVersion) == ARTDefault.version()
```
mattheworiordan: Whilst I appreciate this test is accurate, I think it would be useful to in fact be a more explicit and expect 0.8 for example in the string.  I appreciate when we upgrade to 0.9 it will break, but personally I like that as the test expectation has broken.  For example, if someone mistakenly upgrade to 0.99, this test would continue to pass

ricardopereira: There is already a G4 https://github.com/ably/ably-ios/blob/b3fdead2e828bf472bc51f3364e4118b8459b817/Spec/RestClient.swift#L32.
The current test is verifying if the client conforms with defaults.

tcard:
LGTM, @mattheworiordan please confirm you're OK with having the explicit test at a single place as @ricardopereira says.

Spec/RestClient.swift
mattheworiordan: Fine, thanks for clarifying

mattheworiordan:
@tcard happy for it to be merged

PR #484, opened 2016-08-31
RTN2f

ricardopereira:


mattheworiordan:
LGTM, but I thought we had already added that functionality, although admittedly I can't find the test right now

tcard:
LGTM

@mattheworiordan You're probably referring to https://github.com/ably/ably-ios/blob/master/Spec/RealtimeClient.swift#L39-L40. Admittedly they are testing exactly the same, but from different perspectives.

PR #486, opened 2016-08-31
Fixes 2 crashes that occurs during AblySpec tests execution.

EvgenyKarkan:
As discussed today with @tcard.


mattheworiordan:
If @tcard is happy then good for me

tcard:
LGTM

PR #487, opened 2016-08-31
Rsa8e

EvgenyKarkan:
As a step by step part of https://github.com/ably/docs/pull/113/files

### DIFF
--- a/Source/ARTAuth+Private.h
+++ b/Source/ARTAuth+Private.h
@@ -25,8 +25,6 @@ ART_ASSUME_NONNULL_BEGIN
 - (ARTAuthOptions *)mergeOptions:(ARTAuthOptions *)customOptions;
 - (ARTTokenParams *)mergeParams:(ARTTokenParams *)customParams;
 
-- (ARTAuthOptions *)replaceOptions:(ARTAuthOptions *)customOptions;
-
 - (NSURL *)buildURL:(ARTAuthOptions *)options withParams:(ARTTokenParams *)params;
 - (NSMutableURLRequest *)buildRequest:(ARTAuthOptions *)options withParams:(ARTTokenParams *)params;
 
--- a/Source/ARTAuth+Private.h
+++ b/Source/ARTAuth+Private.h
@@ -25,6 +25,8 @@ ART_ASSUME_NONNULL_BEGIN
 - (ARTAuthOptions *)mergeOptions:(ARTAuthOptions *)customOptions;
 - (ARTTokenParams *)mergeParams:(ARTTokenParams *)customParams;
 
+- (ARTAuthOptions *)replaceOptions:(ARTAuthOptions *)customOptions;
+
 - (NSURL *)buildURL:(ARTAuthOptions *)options withParams:(ARTTokenParams *)params;
 - (NSMutableURLRequest *)buildRequest:(ARTAuthOptions *)options withParams:(ARTTokenParams *)params;
 
--- a/Source/ARTAuth.m
+++ b/Source/ARTAuth.m
@@ -25,6 +25,7 @@
     ARTTokenParams *_tokenParams;
     // Dedicated to Protocol Message
     NSString *_protocolClientId;
+    ARTAuthOptions *_replacedAuthOptions;
 }
 
 - (instancetype)init:(ARTRest *)rest withOptions:(ARTClientOptions *)options {
@@ -34,7 +35,7 @@
         _options = options;
         _logger = rest.logger;
         _protocolClientId = nil;
-        
+        _tokenParams = options.defaultTokenParams ? : [[ARTTokenParams alloc] initWithOptions:self.options];
         [self validate:options];
     }
     
@@ -87,7 +88,8 @@
 }
 
 - (ARTAuthOptions *)replaceOptions:(ARTAuthOptions *)customOptions {
-    return customOptions ? customOptions : self.options;
+    _replacedAuthOptions = [self.options replaceWith:customOptions];
+    return customOptions ? _replacedAuthOptions : self.options;
 }
 
 - (void)storeOptions:(ARTAuthOptions *)customOptions {
@@ -104,6 +106,11 @@
 
 - (ARTTokenParams *)mergeParams:(ARTTokenParams *)customParams {
     return customParams ? customParams : [[ARTTokenParams alloc] initWithOptions:self.options];
+}
+
+- (ARTTokenParams *)replaceParams:(ARTTokenParams *)customParams {
+    ARTTokenParams* tkp = [[ARTTokenParams alloc] initWithOptions: (ARTClientOptions*)_replacedAuthOptions];
+    return customParams ? [_tokenParams replaceWith:customParams] : tkp;
 }
 
 - (void)storeParams:(ARTTokenParams *)customOptions {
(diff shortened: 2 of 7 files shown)


Spec/Auth.swift
```diff
@@ -626,16 +627,24 @@ class Auth : QuickSpec {
...
                     // TokenParams
                     let mergedParams = rest.auth.mergeParams(tokenParams)
                     expect(mergedParams.ttl) == 30.0
+                    
+                    let tokenParams2 = ARTTokenParams()
+                    tokenParams2.ttl = 25.0
+                    
+                    let mergedParams2 = rest.auth.mergeParams(tokenParams2)
+                    expect(mergedParams2.ttl) == 25.0
+                    expect(mergedParams2.clientId).to(beNil())
```
mattheworiordan: I am not really following the test code here.  Thee test code is not testing the `requestToken` method here but instead you are testing private undocumented APIs.  The point of the test is to "should not merge with the configured params and options but instead replace all corresponding values, even when @null@" yet this is not asserting that?

Spec/Auth.swift
```diff
@@ -1885,5 +1894,104 @@ class Auth : QuickSpec {
                 }
             }
         }
+        
+        describe("Reauth") {
+            it("should use authorise({force: true}) to reauth with a token with a different set of capabilities") {
```
mattheworiordan: Can you please include the spec ID here for reference?

Spec/Auth.swift
```diff
@@ -1885,5 +1894,104 @@ class Auth : QuickSpec {
...
+                        expect(reauthTokenDetails?.token).toNot(beNil())
+                        done()
+                    }
+                }
+                
+                // re-attach to the channel
+                waitUntil(timeout: testTimeout) { done in
+                    channel.attach() { error in
+                        expect(error).toNot(beNil())
+                        expect(error!.code).to(equal(40160))
```
mattheworiordan: Why are you testing for an error code?  This should be attaching correctly now???

EvgenyKarkan: I will move this test into separate PR, and it should cover `RTC8` item I guess. 
 
I took spec diffs here https://github.com/ably/docs/pull/113/files and as from my understanding - the `reauth` task contains set of spec items (RSA8e, RSA9h, RSA10k, RSA10j, RSA10d, RSA10g, RTC8, TK2d), so to implement current `reauth` task entirely - I need to implement each item, right?

mattheworiordan: > I will move this test into separate PR, and it should cover RTC8 item I guess.

That's not what I mean.  The attach should success, not fail.  You are testing that it fails following a reauth which is the wrong logic.

> I took spec diffs here https://github.com/ably/docs/pull/113/files and as from my understanding - the reauth task contains set of spec items (RSA8e, RSA9h, RSA10k, RSA10j, RSA10d, RSA10g, RTC8, TK2d), so to implement current reauth task entirely - I need to implement each item, right?

Yup, but this test is wrong as far as I can see and does not demonstrate adherence to the spec.

Spec/Auth.swift
EvgenyKarkan: Many thanks, I'll improve it.

Spec/Auth.swift
EvgenyKarkan: Got it, well, currently it fails in red, because lib code not yet written for test to be green in TDD manner.  In future I won't include anymore such red tests. Thanks.

mattheworiordan: A test that fails for the right reason is fine i.e. if we know you are adding a test that fails because functionality X is not implemented, then that is fine. But adding a test with an incorrect assertion, IMHO, is never right.

EvgenyKarkan: And as for adherence to the spec - yes, we should expect no errors after reattach. Thanks.

Source/ARTAuthOptions.h
```diff
@@ -91,6 +91,7 @@ ART_ASSUME_NONNULL_BEGIN
 - (NSString *)description;
 
 - (ARTAuthOptions *)mergeWith:(ARTAuthOptions *)precedenceOptions;
```
tcard: Is this function called from somewhere now?

Source/ARTAuth+Private.h
```diff
@@ -25,6 +25,8 @@ ART_ASSUME_NONNULL_BEGIN
 - (ARTAuthOptions *)mergeOptions:(ARTAuthOptions *)customOptions;
```
tcard: Is this function called from somewhere now?

Source/ARTAuth.m
```diff
@@ -102,6 +108,11 @@ - (ARTTokenParams *)mergeParams:(ARTTokenParams *)customParams {
     return customParams ? customParams : [[ARTTokenParams alloc] initWithOptions:self.options];
 }
 
+- (ARTTokenParams *)replaceParams:(ARTTokenParams *)customParams {
+    ARTTokenParams* tkp = [[ARTTokenParams alloc] initWithOptions: (ARTClientOptions*)_replacedAuthOptions];
```
tcard: This isn't a type-safe cast. Either ARTTokenParams needs a AuthOptions and not a full ClientOptions, in which case its constructor should be changed to reflect that, or it does need a full ClientOptions and that's what we should be giving it. (I suspect it's the former.)

Source/ARTAuth.m
```diff
@@ -25,6 +25,7 @@ @implementation ARTAuth {
     ARTTokenParams *_tokenParams;
     // Dedicated to Protocol Message
     NSString *_protocolClientId;
+    ARTAuthOptions *_replacedAuthOptions;
```
tcard: I don't get why are we storing this. Isn't it enough to return it?

tcard: (Return it from `replaceOptions`, I mean.)

Source/ARTAuthOptions.m
```diff
@@ -135,4 +135,52 @@ - (BOOL)isMethodGET {
     return [_authMethod isEqualToString:@"GET"];
 }
 
+- (ARTAuthOptions *)replaceWith:(ARTAuthOptions *)options {
```
tcard: Why do we even need to do this? Why not just use the new `options` instead of cloning the old one and copying everything from the new one into it? Unless I'm missing something, that just gives you a clone of the new one that you could've gotten with `[options copy]` in case you really need a copy.

Plus, it's full of unsafe casts from ARTAuthOptions to ARTClientOptions, and I don't see the point of the synchronized block given that the whole library is single-threaded.

tcard: To clarify, I mean that instead of this:

```objective-c
- (void)createTokenRequest:(ARTTokenParams *)tokenParams options:(ARTAuthOptions *)options callback:(void (^)(ARTTokenRequest *, NSError *))callback {
	ARTAuthOptions *mergedOptions = [self replaceOptions:options];
	ARTTokenParams *mergedTokenParams = [self replaceParams:tokenParams];
	// ...
```

We can do this:

```objective-c
- (void)createTokenRequest:(ARTTokenParams *)tokenParams options:(ARTAuthOptions *)options callback:(void (^)(ARTTokenRequest *, NSError *))callback {
	options = options ? options : self.options;
	tokenParams = tokenParams ? tokenParams : self.tokenParams;
	// ...
```

And basically get rid of everything else you added (plus the `merge...` methods laying around). Am I missing something?

Source/ARTAuthOptions.h
EvgenyKarkan: Yes, in `authorise` method.

EvgenyKarkan:
@tcard thanks for feedback, please find time to review.

Source/ARTAuthOptions.h
tcard: OK, I'll guess it will go away later as now `authorize` also replaces rather than merging.

tcard:
LGTM

Source/ARTAuthOptions.h
EvgenyKarkan: Exactly.

PR #488, opened 2016-09-01
Internal listeners test

ricardopereira:


Spec/RealtimeClient.swift
```diff
@@ -333,6 +333,53 @@ class RealtimeClient: QuickSpec {
                     expect(secs).to(beLessThanOrEqualTo(UInt(options.suspendedRetryTimeout)))
                 }
             }
+
+            it("should never register any connection listeners for internal use with the public EventEmitter") {
```
tcard: I don't really get how this test is testing what it claims to be testing.

mattheworiordan: Well it's a best effort to check the number of event listeners registered on the connection.  Do you have a better suggestion?  It's a very hard one to prove really, but at least this ensures common failures may be caught. WDYT?

tcard: What about calling `client.connection.off()` and checking that it doesn't affect any event emitter?

For instance, we use the internal event emitter [when calling `client.close()` while CONNECTING](https://github.com/ably/ably-ios/blob/187513e89ea7d9c7cd7f071528671c0698bd987a/Source/ARTRealtime.m#L154-L155) so that it actually closes when it reaches CONNECTED. So we can take advantage of that here, like this:

```swift
client.connect()
client.close() // Before it connects; this registers a listener on the internal event emitter.
expect(client.connection.status).to(equal(.Connecting))
client.connection.off()
expect(client.connection.status).to(equal(.Connecting)) // Still connecting...
// If we didn't have a separate internal event emitter, the line above would unregister
// the listener, and the next line would fail, because we would never move to 
// CLOSED, because we do that on the internal event listener registered when
// we called close().
expect(client.connection.status).toEventually(equal(.Closed))
```

tcard: I've changed the code to _not_ use an internal event emitter and the tests as they are now were passing anyway, so I've gone ahead and implemented what I said. See 81033b8.

PR #489, opened 2016-09-02
Fix RSC11: ignore environment when host is customized

ricardopereira:
Fix for https://github.com/ably/ably-ios/pull/473/files/2d05f1bdffb5587cdd006a7e0ebefd158c67c8fb#r77165419

### DIFF
--- a/Spec/RestClient.swift
+++ b/Spec/RestClient.swift
@@ -174,7 +174,7 @@ class RestClient: QuickSpec {
                     expect(testHTTPExecutor.requests.first?.URL?.host).toEventually(equal("fake.ably.io"), timeout: testTimeout)
                 }
 
-                it("should ignore an environment when host is not in production") {
+                it("should ignore an environment when restHost is customized") {
                     let options = ARTClientOptions(key: "fake:key")
                     options.environment = "test"
                     options.restHost = "fake.ably.io"
@@ -186,7 +186,7 @@ class RestClient: QuickSpec {
                     expect(testHTTPExecutor.requests.first?.URL?.host).toEventually(equal("fake.ably.io"), timeout: testTimeout)
                 }
                 
-                it("should accept an environment in production") {
+                it("should accept an environment when restHost is left unchanged") {
                     let options = ARTClientOptions(key: "fake:key")
                     options.environment = "myEnvironment"
                     let client = ARTRest(options: options)
--- a/Spec/RestClient.swift
+++ b/Spec/RestClient.swift
@@ -163,7 +163,7 @@ class RestClient: QuickSpec {
 
             // RSC11
             context("endpoint") {
-                it("should accept an options object with a host set") {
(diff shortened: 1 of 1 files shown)


Spec/RestClient.swift
```diff
@@ -173,8 +173,20 @@ class RestClient: QuickSpec {
                     
                     expect(testHTTPExecutor.requests.first?.URL?.host).toEventually(equal("fake.ably.io"), timeout: testTimeout)
                 }
+
+                it("should ignore an environment when host is not in production") {
```
tcard: What does a host being "in production" mean?

Spec/RestClient.swift
```diff
@@ -173,8 +173,20 @@ class RestClient: QuickSpec {
...
+                    let options = ARTClientOptions(key: "fake:key")
+                    options.environment = "test"
+                    options.restHost = "fake.ably.io"
+                    let client = ARTRest(options: options)
+                    client.httpExecutor = testHTTPExecutor
+
+                    publishTestMessage(client, failOnError: false)
+
+                    expect(testHTTPExecutor.requests.first?.URL?.host).toEventually(equal("fake.ably.io"), timeout: testTimeout)
```
tcard: RSC11 leaves this case somewhat unspecified. It says that `environment` set and `restHost` being the default `rest.ably.io` should prepend `environment` to `restHost`, and it says that changing `restHost` overrides the default `rest.ably.io`, but it doesn't say what happens when you set both `restHost` and `environment`.

I really don't see the point of making a special case of ignoring the environment for non-default hosts, ie. I think this should be `fake-test.ably.io`. @mattheworiordan can you clarify this?

tcard: OK, [TOK3k2](http://docs.ably.io/client-lib-development-guide/features/#TO3k2) clarifies it.  It is undefined because it is invalid. I guess, for lack of a more specific meaning for "invalid", that just ignoring the environment here is fine then.

tcard:
LGTM, although a comment abut TOK3k2 would be nice.

Spec/RestClient.swift
mattheworiordan: @tcard not sure if @ricardopereira is around unfortunately.

ricardopereira: @tcard I assumed that "in production" is when restHost ≠ default value.

mattheworiordan: @tcard perhaps you should just take the PR and reword as you see fit as @ricardopereira is away on holiday for a little while still

tcard: OK, I think that "when `environment` is not "production"` is more explicit.

tcard: Sorry, that's not what's happening here at all. Well, da76de0cdf5baa0dc4ebae5a3ba41b92deffce82 it is.

PR #491, opened 2016-09-04
RSA9h

EvgenyKarkan:
7c266e7

### DIFF
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -1206,6 +1206,7 @@ class Auth : QuickSpec {
                 let authOptions = ARTAuthOptions()
                 authOptions.force = true
                 authOptions.queryTime = true
+                authOptions.key = options.key
 
                 var serverDate = NSDate()
                 waitUntil(timeout: testTimeout) { done in
@@ -1318,6 +1319,22 @@ class Auth : QuickSpec {
                 }
                 expect(callbackCalled).to(beTrue())
             }
+            
+            it("should replace defaults if `nil` option's field passed") {
+                let defaultOptions = AblyTests.commonAppSetup()
+                let rest = ARTRest(options: defaultOptions)
+                
+                let customOptions = ARTAuthOptions()
+            
+                waitUntil(timeout: testTimeout) { done in
+                    rest.auth.createTokenRequest(nil, options: customOptions, callback: { tokenRequest, error in
+                        //ARTAuthOptions `key` property must not be nil, otherwise `createTokenRequest` API returns an error
+                        expect(error).toNot(beNil())
+                        expect(error?.domain).to(equal(ARTAblyErrorDomain))
+                        done()
+                    })
+                }
+            }
 
             // RSA9a
             it("should create and sign a TokenRequest") {
@@ -1411,6 +1428,7 @@ class Auth : QuickSpec {
 
                     let authOptions = ARTAuthOptions()
                     authOptions.queryTime = true
+                    authOptions.key = AblyTests.commonAppSetup().key
 
                     waitUntil(timeout: testTimeout) { done in
                         rest.auth.createTokenRequest(nil, options: authOptions, callback: { tokenRequest, error in
@@ -1561,6 +1579,7 @@ class Auth : QuickSpec {
 
                 let authOptions = ARTAuthOptions()
                 authOptions.queryTime = true
+                authOptions.key = AblyTests.commonAppSetup().key
 
                 var serverTime: NSDate?
                 waitUntil(timeout: testTimeout) { done in
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -1176,6 +1176,7 @@ class Auth : QuickSpec {
 
         // RSA9
         describe("createTokenRequest") {
+
             // RSA9h
             it("should not merge with the configured params and options but instead replace all corresponding values, even when @null@") {
                 let options = AblyTests.commonAppSetup()
@@ -1233,11 +1234,11 @@ class Auth : QuickSpec {
                         done()
                     }
                 }
-                
+
                 tokenParams.clientId = "newClientId"
                 tokenParams.ttl = 2000
                 tokenParams.capability = "{ \"test:*\":[\"test\"] }"
-                
+
                 waitUntil(timeout: testTimeout) { done in
                     rest.auth.createTokenRequest(tokenParams, options: authOptions) { tokenRequest, error in
                         expect(error).to(beNil())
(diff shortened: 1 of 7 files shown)


Spec/Auth.swift
```diff
@@ -1106,6 +1106,36 @@ class Auth : QuickSpec {
                         done()
                     }
                 }
+                
+                tokenParams.clientId = "newClientId"
+                tokenParams.ttl = 2000
+                tokenParams.capability = "{ \"test:*\":[\"test\"] }"
+                
+                waitUntil(timeout: testTimeout) { done in
+                    rest.auth.createTokenRequest(tokenParams, options: authOptions) { tokenRequest, error in
```
mattheworiordan: This API does not look right.  Why is this `options:` when the IDL specifies we have two args, `TokenParams` and `AuthOptions`.  Whilst I appreciate this PR is not necessarily about this, it's important we fix things as we go along.  Can you take a look please?

Spec/Auth.swift
```diff
@@ -1106,6 +1106,36 @@ class Auth : QuickSpec {
...
+                        expect(tokenRequest.ttl).to(equal(2000))
+                        expect(tokenRequest.capability).to(equal("{ \"test:*\":[\"test\"] }"))
+                        done()
+                    }
+                }
+                
+                tokenParams.clientId = nil
+                
+                waitUntil(timeout: testTimeout) { done in
+                    rest.auth.createTokenRequest(tokenParams, options: authOptions) { tokenRequest, error in
```
mattheworiordan: My concern with this commit is that at no point have you tested that `nil` AuthOptions does not change the AuthOptions, and at no point have you checked that `AuthOptions` overrides the defaults when provided. The assertions are all based on the token request properties.  You will need to test at least one attribute of AuthOptions please for completeness.

Spec/Auth.swift
EvgenyKarkan: > This API does not look right

Why do you think so? `createTokenRequest` API [has 3 arguments](https://github.com/ably/ably-ios/blob/master/Source/ARTAuth.h#L49-L50) - `ARTTokenParams` and `ARTAuthOptions` objects and a `callback` which provides `ARTTokenRequest` and an error if any.

> Why is this `options:` 

Lib written in `ObjC`, and tests in `Swift`, so `options:` is a part of `ObjC` method signature here and the argument is `authOptions` object of `ARTAuthOptions` class.

Spec/Auth.swift
EvgenyKarkan: Got it, I'll try to find out a way to test without assertions based on he token request properties.

Spec/Auth.swift
mattheworiordan: Ok, this might be an "iOS" thing then that I have misunderstood.  I have never written iOS code so just looking at these PRS with a view to a) does it achieve the spec, b) does it follow the API.  @tcard can you comment on this and confirm you're happy to leave as is i.e. I was wrong :)

Thanks @EvgenyKarkan for the explanation

tcard: Yes @mattheworiordan, we're forced to do this every time we have more than one argument. It's just how Objective-C works. (When the last argument is a callback, as a syntactic convenience Swift lets you omit its name and pass it kind of like a Ruby block, that's why you don't see its name.)

Spec/Auth.swift
mattheworiordan: Ok, thanks

EvgenyKarkan:

tcard:
Nice, but there's still one case missing: the case in which `defaultOptions` has field _non-nil_, but then `options` has that same field as nil. If we were merging the options as we were doing before, the non-nil (ie. the old) version would prevail, but because we're not merging anymore, it should behave as if the field were nil.

For example, you could set `defaultOptions.authCallback` to something, `options.authCallback` to `nil` and check that you get some error like "I have no means to authorize" and that the `defaultOptions.authCallback` wasn't called.

EvgenyKarkan:

ricardopereira:
@EvgenyKarkan What's the state on this one? This PR will merge into `branch:RSA8e` but that branch has been already merged with master. Can I close this and create a new PR pointing to master?

EvgenyKarkan:
Hello @ricardopereira 

> What's the state on this one?

Hope it's done, I asked Toni for review this PR.

ricardopereira:
LGTM. I rebased with master but tests are failing.
I will look what's missing.

Spec/Auth.swift
```diff
@@ -1232,6 +1233,107 @@ class Auth : QuickSpec {
...
+                expect(currentTokenRequest).toEventuallyNot(beNil(), timeout: testTimeout)
+                
+                let options = ARTAuthOptions()
+                options.authCallback = { tokenParams, completion in
+                    secondCallbackCalled = true
+                    completion(currentTokenRequest!, nil)
+                }
+                
+                waitUntil(timeout: testTimeout) { done in
+                    rest.auth.authorise(nil, options: options) { tokenDetails, error in
```
ricardopereira: Shouldn't we testing the `createTokenRequest` (spec `(RSA9) Auth#createTokenRequest`)?

Spec/Auth.swift
```diff
@@ -1232,6 +1233,107 @@ class Auth : QuickSpec {
...
+                }
+                
+                let rest = ARTRest(options: defaultOptions)
+                rest.auth.createTokenRequest(nil, options: nil, callback: { tokenRequest, error in
+                    currentTokenRequest = tokenRequest
+                })
+                expect(currentTokenRequest).toEventuallyNot(beNil(), timeout: testTimeout)
+                
+                waitUntil(timeout: testTimeout) { done in
+                    rest.auth.authorise(nil, options: nil) { tokenDetails, error in
```
ricardopereira: Same, shouldn't we only testing the `createTokenRequest`?

Spec/Auth.swift
ricardopereira: Nevermind. The `createTokenRequest` is being tested.

Spec/Auth.swift
ricardopereira: Nevermind. The `createTokenRequest` is being tested.

ricardopereira:
> Nice, but there's still one case missing: the case in which defaultOptions has field non-nil, but then options has that same field as nil. If we were merging the options as we were doing before, the non-nil (ie. the old) version would prevail, but because we're not merging anymore, it should behave as if the field were nil.
> 
> For example, you could set defaultOptions.authCallback to something, options.authCallback to nil and check that you get some error like "I have no means to authorize" and that the defaultOptions.authCallback wasn't called.

@tcard Are you happy with this 8e63e9c?

ricardopereira:
All other PR's are related with this one, so I will merge it to help me out with my further work.
If anything needs to be adjusted just comment and I will address them into another PR.

tcard:
@ricardopereira I'd be happier if we were checking for a specific error code, since now any failure would make the test pass. But no biggie.

PR #492, opened 2016-09-04
RSA10j

EvgenyKarkan:
c5e461b

### DIFF
--- a/Source/ARTAuth.m
+++ b/Source/ARTAuth.m
@@ -360,13 +360,17 @@
             if (error) {
                 callback(nil, error);
             } else {
-                currentTokenParams.timestamp = time;
+                currentTokenParams.timestamp = [self handleServerTime:time];
                 callback([currentTokenParams sign:replacedOptions.key], nil);
             }
         }];
     } else {
         callback([currentTokenParams sign:replacedOptions.key], nil);
     }
+}
+
+- (NSDate *)handleServerTime:(NSDate *)time {
+    return time;
 }
 
 - (void)setProtocolClientId:(NSString *)clientId {
--- a/Source/ARTAuth.m
+++ b/Source/ARTAuth.m
@@ -392,6 +392,10 @@
     }
 }
 
+- (void)setTokenDetails:(ARTTokenDetails *)tokenDetails {
+    _tokenDetails = tokenDetails;
+}
+
 @end
 
 @implementation NSString (ARTTokenDetailsCompatible)
--- a/Source/ARTAuth.m
+++ b/Source/ARTAuth.m
@@ -277,21 +277,21 @@
 - (void)authorise:(ARTTokenParams *)tokenParams options:(ARTAuthOptions *)authOptions callback:(void (^)(ARTTokenDetails *, NSError *))callback {
     BOOL requestNewToken = NO;
 
-    ARTAuthOptions *mergedOptions;
+    ARTAuthOptions *replacedOptions;
     if ([authOptions isOnlyForceTrue]) {
-        mergedOptions = self.options;
-        mergedOptions.force = YES;
+        replacedOptions = self.options;
+        replacedOptions.force = YES;
     }
     else {
-        mergedOptions = [self mergeOptions:authOptions];
+        replacedOptions = authOptions ? : self.options;
     }
+    [self storeOptions:replacedOptions];
 
-    [self storeOptions:mergedOptions];
     ARTTokenParams *currentTokenParams = [self mergeParams:tokenParams];
     [self storeParams:currentTokenParams];
 
     // Reuse or not reuse the current token
-    if (mergedOptions.force == NO && self.tokenDetails) {
+    if (replacedOptions.force == NO && self.tokenDetails) {
         if (self.tokenDetails.expires == nil) {
             [self.logger verbose:@"RS:%p ARTAuth: reuse current token.", _rest];
             requestNewToken = NO;
@@ -306,7 +306,7 @@
         }
     }
     else {
-        if (mergedOptions.force == YES)
+        if (replacedOptions.force == YES)
             [self.logger verbose:@"RS:%p ARTAuth: forced requesting new token.", _rest];
         else
             [self.logger verbose:@"RS:%p ARTAuth: requesting new token.", _rest];
@@ -314,7 +314,7 @@
     }
 
     if (requestNewToken) {
-        [self requestToken:currentTokenParams withOptions:mergedOptions callback:^(ARTTokenDetails *tokenDetails, NSError *error) {
+        [self requestToken:currentTokenParams withOptions:replacedOptions callback:^(ARTTokenDetails *tokenDetails, NSError *error) {
             if (error) {
                 [self.logger verbose:@"RS:%p ARTAuth: token request failed: %@", _rest, error];
                 if (callback) {
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -1646,7 +1646,10 @@ class Auth : QuickSpec {
 
                 waitUntil(timeout: testTimeout) { done in
                     rest.auth.authorise(ARTTokenParams(), options: ARTAuthOptions(), callback: { tokenDetails, error in
-                        expect(error).to(beNil())
+                        guard let error = error else {
+                            fail("Error is nil"); done(); return
+                        }
+                        expect(error.description).to(contain("no means to renew the token is provided"))
                         done()
                     })
                 }
@@ -1689,12 +1692,12 @@ class Auth : QuickSpec {
 
             // RSA10d
             it("should issue a new token even if an existing token exists when AuthOption.force is true") {
-                let defOptions = AblyTests.commonAppSetup()
-                defOptions.clientId = "defClientId"
-                
-                let rest = ARTRest(options: defOptions)
+                let options = AblyTests.commonAppSetup()
+                options.clientId = "defClientId"
+                let rest = ARTRest(options: options)
 
                 let authOptions = ARTAuthOptions()
+                authOptions.key = options.key
                 authOptions.force = true
 
                 // Current token
@@ -1833,11 +1836,13 @@ class Auth : QuickSpec {
                             expect(authCallbackHasBeenInvoked).to(beTrue())
 
                             authCallbackHasBeenInvoked = false
-                            let options = ARTAuthOptions()
-                            options.force = true
-                            auth.authorise(nil, options: options) { tokenDetails, error in
-                                expect(authCallbackHasBeenInvoked).to(beTrue())
-                                expect(auth.options.useTokenAuth).to(beTrue())
+                            let authOptions2 = ARTAuthOptions()
+
+                            auth.testSuite_forceTokenToExpire()
+
+                            auth.authorise(nil, options: authOptions2) { tokenDetails, error in
+                                expect(authCallbackHasBeenInvoked).to(beFalse())
+                                expect(auth.options.useTokenAuth).to(beFalse())
                                 expect(auth.options.queryTime).to(beFalse())
                                 done()
                             }
@@ -1846,8 +1851,10 @@ class Auth : QuickSpec {
                 }
 
                 it("should not store queryTime") {
-                    let rest = ARTRest(options: AblyTests.commonAppSetup())
+                    let options = AblyTests.commonAppSetup()
+                    let rest = ARTRest(options: options)
                     let authOptions = ARTAuthOptions()
+                    authOptions.key = options.key
                     authOptions.queryTime = true
 
                     var serverTimeRequestWasMade = false
@@ -2105,7 +2112,7 @@ class Auth : QuickSpec {
                     let authOptions = ARTAuthOptions()
                     authOptions.key = AblyTests.commonAppSetup().key //valid key
                     let tokenParams = ARTTokenParams()
-                    tokenParams.ttl = 0.1
+                    tokenParams.ttl = 1.0
 
                     waitUntil(timeout: testTimeout) { done in
                         rest.auth.authorise(tokenParams, options: authOptions) { tokenDetails, error in
@@ -2116,8 +2123,8 @@ class Auth : QuickSpec {
                             guard let expires = tokenDetails?.expires else {
                                 fail("TokenDetails.expires is nil"); done(); return
                             }
-                            expect(issued).to(beCloseTo(expires, within: tokenParams.ttl))
-                            delay(tokenParams.ttl + 1.0) {
+                            expect(issued).to(beCloseTo(expires, within: tokenParams.ttl + 0.1))
+                            delay(tokenParams.ttl + 0.1) {
                                 done()
                             }
                         }
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -16,7 +16,7 @@ class Auth : QuickSpec {
         
         struct ExpectedTokenParams {
             static let clientId = "client_from_params"
-            static let ttl = 5.0
+            static let ttl = 1.0
             static let capability = "{\"cansubscribe:*\":[\"subscribe\"]}"
         }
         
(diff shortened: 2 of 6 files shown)


Spec/Auth.swift
```diff
@@ -1903,6 +1903,32 @@ class Auth : QuickSpec {
                         done()
                     }
                 }
+
```
mattheworiordan: I appreciate the code above was not yours, but I do not follow the logic.  The `queryTime` option is set to true for AuthOptions, I get that.  But the time from the `time` method is stored and then the authorise request is then executed and the issued attribute is checked to be within 1 second.  Firstly, if we have even a minor delay here this test will fail, but I am not sure what it's even testing.  If we want to check the time then we should have a test for http://docs.ably.io/client-lib-development-guide/features/#RSA10k, but this is mixing concerns and is brittle.  Please can you confirm why you think this is even here, and whether RSA10k has suitable coverage proving the request for time is made elsewhere i.e. we can then remove this.

Spec/Auth.swift
```diff
@@ -1903,6 +1903,32 @@ class Auth : QuickSpec {
...
+                        }
+
+                        expect(tokenDetails.clientId).to(equal("testId"))
+                        done()
+                    }
+                }
+                
+                tokenParams.clientId = nil
+                waitUntil(timeout: testTimeout) { done in
+                    rest.auth.authorise(tokenParams, options: authOptions) { tokenDetails, error in
```
mattheworiordan: This has the same issue as https://github.com/ably/ably-ios/pull/491, we are never testing changes to authOptions or `null` auth options.  Also the argument is incorrect.

Spec/Auth.swift
```diff
@@ -1903,6 +1903,32 @@ class Auth : QuickSpec {
...
+                        expect(error).to(beNil())
+                        guard let tokenDetails = tokenDetails else {
+                            XCTFail("TokenDetails is nil"); done(); return
+                        }
+                        
+                        expect(tokenDetails.clientId).to(beNil())
```
mattheworiordan: I do not believe we are truly testing the authorise behaviour correctly, see:

> ... and a TokenParams object is passed in as an argument to #authorize with a null or missing value for ttl, then the ttl used for every subsequent authorisation will be null

We need to test that calling authorise with a new AuthOption or TokenParam set of options will be carried through for all future authorise calls.  Also, method name is `authorize`.  I appreciate that may be part of 0.9, but as you are effectively stating marking this as "done", there is a risk we won't do this later.  We must not merge this without adding a TODO to come back and change the naming, see http://docs.ably.io/client-lib-development-guide/features/#RSA10j

Spec/Auth.swift
EvgenyKarkan: > This has the same issue as #491

Got it, thanks.

> Also the argument is incorrect.

As per our recent conversation in #491 - we agreed that `createTokenRequest` API has correct set of arguments. 
Can we assume that this also correct for `authorise` method [signature](https://github.com/ably/ably-ios/blob/master/Source/ARTAuth.h#L47)?

Spec/Auth.swift
EvgenyKarkan: > Also, method name is `authorize`

Ok, so I need to rename API from `authorise` to `authorize` (`z` letter is a diff) entirely in my current 0.8 scope, got it.

EvgenyKarkan: > We need to test that calling authorise with a new AuthOption or TokenParam set of options will be carried through for all future authorise calls.

Got it.

Spec/Auth.swift
EvgenyKarkan: I guess that previous developer just wanted to test that `issued` time is in 1 second range with server time, but I also cannot understand the pros of doing `time` function call to store server time.

tcard:
Please try to keep independent features separate. This work and RSA9h (#491) are independent and should both be built upon RSA8e (#487). Now if you change RSA9h you also have to change this, and we can't merge this without merging RSA9h.

Spec/Auth.swift
mattheworiordan: Ok, well let's remove then and if this is a concern perhaps add as a separate test such as `localClockInSync` which simply fails with "Local clock is out of sync with Ably so tests may have unpredictable results"

Spec/Auth.swift
mattheworiordan: Yes, if @tcard  is happy, thanks

Spec/Auth.swift
mattheworiordan: > Ok, so I need to rename API from authorise to authorize (z letter is a diff) entirely in my current 0.8 scope, got it.

Well it's a 0.9 change, so we can defer to later, but we need an issue so that we don't mark this as completed mistakenly and never come back to it.  Can you do that please?

EvgenyKarkan:
> Please try to keep independent features separate.  ...
> Now if you change RSA9h you also have to change this, and we can't merge this without merging RSA9h. 

I'll keep it in mind, thanks. I'll merge `RSA9h` into current branch.

Spec/Auth.swift
EvgenyKarkan: I've opened an issue #496.

Source/ARTAuth.m
```diff
@@ -263,9 +263,9 @@ - (void)executeTokenRequest:(ARTTokenRequest *)tokenRequest callback:(void (^)(A
 
 - (void)authorise:(ARTTokenParams *)tokenParams options:(ARTAuthOptions *)authOptions callback:(void (^)(ARTTokenDetails *, NSError *))callback {
     BOOL requestNewToken = NO;
-    ARTAuthOptions *mergedOptions = [self mergeOptions:authOptions];
+    ARTAuthOptions *mergedOptions = authOptions ? authOptions : self.options;
```
mattheworiordan: Name is a bit misleading, perhap just `currentAuthOptions` inline with `currentTokenParams` below?

EvgenyKarkan: Oh, sorry, of course :)

EvgenyKarkan:

Spec/Auth.swift
```diff
@@ -1717,14 +1718,12 @@ class Auth : QuickSpec {
                     waitUntil(timeout: testTimeout) { done in
                         auth.authorise(nil, options: authOptions) { tokenDetails, error in
                             expect(authCallbackHasBeenInvoked).to(beTrue())
-
-                            authCallbackHasBeenInvoked = false
```
tcard: This test shouldn't be changed. It should work, or else we're breaking [RSA10g](http://docs.ably.io/client-lib-development-guide/features/#RSA10g).

Spec/Auth.swift
```diff
@@ -1748,21 +1747,18 @@ class Auth : QuickSpec {
                     }
 
                     waitUntil(timeout: testTimeout) { done in
-                        delay(tokenParams.ttl + 1.0) {
```
tcard: This test also should work as it was before, or else we're breaking [RSA10g](http://docs.ably.io/client-lib-development-guide/features/#RSA10g).

tcard:
@EvgenyKarkan I've rebased this PR on top of master instead of RSA9h, so that they are independent. It was getting pretty hard to review. Sorry for the trouble.

Spec/Auth.swift
mattheworiordan: Because @EvgenyKarkan I believe what @tcard is saying is that you should test that the defaults are reused first (i.e. `Stores the AuthOptions and TokenParams arguments as defaults for subsequent authorisations with the exception of the attributes TokenParams#timestamp and AuthOptions#queryTime`).  The test for http://docs.ably.io/client-lib-development-guide/features/#RSA10g is now missing i.e. we need to ensure that the defaults are carried through to the following request for RSA10g, and for RSA10j (this PR), you need to ensure any provided details override existing defaults.  Added to this, https://github.com/ably/ably-ios/pull/492/files#diff-b5d84fe9bce26902a7481c2ffec84fffR1723 is really not relevant given that param has been removed completely given RSA10d is now removed, see https://github.com/ably/docs/blob/source/content/client-lib-development-guide/versions/features-0-8__0-9.diff#L116.

So in summary:
* Test that defaults are carried through when none are provided (need initial + subsequent request without any arguments)
* Test that defaults are changed when set (need initial + subsequent request without any arguments)
* Remove `force` completely as this just confuses these tests given that option is no longer relevant and should be removed.

Spec/Auth.swift
mattheworiordan: To be fair once authorize is called it forces an auth every time now, so not sure this one does?  However, we still need to test initial defaults are carried forward, and any subsequent details configured are carried forward too

Source/ARTAuth.m
```diff
@@ -275,13 +275,13 @@ - (void)executeTokenRequest:(ARTTokenRequest *)tokenRequest callback:(void (^)(A
...
-    [self storeOptions:mergedOptions];
-    ARTTokenParams *currentTokenParams = [self mergeParams:tokenParams];
+    ARTAuthOptions *currentAuthOptions = authOptions ? authOptions : self.options;
+    [self storeOptions:currentAuthOptions];
+    ARTTokenParams *currentTokenParams = tokenParams ? tokenParams : _tokenParams;
     [self storeParams:currentTokenParams];
 
     // Reuse or not reuse the current token
-    if (mergedOptions.force == NO && self.tokenDetails) {
+    if (currentAuthOptions.force == NO && self.tokenDetails) {
```
mattheworiordan: I think you should just remove `force` from this PR and perhaps bundle the change for RSA10d into this as this is only confusing the review of a PR that is in flux between old & new functionality

Source/ARTAuth.m
```diff
@@ -275,13 +275,13 @@ - (void)executeTokenRequest:(ARTTokenRequest *)tokenRequest callback:(void (^)(A
 
 - (void)authorise:(ARTTokenParams *)tokenParams options:(ARTAuthOptions *)authOptions callback:(void (^)(ARTTokenDetails *, NSError *))callback {
     BOOL requestNewToken = NO;
-    ARTAuthOptions *mergedOptions = [self mergeOptions:authOptions];
-    [self storeOptions:mergedOptions];
-    ARTTokenParams *currentTokenParams = [self mergeParams:tokenParams];
```
mattheworiordan: I believe timestamp and queryTime was not carried forward before, but now it appears they are given these objects are blindly copied?  See http://docs.ably.io/client-lib-development-guide/features/#RSA10g

Spec/Auth.swift
```diff
@@ -1748,21 +1747,18 @@ class Auth : QuickSpec {
...
-                                expect(tokenDetails.capability).to(equal(ExpectedTokenParams.capability))
-                                done()
+                        rest.auth.authorise(nil, options: nil) { tokenDetails, error in
+                            expect(error).to(beNil())
+                            guard let tokenDetails = tokenDetails else {
+                                XCTFail("TokenDetails is nil"); done(); return
                             }
+                            expect(tokenDetails.clientId).to(equal(ExpectedTokenParams.clientId))
+                            expect(tokenDetails.issued!.dateByAddingTimeInterval(ExpectedTokenParams.ttl)).to(beCloseTo(tokenDetails.expires))
+                            expect(tokenDetails.capability).to(equal(ExpectedTokenParams.capability))
```
mattheworiordan: If your have removed the TTL and delay in the test, then how do you know the token is not the same token?  You need to make sure that a new token has ben generated after calling authorise

Spec/Auth.swift
```diff
@@ -1963,40 +1968,43 @@ class Auth : QuickSpec {
...
+                        
+                        let authOptions = ARTAuthOptions()
+                        authOptions.authCallback = nil
+                        
+                        rest.auth.authorise(nil, options: authOptions) { tokenDetails2, error in
+                            expect(error).to(beNil())
+                            guard let tokenDetails2 = tokenDetails2 else {
+                                XCTFail("TokenDetails2 is nil"); done(); return
+                            }
+                            expect(rest.auth.options.authCallback).to(beNil())
```
mattheworiordan: It would be more useful here to check that a documented API is overriden i.e. `auth.options` is undocumented, so instead why not check that the previous callback has not been used perhaps, or in the callback set a specific capability and ensure that capability has now not been used thus proving the callback has not been called.

Spec/Auth.swift
tcard: Oh OK, true, then this `delay` isn't necessary.

Source/ARTAuth.m
EvgenyKarkan: Got it, will improve.

Spec/Auth.swift
EvgenyKarkan: > Test that defaults are changed when set (need initial + subsequent request without any arguments)

I guess you mean  `+ subsequent request`**`with`**`any arguments` here?

> Remove force completely

Remove it only from tests target, or also from the lib target?

Source/ARTAuth.m
EvgenyKarkan: If I remove `force` property from `AuthOptions` object - it will break `RSA10d` & also `RTC8` requirements, because both them needs `force`.
That’s why in my opinion it's impossible to combine both `RSA10j` and `RSA10d` changes in one PR.

Spec/Auth.swift
mattheworiordan: > I guess you mean + subsequent requestwithany arguments here?

No, we need to test that the defaults are used for subsequent authorisations so if we passed in arguments then we'd be changing those defaults?

>> Remove force completely

> Remove it only from tests target, or also from the lib target?

Not sure what you mean by target, but it should be removed as this option is no longer available as part of the tests you are implementing

Source/ARTAuth.m
mattheworiordan: Then combine more stuff into one PR. We asked for as small PRs as possible, but if it's not possible to deliver this without making other changes, then those changes make this PR as small as possible.

Spec/Auth.swift
EvgenyKarkan: > No, we need to test that the defaults are used for subsequent authorisations so if we passed in arguments then we'd be changing those defaults?

I might be missing something, but as per `RSA10j` spec that states: 
>For example, if a client is initialised with `TokenParams#ttl` configured with a custom value, and a `TokenParams` object is passed in as an argument to `authorise` with a null value for `ttl`, then the `ttl` used for every subsequent authorisation will be null. 

Does not this mean that the default parameters need to be replaced?

mattheworiordan: Yes but if no arguments are passed in, then a TokenParams object has not been passed in.

What it actually states is:

>  TokenParams and AuthOptions are optional. When the arguments are present...

The point is that if you don't provide those arguments or they are null (but not empty TokenParam or AuthObtion objects), then the defaults are not overriden.

EvgenyKarkan: > Not sure what you mean by target

Well, Xcode project can contain several build targets - e.g.. main iOS or OSX app, unit tests, apple watch app, app extension, framework, etc. 

In our case we have `Ably` ObjC library target, `AblySpecs` Swift tests target, and `AblyTests` ObjC tests target. 

So my question was, in another words, - should I remove use of `force` from tests only, or should I remove `force` property from `AuthOptions` class in `Ably` library implementation too? 

I guess I just need to remove it from test target only, otherwise it won't be possible to implement `RSA10d` and `RTC8`, because both of them needs `force`.

 Am I right here, could you confirm please @mattheworiordan ?

ricardopereira:
Squashed.

ricardopereira:
@mattheworiordan PTAL

ricardopereira:
@tcard PTAL

Spec/Auth.swift
```diff
@@ -2091,61 +2091,283 @@ class Auth : QuickSpec {
...
-            it("should supersede any configured params and options when TokenParams and AuthOptions were provided") {
-                let options = AblyTests.commonAppSetup()
-                options.clientId = "client_string"
-                let rest = ARTRest(options: options)
+            context("when TokenParams and AuthOptions are provided") {
 
-                let tokenParams = ARTTokenParams(clientId: options.clientId)
-                let defaultTtl = tokenParams.ttl
-                let defaultCapability = tokenParams.capability
+                it("should supersede configured AuthOptions (using key) even if arguments are empty") {
```
mattheworiordan: This should say even argument objects are empty as saying arguments are empty makes it sound like if you don't pass in arguments as opposed to passing in empty objects

Spec/Auth.swift
```diff
@@ -2091,61 +2091,283 @@ class Auth : QuickSpec {
...
+                            expect(rest.auth.options.authUrl).to(beNil())
+                            expect(rest.auth.options.authParams).to(beNil())
+                            expect(rest.auth.options.authHeaders).to(beNil())
+                            done()
+                        }
+                    }
+
+                    // Repeat
+                    waitUntil(timeout: testTimeout) { done in
+                        rest.auth.authorise(nil, options: nil) { tokenDetails, error in
```
tcard: Like here, you're passing  `options: nil` and I guess you should do the same in the test above.

Spec/Auth.swift
```diff
@@ -2091,61 +2091,283 @@ class Auth : QuickSpec {
...
+                    }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        rest.auth.authorise(tokenParams, options: authOptions) { tokenDetails, error in
+                            expect(error).to(beNil())
+                            guard let tokenDetails = tokenDetails else {
+                                XCTFail("TokenDetails is nil"); done(); return
+                            }
+                            expect(tokenDetails.clientId).to(beNil())
+                            expect(tokenDetails.issued).to(beCloseTo(serverDate, within: 1.0)) //1 Second
```
tcard: This doesn't really prove that `queryTime` was used, since it's really likely that the server time and the local time are within one second of each other. Maybe check that a call to `/time` was made? Really testing this would require that we mock the time locally, and I don't think that's worth it just for this.

tcard: Plus, anything could take longer that 1 second in the server for any reason and then `serverDate` would be outdated, especially given you're using it twice. I'd rather not introduce even more reasons for tests to randomly fail because we expect unguaranteed things from the server.

Spec/Auth.swift
ricardopereira: Done e4e97dd.

Spec/Auth.swift
```diff
@@ -2091,61 +2091,283 @@ class Auth : QuickSpec {
...
+                        rest.auth.authorise(nil, options: authOptions) { _, error in
+                            guard let error = error else {
+                                fail("Error is nil"); done(); return
+                            }
+                            expect(error.description).to(contain("no means to renew the token"))
+                            done()
+                        }
+                    }
+
+                    // Second time
```
ricardopereira: Right. Done d069ed9.

Spec/Auth.swift
ricardopereira: 👍

ricardopereira:
@tcard PTAL

Spec/NSObject+TestSuite.h
```diff
@@ -22,6 +22,9 @@ ART_ASSUME_NONNULL_BEGIN
 /// Get argument at index from the identified instance method.
 - (void)testSuite_getArgumentFrom:(SEL)selector atIndex:(NSInteger)index callback:(void (^)(id))callback;
 
+/// Change the returning value using a date from the identified instance method.
+- (id<AspectToken>)testSuite_returnValueFor:(SEL)selector withDate:(NSDate *)value;
```
tcard: Unused.

PR #493, opened 2016-09-05
RSA10d

EvgenyKarkan:


mattheworiordan:
Not entirely sure I understand the `isOnlyForce` logic, but this change sin 0.9 anyway so I think we should merge anyway

EvgenyKarkan:
> Not entirely sure I understand the `isOnlyForce` logic

`RSA10d` spec states: 
> A special case convenience exists for `AuthOption` stating that if all its attributes are null apart from `force`, then when passed to `authorise` as an argument, the previously configured authentication options will remain intact.

So I had to check that all attributes except `force` are `nil`, store result in `isOnlyForce` variable and use inside scope of `ARTAuth` class.

mattheworiordan:
> So I had to check that all attributes except force are nil, store result in isOnlyForce boolean variable and use inside scope of ARTAuth class.

Ok, thanks, I do follow now, sorry about that!

EvgenyKarkan:

Source/ARTAuth.m
```diff
@@ -184,22 +185,23 @@ - (void)requestToken:(ARTTokenParams *)tokenParams withOptions:(ARTAuthOptions *
...
             tokenDetailsFactory = ^(ARTTokenParams *tokenParams, void(^callback)(ARTTokenDetails *__art_nullable, NSError *__art_nullable)) {
                 // Create a TokenRequest and execute it
-                [self createTokenRequest:currentTokenParams options:mergedOptions callback:^(ARTTokenRequest *tokenRequest, NSError *error) {
+                [self createTokenRequest:currentTokenParams options:replacedOptions callback:^(ARTTokenRequest *tokenRequest, NSError *error) {
                     if (error) {
                         callback(nil, error);
                     } else {
                         [self executeTokenRequest:tokenRequest callback:callback];
                     }
+                    _isOnlyForceTrue = NO;
```
tcard: This change doesn't have any effect, does it? The only other place where `_isOnlyForceTrue` is used is in `authorise`, and  there it's always being set before being read.

mattheworiordan:
@EvgenyKarkan I am aware that this change is a 0.9 change that is breaking i.e. we should not be merging this into `master` as that is currently the 0.8 branch.  Before we merge this, I think we need to start a `0.8` branch and maintain that in case any hot fixes are needed on that branch.

It may be worth having a quick catch up today / tomorrow to discuss our 0.8/0.9 strategy

ricardopereira:

ricardopereira:
Squashed.

ricardopereira:
@mattheworiordan @tcard Are you happy with this?

ricardopereira:
@mattheworiordan Just clarifying, i'm confused. Why are you saying that this is a 0.9 change and should not be merged? From what I saw, the 0.9 removes the `force` property, so this change is only for 0.8, right?

mattheworiordan:
LGTM

PR #494, opened 2016-09-05
RSA10g.

EvgenyKarkan:


Spec/Auth.swift
```diff
@@ -1582,6 +1582,7 @@ class Auth : QuickSpec {
                     authOptions.authParams?.append(NSURLQueryItem(name: "body", value: token))
                     authOptions.authHeaders = ["X-Ably":"Test"]
                     authOptions.force = true
+                    authOptions.queryTime = true
```
mattheworiordan: You've added `queryTime` in, yet there is no check to assert whether the time has been requested from Ably. We need to check this else I don't see the point of this.

EvgenyKarkan: Got it, I'll add an assertion to be sure that `time` function has been called.

mattheworiordan: Thanks

EvgenyKarkan:

ricardopereira:

ricardopereira:
Squashed.

PR #495, opened 2016-09-07
RSA10k

EvgenyKarkan:


mattheworiordan:
This PR addresses some of the points from the previous PRs about testing time which is good ,, thanks.

EvgenyKarkan:
Sorry, should I redo something here?

mattheworiordan:
> Sorry, should I redo something here?

I should have been clearer, I think this PR is fine.  However, given the equivalent Java one that I saw was not actually using the offset, can you think of any way to check that the offset is really being used on subsequent requests?

EvgenyKarkan:
Ok, let me describe one case.
- assume `timeOffset` is stored by `Auth` instance yet, and we are expecting subsequent request in future.
- `AuthOptions` argument `queryTime` property set to `false` or missed as an argument (`false` by default)

I guess you want us also to check such case, right?

mattheworiordan:
> I guess you want us also to check such case for subsequent requests, right?

Well yes, some coverage that demonstrably shows that the offset is effecting the time.  You could inject a fake offset and check it effects token, or if you can think of a better way then that would good too, I just want to be sure the offset is actually used as per the spec.

EvgenyKarkan:

Source/ARTAuth.m
```diff
@@ -337,14 +337,25 @@ - (void)createTokenRequest:(ARTTokenParams *)tokenParams options:(ARTAuthOptions
...
-                callback([mergedTokenParams sign:mergedOptions.key], nil);
-            }
-        }];
+        if (self.timeOffset == nil) {
+            [_rest time:^(NSDate *time, NSError *error) {
+                if (error) {
+                    callback(nil, error);
+                } else {
+                    NSDate *dateNow = [NSDate date];
+                    long long nowMillis = [dateNow timeIntervalSince1970]*1000;
```
tcard: Why do you convert the `NSTimeInterval`s to `long long` and then to `NSNumber` instead of just storing the time offset as a `NSTimeInterval`? And why do you multiply by 1000 instead of working with seconds with decimals, which is what `NSTimeInterval` is?

Source/ARTAuth.m
```diff
@@ -369,6 +380,20 @@ - (NSString *)getClientId {
...
 
+- (void)discardTimeOffset {
+    self.timeOffset = nil;
+}
+
+#pragma mark - Helper
+
+- (NSDate*)timeStampDateWithDateNow:(NSDate*)nowDate {
+    long long appNow = [nowDate timeIntervalSince1970]*1000;
+    long long seconds = ([self.timeOffset longLongValue] + appNow)/1000;
```
tcard: You're losing precision here. `seconds` should be a `NSTimeInterval`, which has millisecond precision, while this just has second precision. If you just use `NSTimeInterval` you can get rid of all the `*1000` and `/1000` too.

Spec/Auth.swift
```diff
@@ -1999,6 +1999,154 @@ class Auth : QuickSpec {
...
+                    let hook2 = ARTRest.aspect_hookSelector(rest)
+                    let _ = try? hook2(#selector(ARTRest.time(_:)), withOptions: .PositionAfter, usingBlock:  unsafeBitCast(block2, ARTRest.self))
+                    
+                    waitUntil(timeout: testTimeout) { done in
+                        rest.auth.createTokenRequest(nil, options: authOptions, callback: { tokenRequest, error in
+                            expect(error).to(beNil())
+                            guard let tokenRequest = tokenRequest else {
+                                XCTFail("TokenRequest is nil"); return
+                            }
+                            expect(tokenRequest.timestamp).toNot(beNil())
```
tcard: You should also check that this has the expected value within a margin of error, ie. the first token request's timestamp plus the some offset. (You can wrap this in a `delay(1) { ... }` to make the difference more apparent.)

Spec/Auth.swift
```diff
@@ -1999,6 +1999,154 @@ class Auth : QuickSpec {
...
+                    authOptions.queryTime = true
+                    
+                    waitUntil(timeout: testTimeout) { done in
+                        rest.auth.createTokenRequest(nil, options: authOptions, callback: { tokenRequest, error in
+                            expect(error).to(beNil())
+                            guard let tokenRequest = tokenRequest else {
+                                XCTFail("tokenRequest is nil"); return
+                            }
+                            expect(rest.auth.timeOffset).toNot(beNil())
+                            rest.auth.discardTimeOffset()
```
mattheworiordan: This is not needed.  It's saying the it's possible that the "Client" should be able to discard the local clock offset, not the user.  Anyway, leave now, but this spec item was simply a statement to make it clear that the client library developer when writing the client could under certain circumstances choose to discard the offset internally.

ricardopereira:

ricardopereira:
Squashed.

ricardopereira:
Not passing. I think some pieces are missing.

ricardopereira:
> A client MAY discard the cached local clock offset in situations in which it may have been invalidated, such as if there is a local change to the date, time, or timezone, of the client device

 - [`UIApplicationSignificantTimeChangeNotification `](https://developer.apple.com/reference/uikit/uiapplicationsignificanttimechangenotification) for example, change to a new day (midnight), carrier time update (iOS only!)
 - [`NSCurrentLocaleDidChangeNotification `](https://developer.apple.com/reference/foundation/nscurrentlocaledidchangenotification) when the user’s locale changed

mattheworiordan:
Thanks @ricardopereira. TBH actually handling the changing timezones is not something we worried about hence why it's deliberately unspecified in the lib, but glad you've done it so worth merging in now.

ricardopereira:
@tcard PTAL

mattheworiordan:
@tcard are you happy for this to be merged?

Source/ARTAuth.m
```diff
@@ -355,18 +390,26 @@ - (void)createTokenRequest:(ARTTokenParams *)tokenParams options:(ARTAuthOptions
...
-        [_rest time:^(NSDate *time, NSError *error) {
-            if (error) {
-                callback(nil, error);
-            } else {
-                currentTokenParams.timestamp = [self handleServerTime:time];
-                callback([currentTokenParams sign:replacedOptions.key], nil);
-            }
-        }];
-    } else {
+    if (_timeOffset) {
```
ricardopereira: Fixed 3588ff0.

ricardopereira:
Can I squash and rebase?

Issue #485, opened 2016-08-31
Channel still has FAILED state after .attach() call.


EvgenyKarkan:
As discussed in `#client-libraries` with @ricardopereira
RTL4g spec item implementation is wrong.

ricardopereira:
Fixed.

PR #497, opened 2016-09-13
Fix for RTL4g

EvgenyKarkan:
#485


tcard:
The test for RTL4b for the SUSPENDED state is not passing.

EvgenyKarkan:
Thanks, interesting, yesterday it was green and today after 2 attempts to run tests got green too http://take.ms/tzjsx. 
Will try to find out the reason.

PR #500, opened 2016-09-15
RSC15e as a part of 0.9

EvgenyKarkan:

### DIFF
--- a/Source/ARTRest.m
+++ b/Source/ARTRest.m
@@ -154,7 +154,7 @@
             }
         }
         if (retries < _options.httpMaxRetryCount && [self shouldRetryWithFallback:request response:response error:error]) {
-            if (!blockFallbacks && [request.URL.host isEqualToString:(_prioritizedHost ? _prioritizedHost : [ARTDefault restHost])]) {
+            if (!blockFallbacks && [request.URL.host isEqualToString:(_prioritizedHost ? _prioritizedHost : (_options.restHost ? : [ARTDefault restHost]))]) {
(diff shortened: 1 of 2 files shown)


EvgenyKarkan:

EvgenyKarkan:
@mattheworiordan I was sure that test for what you describing [already exists](https://github.com/ably/ably-ios/blob/master/Spec/RestClient.swift#L618-L648), no?
In my PR I only added test for custom rest host, to be sure that request will try it instead of default `rest.ably.io`. 
Could you please take a look again, do I still need any changes here?

mattheworiordan:
> In my PR I only added test for custom rest host, to be sure that request will try it instead of default rest.ably.io. 

But I am not really clear why this test is needed then given providing a custom `restHost` is covered in many places, including of course http://docs.ably.io/client-lib-development-guide/features/#TO3k2.  We know that restHost is used by those tests, so I am not clear what we are achieving then if a test for http://docs.ably.io/client-lib-development-guide/features/#TO3k2 already exists?

Can you explain as I feel like we're just duplicating tests unnecessarily now?

EvgenyKarkan:
You are right, I see that there are couple of places to test overridden `restHost`.
But please understand me, I just wanted to test it after I read this diff from 0.8:
`(unless overriden in ClientOptions#restHost)`.
So I will close this useless PR, thanks for the feedback @mattheworiordan. :)

mattheworiordan:
Ok, no problem. I am hoping to set up a call with Dmitry and all of you shortly so that we can discuss next steps as perhaps some guidance of expected test coverage would be sensible moving forwards.

EvgenyKarkan:
Closed as discussed above.

PR #501, opened 2016-09-15
RSC15a as a part of 0.9

EvgenyKarkan:

### DIFF
--- a/Spec/RestClient.swift
+++ b/Spec/RestClient.swift
@@ -648,7 +648,7 @@ class RestClient: QuickSpec {
                 }
                 
                 // RSC15b
-                it("won't apply custom fallback hosts if ClientOptions#fallbackHosts array is empty, use defaults instead") {
+                it("won't apply fallback hosts if ClientOptions#fallbackHosts array is empty") {
                     let options = ARTClientOptions(key: "xxxx:xxxx")
                     options.fallbackHosts = [] //to test TO3k6
                     let client = ARTRest(options: options)
@@ -659,11 +659,6 @@ class RestClient: QuickSpec {
                     var capturedURLs = [String]()
                     testHTTPExecutor.afterRequest = { request, callback in
                         capturedURLs.append(request.URL!.absoluteString)
-                        if testHTTPExecutor.requests.count == 2 {
-                            // Stop
-                            testHTTPExecutor.http = nil
-                            callback!(nil, nil, nil)
-                        }
                     }
                     
                     waitUntil(timeout: testTimeout) { done in
@@ -672,12 +667,8 @@ class RestClient: QuickSpec {
                         }
                     }
                     
-                    expect(testHTTPExecutor.requests).to(haveCount(2))
-                    if testHTTPExecutor.requests.count < 2 {
-                        return
-                    }
-                    
-                    expect(NSRegularExpression.match(capturedURLs[1], pattern: "//[a-e].ably-realtime.com")).to(beTrue())
+                    expect(testHTTPExecutor.requests).to(haveCount(1))
+                    expect(NSRegularExpression.match(capturedURLs[0], pattern: "//rest.ably.io")).to(beTrue())
                 }
                 
                 // RSC15b
--- a/Spec/RestClient.swift
+++ b/Spec/RestClient.swift
@@ -876,7 +876,7 @@ class RestClient: QuickSpec {
                         expect(resultFallbackHosts).to(equal(expectedFallbackHosts))
                     }
                     
-                    it("until httpMaxRetryCount has been reached, if an empty array of fallback hosts is provided, then fallback host functionality is disabled, use defaults instead") {
+                    it("if an empty array of fallback hosts is provided, then fallback host functionality is disabled") {
                         let options = ARTClientOptions(key: "xxxx:xxxx")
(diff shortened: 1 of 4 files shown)


Spec/RestClient.swift
```diff
@@ -824,6 +822,128 @@ class RestClient: QuickSpec {
...
+                        let extractHostname = { (request: NSMutableURLRequest) in
+                            NSRegularExpression.extract(request.URL!.absoluteString, pattern: "[f-j].ably-realtime.com")
+                        }
+                        let resultFallbackHosts = testHTTPExecutor.requests.flatMap(extractHostname)
+                        let expectedFallbackHosts = expectedHostOrder.map { _fallbackHosts[$0] }
+                        
+                        expect(resultFallbackHosts).to(equal(expectedFallbackHosts))
+                    }
+                    
+                    it("until httpMaxRetryCount has been reached, if an empty array of fallback hosts is provided, then fallback host functionality is disabled, use defaults instead") {
```
mattheworiordan: I don't think is right.  If an empty array or fallback hosts is configured, then according to http://docs.ably.io/client-lib-development-guide/features/#RSC15a, ` If an empty array of fallback hosts is provided, then fallback host functionality is disabled`. It appears that you are instead using the defaults?

Spec/RestClient.swift
```diff
@@ -824,6 +822,128 @@ class RestClient: QuickSpec {
...
+                        let extractHostname = { (request: NSMutableURLRequest) in
+                            NSRegularExpression.extract(request.URL!.absoluteString, pattern: "[a-e].ably-realtime.com")
+                        }
+                        let resultFallbackHosts = testHTTPExecutor.requests.flatMap(extractHostname)
+                        let expectedFallbackHosts = Array(expectedHostOrder.map({ ARTDefault.fallbackHosts()[$0] as! String })[0..<Int(options.httpMaxRetryCount)])
+                        
+                        expect(resultFallbackHosts).to(equal(expectedFallbackHosts))
+                    }
+                    
+                    it("until all fallback hosts have been tried, if an empty array of fallback hosts is provided, then fallback host functionality is disabled, use defaults instead") {
```
mattheworiordan: Same problem here, it appears you are using the defaults instead of disabling the fallback functionality

Spec/RestClient.swift
mattheworiordan: Sorry, I see this is fixed in b0ae3c6

EvgenyKarkan: > I see this is fixed in b0ae3c6  

Yes, exactly, thanks for feedback.

Spec/RestClient.swift
```diff
@@ -815,6 +813,94 @@ class RestClient: QuickSpec {
 
                         expect(resultFallbackHosts).to(equal(expectedFallbackHosts))
                     }
+                    
+                    let _fallbackHosts = ["f.ably-realtime.com", "g.ably-realtime.com", "h.ably-realtime.com", "i.ably-realtime.com", "j.ably-realtime.com"]
+                    
+                    it("until httpMaxRetryCount has been reached, if custom fallback hosts are provided in ClientOptions#fallbackHosts, then they will be used instead") {
+                        let options = ARTClientOptions(key: "xxxx:xxxx")
+                        options.httpMaxRetryCount = 5
```
mattheworiordan: Given you've provided 5, why not set this to 4 for example to ensure that httpMaxRetryCount is working as opposed to the max of hosts available is being used?

EvgenyKarkan:

EvgenyKarkan:
@mattheworiordan could you please find some spare time to take a look at 1cef166 ?

Spec/RestClient.swift
```diff
@@ -839,6 +839,7 @@ class RestClient: QuickSpec {
                         }
                         
                         expect(testHTTPExecutor.requests).to(haveCount(Int(1 + options.httpMaxRetryCount)))
+                        expect((testHTTPExecutor.requests.count) < (_fallbackHosts.count + 1)).to(beTrue())
```
mattheworiordan: Why not `expect(testHTTPExecutor.requests).to(haveCount(Int(1 + options.httpMaxRetryCount)))` though?

EvgenyKarkan: Agree, I'm using local variable `_fallbackHosts` and its `count` property. 
Assert based on that hard-coded array is useless, because who knows, how big an array can be. 

If I use assert based on `options.fallbackHosts.count` like that: 
`expect((testHTTPExecutor.requests.count) < (options.fallbackHosts.count + 1)).to(beTrue())` 

I can check that count of executed requests (which is 5 = 1 to primary host + 4 to fallback hosts, cuz I use `httpMaxRetryCount = 4` value) is less than max available count (which is 6 = 1 to primary + 5 to fallback hosts).

So I just wanted to be sure that reduce in retry count from 5 to 4 - works as expected, does it make sense @mattheworiordan ?

ricardopereira: @EvgenyKarkan on that line you're checking the `_fallbackHosts`, not the `options.fallbackHosts`. You already verifying the number of requests of the `testHTTPExecutor` and the `_fallbackHosts` is a local var of the test suite that we control, so I think that line is useless.

ricardopereira:
Rebased and tests are passing.

PR #502, opened 2016-09-16
RTN17b as a part of 0.9

EvgenyKarkan:


EvgenyKarkan:
Yes, there is already a test for that case.

PR #503, opened 2016-09-19
RTN17a as a part of 0.9

EvgenyKarkan:


mattheworiordan:
Ok, this looks fine, but again the change in text for that spec item RTN17a was actually not a change in behaviour. We simply added it to provide clarity on the spec as that requirement was already covered by http://docs.ably.io/client-lib-development-guide/versions/features-0-8/#RTC1d and http://docs.ably.io/client-lib-development-guide/versions/features-0-8/#TO3k3, and by the original spec http://docs.ably.io/client-lib-development-guide/versions/features-0-8/#RTN17a

EvgenyKarkan:
Ok, if you don't mind I'll close this PR , because such behaviour already tested in `RTC1d`, where "fake.ably.io" used as `realtimeHost` value.

mattheworiordan:
OK

PR #513, opened 2016-10-13
TO3k7

ricardopereira:


Spec/RestClient.swift
```diff
@@ -565,6 +565,72 @@ class RestClient: QuickSpec {
...
             context("Host Fallback") {
 
+                // TO3k7
+                context("fallbackHostsUseDefault option") {
+
+                    it("allows the default fallback hosts to be used when @environment@ is not production") {
+                        let options = ARTClientOptions(key: "xxxx:xxxx")
+                        options.environment = "not-production"
+                        options.fallbackHosts = ["fakeA.ably.io", "fakeB.ably.io"]
+                        options.fallbackHostsUseDefault = true
```
mattheworiordan: It is never valid to use default and provide fallback hosts.  Why are these both set?  You are either using the defaults or using the provided fallbacks, but never both

Spec/RestClient.swift
```diff
@@ -565,6 +565,72 @@ class RestClient: QuickSpec {
...
+                            expect(fallback.hosts).to(contain($0))
+                        }
+                    }
+
+                    it("allows the default fallback hosts to be used when a custom Realtime or REST host endpoint is being used") {
+                        let options = ARTClientOptions(key: "xxxx:xxxx")
+                        options.restHost = "fake1.ably.io"
+                        options.realtimeHost = "fake2.ably.io"
+                        options.fallbackHosts = ["fakeA.ably.io", "fakeB.ably.io"]
+                        options.fallbackHostsUseDefault = true
```
mattheworiordan: As above, this says "allows the default fallback hosts to be used" yet you're providing fallback hosts in the Array

Spec/RestClient.swift
```diff
@@ -565,6 +565,72 @@ class RestClient: QuickSpec {
...
+                    }
+
+                    it("should never accept to configure @fallbackHost@ and set @fallbackHostsUseDefault@ to @true@") {
+                        let options = ARTClientOptions(key: "xxxx:xxxx")
+                        expect(options.fallbackHosts).to(beNil())
+                        expect(options.fallbackHostsUseDefault).to(beFalse())
+
+                        options.fallbackHostsUseDefault = true
+                        options.fallbackHosts = []
+                        expect(options.fallbackHosts).toNot(beNil())
```
mattheworiordan: I think this should raise an exception rather than silently switch out the values.  If the user is providing both, they have misunderstood the options and should be shown an exception

ricardopereira: Done 255ff0c.

ricardopereira:
@mattheworiordan PTAL

PR #518, opened 2016-10-17
RSA4b

ricardopereira:


ricardopereira:
🚦 If the spec is well implemented then I will fix the client code.

Spec/Auth.swift
```diff
@@ -185,6 +185,67 @@ class Auth : QuickSpec {
                         expect(client.auth.method).to(equal(ARTAuthMethod.Token))
                     }
                 }
+
+                // RSA4b
+                it("in REST, if the token creation failed or the subsequent request with the new token failed due to a token error, then the request should result in an error") {
+                    let options = AblyTests.clientOptions()
+                    options.key = AblyTests.commonAppSetup().key
+                    options.token = getTestToken()
```
mattheworiordan: I am not following why you provide a token.  This spec states "When the client does have a means to renew the token automatically" yet you've provided an explicit token and thus no means to renew the token?

Spec/Auth.swift
```diff
@@ -185,6 +185,67 @@ class Auth : QuickSpec {
...
+                            done()
+                        }
+                    }
+                }
+
+                // RSA4b
+                it("in Realtime, if the token creation failed or the subsequent request with the new token failed due to a token error, then the connection should move to the FAILED state") {
+                    let options = AblyTests.clientOptions()
+                    options.key = AblyTests.commonAppSetup().key
+                    options.token = getTestToken(ttl: 0.5)
```
mattheworiordan: As above, providing a token explicitly is not a means to renew (authCallback or authUrl should be used ideally to demonstrate the point)

Spec/Auth.swift
ricardopereira: Fixed 5687309.

Spec/Auth.swift
ricardopereira: Fixed 5687309.

ricardopereira:
@mattheworiordan Still failing. PTAL and if it's ok, then I fix the client.

Spec/Auth.swift
```diff
@@ -188,9 +188,7 @@ class Auth : QuickSpec {
 
                 // RSA4b
                 it("in REST, if the token creation failed or the subsequent request with the new token failed due to a token error, then the request should result in an error") {
-                    let options = AblyTests.clientOptions()
-                    options.key = AblyTests.commonAppSetup().key
-                    options.token = getTestToken()
+                    let options = AblyTests.commonAppSetup()
```
tcard: This still won't have a means to renew the token or even use token auth, as it will use the key.

Spec/Auth.swift
```diff
@@ -215,35 +213,24 @@ class Auth : QuickSpec {
...
-
-                        channel.publish("testing", data: nil) { error in
-                            guard let error = error else {
-                                fail("Error is nil"); done(); return
+                        realtime.connection.once(.Failed) { stateChange in
+                            guard let errorInfo = stateChange?.reason else {
+                                fail("ErrorInfo is nil"); done(); return
                             }
-                            expect(error.code).to(equal(40141))
+                            print(errorInfo.code)
```
tcard: Print?

Spec/Auth.swift
```diff
@@ -185,6 +185,54 @@ class Auth : QuickSpec {
...
+                                fail("Error is nil"); done(); return
+                            }
+                            expect(error.code).to(equal(40141))
+                            done()
+                        }
+                    }
+                }
+
+                // RSA4b
+                it("in Realtime, if the token creation failed or the subsequent request with the new token failed due to a token error, then the connection should move to the FAILED state") {
```
mattheworiordan: This description is wrong it's the wrong spec:
```
If the token creation failed or the subsequent request with the new token failed due to a token error, then the request should result in an error
```
It does not become failed, it only becomes failed if the connection fails with the provided token. If the token generation fails, it moves to disconnected and reports the error.  Unfortunately this is entirely wrong.

ricardopereira:
@mattheworiordan PTAL. If it's ok then I'll fix the client.

mattheworiordan:
Thanks, LGTM, 👍

ricardopereira:
Tests are passing now but I'm not sure about one thing.

Source/ARTWebSocketTransport.m
```diff
@@ -114,7 +114,14 @@ - (void)connectForcingNewToken:(BOOL)forceNewToken {
...
                 if (error) {
                     [selfStrong.logger error:@"R:%p WS:%p ARTWebSocketTransport: token auth failed with %@", _delegate, self, error.description];
-                    [selfStrong.delegate realtimeTransportFailed:selfStrong withError:[[ARTRealtimeTransportError alloc] initWithError:error type:ARTRealtimeTransportErrorTypeAuth url:self.websocketURL]];
+                    if (error.code == 40102 /*incompatible credentials*/) {
+                        // RSA15c
+                        [selfStrong.delegate realtimeTransportFailed:selfStrong withError:[[ARTRealtimeTransportError alloc] initWithError:error type:ARTRealtimeTransportErrorTypeAuth url:self.websocketURL]];
+                    }
+                    else {
+                        // RSA4b
+                        [selfStrong.delegate realtimeTransportDisconnected:selfStrong withError:[[ARTRealtimeTransportError alloc] initWithError:error type:ARTRealtimeTransportErrorTypeAuth url:self.websocketURL]];
```
ricardopereira: @mattheworiordan I'm not sure about this change I made.

Regarding [one of your comments](https://github.com/ably/ably-ios/pull/518#discussion_r85763133), I'm confused with [`RSA15c`](http://docs.ably.io/client-lib-development-guide/features/#RSA15c), it states that the Realtime should transition the connection state to FAILED which is incoherent with what you said.

mattheworiordan: If you're referring to https://github.com/ably/ably-ios/pull/518#discussion-diff-85763133R216 then that's a different spec

Spec/Auth.swift
```diff
@@ -245,6 +245,89 @@ class Auth : QuickSpec {
...
+                            }
+                            expect(errorInfo.message).to(contain("server with the specified hostname could not be found"))
+                            done()
+                        }
+                        realtime.connect()
+                    }
+                }
+
+                // RSA4b
+                it("in Realtime, if the connection fails to a token error, then the connection should move to the FAILED state and reports the error") {
```
mattheworiordan: I think this should say "fails due to a terminal token error" to make it clear this is different to https://github.com/ably/ably-ios/pull/518/files#diff-b5d84fe9bce26902a7481c2ffec84fffR279

ricardopereira: Done.

PR #519, opened 2016-10-18
RSA4c

ricardopereira:


ricardopereira:
🚦 If the spec is well implemented then I will fix the client code.

Spec/Auth.swift
```diff
@@ -205,6 +205,272 @@ class Auth : QuickSpec {
...
+                            // Token should renew and fail
+                            realtime.options.authParams = [NSURLQueryItem]()
+
+                            expect(realtime.connection.errorReason).toEventuallyNot(beNil(), timeout: testTimeout)
+                            guard let errorInfo = realtime.connection.errorReason else {
+                                fail("ErrorInfo is empty"); return
+                            }
+                            expect(errorInfo.code) == 80019
+
+                            expect(realtime.connection.state).to(equal(ARTRealtimeConnectionState.Connected))
```
mattheworiordan: I don't think this test can pass until http://docs.ably.io/client-lib-development-guide/features/#RTC8a is implemented. Currently the token would expire, Ably would disconnect the connection, and then the connection would become disconnected.  Instead what you want is for the connection be made with a token that does not expire, you then inject an `AUTH ProtocolMessage` so the client thinks the server has requested it to authorise, it authorises fails and then just leaves the connection in the CONNECTED state waiting for the server to disconnect it eventually.

Spec/Auth.swift
```diff
@@ -205,6 +205,272 @@ class Auth : QuickSpec {
...
+                            }
+
+                            guard let errorInfo = realtime.connection.errorReason else {
+                                fail("ErrorInfo is empty"); return
+                            }
+                            expect(errorInfo.code) == 80019
+                        }
+
+                        // RSA4c3
+                        it("if the connection is CONNECTED, then the connection should remain CONNECTED") {
```
mattheworiordan: As above, not sure this will work now, perhaps save for later as part of RTC8a

Spec/Auth.swift
```diff
@@ -205,6 +205,272 @@ class Auth : QuickSpec {
...
+                                    guard let errorInfo = stateChange.reason else {
+                                        fail("ErrorInfo is nil"); done(); return
+                                    }
+                                    expect(errorInfo.code) == 80019
+                                    done()
+                                }
+                            }
+
+                            guard let errorInfo = realtime.connection.errorReason else {
+                                fail("ErrorInfo is empty"); return
```
mattheworiordan: Might be nice to check for a reason that indicates a timeout here

Spec/Auth.swift
```diff
@@ -205,6 +205,272 @@ class Auth : QuickSpec {
...
+                                    done()
+                                }
+                            }
+
+                            // Token should renew and fail
+                            realtime.options.authCallback = { tokenParams, completion in
+                                completion(nil, NSError(domain: NSURLErrorDomain, code: -1003, userInfo: [NSLocalizedDescriptionKey: "A server with the specified hostname could not be found."]))
+                            }
+
+                            expect(realtime.connection.errorReason).toEventuallyNot(beNil(), timeout: testTimeout)
```
tcard: As @mattheworiordan says above, this should happen because of an incoming AUTH and not because the token expires, because the token expiring, as per [RTN15a](http://docs.ably.io/client-lib-development-guide/features/#RTN15a), causes a transition to DISCONNECTED right away instead of staying CONNECTED.

Spec/Auth.swift
```diff
@@ -205,6 +205,272 @@ class Auth : QuickSpec {
...
+                                }
+                            }
+
+                            // Token should renew and fail
+                            let invalidToken = String(token.characters.reverse())
+                            realtime.options.authParams = [NSURLQueryItem]()
+                            realtime.options.authParams?.append(NSURLQueryItem(name: "type", value: "text"))
+                            realtime.options.authParams?.append(NSURLQueryItem(name: "body", value: invalidToken))
+
+                            expect(realtime.connection.errorReason).toEventuallyNot(beNil(), timeout: testTimeout)
```
tcard: Same here:

> As @mattheworiordan says above, this should happen because of an incoming AUTH and not because the token expires, because the token expiring, as per RTN15a, causes a transition to DISCONNECTED right away instead of staying CONNECTED.

Spec/Auth.swift
```diff
@@ -205,6 +205,272 @@ class Auth : QuickSpec {
...
+                            expect(realtime.connection.state).to(equal(ARTRealtimeConnectionState.Connected))
+                        }
+                    }
+
+                    context("the attempt times out after realtimeRequestTimeout") {
+                        // RSA4c1 & RSA4c2
+                        it("if the connection is CONNECTING, then the connection attempt should be treated as unsuccessful") {
+                            let previousRealtimeRequestTimeout = ARTDefault.realtimeRequestTimeout()
+                            defer { ARTDefault.setRealtimeRequestTimeout(previousRealtimeRequestTimeout) }
+                            ARTDefault.setRealtimeRequestTimeout(0.5)
```
tcard: Why 0.5? It could even be 0.0.

Spec/Auth.swift
```diff
@@ -205,6 +205,272 @@ class Auth : QuickSpec {
...
+                                    done()
+                                }
+                            }
+
+                            // Token should renew and fail
+                            realtime.options.authCallback = { tokenParams, completion in
+                                // Ignore `completion` closure to force a time out
+                            }
+
+                            expect(realtime.connection.errorReason).toEventuallyNot(beNil(), timeout: testTimeout)
```
tcard: Same here:

> As @mattheworiordan says above, this should happen because of an incoming AUTH and not because the token expires, because the token expiring, as per RTN15a, causes a transition to DISCONNECTED right away instead of staying CONNECTED.

mattheworiordan:
@ricardopereira this looks like there are still outstanding tasks?

ricardopereira:
I need to rebase this because it's missing the `RTC8a`.

Spec/Auth.swift
ricardopereira: Rebased.

Spec/Auth.swift
ricardopereira: Done ab3e321.

Source/ARTRealtime.m
```diff
@@ -268,11 +282,12 @@ - (void)transition:(ARTRealtimeConnectionState)state withErrorInfo:(ARTErrorInfo
 
 - (void)transitionSideEffects:(ARTConnectionStateChange *)stateChange {
     ARTStatus *status = nil;
+    __weak __typeof(self) weakSelf = self;
```
tcard: This kind of things could really use an explanatory comment. I have no idea why this change. I mean, I assume you don't want to increase the reference count in all those callback blocks but I'm not sure if there's some subtle reason for that.

Source/ARTRealtime.m
```diff
@@ -536,13 +569,26 @@ - (void)onClosed {
...
 
+- (void)onAuth {
+    [self.logger info:@"R:%p server has requested an authorise", self];
+    switch (self.connection.state) {
+        case ARTRealtimeConnecting:
+        case ARTRealtimeConnected:
+            [self transportReconnectWithRenewedToken];
+            break;
+        default:
+            NSAssert(false, @"Invalid Realtime state: expected Connecting or Connected");
```
tcard: I would reserve those assertions for bugs _on the library_. If e. g. the server sends an AUTH while the connection is closing, that would be a bug in the server, and we probably should treat those more gracefully than crashing, e. g. log and ignore.

Source/ARTRealtime.m
```diff
@@ -567,12 +613,43 @@ - (void)transportReconnectWithHost:(NSString *)host {
...
+                return;
+            default:
+                break;
+            }
+            [[weakSelf getLogger] debug:__FILE__ line:__LINE__ message:@"R:%p WS:%p authorised: %@ error: %@", weakSelf, [weakSelf getTransport], tokenDetails, error];
+            [weakSelf transportReconnectWithTokenDetails:tokenDetails error:error];
+        }];
+    }
+    @finally {
+        self.auth.delegate = self;
```
tcard: Also no idea why this.

Source/ARTRealtime.m
```diff
@@ -536,13 +569,26 @@ - (void)onClosed {
...
+}
+
 - (void)onError:(ARTProtocolMessage *)message {
-    // TODO work out which states this can be received in
     if (message.channel) {
         [self onChannelMessage:message];
     } else {
         ARTErrorInfo *error = message.error;
         if ([self shouldRenewToken:&error]) {
+            [self.transport close];
```
tcard: Then it's closed again at 621?

Source/ARTRealtime.m
```diff
@@ -567,12 +613,43 @@ - (void)transportReconnectWithHost:(NSString *)host {
...
+        return;
+    }
     [_transport close];
     _transport = [[_transportClass alloc] initWithRest:self.rest options:self.options resumeKey:_transport.resumeKey connectionSerial:_transport.connectionSerial];
     _transport.delegate = self;
-    [_transport connectForcingNewToken:true];
+    [_transport connectWithToken:tokenDetails.token error:error];
+}
+
+- (void)transportReconnectWithRenewedToken {
```
tcard: Unless I'm missing something, all this logic will only apply when `transportReconnectWithRenewedToken` is called, ie. when an AUTH or ERROR is received, but not when e. g. first connection or manually calling `authorize`. The spec says "If an attempt by the realtime client library to authenticate is made using...", so _every_ auth attempt should follow this logic. I'd say this logic should be inside `Auth.authorize` itself, don't you think?

Spec/RealtimeClientConnection.swift
```diff
@@ -3583,6 +3623,51 @@ class RealtimeClientConnection: QuickSpec {
                     }
                 }
             }
+
+            it("should abort reconnection with new token if the server has requested it to authorise and after it the connection has been closed") {
```
tcard: Which spec lines is this testing?

Source/ARTRealtime.m
ricardopereira: Yes, you're right, I don't want to increase the ref count on those blocks because, for example, the `unlessStateChangesBefore` is setting a timer and if the `ARTRealtime` instance is released before that timer, then it could create a retain cycle and then we have a leak. That could also bring some instability in the test suite because we are creating a lot of `ARTRealtime` instances. I will add a comment about it. Thanks.

Source/ARTRealtime.m
ricardopereira: I can't precise the problem that was occurring if the `close` wasn't called explicitly but the main reason I added the `close` call was because the `authorize` from `transportReconnectWithRenewedToken` takes time and the client shouldn't use the current token.

Source/ARTRealtime.m
ricardopereira: `ARTAuth` doesn't know about `ARTRealtime` so it has a delegate (`ARTAuthDelegate`). `ARTRealtime` is using that delegate but in that case it should not use the delegate functionalities.

ricardopereira: Actually, your [comment above](https://github.com/ably/ably-ios/pull/519/files/0c86eb8424812f2fa2cbe8fd092b8341fbaacd85#r91504530) is relevant about this case.

Spec/RealtimeClientConnection.swift
ricardopereira: None. That was something I notice on the test suite where the client gets connected when it's waiting for the `authorize` and the client is explicitly closed. It should ignore it and remain closed.

Source/ARTRealtime.m
ricardopereira: Done 6939c90.

Source/ARTRealtime.m
ricardopereira: Done b6d6f92.

Source/ARTRealtime.m
mattheworiordan: @ricardopereira have you made the change on this?

ricardopereira: @mattheworiordan No, working on it right now.

Source/ARTRealtime.m
ricardopereira: Done efa57ab.

Source/ARTRealtime.m
ricardopereira: Done efa57ab.

tcard: That one broke some tests, unfortunately...

ricardopereira: 😔 Yes, I'm on it. Thanks

ricardopereira: Fixed 3462531.

ricardopereira:
The `authorize` should be called before the transport connects because of the _**(`RSA4c3`) If the connection is CONNECTED, then the connection should remain CONNECTED**_. So, I removed the `ARTAuth` dependency from the `ARTWebSocketTransport` and for now on the `ARTRealtime` manages the authentication.

@mattheworiordan @tcard Can I squash and merge?

mattheworiordan:
Seeing as though it's passing LGTM

PR #520, opened 2016-10-18
Update RSA10a for 0.9

ricardopereira:


Spec/Auth.swift
```diff
@@ -1617,7 +1617,7 @@ class Auth : QuickSpec {
         describe("authorise") {
 
             // RSA10a
-            it("should create a token if needed and use it") {
+            it("should reuse the valid token and ensure Token Auth is used for all future requests") {
```
mattheworiordan: Nope, it should not reused the valid token, but should create a token and ensure Token Auth...

ricardopereira: Fixed ed52009.

ricardopereira:
If the spec looks good then I'll fix the client.

tcard:
The spec LGTM.

PR #521, opened 2016-10-18
Update RSA10j for 0.9

ricardopereira:


Spec/Auth.swift
```diff
@@ -2441,6 +2441,26 @@ class Auth : QuickSpec {
                     }
                 }
 
+                it("example: if is initialised with custom ttl and a TokenParams object is passed with null, then the @ttl@ used for every subsequent authorization will be null") {
+                    let options = AblyTests.commonAppSetup()
+                    options.defaultTokenParams = { $0.ttl = 0.1; return $0 }(ARTTokenParams())
```
tcard: Cute :)

Spec/Auth.swift
```diff
@@ -2441,6 +2441,26 @@ class Auth : QuickSpec {
...
+                    let rest = ARTRest(options: options)
+
+                    waitUntil(timeout: testTimeout) { done in
+                        rest.auth.authorise(nil, options: nil) { tokenDetails, error in
```
tcard: The spec says: "and a TokenParams object is passed in as an argument to #authorize with a null or missing value for ttl" but this is using a null TokenParams. It's the TTL which should be null or missing. Of course, that's not possible in iOS, so you should probably use other TokenParams field which can be null, or define zero as "missing" for the TTL field and test that.

This is also missing the "every subsequent authorization" part. You're testing no subsequent authorizations; at least one would be good.

Spec/Auth.swift
ricardopereira: Nice right? 😁 I liked it too.

Spec/Auth.swift
ricardopereira: Done 6a99e08.

Now the test fails with `<Error Domain=io.ably.cocoa Code=40101 "" UserInfo={ARTErrorStatusCode=401, NSLocalizedDescription=, NSLocalizedFailureReason=Request mac does not match}>`.

The `ttl` spec says that _when the `ttl` is omitted, the REST API default of 60 minutes is applied by Ably_, so should the client set the default value if the `ttl` is missing (ttl=0)  or should the attribute be omitted on the signing method? I'm mean [here](https://github.com/ably/ably-ios/blob/6a99e08d5bba78dcdcbc9242601bd0ad6bda37a7/Source/ARTTokenParams.m#L157).

ricardopereira: @tcard Any thoughts?

tcard: @ricardopereira It should use the default, of course.

I'm not sure why the MAC signing doesn't work. Can you arrange [this](https://github.com/ably/ably-ios/blob/6a99e08d5bba78dcdcbc9242601bd0ad6bda37a7/Source/ARTTokenParams.m#L64-L65) so that it adds an empty clientId param there (`&clientId=&otherParam=...`) and try again?

mattheworiordan: Ignore my previous comment about not setting the TTL, in fact I do do that in the Ruby lib as well :/ https://github.com/ably/ably-ruby/blob/master/lib/ably/auth.rb#L34

Spec/Auth.swift
```diff
@@ -2441,6 +2441,59 @@ class Auth : QuickSpec {
                     }
                 }
 
+                it("example: if is initialised with custom ttl and a TokenParams object is passed with null, then the @ttl@ used for every subsequent authorization will be null") {
```
mattheworiordan: This wording is just weird and does not reflect what http://docs.ably.io/client-lib-development-guide/features/#RSA10j says.

Why not just use the actual wording from the spec? "if a client is initialised with TokenParams#ttl configured with a custom value, and a TokenParams object is passed in as an argument to #authorize with a null value for ttl, then the ttl used for every subsequent authorization will be null"

PR #526, opened 2016-10-19
RTC8

ricardopereira:


ricardopereira:
🚦 I'll implement the new features when all the implemented specs are approved.

ricardopereira:
@mattheworiordan @tcard PTAL

Spec/RealtimeClient.swift
```diff
@@ -352,6 +352,478 @@ class RealtimeClient: QuickSpec {
...
+
+                    waitUntil(timeout: testTimeout) { done in
+                        let partialDone = AblyTests.splitDone(2, done: done)
+
+                        client.connection.once(.Connected) { stateChange in
+                            guard let stateChange = stateChange else {
+                                fail("ConnectionStateChange is nil"); partialDone(); return
+                            }
+                            expect(stateChange.previous).to(equal(ARTRealtimeConnectionState.Connected))
+                            expect(stateChange.reason).to(beNil())
```
mattheworiordan: I don't think you're checking sufficiently that the upgrade has occurred.  I would attempt to attach to channel `foo` with the original token, that should fail because of lack of permissions, then upgrade the token and clientId (we allow upgrade from anonymous to identified), then attempt attach to `foo` which should now work and also check the CONNECTED ConnectionDetails clientId is set correctly.

Spec/RealtimeClient.swift
```diff
@@ -352,6 +352,478 @@ class RealtimeClient: QuickSpec {
...
+                            guard let errorMessage = errorMessages.first else {
+                                fail("Missing ERROR protocol message"); partialDone(); return
+                            }
+                            expect(errorMessage.channel).to(contain("test"))
+                            expect(errorMessage.error?.code).to(equal(error.code))
+                            partialDone()
+                        }
+
+                        let tokenParams = ARTTokenParams()
+                        tokenParams.capability = "{\"test\":[\"subscribe\"]}"
```
mattheworiordan: If the new token capability provides access to `test`, and you are attached to `test`, this should not fail?  Something appears to be wrong here.

Spec/RealtimeClient.swift
```diff
@@ -352,6 +352,478 @@ class RealtimeClient: QuickSpec {
...
+
+                    waitUntil(timeout: testTimeout) { done in
+                        let partialDone = AblyTests.splitDone(2, done: done)
+
+                        client.connection.once(.Failed) { stateChange in
+                            guard let stateChange = stateChange else {
+                                fail("ConnectionStateChange is nil"); partialDone(); return
+                            }
+                            expect(stateChange.previous).to(equal(ARTRealtimeConnectionState.Connected))
+                            expect(stateChange.reason).toNot(beNil())
```
mattheworiordan: Can we check for a particular error code or type here please so that we know we're testing the correct thing?

Spec/RealtimeClient.swift
```diff
@@ -352,6 +352,478 @@ class RealtimeClient: QuickSpec {
...
+                                guard let tokenDetails = tokenDetails else {
+                                    fail("TokenDetails is nil"); done(); return
+                                }
+                                expect(tokenDetails.token).toNot(beNil())
+                                expect(client.connection.state).to(equal(ARTRealtimeConnectionState.Connected))
+
+                                guard let transport = client.transport as? TestProxyTransport else {
+                                    fail("TestProxyTransport is not set"); done(); return
+                                }
+                                expect(transport.protocolMessagesReceived.filter({ $0.action == .Connected })).to(haveCount(1))
```
mattheworiordan: This seems fine, but I'm not sure how you are confirming that the original connection attempt is cancelled.  I wouldn't worry too much about it, but I am not sure I see how this tests confirms that?

Spec/RealtimeClient.swift
```diff
@@ -352,6 +352,478 @@ class RealtimeClient: QuickSpec {
...
+
+                            let tokenParams = ARTTokenParams()
+                            tokenParams.clientId = "john"
+
+                            let authOptions = ARTAuthOptions()
+                            authOptions.authCallback = { tokenParams, completion in
+                                completion(getTestTokenDetails(clientId: "tester"), nil)
+                            }
+
+                            client.auth.authorize(tokenParams, options: authOptions) { tokenDetails, error in
```
mattheworiordan: I am not sure you have tested this correctly, the connection is becoming connected anyway given you have called connect.  Why not instead call `authorize` on the connection which is not configured to autoConnect, and simply check that the connection is `CONNECTED` once the callback is called?  Looking below, it seems you are checking that the state is `CONNECTING` which is wrong according to the spec "authorize call should complete with the new token once the connection has moved to the CONNECTED".

Also, you are not testing that it's called only once moved to the FAILED state, can you test that or do you feel it's covered sufficiently elsewhere?  If so, perhaps add a comment to that test.

Spec/RealtimeClient.swift
```diff
@@ -352,6 +352,478 @@ class RealtimeClient: QuickSpec {
...
+                            }
+                            expect(transport.protocolMessagesSent.filter({ $0.action == .Auth })).to(haveCount(0))
+                            expect(transport.protocolMessagesReceived.filter({ $0.action == .Connected })).to(haveCount(1)) //New transport
+                            partialDone()
+                        }
+
+                        guard let transport = client.transport as? TestProxyTransport else {
+                            fail("First transport is not TestProxyTransport"); done(); return
+                        }
+                        transport.simulateIncomingNormalClose()
```
mattheworiordan: Surely the `authorize` request should only be called once you have confirmed it is indeed in the `CLOSED` state as according to the spec "when the connection is in the DISCONNECTED, SUSPENDED, FAILED, or CLOSED state when auth#authorize...".

Spec/RealtimeClient.swift
```diff
@@ -352,6 +352,478 @@ class RealtimeClient: QuickSpec {
...
+                    }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        let partialDone = AblyTests.splitDone(3, done: done)
+
+                        client.connection.once(.Connecting) { stateChange in
+                            guard let stateChange = stateChange else {
+                                fail("ConnectionStateChange is nil"); partialDone(); return
+                            }
+                            expect(stateChange.previous).to(equal(ARTRealtimeConnectionState.Closed))
```
mattheworiordan: I appreciate you are checking that the connection is `CLOSED`, but you are not making it `CLOSED` before calling `authorize` so there is a race condition here.

Spec/RealtimeClient.swift
ricardopereira: Done 2ff1b41.

Spec/RealtimeClient.swift
ricardopereira: Done 4b35e6a.

Spec/RealtimeClient.swift
ricardopereira: Done 81e48b4.

Spec/RealtimeClient.swift
ricardopereira: Done 77d77fe.

Spec/RealtimeClient.swift
ricardopereira: Done 7639ab6.

Spec/RealtimeClient.swift
ricardopereira: Done ab5b545.

ricardopereira:
@mattheworiordan PTAL

Spec/RealtimeClientConnection.swift
```diff
@@ -2318,7 +2321,7 @@ class RealtimeClientConnection: QuickSpec {
                                     guard let errorInfo = errorInfo else {
                                         fail("ErrorInfo is nil"); done(); return
                                     }
-                                    expect(errorInfo.code).to(equal(40142)) //Token expired
+                                    expect(errorInfo.code).to(equal(-1003))
```
mattheworiordan: Sorry, why -1003?  Where did that error code come from?  Here's our list of error codes, https://github.com/ably/ably-common/blob/master/protocol/errors.json

mattheworiordan:
Looks good, 2 small comments

Spec/RealtimeClientConnection.swift
ricardopereira: @mattheworiordan If it's ok then please accept the PR changes.

mattheworiordan: Sorry, I don't understand, that link links back here?

ricardopereira: Ups, wrong link. @mattheworiordan https://github.com/ably/ably-ios/blob/ff2fd386cff34b807a97964281ef9a69c493cfe4/Spec/RealtimeClientConnection.swift#L2293

ricardopereira: @mattheworiordan Did you saw my last comment? ^
BTW, It's good to merge?

ricardopereira: @mattheworiordan Can I squash everything and merge?

PR #527, opened 2016-10-19
Remove `AuthOptions.force`

ricardopereira:


ricardopereira:
There are broken tests. I'll fix them now.

ricardopereira:
@tcard PTAL

Spec/Auth.swift
```diff
@@ -1679,6 +1675,7 @@ class Auth : QuickSpec {
             // RSA10a
             it("should create a token if needed and use it") {
```
mattheworiordan: This description is wrong, it will always create a token?

mattheworiordan: Also, https://github.com/ably/ably-ios/pull/527/files#diff-b5d84fe9bce26902a7481c2ffec84fffR1673 should be `authorize`

Spec/Auth.swift
```diff
@@ -1692,7 +1689,7 @@ class Auth : QuickSpec {
                             guard let tokenDetails = tokenDetails else {
                                 XCTFail("TokenDetails is nil"); done(); return
                             }
-                            expect(tokenDetails.token).to(equal(options.token))
+                            expect(tokenDetails.token).toNot(equal(options.token))
```
mattheworiordan: The comment above says `Reuse the valid token` yet that's not correct.  Also, I would like to see that implicit publish operations don't now create a new token every time.  Could we therefore change the logic for this test to:

* Publish a message
* Check that token exists
* Publish another message
* Check that token has not changed
* Call authorize
* Check that token has changed

ricardopereira: Done 6bef509 👍

PR #533, opened 2016-11-14
TE6, TD7: Token{Request, Details}::fromJson

### LINKED ISSUE #529: Token Request 'ttl' should be in milliseconds
Please represent 'ttl' property of ARTTokenRequest in milliseconds to maintain consistency with other client libraries.  For instance, our servers use the Ably-Java lib 0.8.5, to return a token request during the authCallback in the iOS client.  The java lib respects the 'ttl' as milliseconds, but the iOS client uses seconds.

tcard:
Implements ably/docs#193.

Fixes #529.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,18 +308,11 @@ let clientOptions = ARTClientOptions()
 clientOptions.authCallback = { params, callback in
     getTokenRequestJSONFromYourServer(params) { json, error in
         //handle error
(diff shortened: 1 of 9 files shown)


mattheworiordan:
@tcard unfortunately this needs to be `fromJson` instead of `fromJSON`. Can you amend please @tcard or @ricardopereira?

ricardopereira:
Ok, I'll do it.

Issue #531, opened 2016-11-11
CFRunLoopPerformBlock never executes


aquigriffin:
In ARTHttp.m, starting at line 175.  CFRunLoopRef fails to perform block if CurrentRunLoopRef ignores wakeup.  Reproduced with performing token request on application startup.

ricardopereira:
@aquigriffin Could you add a short runnable sample to reproduce this issue?

ricardopereira:
Even so, I prefer using GCD instead of run loops. It's much safer. No need to deal with run loop management and we can avoid issues like this.

aquigriffin:
@ricardopereira Here is an example of my ClientOptions authCallback:
 `clientOptions.authCallback = ^(ARTTokenParams *params, void(^callback)(id<ARTTokenDetailsCompatible>, NSError*)) {
//interaction is our network layer communicating with our backend for the token request
        [_interaction startTaskOnSuccess:^(OUPAblyTokenRequestResponseDTO *_Nullable responseDTO) {
            //NSLog(@"received token params from our BND: %@", responseDTO);
            [weakSelf updateChannelNamesWithResponseDTO:responseDTO];
            ARTTokenRequest *tokenRequest = [self tokenRequestFromTokenRequestResponseDTO:responseDTO];
            callback(tokenRequest,nil);
        } onFailure:^(NSError *_Nullable error){
            callback(nil,error);
        }];
    };`

The client is being created on AppDelegate applicationDidLaunchWithOptions: with the following code:

`- (void)configureClient {
    _client = [[ARTRealtime alloc] initWithOptions:[self clientOptions]];
    
    [_client.auth.logger debugMode];
    [_client.auth.logger errorMode];
    [_client.auth.logger verboseMode];

    __weak typeof(self) weakSelf = self;
    [_client.connection on:^(ARTConnectionStateChange *stateChange) {
        [weakSelf handleConnectionStateChange:stateChange];
    }];
}`

If I force the callback(ARTTokenRequest *, NSError *) in clientOptions.authCallback to be run on the main thread, the callback will perform the block at line 180 in ARTHttp.m, and connect the client.  If I am not running on main the block will not fire.  There may be an issue with the connection retry happening before the callback can fire if not on the main thread, but I have not been able to verify this.  I have verified I can repro this on device and simulator ios9.0+, iphone6/6s.

ricardopereira:
[...]

PR #536, opened 2016-11-17
Fix: check if authCallback is running on a background queue

ricardopereira:
#531
 - interacting with certain frameworks like UIKit, we need to ensure that all calls happen from the main thread.

### DIFF
--- a/Ably.xcodeproj/project.pbxproj
+++ b/Ably.xcodeproj/project.pbxproj
@@ -126,6 +126,8 @@
 		D7534C321D79E5C20054C182 /* Ably.h in Headers */ = {isa = PBXBuildFile; fileRef = D7534C311D79E5C20054C182 /* Ably.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		D7588AF31BFF91B800BB8279 /* ARTURLSessionServerTrust.h in Headers */ = {isa = PBXBuildFile; fileRef = D7588AF11BFF91B800BB8279 /* ARTURLSessionServerTrust.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		D7588AF41BFF91B800BB8279 /* ARTURLSessionServerTrust.m in Sources */ = {isa = PBXBuildFile; fileRef = D7588AF21BFF91B800BB8279 /* ARTURLSessionServerTrust.m */; };
+		D75A3F1B1DDE5B62002A4AAD /* ARTGCD.h in Headers */ = {isa = PBXBuildFile; fileRef = D75A3F191DDE5B62002A4AAD /* ARTGCD.h */; };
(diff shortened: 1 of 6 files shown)


mattheworiordan:
@ricardopereira does this PR fix https://github.com/ably/ably-ios/issues/531?

ricardopereira:
@mattheworiordan Yes. I reproduced the issue and now it's working.

ricardopereira:
BTW, for the record:

```
@implementation AppDelegate {
    ARTClientOptions *_clientOptions;
    ARTRealtime *_client;
}

- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
    _clientOptions = [[ARTClientOptions alloc] init];
    _clientOptions.authCallback = ^(ARTTokenParams *params, void(^callback)(id<ARTTokenDetailsCompatible>, NSError*)) {
        callback(nil, [NSError errorWithDomain:@"localhost" code:0 userInfo:nil]);
    };
    _client = [[ARTRealtime alloc] initWithOptions:_clientOptions];
    [_client.connection on:^(ARTConnectionStateChange *stateChange) {
        NSLog(@"%@", stateChange);
    }];
    return YES;
}

@end
```

The `callback(nil, [NSError errorWithDomain:@"localhost" code:0 userInfo:nil]);` was not called.

tcard:
I'm not sure this solves the problem as generally as it could. What's so special about `authCallback`? Shouldn't this be done deeper, like at the ARTHttp level? Or all the places where the CFRunLoopPerformBlock thing is used?

ricardopereira:
>I'm not sure this solves the problem as generally as it could.

You're right.

> What's so special about `authCallback`?

I don't have an answer. What I know is that the `application:didFinishLaunchingWithOptions` is running in the main queue and you can test it quite easily (just create a UIView and present it... it works). I debug it line by line and the only case where the execution transitions to a background queue is when the HTTP request is made. The `completion` is called in the background. Maybe because the `authCallback` is created in the main queue.

@tcard After I opened the PR I started thinking on generalising the solution. I though checking every callback that is accessible by the user (like, `authCallback`, state changes, publish message callback) and add the code to execute the callback on the main queue if needed. What do you think?

tcard:
> I though checking every callback that is accessible by the user (like, authCallback, state changes, publish message callback) and add the code to execute the callback on the main queue if needed. What do you think?

Sure, sounds better to me. Although honestly I'm flying a bit blind here, I don't really know how GCD works, etc. so you'll probably know better.

Just please let's avoid duplicating code as we do here, we can put this if-else logic in a generic function that takes a block.

tcard:
@ricardopereira Is there anything missing here?

ricardopereira:
@tcard No, it's done.

tcard:
@ricardopereira Cool, thanks. I still don't get what's special about authCallback, but well...

ricardopereira:
@tcard I explained [here](https://github.com/ably/ably-ios/issues/531#issuecomment-261541444).

PR #537, opened 2016-11-22
RTN22

ricardopereira:


Spec/RealtimeClientConnection.swift
```diff
@@ -3320,6 +3320,52 @@ class RealtimeClientConnection: QuickSpec {
...
+
+                    var authorizeMethodCallCount = 0
+                    let hook = client.auth.testSuite_injectIntoMethodAfter(#selector(client.auth.authorize)) {
+                        authorizeMethodCallCount += 1
+                    }
+                    defer { hook.remove() }
+
+                    let authMessage = ARTProtocolMessage()
+                    authMessage.action = .Auth
+                    transport.receive(authMessage)
```
mattheworiordan: We may in fact not need to stub which makes the tests more resilient.  Why not issue a token with a `ttl` of 35 seconds, and within 5 seconds you will receive the `AUTH` protocol message naturally.  That way all of this stubbing becomes unnecessary

Spec/RealtimeClientConnection.swift
```diff
@@ -3366,6 +3366,65 @@ class RealtimeClientConnection: QuickSpec {
...
+                    var authorizeMethodCallCount = 0
+                    let hook = client.auth.testSuite_injectIntoMethodAfter(#selector(client.auth.authorize)) {
+                        authorizeMethodCallCount += 1
+                    }
+                    defer { hook.remove() }
+
+                    let disconnectedMessage = ARTProtocolMessage()
+                    disconnectedMessage.action = .Disconnected
+                    disconnectedMessage.error = ARTErrorInfo.createWithCode(40140, message: "token error (unspecified)")
+                    transport.receive(disconnectedMessage)
```
mattheworiordan: Again, why stub, why not simply issue  a token with a short TTL of say 10 seconds, wait for the disconnect, and then follow the rest of the steps?  I really think we should be using stubbing more sparingly generally, unless it's unavoidable.  In most cases I do not believe it is.

Spec/RealtimeClientConnection.swift
```diff
@@ -3320,6 +3320,118 @@ class RealtimeClientConnection: QuickSpec {
...
+                    guard let initialToken = client.auth.tokenDetails?.token else {
+                        fail("Initial token is nil"); return
+                    }
+
+                    guard let transport = client.transport as? TestProxyTransport else {
+                        fail("TestProxyTransport is not set"); return
+                    }
+
+                    var authorizeMethodCallCount = 0
+                    let hook = client.auth.testSuite_injectIntoMethodAfter(#selector(client.auth.authorize)) {
```
mattheworiordan: I would prefer to not see this injection as it's not necessary

ricardopereira: Done dbcaec8.

mattheworiordan:
Are you awaiting final review?  Seems good, but would be good to make sure connections are actually resumed correctly in both cases.

Spec/RealtimeClientConnection.swift
```diff
@@ -3361,6 +3361,117 @@ class RealtimeClientConnection: QuickSpec {
...
+                    }
+
+                    guard let transport = client.transport as? TestProxyTransport else {
+                        fail("TestProxyTransport is not set"); return
+                    }
+
+                    expect(transport.protocolMessagesReceived.filter{ $0.action == .Auth }).toEventually(haveCount(1), timeout: testTimeout)
+
+                    waitUntil(timeout: testTimeout) { done in
+                        client.connection.once(.Connected) { stateChange in
```
tcard: This is racy. The expectation above can pass _after_ this event is fired. The event listener should be set up before. In fact, you can just remove the expectation above and check that the AUTH protocol message was received inside the CONNECTED event listener.

Spec/RealtimeClientConnection.swift
```diff
@@ -3361,6 +3361,117 @@ class RealtimeClientConnection: QuickSpec {
...
+                        }
+                    }
+                }
+
+                // RTN22a
+                it("re-authenticate and resume the connection when the client is forcibly disconnected following a DISCONNECTED message containing an error code in the range 40140 <= code < 40150") {
+                    let options = AblyTests.commonAppSetup()
+                    options.autoConnect = false
+                    options.disconnectedRetryTimeout = 1.0
+                    options.token = getTestToken(key: options.key!, ttl: 5.0)
```
tcard: Five seconds isn't a bit too much?

Spec/RealtimeClientConnection.swift
```diff
@@ -3361,6 +3361,117 @@ class RealtimeClientConnection: QuickSpec {
                     }
                 }
 
+                // RTN22
+                it("Ably can request that a connected client re-authenticates by sending the client an AUTH ProtocolMessage") {
+                    let options = AblyTests.commonAppSetup()
+                    options.autoConnect = false
+                    options.token = getTestToken(ttl: 5.0)
```
tcard: Five seconds isn't a bit too much?

Spec/RealtimeClientConnection.swift
```diff
@@ -3361,6 +3361,117 @@ class RealtimeClientConnection: QuickSpec {
...
+                            expect(stateChange?.reason).to(beNil())
+                            expect(initialToken).toNot(equal(client.auth.tokenDetails?.token))
+                            done()
+                        }
+                    }
+
+                    expect(client.connection.id).to(equal(initialConnectionId))
+                    expect(authorizeMethodCallCount) == 1
+
+                    waitUntil(timeout: testTimeout) { done in
```
tcard: What is this whole block checking?

ricardopereira: This block is checking if the channel was resumed successfully (recommended by @mattheworiordan).

> test that uses a low value TTL for the token, attaches to a channel, then waits for the auth to complete renewing the token, and then publish over REST to that channel to make sure it's received?

Spec/RealtimeClientConnection.swift
ricardopereira: Recommended by @mattheworiordan 

> This looks a bit brittle, 2s to connect & attach in CI may fail, can we not just give it a bit of buffer, say 6 seconds to make sure we don't get test failures because of delays in running code?

Spec/RealtimeClientConnection.swift
```diff
@@ -3366,6 +3366,65 @@ class RealtimeClientConnection: QuickSpec {
...
+                    disconnectedMessage.action = .Disconnected
+                    disconnectedMessage.error = ARTErrorInfo.createWithCode(40140, message: "token error (unspecified)")
+                    transport.receive(disconnectedMessage)
+
+                    expect(client.connection.state).toEventually(equal(ARTRealtimeConnectionState.Connecting), timeout: testTimeout)
+
+                    waitUntil(timeout: testTimeout) { done in
+                        client.connection.once(.Connected) { stateChange in
+                            expect(stateChange?.reason).to(beNil())
+                            expect(initialToken).toNot(equal(client.auth.tokenDetails?.token))
```
ricardopereira: Done.

Spec/RealtimeClientConnection.swift
ricardopereira: As above.

Spec/RealtimeClientConnection.swift
ricardopereira: Yes, you're right. Will fix.

Spec/RealtimeClientConnection.swift
tcard: Well, OK then. I'd say you can just mock the AUTH message from the server, but well, I guess we want the real thing.

ricardopereira:
@mattheworiordan passing. PTAL

Spec/RealtimeClientConnection.swift
mattheworiordan: It's just a fail safe to make sure that the reauth is in fact working and there is continuity over the channels

Issue #538, opened 2016-11-22
Connection resume does not fire channel detached


mattheworiordan:
@ricardopereira please can we write a test to confirm what happens to a channel when a connection resume fails.  What should happen is that all channels that were attached become detached, the state change is emitted along with an error. Please see http://docs.ably.io/client-lib-development-guide/versions/v0.8/features/#RTN15c3 for 0.8 spec.

A customer has reported the detached event not firing.

mattheworiordan:
@ricardopereira whilst reviewing the verbose logs, I can unfortunately see that the client library does not output the URL it is connecting to for new WS connections.  Can you please ensure all HTTP and WebSocket connections / requests output the complete URL with headers if appropriate when verbose logging is on?

mattheworiordan:
@ricardopereira unfortunately I also now suspect (but do not have proof as the logs are concatenated meaning I cannot prove this) that the resume following an auth token change is not working.  I suspect the resume arguments for the WS connection are not included.

Can you additionally write a test that uses a low value TTL for the token, attaches to a channel, then waits for the auth to complete renewing the token, and then publish over REST to that channel to make sure it's received?

mattheworiordan:
FYI, in the connected message following auth, I see:

```
{
    action = 4;
    connectionDetails =     {
        clientId = "REDACTED";
        connectionKey = "REDACTED";
        connectionStateTtl = 120000;
        maxFrameSize = 262144;
        maxIdleInterval = 15000;
        maxInboundRate = 50;
        maxMessageSize = 65536;
        serverId = "frontend.d8d9.1.eu-west-1-A.i-5a33dacc";
    };
    connectionId = REDACTED;
    connectionKey = "REDACTED";
    connectionSerial = "-1";
}
```

If the resume failed, we should have received an error with the connected message with an `80008` error code.  As you can see above, that did not happen.

So we need to check that:

* Even with the absence of an error, if the connection ID changes, then all channels should become detached (the error is only informational and not the reason channels are detached)
* Auth renewal triggers a reconnect with resume and resumes connections
* Auth renewal triggers a reconnect with (fabricated) faulty resume, resumes connection, emits an error, and detaches all channels.

ricardopereira:
[...]

PR #539, opened 2016-11-23
Fix: connection resume does not fire channel detached

ricardopereira:
#538

### DIFF
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -2093,6 +2093,72 @@ class RealtimeClientConnection: QuickSpec {
                         expect(channel.errorReason).to(beIdenticalTo(protocolError.error))
                     }
 
+                    it("should detach all channels if the connectionId has changed") {
+                        let options = AblyTests.commonAppSetup()
+                        options.disconnectedRetryTimeout = 1.0
+                        options.tokenDetails = getTestTokenDetails(ttl: 2.0)
+                        let client = AblyTests.newRealtime(options)
+                        defer { client.dispose(); client.close() }
+
+                        let channel = client.channels.get("test")
+                        waitUntil(timeout: testTimeout) { done in
+                            channel.attach() { error in
+                                expect(error).to(beNil())
+                                done()
+                            }
+                        }
+
+                        let initialConnectionId = client.connection.id
+
+                        guard let firstTransport = client.transport as? TestProxyTransport else {
+                            fail("TestProxyTransport is not set"); return
+                        }
+
+                        // Wait for token to expire
+                        waitUntil(timeout: testTimeout) { done in
+                            let partialDone = AblyTests.splitDone(2, done: done)
+                            channel.once(.Detached) { error in
+                                expect(error).to(beNil())
+                                partialDone()
+                            }
+                            // Wait for connection resume
+                            client.connection.once(.Connected) { stateChange in
+                                expect(stateChange?.reason).to(beNil())
+                                partialDone()
+                            }
+                        }
+
+                        expect(firstTransport.protocolMessagesReceived.filter{ $0.action == .Disconnected }).to(haveCount(1))
+
+                        guard let secondsTransport = client.transport as? TestProxyTransport else {
+                            fail("TestProxyTransport is not set"); return
+                        }
+
+                        let connectedMessages = secondsTransport.protocolMessagesReceived.filter{ $0.action == .Connected }
+                        expect(connectedMessages).to(haveCount(1)) //New transport connected
+                        guard let receivedConnectionId = connectedMessages.first?.connectionId else {
+                            fail("ConnectionID is nil"); return
+                        }
+                        expect(client.connection.id).to(equal(receivedConnectionId))
+                        expect(client.connection.id).toNot(equal(initialConnectionId))
+
+                        waitUntil(timeout: testTimeout) { done in
+                            let partialDone = AblyTests.splitDone(2, done: done)
+                            let expectedMessage = ARTMessage(name: "ios", data: "message1")
+
+                            channel.subscribe() { message in
+                                expect(message.name).to(equal(expectedMessage.name))
+                                expect(message.data as? String).to(equal(expectedMessage.data as? String))
+                                partialDone()
+                            }
+
+                            let rest = ARTRest(options: AblyTests.clientOptions(key: options.key!))
+                            rest.channels.get("test").publish([expectedMessage]) { error in
+                                expect(error).to(beNil())
+                                partialDone()
+                            }
+                        }
+                    }
                 }
 
                 // RTN15d
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -1949,7 +1949,7 @@ class RealtimeClientConnection: QuickSpec {
                 context("System's response to a resume request") {
 
                     // RTN15c1
-                    pending("CONNECTED ProtocolMessage with the same connectionId as the current client, and no error") {
+                    it("CONNECTED ProtocolMessage with the same connectionId as the current client, and no error") {
                         let options = AblyTests.commonAppSetup()
                         options.disconnectedRetryTimeout = 1.0
                         let client = AblyTests.newRealtime(options)
@@ -1972,12 +1972,12 @@ class RealtimeClientConnection: QuickSpec {
                                 done()
                             }
                         }
-                        expect(channel.state).to(equal(ARTRealtimeChannelState.Attached))
+                        expect(channel.state).toEventually(equal(ARTRealtimeChannelState.Attached), timeout: testTimeout)
                         expect(client.queuedMessages).toEventually(haveCount(0), timeout: testTimeout)
                     }
 
                     // RTN15c2
-                    pending("CONNECTED ProtocolMessage with the same connectionId as the current client and an non-fatal error") {
+                    it("CONNECTED ProtocolMessage with the same connectionId as the current client and an non-fatal error") {
                         let options = AblyTests.commonAppSetup()
                         options.disconnectedRetryTimeout = 1.0
                         let client = AblyTests.newRealtime(options)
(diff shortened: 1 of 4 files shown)


ricardopereira:
@mattheworiordan PTAL and if it's good then I fix the client.

Spec/RealtimeClientConnection.swift
```diff
@@ -2093,6 +2093,39 @@ class RealtimeClientConnection: QuickSpec {
...
+                        disconnectedMessage.connectionId = "xxxxxxxxx"
+                        transport.receive(disconnectedMessage)
+
+                        waitUntil(timeout: testTimeout) { done in
+                            // Resumed
+                            client.connection.once(.Connected) { stateChange in
+                                guard let stateChange = stateChange else {
+                                    fail("ConnectionStateChange is nil"); done(); return
+                                }
+                                expect(client.connection.id).to(equal(disconnectedMessage.connectionId))
```
mattheworiordan: Why would this be equal to the fabricated message that was received?  Realtime will assign a new `connectionId` as that ID is now invalid

Spec/RealtimeClientConnection.swift
```diff
@@ -2126,6 +2126,33 @@ class RealtimeClientConnection: QuickSpec {
...
+
+                    it("should resume the connection after an auth renewal") {
+                        let options = AblyTests.commonAppSetup()
+                        options.disconnectedRetryTimeout = 1.0
+                        options.useTokenAuth = true
+                        let client = AblyTests.newRealtime(options)
+                        defer { client.dispose(); client.close() }
+                        let channel = client.channels.get("test")
+
+                        expect(client.connection.state).toEventually(equal(ARTRealtimeConnectionState.Connected), timeout: testTimeout)
```
mattheworiordan: This is not correct, this is not the behaviour we'd expect as a channel ATTACH that does not become ATTACHED would be resent or timeout.

The process I said we needed to test is:

* Connect
* Attach a channel (i.e wait for it to be attached)
* Token expires and connection is disconnected by Ably
* Resume happens automatically, wait until connected
* Publish a message over REST on the channel
* Ensure the message is received on the channel and the channel is still attached

Spec/RealtimeClientConnection.swift
```diff
@@ -2103,36 +2103,59 @@ class RealtimeClientConnection: QuickSpec {
...
-                                }
-                                expect(client.connection.id).to(equal(disconnectedMessage.connectionId))
-                                done()
+                                expect(stateChange?.reason).to(beNil())
+                                partialDone()
                             }
                         }
 
-                        expect(channel.state).to(equal(ARTRealtimeChannelState.Detached))
+                        expect(client.connection.id).toNot(equal(initialConnectionId))
```
ricardopereira: From the logs, when the token expires the `connectionId` always changes, so I decided to add that and additionally it checks if the channel gets detached.

ricardopereira:
@mattheworiordan PTAL at the last commit.

ricardopereira:
The problems I noticed while debugging:
 - `ARTRealtimeConnectionState.Connected` event is not emmited when the connection renews the token
 - The connection resume only detaches when the `connectionId` is different from the last one and an error exists

mattheworiordan:
@ricardopereira can you address `whilst reviewing the verbose logs, I can unfortunately see that the client library does not output the URL it is connecting to for new WS connections. Can you please ensure all HTTP and WebSocket connections / requests output the complete URL with headers if appropriate when verbose logging is on?` as well please?  Or is that covered?

Spec/RealtimeClientConnection.swift
```diff
@@ -2093,6 +2101,73 @@ class RealtimeClientConnection: QuickSpec {
                         expect(channel.errorReason).to(beIdenticalTo(protocolError.error))
                     }
 
+                    it("should resume the connection after an auth renewal and detach all channels if the connectionId changes") {
+                        let options = AblyTests.commonAppSetup()
+                        options.disconnectedRetryTimeout = 1.0
+                        options.tokenDetails = getTestTokenDetails(ttl: 2.0)
```
mattheworiordan: This looks a bit brittle, 2s to connect & attach in CI may fail, can we not just give it a bit of buffer, say 6 seconds to make sure we don't get test failures because of delays in running code?

Spec/RealtimeClientConnection.swift
```diff
@@ -2093,6 +2101,73 @@ class RealtimeClientConnection: QuickSpec {
...
+                        let initialConnectionId = client.connection.id
+
+                        guard let firstTransport = client.transport as? TestProxyTransport else {
+                            fail("TestProxyTransport is not set"); return
+                        }
+
+                        // Wait for token to expire
+                        waitUntil(timeout: testTimeout) { done in
+                            let partialDone = AblyTests.splitDone(2, done: done)
+                            channel.once(.Detached) { error in
```
mattheworiordan: Why is the channel becoming detached?  If a connection is closed because of an auth failure, the channel does not become detached, it remains attached until the connection is closed or moves to the suspended state?

Spec/RealtimeClientConnection.swift
```diff
@@ -2093,6 +2101,73 @@ class RealtimeClientConnection: QuickSpec {
...
+                            fail("TestProxyTransport is not set"); return
+                        }
+
+                        let connectedMessages = secondsTransport.protocolMessagesReceived.filter{ $0.action == .Connected }
+                        expect(connectedMessages).to(haveCount(1)) //New transport connected
+                        guard let receivedConnectionId = connectedMessages.first?.connectionId else {
+                            fail("ConnectionID is nil"); return
+                        }
+                        expect(client.connection.id).to(equal(receivedConnectionId))
+                        expect(client.connection.id).toNot(equal(initialConnectionId))
```
mattheworiordan: I don't follow why the connectionId has changed?  Why?  In a normal operation it won't, and as far as I can tell, you are not doing anything that would force it to change?

mattheworiordan:
This is not correct, please review.

Also, can you confirm that we are testing:

* [ ] Even with the absence of an error, if the connection ID changes, then all channels should become detached (the error is only informational and not the reason channels are detached)
* [ ] Auth renewal triggers a reconnect with resume and resumes connections
* [ ] Auth renewal triggers a reconnect with (fabricated) faulty resume, resumes connection, emits an error, and detaches all channels.

Spec/RealtimeClientConnection.swift
ricardopereira: @mattheworiordan the `connectionId` changes and that's why the channel gets detached.

First CONNECTED:
```
2016-11-23 13:16:04.508 xctest[32837:4092858] VERBOSE: RS:0x7f955c623000 ARTJsonLikeEncoder<msgpack>: protocolMessageFromDictionary {
    action = 4;
    connectionDetails =     {
        connectionKey = "e91RTOWlwA0vIq!tCwrawkZGCTS118T-ca6e91RTOWlwA0vIq";
        connectionStateTtl = 120000;
        maxFrameSize = 262144;
        maxIdleInterval = 15000;
        maxInboundRate = 250;
        maxMessageSize = 16384;
        serverId = "frontend.64d2.1.eu-central-1-A.i-d35f816e";
    };
    connectionId = tCwrawkZGC;
    connectionKey = "e91RTOWlwA0vIq!tCwrawkZGCTS118T-ca6e91RTOWlwA0vIq";
    connectionSerial = "-1";
}
```

It gets the DISCONNECTED from Ably:
```
2016-11-23 13:16:05.998 xctest[32837:4092858] VERBOSE: RS:0x7f955c623000 ARTJsonLikeEncoder<msgpack>: protocolMessageFromDictionary {
    action = 6;
    error =     {
        code = 40142;
        message = "Key/token status changed (expire)";
        statusCode = 401;
    };
}
```

The second CONNECTED:
```
2016-11-23 13:16:06.614 xctest[32837:4092858] VERBOSE: RS:0x7f955c623000 ARTJsonLikeEncoder<msgpack>: protocolMessageFromDictionary {
    action = 4;
    connectionDetails =     {
        connectionKey = "e91NTgaIgA0vIq!guMApg59VswaKAc4-d86e91NTgaIgA0vIq";
        connectionStateTtl = 120000;
        maxFrameSize = 262144;
        maxIdleInterval = 15000;
        maxInboundRate = 250;
        maxMessageSize = 16384;
        serverId = "frontend.25af.1.eu-central-1-A.i-305dca8c";
    };
    connectionId = guMApg59Vs;
    connectionKey = "e91NTgaIgA0vIq!guMApg59VswaKAc4-d86e91NTgaIgA0vIq";
    connectionSerial = "-1";
}
```

Full log: https://gist.github.com/ricardopereira/15a80610b22c9216d5587f2375359921

ricardopereira:
>@ricardopereira can you address `whilst reviewing the verbose logs, I can unfortunately see that the client library does not output the URL it is connecting to for new WS connections. Can you please ensure all HTTP and WebSocket connections / requests output the complete URL with headers if appropriate when verbose logging is on?` as well please?  Or is that covered?

@mattheworiordan Sorry, can you point me to a log entry where it is missing the URL ouput?
On each WebSocket `connect` it prints the current URL.

Spec/RealtimeClientConnection.swift
ricardopereira: Sorry, I answered in the other comment https://github.com/ably/ably-ios/pull/539#discussion_r89314960

mattheworiordan:
> @mattheworiordan Sorry, can you point me to a log entry where it is missing the URL ouput?
On each WebSocket connect it prints the current URL.

The log I sent you had no WebSocket URLs?

ricardopereira:
>The log I sent you had no WebSocket URLs?

@mattheworiordan i.e. line **467** `R:0x600000132b60 WS:0x6080000b2300 url /?accessToken=FretAw.EuwDunZs3QlP772pw4M6sWJs_kJ8lsmgyTVuReyckGYTv3rT19ecO810t8WhFtIQGlDfPXGseF_eSB0rO-zHXEMP_ZIPppXeFH-oNWRoQ1qV5cruvBwOXN7giWgLPbSEzgiZwYLbNK7lPWxO466ls1A&clientId=development.client/167420&echo=true&format=msgpack&v=0.8&lib=ios-(null) -- wss://realtime.ably.io:443`

mattheworiordan:
Ah, my bad, mind including the host in there too though?  I was scanning the logs for the host, my mistake, but I am sure it will happen to others.

Spec/RealtimeClientConnection.swift
```diff
@@ -2057,13 +2060,16 @@ class RealtimeClientConnection: QuickSpec {
                             }
                         }
                         expect(channel.state).to(equal(ARTRealtimeChannelState.Detached))
-                        expect(channel.errorReason!.code).to(equal(80008))
-                        expect(channel.errorReason!.message).to(contain("Unable to recover connection"))
+                        guard let channelError = channel.errorReason else {
+                            fail("Channel error is nil"); return
```
mattheworiordan: Btw. Don't change this, but in future we should have an error message that explains the problem rather than the boolean logic i.e. anyone can see from this code that channel error is nil, but what the reader cares about is why `nil` is a problem i.e. this should instead be `fail("Connection resume failed and error should be propagated to the cannel")`.   This helps the reader of the tests in future understand why this is a problem as opposed to just seeing that it is expected to be `nil`.

ricardopereira:
@mattheworiordan Yes, sure. If it passes then it's good to go. If not then I'll see what's wrong.

PR #540, opened 2016-11-23
RTN4g

ricardopereira:


mattheworiordan:
I assume this will be closed as we no longer expect ERROR event?

ricardopereira:
Yes, you right. I will implement the `RTL4h`.

Issue #541, opened 2016-11-23
Pending tests


ricardopereira:
```
Pending: when publishing a Message with an explicit clientId that is incompatible with the identified client’s clientId
Pending: should be used a PresenceMap to maintain a list of members
Pending: should result in an error if the channel is in the FAILED state
Pending: should result in an error if the channel moves to the FAILED state
```

mattheworiordan:
Are you looking at these?

ricardopereira:
@mattheworiordan Yes, I am.

PR #542, opened 2016-11-27
Remove pending tests

ricardopereira:
#541

### DIFF
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -2620,104 +2620,98 @@ class Auth : QuickSpec {
                 }
             }
         }
-        
+
         describe("Reauth") {
+
             // RTC8
-            pending("should use authorise({force: true}) to reauth with a token with a different set of capabilities") {
-                // init ARTRest
-                let restOptions = AblyTests.setupOptions(AblyTests.jsonRestOptions)
-                let rest = ARTRest(options: restOptions)
-
-                // get first token
-                let tokenParams = ARTTokenParams()
-                tokenParams.capability = "{\"wrongchannel\": [\"*\"]}"
-                tokenParams.clientId = "testClientId"
-
-                var firstToken = ""
-
-                waitUntil(timeout: testTimeout) { done in
-                    rest.auth.requestToken(tokenParams, withOptions: nil) { tokenDetails, error in
-                        expect(error).to(beNil())
-                        expect(tokenDetails).toNot(beNil())
-                        firstToken = tokenDetails!.token
-                        done()
-                    }
-                }
-                expect(firstToken).toNot(beNil())
-                expect(firstToken.characters.count > 0).to(beTrue())
-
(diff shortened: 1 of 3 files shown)


Spec/Auth.swift
```diff
@@ -2620,104 +2620,84 @@ class Auth : QuickSpec {
...
 
-                // get first token
-                let tokenParams = ARTTokenParams()
-                tokenParams.capability = "{\"wrongchannel\": [\"*\"]}"
-                tokenParams.clientId = "testClientId"
+        describe("Reauth") {
 
-                var firstToken = ""
+            // RTC8
+            it("should use authorise({force: true}) to reauth with a token with a different set of capabilities") {
```
mattheworiordan: force: true is no longer supported

Spec/Auth.swift
```diff
@@ -2620,104 +2620,84 @@ class Auth : QuickSpec {
...
                 }
-                expect(firstToken).toNot(beNil())
-                expect(firstToken.characters.count > 0).to(beTrue())
 
-                // init ARTRealtime
-                let realtimeOptions = AblyTests.commonAppSetup()
-                realtimeOptions.token = firstToken
-                realtimeOptions.clientId = "testClientId"
+                let authOptions = ARTAuthOptions()
+                authOptions.force = true
```
mattheworiordan: force is not supported?

Spec/Auth.swift
```diff
@@ -2620,104 +2620,84 @@ class Auth : QuickSpec {
...
+                    realtime.auth.authorise(tokenParams, options: authOptions) { tokenDetails, error in
+                        expect(error).to(beNil())
+                        expect(tokenDetails).toNot(beNil())
                         done()
                     }
-                    realtime.connect()
                 }
 
-                // create a `rightchannel` channel and check can't attach to it
-                let channel = realtime.channels.get("rightchannel")
```
mattheworiordan: Why did you remove this?  This is the point of the test no, i.e. attach with one set of permissions, change them, check that permissions have actually changed by checking the channel attach success?

Spec/Auth.swift
mattheworiordan: Ok, but I'm not sure why we are doing any work on `0.8` now other than critical fixes?

mattheworiordan:
👍 for merging.

ricardopereira:
I still need to see why the tests are no passing 🙁

mattheworiordan:
> I still need to see why the tests are not passing 🙁

Any update?

Spec/Auth.swift
ricardopereira: Ok, I agree.

Spec/Auth.swift
ricardopereira: I will rollback 👍

ricardopereira:
I'm finishing this now.

PR #545, opened 2016-11-28
Update RTL4 for 0.9

ricardopereira:


Spec/RealtimeClientChannel.swift
```diff
@@ -656,6 +658,48 @@ class RealtimeClientChannel: QuickSpec {
                         }
                     }
                 }
+
+                // RTL4i
```
tcard: This will hurt: RTL4i is exactly what RTL4h was before and it's already tested a few  lines above. I don't know why the letter was changed.

So these tests can be removed, previous RTL4h tests above should just be renamed to RTL4i, and the new RTLh ("If the channel is in a pending state DETACHING or ATTACHING, do the attach operation after the completion of the pending request") should be tested instead.

mattheworiordan: Agreed

ricardopereira:
Missing #549 to pass.

ricardopereira:
@mattheworiordan @tcard PTAL

tcard:
LGTM, pending #549.

ricardopereira:
Green.

PR #543, opened 2016-11-28
Update RTL2 for 0.9 

ricardopereira:

PR #548, opened 2016-11-28
Update RTL12 for 0.9

ricardopereira:


ricardopereira:
"Note that UPDATE should be the only event emitted: in particular, the library must not emit an ATTACHED event if the channel was already attached" is done: https://github.com/ably/ably-ios/pull/548/commits/dfe518429f5fd464169eb37321da85b8683cce27#diff-c3246c6178b672c3312c1467319b835eR2289

ricardopereira:
I updated it with `0-9-master` but there are no changes because #543 already updated this spec.

PR #550, opened 2016-11-28
Update RTL14 for 0.9

ricardopereira:


Spec/RealtimeClientChannel.swift
```diff
@@ -2325,6 +2325,48 @@ class RealtimeClientChannel: QuickSpec {
...
+                    }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        let errorProtocolMessage = AblyTests.newErrorProtocolMessage()
+                        errorProtocolMessage.action = .Error
+                        errorProtocolMessage.channel = "foo"
+
+                        let partialDone = AblyTests.splitDone(2, done: done)
+
+                        channel.once(.Error) { stateChange in
```
tcard: `.Error` doesn't exist anymore. Only FAILED should be emitted as per RTL14.

tcard:
Why was this merged?

tcard:
Ah OK, because of https://github.com/ably/ably-ios/pull/544#issuecomment-267021928.

ricardopereira:
@tcard Because I removed the `.Error` event and I thought it would be the only thing to change.

tcard:
Ah right, sorry, I was misreading my own change request...

PR #558, opened 2016-12-13
Xcode 8.2 (Swift 2.3)

ricardopereira:
Far more stable than Xcode 7.3.

Fixes as well a wierd problem while debugging Swift specs:
 - `Couldn't Generate Swift Representation. Error (from SourceKit): "Could not load module: Ably"`
 - `Umbrella header for module 'Ably' does not include header ARTGCD.h'`

**`CLANG_WARN_INFINITE_RECURSION`** and **`CLANG_WARN_SUSPICIOUS_MOVE`** turned on.


mattheworiordan:
LGTM

Issue #557, opened 2016-12-13
ART* should match IDL definition


ricardopereira:
As agreed on the call, we will update the ART* classes to match the IDL. The ones we know need changing are below, but @ricardopereira please review all classes before submitting a PR for this issue:

* [ ] ARTRealtimeConnectionState -> ARTConnectionState
* [ ] ARTRealtimeConnectionEvent -> ARTConnectionEvent
* [ ] ARTRealtimeChannelState -> ARTChannelState
* [ ] ARTRealtimeChannelEvent -> ARTChannelEvent

The IDL is at http://docs.ably.io/client-lib-development-guide/features/#idl

mattheworiordan:
@ricardopereira @tcard further to the conversation at https://github.com/ably/ably-ios/pull/543#pullrequestreview-11443102, I have given this some thought, and I am in fact now against the change.  My thinking is:

* It's an unnecessary breaking change 
* It means we have to update all tutorials and documentation, this is a complex task for tutorials as each tutorial's commits need to be updated and then each tutorial needs updating, see https://github.com/ably/tutorials/pulls and https://github.com/ably/docs/pull/213#pullrequestreview-12620259 for reasons why this is hard.
* We have a ARTChannel and ARTRealtimeChannel which reflects the IDL which has `Channel` and `RealtimeChannel`.  We may later have `ARTConnection` for REST (although unlikely), so if we did having `ARTRealtimeConnection` is not such a bad thing.
* If we have `ARTChannel` and `ARTRealtimeChannel`, then when considering https://github.com/ably/ably-ios/issues/556, `ARTChannelState` should indeed stay as ARTRealtimeChannelState`.

In short whilst I think the naming is not perfect, changing it does not really achieve much, but will cause pain for customers and us in regards to documentation, examples we have, tutorials etc.

Is there really any compelling reason to change this now?

FYI @paddybyers @simonwoolf

ricardopereira:
@mattheworiordan Ok, I agree with most of it. We should at least rename each cases from `ARTRealtimeConnectionState`:

```
typedef NS_ENUM(NSUInteger, ARTRealtimeConnectionState) {
    ARTRealtimeInitialized, //<---- ARTRealtimeInitialized to ARTRealtimeConnectionInitialized
    ARTRealtimeConnecting,
    ARTRealtimeConnected,
    ARTRealtimeDisconnected,
    ARTRealtimeSuspended,
    ARTRealtimeClosing,
[...]

PR #559, opened 2016-12-14
UPDATE event

ricardopereira:
#551

### DIFF
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -304,6 +304,41 @@ class RealtimeClientConnection: QuickSpec {
                     expect(events[7].rawValue).to(equal(ARTRealtimeConnectionState.Failed.rawValue), description: "Should be FAILED state")
                 }
 
+                // RTN4h
+                it("should never emit a ConnectionState event for a state equal to the previous state") {
+                    let options = AblyTests.commonAppSetup()
+                    let client = ARTRealtime(options: options)
+                    defer { client.dispose(); client.close() }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        client.connection.once(.Connected) { stateChange in
+                            expect(stateChange?.reason).to(beNil())
+                            done()
+                        }
+                    }
+
+                    client.connection.once(.Connected) { stateChange in
+                        fail("Should not emit a Connected state")
+                    }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        client.connection.once(.Update) { stateChange in
+                            guard let stateChange = stateChange else {
+                                fail("ConnectionStateChange is nil"); done(); return
+                            }
+                            expect(stateChange.reason).to(beNil())
+                            expect(client.connection.state).to(equal(ARTRealtimeConnectionState.Connected))
+                            expect(stateChange.current).to(equal(ARTRealtimeConnectionState.Connected))
+                            expect(stateChange.current).to(equal(stateChange.previous))
+                            done()
+                        }
+
+                        let authMessage = ARTProtocolMessage()
+                        authMessage.action = .Auth
+                        client.transport.receive(authMessage)
+                    }
+                }
+
                 // RTN4b
                 it("should emit states on a new connection") {
                     let options = AblyTests.commonAppSetup()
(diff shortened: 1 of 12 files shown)


Tests/ARTRealtime+TestSuite.m
```diff
@@ -23,7 +23,7 @@ - (void)testSuite_waitForConnectionToClose:(XCTestCase *)testCase {
         }];
         [self.connection off];
 
-        [self.connection once:ARTRealtimeClosed callback:^(ARTConnectionStateChange *stateChange) {
+        [self.connection once:ARTRealtimeConnectionEventClosed callback:^(ARTConnectionStateChange *stateChange) {
```
ricardopereira: Should I rename `ARTRealtimeClosed` to `ARTRealtimeConnectionEventStateClosed`?

ricardopereira:
@mattheworiordan @tcard Tests are not passing because of the RTN24. Please take a look and see if it's well implemented. If so, I'll fix the client. I want this merged to rebase all the remaining PR's and finish them. Please, have in mind that I only renamed the ERROR event but this should be a good point to rename any Enum State or Enum Event.

Spec/RealtimeClientConnection.swift
```diff
@@ -3468,6 +3468,88 @@ class RealtimeClientConnection: QuickSpec {
...
+                            fail("ConnectionStateChange is nil"); done(); return
+                        }
+                        expect(stateChange.reason).to(beNil())
+                        expect(client.connection.state).to(equal(ARTRealtimeConnectionState.Connected))
+                        expect(stateChange.current).to(equal(ARTRealtimeConnectionState.Connected))
+                        expect(stateChange.current).to(equal(stateChange.previous))
+                        done()
+                    }
+
+                    client.auth.authorize(nil, options: nil) { tokenDetails, error in
```
mattheworiordan: I am not sure that an authorize will always respond with a `CONNECTED` message to be honest. Perhaps what you should do instead, so that the test is closer to the behaviour we expect is:

* Initially connect with token auth callback but no `clientId`, setting a lot TTL of say 35 seconds (I think if > 30s the server will send a AUTH protocol message at 30s before expiry)
* Ensure you are connected
* Wait for the auth callback to be called again but this time auth with a clientId.  We can then be 100% sure the CONNECTED message with new details is expected
* Check that the UPDATE is emitted as you are now and the clientId has changed

Spec/RealtimeClientConnection.swift
```diff
@@ -3468,6 +3468,88 @@ class RealtimeClientConnection: QuickSpec {
...
+                            fail("Reason error is nil"); done(); return
+                        }
+                        expect(error.code) == 40102
+                        expect(client.connection.state).to(equal(ARTRealtimeConnectionState.Connected))
+                        expect(stateChange.current).to(equal(ARTRealtimeConnectionState.Connected))
+                        expect(stateChange.current).to(equal(stateChange.previous))
+                        done()
+                    }
+
+                    client.auth.authorize(nil, options: nil) { tokenDetails, error in
```
mattheworiordan: This will not result in a new CONNECTED message unfortunately.  The authorize may fail, but the connection does not change so no new CONNECTED message is sent.  Instead, in this instance you may need to fabricate a CONNECTED message I am afraid with an embedded error and you will simply need to check that the error reason is set from that.  I know we try not to fabricate, but in this instance I cannot think of a way around that.

Tests/ARTRealtime+TestSuite.m
ricardopereira: To be done on #557.

Spec/RealtimeClientConnection.swift
ricardopereira: Done 502732e.

Spec/RealtimeClientConnection.swift
ricardopereira: Done 59ca18f.

ricardopereira:
@mattheworiordan PTAL

Spec/RealtimeClientConnection.swift
```diff
@@ -3513,8 +3513,8 @@ class RealtimeClientConnection: QuickSpec {
             // RTN24
             it("should set the Reason attribute based on the Error member of the CONNECTED ProtocolMessage") {
```
mattheworiordan: This should say "should set the Connection reason attribute based on the..."

ricardopereira: Done 👍

PR #563, opened 2016-12-28
Update RTP2 for 0.9

ricardopereira:


ricardopereira:
I'll change the client after your validation.

Spec/RealtimeClientPresence.swift
```diff
@@ -957,6 +957,94 @@ class RealtimeClientPresence: QuickSpec {
...
+                        defer { clientSubscribed.dispose(); clientSubscribed.close() }
+                        let channelSubscribed = clientSubscribed.channels.get("foo")
+                        channelSubscribed.attach()
+
+                        let clientPresentMember = ARTRealtime(options: options)
+                        defer { clientPresentMember.dispose(); clientPresentMember.close() }
+                        let channelPresentMember = clientPresentMember.channels.get("foo")
+
+                        var hasInconsistentConnectionIdMethodCalls = 0
+                        let hook = channelSubscribed.presenceMap.testSuite_injectIntoMethodAfter(NSSelectorFromString("hasInconsistentConnectionId")) {
```
mattheworiordan: This type of test condition, IMHO, is bad.  You're testing that this test passes because a private method is being called, instead of simply testing that the behaviour and outcome is correct.  I would prefer that you don't check these types of things unless there is no other way (i.e. you want to confirm that an internal event has happened perhaps).  In this case however, there are plenty of ways to confirm the intended functionality - you could enter a few members in the client, then inject synthesized `ProtocolMessage` events from Ably so that you get a realistic set of protocol messages sent to the client which should prove that compare by timestamp is working i.e. some are discarded, and some are applied.  I implemented all of this yesterday in fact, see https://github.com/ably/ably-ruby/pull/99/commits/e699106b0c55a82040336b4aa3a58a95f620546f#diff-c9dcfc1419f386559fc4a8a0ef368c99R597

Spec/RealtimeClientPresence.swift
```diff
@@ -957,6 +957,94 @@ class RealtimeClientPresence: QuickSpec {
...
+                        let options = AblyTests.commonAppSetup()
+                        let client = ARTRealtime(options: options)
+                        defer { client.dispose(); client.close() }
+                        let channel = client.channels.get("foo")
+
+                        var compareByMsgSerialAndOrIndexMethodCalls = 0
+                        let hook = channel.presenceMap.testSuite_injectIntoMethodAfter(NSSelectorFromString("compareByMsgSerialAndOrIndex")) {
+                            compareByMsgSerialAndOrIndexMethodCalls += 1
+                        }
+                        defer { hook.remove() }
```
mattheworiordan: This test is better than the one above, but it does not confirm that if say the timestamp for a new message is older yet the ID check indicates it's newer (and the connection ID matches the first part of the message), then it works as expected.  I would consider determining the connectionId and injecting messages at a protocol level to test the permutations at https://github.com/ably/ably-ruby/pull/99/commits/e699106b0c55a82040336b4aa3a58a95f620546f#diff-c9dcfc1419f386559fc4a8a0ef368c99R617

Spec/RealtimeClientPresence.swift
```diff
@@ -1045,6 +1045,49 @@ class RealtimeClientPresence: QuickSpec {
...
+                        fail("TestProxyTransport is not set"); return
+                    }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        let partialDone = AblyTests.splitDone(2, done: done)
+                        transport.beforeProcessingReceivedMessage = { protocolMessage in
+                            if protocolMessage.action == .Sync {
+                                let injectLeave = ARTPresenceMessage()
+                                injectLeave.action = .Leave
+                                injectLeave.clientId = "user110"
```
mattheworiordan: I am not sure I follow this.  If the PresenceMessage has no `connectionId`, then how will it  be compared against the members added by `addMembersSequentiallyToChannel`.  Surely you need to make sure at least that the `connectionId` is the same.  Secondly, it would be good to test both scenarios i.e. messages that have effectively arrived before the SYNC (and thus discarded) and those that arrive after the SYNC (and thus used)

Spec/RealtimeClientPresence.swift
```diff
@@ -957,6 +957,94 @@ class RealtimeClientPresence: QuickSpec {
...
+
+                        waitUntil(timeout: testTimeout) { done in
+                            channelPresentMember.presence.enterClient("tester", data: nil) { error in
+                                expect(error).to(beNil())
+                                done()
+                            }
+                        }
+
+                        waitUntil(timeout: testTimeout) { done in
+                            channelSubscribed.presence.get { presences, error in
```
ricardopereira: Done bac7e91.

Spec/RealtimeClientPresence.swift
```diff
@@ -957,6 +957,94 @@ class RealtimeClientPresence: QuickSpec {
...
+                                expect(error).to(beNil())
+                                expect(presences).to(haveCount(1))
+                                done()
+                            }
+                        }
+
+                        waitUntil(timeout: testTimeout) { done in
+                            channelSubscribed.presence.subscribe(.Leave) { presence in
+                                // Check `synthesized leave` event
```
ricardopereira: Done bac7e91.

Spec/RealtimeClientPresence.swift
ricardopereira: @mattheworiordan Yes you're right. I should test behaviour and not implementation details.  I've looked at how Ruby spec is implemented and I changed it: bac7e91.

Spec/RealtimeClientPresence.swift
```diff
@@ -957,6 +957,94 @@ class RealtimeClientPresence: QuickSpec {
                     expect(compareForNewnessMethodCalls) == 1
                 }
 
+                // RTP2b
+                context("compare for newness") {
+
+                    // RTP2b1
+                    it("presence message has a connectionId which is not an initial substring of its id") {
```
ricardopereira: Done bac7e91.

Spec/RealtimeClientPresence.swift
ricardopereira: Done 23e083d.

Spec/RealtimeClientPresence.swift
ricardopereira: Done debb6c6.

ricardopereira:
@mattheworiordan Please check if my changes are correct. If so, I'll fix the client code.

PR #567, opened 2017-01-04
RTP18

ricardopereira:


Spec/RealtimeClientPresence.swift
```diff
@@ -122,6 +122,139 @@ class RealtimeClientPresence: QuickSpec {
...
+                        fail("TestProxyTransport is not set"); return
+                    }
+
+                    expect(channel.presenceMap.syncInProgress).to(beFalse())
+                    expect(channel.presenceMap.members).to(beEmpty())
+
+                    waitUntil(timeout: testTimeout) { done in
+                        let partialDone = AblyTests.splitDone(2, done: done)
+                        channel.presence.subscribe(.Present) { error in
+                            expect(channel.presence.syncComplete).to(beFalse())
```
mattheworiordan: Don't we want to wait until the sync is complete before sending the server-initiated SYNC message?

ricardopereira: @mattheworiordan but the channel has no members at this point. No sync message will be received.

PR #570, opened 2017-01-10
Update RTP5 for 0.9

ricardopereira:


Spec/RealtimeClientPresence.swift
```diff
@@ -316,6 +380,426 @@ class RealtimeClientPresence: QuickSpec {
...
+                        guard let transport = client.transport as? TestProxyTransport else {
+                            fail("TestProxyTransport is not set"); return
+                        }
+
+                        waitUntil(timeout: testTimeout) { done in
+                            let partialDone = AblyTests.splitDone(2, done: done)
+                            transport.beforeProcessingReceivedMessage = { protocolMessage in
+                                if protocolMessage.action == .Attached {
+                                    // No Sync
+                                    expect(protocolMessage.hasPresence).to(beFalse())
```
mattheworiordan: RTPC1 states:

>  If the resumed flag is true and no SYNC is initiated as part of the attach, do nothing as there is no loss of continuity on the channel and no change to presence.

You are testing it to be false.  You may need to inject an ATTACHED ProtocolMessage with has presence flag true for now as the realtime system does not adhere to this yet, hopefully by the end of this week it will be in sandbox so you can perhaps add a TODO to remove the fabricated message.

Spec/RealtimeClientPresence.swift
```diff
@@ -316,6 +380,426 @@ class RealtimeClientPresence: QuickSpec {
...
+                                    else if protocolMessage.action == .Sync {
+                                        transport.beforeProcessingReceivedMessage = nil
+                                        partialDone()
+                                    }
+                                }
+
+                                // Before the sync ends
+                                channel.presenceMap.testSuite_injectIntoMethodBefore(#selector(ARTPresenceMap.endSync)) {
+                                    expect(channel.presenceMap.members).to(haveCount(4))
+                                    expect(channel.presenceMap.localMembers).to(haveCount(1))
```
mattheworiordan: I am not sure this demonstrates:

> . The members re-entered automatically must be removed from the internal PresenceMap ensuring that members present on the channel are constructed from presence events sent from Ably since the channel became ATTACHED

Can you check that when the `Enter` event is sent to Ably that the member has been removed from `localMembers` and is only later added back in when the lib receives the ACK please?

Spec/RealtimeClientPresence.swift
```diff
@@ -316,6 +380,426 @@ class RealtimeClientPresence: QuickSpec {
...
+                                transport.beforeProcessingReceivedMessage = { protocolMessage in
+                                    if protocolMessage.action == .Attached {
+                                        // No Sync
+                                        expect(protocolMessage.hasPresence).to(beFalse())
+                                        // Force Loss of continuity
+                                        protocolMessage.flags = protocolMessage.flags - Int64(ARTProtocolMessageFlag.Resumed.rawValue)
+                                        partialDone()
+                                    }
+                                }
+                                // Re-entered automatically
```
mattheworiordan: As above, please check the the local members are flushed

Spec/RealtimeClientPresence.swift
```diff
@@ -316,6 +380,426 @@ class RealtimeClientPresence: QuickSpec {
...
+                        let options = AblyTests.commonAppSetup()
+                        let client = AblyTests.newRealtime(options)
+                        defer { client.dispose(); client.close() }
+                        let channel = client.channels.get("foo")
+
+                        waitUntil(timeout: testTimeout) { done in
+                            let partialDone = AblyTests.splitDone(3, done: done)
+                            channel.presence.enterClient("tester", data: nil) { error in
+                                guard let error = error else {
+                                    fail("Error is nil"); partialDone(); return
```
mattheworiordan: Do you need to call `partialDone` when `fail` has been called?

mattheworiordan: Bit odd because you're expecting 3 dones, yet there are 4

Spec/RealtimeClientPresence.swift
ricardopereira: Done 5af15b8.

Spec/RealtimeClientPresence.swift
ricardopereira: Done dafe263.

Spec/RealtimeClientPresence.swift
ricardopereira: Done e35ff75.

ricardopereira:
screen shot 2017-02-02 at 14 27 35

```
Test Suite 'All tests' passed at 2017-02-02 14:26:46.318.
	 Executed 667 tests, with 0 failures (0 unexpected) in 449.362 (450.244) seconds
```

I don't know why the hell the test suite doesn't pass 😠

ricardopereira:
Well, seems Travis is having a rough day https://www.traviscistatus.com/incidents/k79mjcv403c4

PR #571, opened 2017-01-10
RTP17

ricardopereira:


Spec/RealtimeClientPresence.swift
```diff
@@ -1401,6 +1401,164 @@ class RealtimeClientPresence: QuickSpec {
...
+                            partialDone()
+                        }
+                        channelB.presence.leave("bye")
+                    }
+                }
+
+                // RTP17a
+                it("all members belonging to the current connection are published as a PresenceMessage on the Channel by the server irrespective of whether the client has permission to subscribe or the Channel is configured to publish presence events") {
+                    let options = AblyTests.commonAppSetup()
+                    options.tokenDetails = getTestTokenDetails(capability: "{\"foo\":[\"presence\",\"publish\"]}")
```
mattheworiordan: BTW. I don't think this test will pass now, you may need to add a TODO to fix one @paddybyers releases this functionality to sandbox

PR #573, opened 2017-01-11
Update RSA8e for 0.9

ricardopereira:


Source/ARTAuth.h
```diff
@@ -43,13 +43,17 @@ ART_ASSUME_NONNULL_BEGIN
  */
 - (void)requestToken:(art_nullable ARTTokenParams *)tokenParams withOptions:(art_nullable ARTAuthOptions *)authOptions
             callback:(void (^)(ARTTokenDetails *__art_nullable, NSError *__art_nullable))callback;
+- (void)requestToken:(void (^)(ARTTokenDetails *__art_nullable, NSError *__art_nullable))callback;
```
ricardopereira: Objective-C doesn't allow optional values on arguments, so we need to have another method.

mattheworiordan: Ok, so this is overloaded?

ricardopereira: Yes.

ricardopereira: Yes.

PR #574, opened 2017-01-11
Update RSA9h for 0.9

ricardopereira:


Spec/Auth.swift
```diff
@@ -1801,6 +1801,55 @@ class Auth : QuickSpec {
...
+                        }
+                        expect(tokenRequest.clientId) == tokenParams.clientId
+                        expect(tokenRequest.ttl) == tokenParams.ttl
+                        expect(tokenRequest.capability) == tokenParams.capability
+                        done()
+                    }
+                }
+
+                waitUntil(timeout: testTimeout) { done in
+                    rest.auth.createTokenRequest { tokenRequest, error in
```
mattheworiordan: I am not sure this is right.  Unlike `authorize`, the defaults are not stored for subsequent requests, yet with this test it appears you assume they are?

PR #575, opened 2017-01-11
Update RSA10g for 0.9

ricardopereira:


Spec/Auth.swift
```diff
@@ -2388,6 +2388,44 @@ class Auth : QuickSpec {
                     }
                 }
 
+                it("should use configured defaults if the object arguments are omitted") {
```
mattheworiordan: Why no error code reference like you normally do in the spec?

PR #582, opened 2017-02-09
Push notifications

ricardopereira:
## Installation

You can install Ably for iOS through CocoaPods, Carthage or manually.

### Installing through [CocoaPods](https://cocoapods.org/) (recommended)

Add this line to your application's Podfile:

    # For Xcode 7.3 and newer
    pod 'Ably', :git => 'https://github.com/ably/ably-ios.git', :branch => 'push'

And then install the dependency:

    $ pod install

Then you can:

     // On Objective-C
    #import <Ably/Ably.h>

    // On Swift
    import Ably

### Installing through [Carthage](https://github.com/Carthage/Carthage/)

Add this line to your application's Cartfile:

    # For Xcode 7.3 and newer
    github "ably/ably-ios" "push"

And then run `carthage update` to build the framework and drag the built Ably.framework into your Xcode project.

### Manual installation 

1. Clone the `push` branch: `git clone -b push git@github.com:ably/ably-ios.git`
2. Drag the directory `ably-ios/ably-ios` into your project as a group.
3. Ably depends on [SocketRocket](https://github.com/facebook/SocketRocket) 0.5.1; get it [from the releases page](https://github.com/facebook/SocketRocket/releases/tag/0.5.1) and follow [its manual installation instructions](https://github.com/facebook/SocketRocket#installing-ios).
4. Ably also depends on [msgpack](https://github.com/rvi/msgpack-objective-C) 0.1.8; get it [from the releases page](https://github.com/rvi/msgpack-objective-C/releases/tag/0.1.8) and link it into your project.


Source/ARTPush.m
```diff
@@ -0,0 +1,132 @@
...
+    if (!deviceToken) {
+        // Waiting for device token
+        [_deviceTokenEmitter once:^(ARTDeviceToken *deviceToken) {
+            [self activate:deviceDetails registerCallback:registerCallback];
+        }];
+        return;
+    }
+
+    if (registerCallback) {
+        self.device.updateToken = registerCallback(deviceDetails, nil);
```
ricardopereira: @tcard maybe this should be async and not waiting for a return.

ricardopereira: In which situation you want to pass an error to the `registerCallback`? I don't see any case.

Source/ARTDeviceDetails.m
```diff
@@ -0,0 +1,50 @@
...
+}
+
+@implementation ARTDeviceDetails
+
++ (instancetype)fromLocalDevice {
+    return [[ARTDeviceDetails alloc] init];
+}
+
+- (instancetype)init {
+    return [self initWithId:[[NSUUID new] UUIDString]];
```
tcard: OK for now, but I did mention ULID, right? It's not crucial but it's also straightforward to use instead. It's what I use in Java so it'd be nice for IDs to be uniform.

Source/ARTDeviceDetails.m
```diff
@@ -0,0 +1,50 @@
...
+            return @"desktop"; //2
+        case ARTDeviceFormFactorEmbedded:
+            return @"embedded"; //3
+    }
+}
+
+@implementation ARTDeviceDetails
+
++ (instancetype)fromLocalDevice {
+    return [[ARTDeviceDetails alloc] init];
```
tcard: This is missing all device info, so this is TODO, right?

Source/ARTJsonLikeEncoder.m
```diff
@@ -479,6 +496,50 @@ - (NSDictionary *)tokenDetailsToDictionary:(ARTTokenDetails *)tokenDetails {
...
+    return deviceDetails;
+}
+
+- (NSDictionary *)devicePushDetailsToDictionary:(ARTDevicePushDetails *)devicePushDetails {
+    NSMutableDictionary *dictionary = [NSMutableDictionary dictionary];
+
+    dictionary[@"transportType"] = devicePushDetails.transportType;
+
+    if (devicePushDetails.deviceToken) {
+        // Normalizing token by removing symbols and spaces
```
tcard: Why? In tests I've used the device token as-is just fine.

Source/ARTPush.h
```diff
@@ -0,0 +1,70 @@
...
+@interface ARTDeviceToken : NSData
+@end
+
+@interface ARTUpdateToken : NSString
+@end
+
+
+#pragma mark ARTPushNotifications interface
+
+#ifdef TARGET_OS_IPHONE
```
tcard: Does this apply to iOS in general?

Source/ARTPush.h
```diff
@@ -0,0 +1,70 @@
...
+- (instancetype)init NS_UNAVAILABLE;
+- (instancetype)init:(ARTRest *)rest;
+
+/// Publish a push notification.
+- (void)publish:(id<ARTPushRecipient>)recipient jsonObject:(ARTJsonObject *)jsonObject;
+
+#ifdef TARGET_OS_IPHONE
+/// Register a device, including the information necessary to deliver push notifications to it.
+- (void)activate;
+- (void)activate:(ARTDeviceDetails *)deviceDetails;
```
tcard: This doesn't take deviceDetails now. It's hard-wired to `rest.push.device`.

Source/ARTPush.h
```diff
@@ -0,0 +1,70 @@
...
+- (void)activate:(ARTDeviceDetails *)deviceDetails registerCallback:(nullable ARTUpdateToken* (^)(ARTDeviceDetails * _Nullable, ARTErrorInfo * _Nullable))registerCallback;
```
tcard: So a problem I've encountered in Java is that this callback doesn't survive closing the app. So if you call activate, then the app closes, then the device token arrives (starting the app, I guess?) and the activation completes, the callback won't exist at that point, will it?

In Java I'm solving this with an Android's BroadcastReceiver. I don't know about iOS.

Source/ARTPush.m
```diff
@@ -0,0 +1,148 @@
...
+}
+
+- (instancetype)init:(ARTRest *)rest {
+    if (self = [super init]) {
+        _rest = rest;
+        _logger = rest.logger;
+        _device = [ARTDeviceDetails fromLocalDevice];
+        _state = ARTPushStateDeactivated;
+        _deviceTokenEmitter = [[ARTEventEmitter alloc] init];
+        _jsonEncoder = [[ARTJsonLikeEncoder alloc] init];
```
tcard: This should respect `rest.options.useBinaryProtocol` and use MsgPack if true.

Source/ARTPush.m
```diff
@@ -0,0 +1,148 @@
...
+
+- (void)activate:(ARTDeviceDetails *)deviceDetails registerCallback:(ARTUpdateToken* (^)(ARTDeviceDetails * _Nullable, ARTErrorInfo * _Nullable))registerCallback {
+    if (self.state == ARTPushStateActivated) {
+        return;
+    }
+
+    NSData *deviceToken = [[NSUserDefaults standardUserDefaults] dataForKey:ARTDeviceTokenKey];
+    if (!deviceToken) {
+        // Waiting for device token
+        [_deviceTokenEmitter once:^(ARTDeviceToken *deviceToken) {
```
tcard: As I said about registerCallback, this won't survive closing the app.

Source/ARTPush.m
tcard: > In which situation you want to pass an error to the registerCallback? I don't see any case.

In none that I can think of, really. Maybe we'll end up removing it.

> maybe this should be async and not waiting for a return.

Yes, it should. It returns `io T`. In iOS, `) -> io T` translates roughly into `, callback: (T, Error))`. Just like  `authCallback`.

Source/ARTPushRecipient.h
```diff
@@ -0,0 +1,47 @@
+//
+//  ARTPushRecipient.h
```
tcard: Sorry you took all the effort to do this, but we really want to keep this "stringly-typed" so that adding new recipient types on the server don't require changing anything in the client libraries.

So ARTPushRecipient should just be NSDictionary.

Source/ARTPush.h
ricardopereira: Yes, it is an old conditional but it specifies the iOS system.

Source/ARTPush.m
ricardopereira: Fixed 08f9d01.

Source/ARTPush.h
ricardopereira: Fixed 08f9d01.

Source/ARTPush.h
ricardopereira: I need to check that on a real device but from what I know is that the app remains in the background when the token is getting retrieved from APNS.

Source/ARTPush.h
```diff
@@ -0,0 +1,70 @@
...
+@interface ARTJsonObject : NSDictionary
+@end
+
+@interface ARTDeviceId : NSString
+@end
+
+@interface ARTDeviceToken : NSData
+@end
+
+@interface ARTUpdateToken : NSString
```
ricardopereira: Yes, that was an initial solution but I changed it to a `typedef` (https://github.com/ably/ably-ios/pull/582/commits/581e76f4d704b6fc69eaeb1ab788d3cccaf15f18#diff-e040b62ac54206ef54da9e6409623314R18). Now it's working fine. 👍

Source/ARTDeviceDetails.h
```diff
@@ -0,0 +1,43 @@
...
+#import <Foundation/Foundation.h>
+#import "ARTPush.h"
+
+@class ARTDevicePushDetails;
+
+NS_ASSUME_NONNULL_BEGIN
+
+extern NSString *const ARTDevicePlatform;
+
+typedef NS_ENUM(NSUInteger, ARTDeviceFormFactor) {
```
ricardopereira: Fixed 64e6dab.

Source/ARTDeviceDetails.m
ricardopereira: @tcard I forgot about it, 👍 . So I should initialise it with the `deviceId` that is stored and the current `updateToken`, if it's in memory?

Source/ARTDeviceDetails.m
ricardopereira: @tcard Yes, you mention it but there is no standard ULID generator. I will look into the paper of ULID and try to implement it.

ricardopereira: BTW, does the Java has it already implemented in the language itself?

Source/ARTJsonLikeEncoder.m
ricardopereira: @tcard You tested it converting it into a string type? The token is a Hex string, something like `<12ce7dda 8032c423 8f8bd40f 3484e5bb f4698da5 8b7fdf8d 5c55e0a2 XXXXXXXX>`. So I should let it be as it is.

ricardopereira: Fixed d67ba78.

Source/ARTPush.m
ricardopereira: Fixed bd73596.

Source/ARTPushRecipient.h
ricardopereira: Done 58f7eab.

Source/ARTDeviceDetails.m
```diff
@@ -0,0 +1,41 @@
...
+//
+//  Created by Ricardo Pereira on 07/02/2017.
+//  Copyright © 2017 Ably. All rights reserved.
+//
+
+#import "ARTDeviceDetails.h"
+#import "ARTDevicePushDetails.h"
+
+NSString *const ARTDevicePlatform = @"ios";
+NSString *const ARTDeviceFormFactor = @"mobile";
```
mattheworiordan: Really hard-coded const?  Given this could run on tablet or watch, I don't think this is right.  Equally, we should really support MacOS at some point too.

Source/ARTDeviceDetails.m
mattheworiordan: > I will look into the paper of ULID and try to implement it.

I don't think we should be writing our own implementation if we can help it.  Is that really necessary if we can't find a lib that does it for us?

Source/ARTPushChannel.m
```diff
@@ -0,0 +1,122 @@
...
+        if (error) {
+            [_logger error:@"%@: subscribe notifications for device %@ in channel %@ failed (%@)", NSStringFromClass(self.class), deviceId, _channel.name, error.localizedDescription];
+        }
+        else {
+            [_logger error:@"%@: subscribe notifications for device %@ in channel %@ failed with status code %ld", NSStringFromClass(self.class), deviceId, _channel.name, (long)response.statusCode];
+        }
+    }];
+}
+
+- (void)subscribeForClientId:(NSString *)clientId {
```
mattheworiordan: Would `clientId` not be implicit @tcard?  Or do you think it's better to leave it as explicit as it's more flexible in future?

Source/ARTDeviceDetails.m
tcard: No, I used a library. Sorry, I assumed there was a library for Obj-C! It doesn't seem to be the case so you can be the first to make a CocoaPod :) Or I will if you don't want to!

Source/ARTDeviceDetails.m
tcard: Right.

Source/ARTPushChannel.m
tcard: In the IDL it is implicit. But I don't know, I guess it can work like Presence#enterClient?

mattheworiordan: > In the IDL it is implicit. But I don't know, I guess it can work like Presence#enterClient?

Not sure I see a use case for a client device to specify `clientId` other than the one implicit in the token.  Servers on the other hand would do this frequently, is this API intended for server-side access as well?

tcard: > Servers on the other hand would do this frequently, is this API intended for server-side access as well?

No, they would use the admin API most likely.

mattheworiordan: Well I would say that we shouldn't encourage people to try and register their device for a clientId that doesn't match their token, so if we want to offer this we should perhaps have an overloaded method where by default `clientId` is not required.

Source/ARTDeviceDetails.m
ricardopereira: There is an C++ library, so I created an Objective-C wrapper of it: https://github.com/whitesmith/ulid. It still need some work but it's working. I'll improve it later to continue this work.

tcard: Nice!

Source/ARTPushChannel.m
ricardopereira: > we should perhaps have an overloaded method where by default clientId is not required

Done 2c68fa9.

ricardopereira:
@tcard `PushAdmin` is done. So, the final step is activation state machine? /cc @mattheworiordan

Source/ARTJsonLikeEncoder.m
tcard: Ah, sorry, I was wrong, this is what I did: `[deviceToken hexString]` using some random `hexString` method I found googling. So yes, removing symbols and spaces should do the same. Sorry!

ricardopereira: Done 4ab1d74.

Source/ARTDeviceDetails.m
ricardopereira: Done 002c882.

Source/ARTPush.m
```diff
@@ -54,6 +55,16 @@ - (void)didFailToRegisterForRemoteNotificationsWithError:(NSError *)error {
     [_logger error:@"ARTPush: device token not received (%@)", [error localizedDescription]];
 }
 
++ (ARTPushActivationStateMachine *)activationMachine {
+    static dispatch_once_t once;
+    static id activationMachineInstance;
+    dispatch_once(&once, ^{
+        // Error: ARTPush.m:62:81: Instance variable '_httpExecutor' accessed in class method
+        //activationMachineInstance = [[ARTPushActivationStateMachine alloc] init:_httpExecutor];
+    });
```
ricardopereira: @tcard `ARTPushActivationStateMachine` should be a singleton right? I need to have in mind which data the state machine needs to perform is work. Does it need more than the local device?

tcard: Yes, it should be a singleton. The persistent data it needs is the current state, the pending events, the device ID (which, once generated, doesn't change), the device's update token and the APNs device token. Also, whether activate/deactivate were called with custom registration or not.

ricardopereira: @tcard I can do requests without the auth token or key, right? I don't really need the Rest instance. So, I can create the singleton with no dependencies.

tcard: @ricardopereira No, you can't, you need to authenticate in order to register, update the registration, and unregister. Unless the user provides the custom callback, of course.

ricardopereira:
@tcard I think this is done.

Source/ARTPushChannel.h
```diff
@@ -0,0 +1,37 @@
...
+@protocol ARTHTTPAuthenticatedExecutor;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ARTPushChannel : NSObject
+
+- (instancetype)init NS_UNAVAILABLE;
+- (instancetype)init:(id<ARTHTTPAuthenticatedExecutor>)httpExecutor withChannel:(ARTChannel *)channel;
+
+- (void)subscribe;
```
tcard: This isn't in the spec.

Source/ARTPushChannel.h
```diff
@@ -0,0 +1,37 @@
...
+- (void)subscribeDevice:(ARTDeviceId *)deviceId;
+- (void)subscribeClient:(NSString *)clientId;
+
+- (void)unsubscribe;
```
tcard: Not in the spec.

Source/ARTPushChannel.h
```diff
@@ -0,0 +1,37 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ARTPushChannel : NSObject
+
+- (instancetype)init NS_UNAVAILABLE;
+- (instancetype)init:(id<ARTHTTPAuthenticatedExecutor>)httpExecutor withChannel:(ARTChannel *)channel;
+
+- (void)subscribe;
+- (void)subscribeDevice:(ARTDeviceId *)deviceId;
```
tcard: This and `unsubscribeDevice` shouldn't take an argument but implicitly use the local device.

Source/ARTPushChannelSubscription.h
```diff
@@ -0,0 +1,25 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ARTPushChannelSubscription : NSObject
+
+@property (nonatomic, readonly) NSString *deviceId;
+@property (nonatomic, readonly) NSString *clientId;
+@property (nonatomic, readonly) NSString *channelName;
+
+- (instancetype)init NS_UNAVAILABLE;
+- (instancetype)initWithDeviceId:(NSString *)deviceId andChannel:(NSString *)channelName;
```
tcard: In the spec, those are static methods named `forDevice` and `forClientId`, so we should stick to those.

Source/ARTPushChannelSubscription.h
```diff
@@ -0,0 +1,25 @@
...
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ARTPushChannelSubscription : NSObject
+
+@property (nonatomic, readonly) NSString *deviceId;
+@property (nonatomic, readonly) NSString *clientId;
+@property (nonatomic, readonly) NSString *channelName;
```
tcard: Should be just `channel`.

Source/ARTPushChannelSubscription.h
```diff
@@ -0,0 +1,25 @@
...
+//  Copyright © 2017 Ably. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ARTPushChannelSubscription : NSObject
+
+@property (nonatomic, readonly) NSString *deviceId;
```
tcard: Nullable, and `clientId` too.

Source/ARTPushDeviceRegistrations.m
```diff
@@ -0,0 +1,98 @@
...
+    if (!deviceDetails.updateToken) {
+        [_logger error:@"%@: update token is missing", NSStringFromClass(self.class)];
+        return;
+    }
+    NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[[NSURL URLWithString:@"/push/deviceRegistrations"] URLByAppendingPathComponent:deviceDetails.id]];
+    NSData *tokenData = [deviceDetails.updateToken dataUsingEncoding:NSUTF8StringEncoding];
+    NSString *tokenBase64 = [tokenData base64EncodedStringWithOptions:0];
+    [request setValue:[NSString stringWithFormat:@"Bearer %@", tokenBase64] forHTTPHeaderField:@"Authorization"];
+    request.HTTPMethod = @"PUT";
+    request.HTTPBody = [[_httpExecutor defaultEncoder] encode:@{
```
tcard: This will just save the device token. It should send the whole thing.

Source/ARTDeviceDetails.m
```diff
@@ -0,0 +1,54 @@
...
+
+#if TARGET_OS_IOS
+#import <UIKit/UIKit.h>
+NSString *const ARTDeviceFormFactor = @"phone";
+#elif TARGET_OS_TV
+NSString *const ARTDeviceFormFactor = @"tv";
+#elif TARGET_OS_WATCH
+NSString *const ARTDeviceFormFactor = @"watch";
+#elif TARGET_OS_SIMULATOR
+NSString *const ARTDeviceFormFactor = @"simulator";
```
tcard: Doesn't the simulator simulate other form factors?

tcard:
@ricardopereira This was getting urgent last week so I had to take over this, so don't worry about it. I'll push my changes (hopefully) shortly.

ricardopereira:
@tcard What about tests? Should I add them?

tcard:
@ricardopereira Not yet, I think. The code we want to test is still in flux.

mattheworiordan:
> The code we want to test is still in flux.

@tcard could we not add test coverage over the REST API at least?  That's not changing much is it? See https://github.com/ably/ably-ruby/pull/115/files for a reasonable set of tests over this functionality or of course https://github.com/ably/ably-js/pull/395/commits/6294597ce0a9688cadda4d4885b392468afcb18a

ricardopereira:
@tcard I will change the way how the user adds his custom register/deregister callback because there is no way to test it nicely using the `AppDelegate`: https://github.com/ably/ably-ios/blob/push/Source/ARTPushActivationStateMachine.m#L207-L224. I didn't thought about it while I was coding 😒

I will change it and use the `NSNotificationCenter`. The user can implement it like:

```swift
NSNotificationCenter.default.addObserver(
    UIApplication.shared,
    selector: #selector(AppNotifications.refreshNotification(_:)), //<-- Custom callback
    name: ARTCustomRegisterCallbackNotificationName, //<-- identifier
    object: nil
)
```

It gives the user more flexibility and a clean `AppDelegate`.

tcard:
@ricardopereira Does that allow you to check whether the user added an observer or?

In any case, I don't think the former way is so hard to test. Instead of hardcoding `[UIApplication sharedApplication].delegate`, you can define a method that returns it and you can replace for testing using the hooks thing.

ricardopereira:
Nevermind, I can't detect if an `ARTCustomRegisterCallbackNotificationName` as registered observers.

ricardopereira:
Test suite is failing because of #669.

funkyboy:
@ricardopereira Please let me know if tests pass also on iOS10 and 11.
Working on [this](https://github.com/ably/ably-ios/pull/681) now, so hopefully things will be more streamlined soon.

ricardopereira:
Done.

screen shot 2018-02-01 at 10 14 35

screen shot 2018-02-01 at 10 09 55

Source/ARTPushActivationStateMachine.m
```diff
@@ -101,7 +101,7 @@ - (void)persist {
 
 - (void)deviceRegistration:(ARTErrorInfo *)error {
     #ifdef TARGET_OS_IOS
-    ARTLocalDevice *local = _rest.device;
+    ARTLocalDevice *local = _rest.device_nosync;
 
     if (![[UIApplication sharedApplication].delegate conformsToProtocol:@protocol(ARTPushRegistererDelegate)]) {
         [NSException raise:@"ARTPushRegistererDelegate must be implemented on AppDelegate" format:@""];
```
funkyboy: Wondering: can this fix or impact on this? https://github.com/ably/ably-ios/issues/673

ricardopereira: @funkyboy If the user is using the `push` branch, then yes.

ricardopereira: That was a fix for #654.

PR #584, opened 2017-03-06
0.8: Sandbox changes

ricardopereira:


mattheworiordan:
Looks good.  Will the 0.9 branch be rebased on top of this then?  We would like to release 0.9 as 1.0 ASAP in fact.  When can we do that?

PR #586, opened 2017-03-10
Thread safety

ricardopereira:


mattheworiordan:
@tcard I think you're going to have to handle this one as I don't really understand the environment / platform well enough.

Source/ARTHttp.m
```diff
@@ -175,18 +179,19 @@ - (void)executeRequest:(NSMutableURLRequest *)request completion:(void (^)(NSHTT
     [self.logger debug:@"%@ %@", request.HTTPMethod, request.URL.absoluteString];
     [self.logger verbose:@"Headers %@", request.allHTTPHeaderFields];
 
+    __weak typeof(self) weakSelf = self;
```
tcard: Why weak?

ricardopereira: When an `ARTHttp` object gets deallocated but the request is still active.

Examples/Tests/TestsTests/TestsTests.swift
```diff
@@ -64,17 +64,19 @@ class TestsTests: XCTestCase {
         self.waitForExpectationsWithTimeout(10, handler: nil)
 
         let backgroundRealtimeExpectation = self.expectationWithDescription("Realtime in a Background Queue")
+        var realtime: ARTRealtime! //strong reference
```
ricardopereira: I'll take a look 👍

Source/ARTHttp.m
ricardopereira: Well, you're right. It doesn't make sense now because of the `[_urlSession invalidateAndCancel];` in the `dealloc()`.

tcard: > When an ARTHttp object gets deallocated but the request is still active.

Yes, but why? I mean, if we use a normal strong reference in the async callback as we did before, the ARTHttp object won't be deallocated until the callback is called, ie. while the request is active. Why would you want to prevent that?

ricardopereira: @tcard In case where you release a Rest client and you don't want to receive any network answers for instances that you don't care anymore. Does it make sense?

ricardopereira: That was something I noticed in a test that creates 50 clients and stores them in an array and when the `removeAll` (items) is called, those clients remained in the memory because they were waiting for the response.

tcard: @ricardopereira But this way it's easy that you e. g. create a ARTRest instance, fire some request expecting a result (history, etc.), then release your instance since you don't need it anymore, but you never get your results. This way you're forcing users to keep instances around not as long as _they_ are using them but as long as requests using it may take, etc. which pretty much amounts to manually managed memory. This behavior goes against what all other libraries do and what you'd expect from a garbage-collected language.

ricardopereira: Done: 94f0f2e

Examples/Tests/TestsTests/TestsTests.swift
ricardopereira: Done: 94f0f2e

Source/ARTHttp.m
ricardopereira: REST is released when the last request finishes. Realtime is release when the `close()` is called. @tcard Do you agree with the Realtime behaviour?

ricardopereira: @tcard When the reference count is 0, the REST will live until the a HTTP request is active. The same goes for the Realtime. It will live until a connection is active.

tcard: OK, but you just said that Realtime will live until `close()`  is called. You mean called internally then? I thought you meant called manually by the user, which didn't make sense to me, as obviously the user still has a reference afterwards, namely the one it just called `closed()` on!

Anyway, if that's what's happening then yes, sounds good.

Spec/RealtimeClientConnection.swift
```diff
@@ -1859,9 +1870,11 @@ class RealtimeClientConnection: QuickSpec {
                     let options = AblyTests.commonAppSetup()
                     options.autoConnect = false
                     options.authCallback = { tokenParams, callback in
-                        callback(getTestTokenDetails(key: options.key, capability: tokenParams.capability, ttl: tokenParams.ttl), nil)
+                        delay(0) {
```
tcard: These things could really use a comment. I have no idea why this delay. Our users also need to make sure they don't return right away from their authCallbacks? If so, we probably should put the delay in our code, not on theirs.

Spec/RealtimeClientConnection.swift
```diff
@@ -1871,61 +1884,45 @@ class RealtimeClientConnection: QuickSpec {
...
-                        }
-                    }
-
-                    var transport: TestProxyTransport!
+                        // Let the token expire
+                        client.connection.once(.Disconnected) { stateChange in
+                            guard let reason = stateChange?.reason else {
+                                fail("Token error is missing"); done(); return
+                            }
+                            reason.code == 40142
```
tcard: Missing an assertion here?

Source/ARTHttp.m
ricardopereira: If a Realtime client has an active connection then that object is only released when the connection is closed.

Spec/RealtimeClientConnection.swift
```diff
@@ -1859,9 +1870,11 @@ class RealtimeClientConnection: QuickSpec {
                     let options = AblyTests.commonAppSetup()
                     options.autoConnect = false
                     options.authCallback = { tokenParams, callback in
-                        callback(getTestTokenDetails(key: options.key, capability: tokenParams.capability, ttl: tokenParams.ttl), nil)
+                        delay(0) {
+                            callback(getTestTokenDetails(key: options.key, capability: tokenParams.capability, ttl: tokenParams.ttl), nil)
+                        }
                     }
-                    let tokenTtl = 1.0
+                    let tokenTtl = 3.0
```
tcard: Why?

Spec/RealtimeClientConnection.swift
```diff
@@ -1871,61 +1884,45 @@ class RealtimeClientConnection: QuickSpec {
...
+                                let stateChange = stateChange!
+                                let state = stateChange.current
+                                let errorInfo = stateChange.reason
+                                switch state {
+                                case .Connected:
+                                    expect(errorInfo).to(beNil())
+                                    // New token
+                                    expect(client.auth.tokenDetails!.token).toNot(equal(options.token))
+                                    done()
+                                case .Failed, .Suspended:
```
tcard: Why did you remove the `.Disconnected`  case?

Spec/RealtimeClientConnection.swift
```diff
@@ -3368,27 +3368,23 @@ class RealtimeClientConnection: QuickSpec {
...
                     }
 
                     waitUntil(timeout: testTimeout) { done in
-                        transport.ignoreSends = true
                         channel.publish(nil, data: "message") { error in
                             expect(error).to(beNil())
                             guard let newTransport = client.transport as? TestProxyTransport else {
                                 fail("Transport is nil"); done(); return
                             }
+                            expect(newTransport).toNot(beIdenticalTo(transport))
```
ricardopereira: The old transport is the `transport`.

Spec/RealtimeClientConnection.swift
```diff
@@ -3368,27 +3368,23 @@ class RealtimeClientConnection: QuickSpec {
...
                             expect(transport.protocolMessagesReceived.filter{ $0.action == .Connected }).to(haveCount(1))
                             expect(newTransport.protocolMessagesReceived.filter{ $0.action == .Connected }).to(haveCount(1))
-                            expect(transport.protocolMessagesSent.filter{ $0.action == .Message }).to(haveCount(0))
-                            expect(transport.protocolMessagesSentIgnored.filter{ $0.action == .Message }).to(haveCount(1))
+                            expect(transport.protocolMessagesReceived.filter{ $0.action == .Connected }).to(haveCount(1))
```
ricardopereira: I removed the `transport.ignoreSends = false` to give to the test a more real case scenario. The message is published and the connection gets broken right after. The test should check the `protocolMessagesSent` for MESSAGE, you're right.

Spec/RealtimeClientConnection.swift
ricardopereira: The `authCallback` shouldn't be blocking the execution because of the internal timers for time outs. You're right, this should be done on our side.

Spec/RealtimeClientConnection.swift
ricardopereira: 1 second is too soon. The response can take a while and the token is no longer accepted by the client at that time.

Spec/RealtimeClientConnection.swift
ricardopereira: The `Disconnected` is always called because of the token.

Spec/RealtimeClientConnection.swift
ricardopereira: Actually, this is already done: https://github.com/ably/ably-ios/blob/e812cda4bde111b2c55a96d9198c6b4812758665/Source/ARTAuth.m#L254

ricardopereira:
@mattheworiordan Finally, it's passing. @tcard Could you please take a look at the most recent commits? Thanks.

Spec/RealtimeClientConnection.swift
ricardopereira: Done 257454a.

Issue #589, opened 2017-03-28
Crash in ARTJsonLikeEncoder


matt-OfferUp:
Our iOS client has been crashing because of the Ably SDK (0.8.8). In the protocolMessageFromDictionary method in ARTJsonLikeEncoder, if an error is received, an ARTErrorInfo is being created. When this happens, if [error artString:@"message"] returns nil the app will crash. This has happened to us quite a bit in the last 24 hours.
I also updated the app to use version 1.0.0 (for my testing) but this still appears to be an issue.

ricardopereira:
@matt-OfferUp Sorry for the inconvenience. It was easy to reproduce the issue from your description, thanks. Can you test the fix by using `pod 'Ably', :git => 'https://github.com/ably/ably-ios.git', :branch => 'fix-589'`? Thanks again.

matt-OfferUp:
This looks like it will work. I dont have a way to force this to happen so I am just looking at the code but this should fix the crash.

ricardopereira:
Updates released:
 - [v1.0.1](https://github.com/ably/ably-ios/releases/tag/1.0.1)
 - [v0.8.11](https://github.com/ably/ably-ios/releases/tag/0.8.11)

mattheworiordan:
Ok, thanks

PR #590, opened 2017-03-28
Fix: should decode a protocol message that has an error without a message

ricardopereira:
#589

### DIFF
--- a/Source/ARTRest.m
+++ b/Source/ARTRest.m
@@ -176,22 +176,52 @@
     }
 }
 
-- (void)executeRequest:(NSMutableURLRequest *)request completion:(void (^)(NSHTTPURLResponse *__art_nullable, NSData *__art_nullable, NSError *__art_nullable))callback {
+- (void)executeRequest:(NSURLRequest *)request completion:(void (^)(NSHTTPURLResponse *__art_nullable, NSData *__art_nullable, NSError *__art_nullable))callback {
     return [self executeRequest:request completion:callback fallbacks:nil retries:0];
 }
 
-- (void)executeRequest:(NSMutableURLRequest *)request completion:(void (^)(NSHTTPURLResponse *__art_nullable, NSData *__art_nullable, NSError *__art_nullable))callback fallbacks:(ARTFallback *)fallbacks retries:(NSUInteger)retries {
+- (void)executeRequest:(NSURLRequest *)request completion:(void (^)(NSHTTPURLResponse *__art_nullable, NSData *__art_nullable, NSError *__art_nullable))callback fallbacks:(ARTFallback *)fallbacks retries:(NSUInteger)retries {
     __block ARTFallback *blockFallbacks = fallbacks;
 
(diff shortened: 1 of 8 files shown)


mattheworiordan:
@ricardopereira thanks for this, however I am not sure this quite addresses the problem correctly. It appears that the library currently expects a JSON / MsgPack response in all cases. However, if for example one of our ELBs returns an error, it will typically not even have a content type and be a simple text response.  Can you add a test that makes sure the library copes with an error message in plaintext and HTML format from ELB?

mattheworiordan:
PS. If the response from the server is not of the correct format, it would be useful if you could artificially construct an ErrorInfo object with the message being the response body (perhaps limit to 1kb in case a large HTML page response is returned) and the `statusCode` property populated with the HTTP status code.

Spec/RestClient.swift
```diff
@@ -1322,6 +1322,19 @@ class RestClient: QuickSpec {
                 }
             }
 
+            // https://github.com/ably/ably-ios/issues/589
+            it("client should handle error messages in plaintext and HTML format") {
```
ricardopereira: @mattheworiordan PTAL

(BTW, this will fail. Just ensuring this is enough.)

mattheworiordan: Sure, test looks good, just need the code now :)

mattheworiordan:
@ricardopereira thanks, I see the test are still failing though. Are those failing tests related to this issue?

ricardopereira:
@tcard Still failing. Should I handle this?

Source/ARTRest.m
```diff
@@ -213,31 +213,27 @@ - (void)executeRequest:(NSURLRequest *)request completion:(void (^)(NSHTTPURLRes
...
-            NSError *dataError = [self->_encoders[response.MIMEType] decodeError:data];
-            if ([self shouldRenewToken:&dataError]) {
-                [self.logger debug:__FILE__ line:__LINE__ message:@"RS:%p retry request %@", self, request];
-                // Make a single attempt to reissue the token and resend the request
-                if (_tokenErrorRetries < 1) {
-                    [self executeRequest:request withAuthOption:ARTAuthenticationTokenRetry completion:callback];
-                    return;
+            if (data) {
+                NSError *dataError = [self->_encoders[response.MIMEType] decodeError:data];
+                if ([self shouldRenewToken:&dataError]) {
```
ricardopereira: @tcard I think you're missing the else case?

tcard: Right, fixed at c595705.

ricardopereira: @tcard the `shouldRenewToken` changes the `dataError` but the right now it ignores that change.

tcard:
Test are passing except for a few that have nothing to do with this. @ricardopereira can you do a release for 0.8 and 1.0?

ricardopereira:
Yes, sure. But I'm AFTK now.

Issue #591, opened 2017-04-11
JSON encoding exception - handle encoding failures


mattheworiordan:
See log below:

```
Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x2bf6849f __exceptionPreprocess
1  libobjc.A.dylib                0x39722c8b objc_exception_throw
2  CoreFoundation                 0x2bf683e5 -[NSException initWithCoder:]
3  Foundation                     0x2ccdadc1 _writeJSONValue
4  Foundation                     0x2ccdc953 ___writeJSONObject_block_invoke
5  CoreFoundation                 0x2be863ed __65-[__NSDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:]_block_invoke
6  CoreFoundation                 0x2be8630f -[__NSDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:]
7  Foundation                     0x2ccdc5d3 _writeJSONObject
8  Foundation                     0x2ccdacff _writeJSONValue
9  Foundation                     0x2ccdab15 -[_NSJSONWriter dataWithRootObject:options:error:]
10 Foundation                     0x2ccdb9af +[NSJSONSerialization dataWithJSONObject:options:error:]
11 Acme                           0x3b3923 -[ARTJsonEncoder encode:] (ARTJsonEncoder.m:31)
12 Acme                           0x3b7e45 -[ARTJsonLikeEncoder encode:] (ARTJsonLikeEncoder.m:708)
13 Acme                           0x3b3cc7 -[ARTJsonLikeEncoder encodeProtocolMessage:] (ARTJsonLikeEncoder.m:122)
14 Acme                           0x3d0a0b -[ARTWebSocketTransport send:] (ARTWebSocketTransport.m:73)
15 Acme                           0x3bec57 -[ARTRealtime sendImpl:callback:] (ARTRealtime.m:595)
16 Acme                           0x3becd3 -[ARTRealtime send:callback:] (ARTRealtime.m:600)
17 Acme                           0x3c1e9b -[ARTRealtimeChannel sendMessage:callback:] (ARTRealtimeChannel.m:222)
18 Acme                           0x3c183b -[ARTRealtimeChannel publishProtocolMessage:callback:] (ARTRealtimeChannel.m:187)
19 Acme                           0x3c0f0b -[ARTRealtimeChannel internalPostMessages:callback:] (ARTRealtimeChannel.m:86)
[...]

PR #592, opened 2017-04-11
Encoder: handle invalid types gracefully

ricardopereira:
#591

### DIFF
--- a/Source/ARTChannel.m
+++ b/Source/ARTChannel.m
@@ -89,9 +89,13 @@
     if (!self.dataEncoder) {
         return message;
     }
-    message = [message encodeWithEncoder:self.dataEncoder error:error];
+    NSError *e = nil;
+    message = [message encodeWithEncoder:self.dataEncoder error:&e];
+    if (e) {
+        [self.logger error:@"ARTChannel: error encoding data: %@", e];
+    }
     if (error) {
-        [self.logger error:@"ARTChannel: error encoding data: %@", *error];
+        *error = e;
     }
     return message;
 }
--- a/Source/ARTChannel.m
+++ b/Source/ARTChannel.m
@@ -47,8 +47,8 @@
 - (void)publish:(art_nullable NSString *)name data:(art_nullable id)data callback:(art_nullable void (^)(ARTErrorInfo *__art_nullable error))callback {
     NSError *error;
     ARTMessage *messagesWithDataEncoded = [self encodeMessageIfNeeded:[[ARTMessage alloc] initWithName:name data:data] error:&error];
-    if (callback && error) {
-        callback([ARTErrorInfo createFromNSError:error]);
+    if (error) {
+        if (callback) callback([ARTErrorInfo createFromNSError:error]);
         return;
     }
     [self internalPostMessages:messagesWithDataEncoded callback:callback];
@@ -62,7 +62,7 @@
     NSError *error;
     ARTMessage *messagesWithDataEncoded = [self encodeMessageIfNeeded:[[ARTMessage alloc] initWithName:name data:data clientId:clientId] error:&error];
     if (error) {
-        callback([ARTErrorInfo createFromNSError:error]);
+        if (callback) callback([ARTErrorInfo createFromNSError:error]);
         return;
     }
     [self internalPostMessages:messagesWithDataEncoded callback:callback];
--- a/Source/ARTChannel.m
+++ b/Source/ARTChannel.m
@@ -47,7 +47,7 @@
 - (void)publish:(art_nullable NSString *)name data:(art_nullable id)data callback:(art_nullable void (^)(ARTErrorInfo *__art_nullable error))callback {
     NSError *error;
     ARTMessage *messagesWithDataEncoded = [self encodeMessageIfNeeded:[[ARTMessage alloc] initWithName:name data:data] error:&error];
-    if (error) {
+    if (callback && error) {
         callback([ARTErrorInfo createFromNSError:error]);
         return;
     }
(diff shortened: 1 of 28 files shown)


Source/ARTJsonEncoder.m
```diff
@@ -22,12 +22,20 @@ - (NSString *)formatAsString {
...
 
-- (NSData *)encode:(id)obj {
-    return [NSJSONSerialization dataWithJSONObject:obj options:0 error:nil];
+- (NSData *)encode:(id)obj error:(NSError **)error {
+    @try {
+        return [NSJSONSerialization dataWithJSONObject:obj options:0 error:error];
+    }
+    @catch (NSException *exception) {
+        if (error) {
+            *error = [[NSError alloc] initWithDomain:ARTAblyErrorDomain code:ARTClientCodeErrorInvalidType userInfo:@{NSLocalizedDescriptionKey: exception.reason}];
```
ricardopereira: @mattheworiordan I decided to not use the `isValidJSONObject` method because it doesn't tell us why it isn't valid. On the other hand, catching the exception gives us more details about the failure.

Source/ARTChannel.m
```diff
@@ -45,38 +45,53 @@ - (void)publish:(NSString *)name data:(id)data {
...
     if (!self.dataEncoder) {
         return message;
     }
-    NSError *error = nil;
-    message = [message encodeWithEncoder:self.dataEncoder error:&error];
-    if (error != nil) {
-        [self.logger error:@"ARTChannel: error encoding data: %@", error];
-        [NSException raise:NSInvalidArgumentException format:@"ARTChannel: error encoding data: %@", error];
+    message = [message encodeWithEncoder:self.dataEncoder error:error];
+    if (error) {
```
tcard: `error && *error`; `error` alone will be true just by passing a non-null `NSError **`.

In fact, so that we always log this, and not only when `error` is passed, we should do something like:

```
NSError *e = nil;
message = [message encodeWithEncoder:self.dataEncoder error:&e];
if (e) {
    [self.logger error:@"ARTChannel: error encoding data: %@", e];
}
 if (error) {
    *error = e;
}
```

Source/ARTEncoder.h
```diff
@@ -43,8 +43,8 @@ ART_ASSUME_NONNULL_BEGIN
 - (art_nullable NSData *)encodePresenceMessage:(ARTPresenceMessage *)message;
 - (art_nullable NSData *)encodePresenceMessages:(NSArray *)messages;
 
-- (art_nullable NSData *)encodeProtocolMessage:(ARTProtocolMessage *)message;
-- (art_nullable ARTProtocolMessage *)decodeProtocolMessage:(NSData *)data;
+- (art_nullable NSData *)encodeProtocolMessage:(ARTProtocolMessage *)message error:(NSError * __autoreleasing *)error;
```
tcard: I'm not sure if `__autoreleasing` is necessary, but I think `__Nullable` is: `NSError *_Nullable *_Nullable`

Source/ARTRealtime.m
```diff
@@ -810,12 +811,28 @@ - (BOOL)isActive {
 - (void)sendImpl:(ARTProtocolMessage *)msg callback:(void (^)(ARTStatus *))cb {
     if (msg.ackRequired) {
         msg.msgSerial = [NSNumber numberWithLongLong:self.msgSerial++];
```
tcard: This should be moved below the error handling, so that errors don't increase msgSerial.

Spec/Utilities.swift
```diff
@@ -31,14 +31,81 @@ class Utilities: QuickSpec {
...
+                        channel.publish([ARTMessage(name: nil, data: NSDate()), ARTMessage(name: nil, data: NSDate())]) { error in
+                            guard let error = error else {
+                                fail("Error shouldn't be nil"); done(); return
+                            }
+                            expect(error.message).to(contain("encoding failed"))
+                            expect(error.reason).to(contain("must be NSString, NSData, NSArray or NSDictionary"))
+                            done()
+                        }
+                    }
+                }
```
tcard: Missing a test for incoming malformed JSON.

Source/ARTEncoder.h
ricardopereira: I did that because I took a look at the `NSJSONSerialization` header file that returns errors as well 😅 `[NSJSONSerialization JSONObjectWithData:<#(nonnull NSData *)#> options:<#(NSJSONReadingOptions)#> error:<#(NSError * _Nullable __autoreleasing * _Nullable)#>]`

Source/ARTRealtime.m
ricardopereira: @tcard Ah you're right. So I need to encode it twice?

ricardopereira: What about decrementing the msgSerial if the encode fails?

Spec/Utilities.swift
tcard: (Also for malformed MsgPack, BTW.)

Source/ARTRealtime.m
tcard: Ah right, I didn't notice we need the updated value. I think `msg.msgSerial = [NSNumber numberWithLongLong:self.msgSerial + 1]` here and then below just `self.msgSerial++;` would be the simplest.

Spec/Utilities.swift
ricardopereira: Done: c4de4b9

Source/ARTChannel.m
ricardopereira: Done f4619aa

Source/ARTRealtime.m
ricardopereira: Done f4619aa. Thanks

mattheworiordan:
Ok, so we would like to extend these tests so that they additionally cover incoming invalid/non-encodeable payloads for both binary (msgpack) and text (JSON) transports i.e. if the client library receives a payload that causes a native encoding fault / exception, then the library must handle that in a way that does not cause it to crash.  We need to test that both for REST and Realtime.

Issue #596, opened 2017-05-02
Crash strategy


mattheworiordan:
@ricardopereira @tcard as discussed, we are going to approach the reported crashes in the library in four ways:

* [x] Wrap all public methods with error handling such that an error is handled, logged, and the connection is moved to the failed state (if appropriate).  The same generic handler should also apply to any incoming events that generate exceptions, and I expect in almost all cases should move the connection to the failed state.
* [x] Like a lot of more traditional lower level libraries, they often make assertions about state/properties/context before executing code.  Whilst this won't add any more resilience to the code i.e. it will still fail, it will hopefully help us to diagnose issues more quickly as we can add more context to the failure i.e. a field was expected, an non-null value was expected, and include other details.
* [x] The logger should store the last X log lines at verbose level (we need to think about the implications of this from a performance perspective though, can we cache the log entries but perhaps not interpolate the log line until needed?) so that when we generate error reports, we have more context to the events leading up to this crash.
* [x] We will consider how we can introduce an Ably client-library exception reporting service where we can send errors when an unexpected error in the library occurs.

PS. Can we double check that https://docs.ably.io/client-lib-development-guide/features/#RTE6 is working.  I am not sure we would need to log these errors to the central service as these are errors in customer code, not ours.  However, we should definitely output that to the local logger and handle the error gracefully.

tcard:
All done.

PR #597, opened 2017-05-02
Log history

ricardopereira:
Regarding #596.

### DIFF
--- a/Source/ARTLog.m
+++ b/Source/ARTLog.m
@@ -43,23 +43,54 @@ static const char *logLevelName(ARTLogLevel level) {
     return [NSString stringWithFormat:@"%s: %@", logLevelName(self.level), self.message];
 }
 
+- (NSString *)description {
+    return [self toString];
+}
+
+#pragma mark - NSCoding
+
+- (id)initWithCoder:(NSCoder *)decoder {
+    self = [super init];
+    if (!self) {
+        return nil;
+    }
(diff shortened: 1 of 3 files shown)


Source/ARTLog.m
```diff
@@ -43,23 +43,54 @@ - (NSString *)toString {
     return [NSString stringWithFormat:@"%s: %@", logLevelName(self.level), self.message];
 }
 
+- (NSString *)description {
+    return [self toString];
+}
+
+#pragma mark - NSCoding
```
tcard: What is this for?

ricardopereira: To save the object in disk with `NSUserDefaults`.

ricardopereira: If necessary. Currently, the logger does not store the history.

Source/ARTLog.m
```diff
@@ -43,23 +43,54 @@ - (NSString *)toString {
...
+    NSUInteger _historyLines;
 }
 
 - (instancetype)init {
-    return [self initCapturingOutput:false];
+    return [self initCapturingOutput:true];
 }
 
 - (instancetype)initCapturingOutput:(BOOL)capturing {
+    return [self initCapturingOutput:true historyLines:10];
```
mattheworiordan: 10 seems awfully low, should we bump to say 50 or 100?

ricardopereira: Done c3afa25.

PR #598, opened 2017-05-03
Crashes

ricardopereira:


tcard:
LGTM so far.

mattheworiordan:
@ricardopereira @tcard I must admit I am still a little nervous about this fix because as far as I can tell, we don't yet understand how this could have happened.  What worries me is that it's plausible we're just covering up a bigger underlying issue, and as far as I can tell, we've not seen any indication that the realtime system is delivering ACK/NACKs out of order for example.

Few things:

* Have we investigated possible causes for this thoroughly enough?  This feels like sticky tape to hide the issue as opposed to fixing the issue.
* If we add exception reporting, can we send more info when this happens (and other issues like this  that we have perhaps recently or some time ago also fixed) to help us diagnose the root cause?

tcard:
> Have we investigated possible causes for this thoroughly enough?

No. We're focusing first on handling unexpected conditions gracefully.

> If we add exception reporting, can we send more info when this happens (and other issues like this that we have perhaps recently or some time ago also fixed) to help us diagnose the root cause?

That's a good idea, I'll add support for sending Sentry reports without crashing.

ricardopereira:
I was afraid that the incoming messages where processed with a different order but I did some tests and the dispatch queue is well implemented. The queue is serial so each task is executed one at a time in the order in which they were added to the queue.

mattheworiordan:
> The queue is serial so each task is executed one at a time in the order in which they were added to the queue.

Ok, well apparently not if it's crashing 😏 

On a serious note though, no other library has reported this problem so it seems unlikely this is a realtime issue.  As such, the iOS lib is probably doing something odd in this area.

ricardopereira:
> On a serious note though, no other library has reported this problem so it seems unlikely this is a realtime issue. As such, the iOS lib is probably doing something odd in this area.

Yes, I agree but I don't know where to look. I was suspicious about the incoming messages order but that's something that is working as expected.

@tcard Do you have any suggestion?

PR #599, opened 2017-05-04
WIP: Report crashes to Sentry.

tcard:
To-do:

- [ ] Tests.
- [x] Catch signals caused by the library (in progress).
- [x] Send device info with report.
- [ ] Get stack trace line numbers to work in Sentry.

### DIFF
--- a/Ably.podspec
+++ b/Ably.podspec
@@ -18,5 +18,5 @@ Pod::Spec.new do |s|
   s.module_map        = 'Source/Ably.modulemap'
   s.dependency 'SocketRocket', '0.5.1'
   s.dependency 'msgpack', '0.1.8'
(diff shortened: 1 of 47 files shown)


tcard:
So the only problem now is that KSCrash has a silly size limit on the data it stores on a crash, and we're sending so much context (log lines, and protocol messages sent back and forth) that it doesn't fit. We can trivially fork KSCrash and increase this limit. I'll try and do that tomorrow morning, and then we've got a pretty complete release.

mattheworiordan:
>  I'll try and do that tomorrow morning, and then we've got a pretty complete release.

Thanks @tcard

tcard:
This is now released as 1.0.4 and 0.8.14.

PR #609, opened 2017-05-31
Test suite

ricardopereira:


Source/ARTRest.m
```diff
@@ -530,6 +530,9 @@ BOOL ARTstartHandlingUncaughtExceptions(ARTRest *self) {
 }
 
 void ARTstopHandlingUncaughtExceptions(ARTRest *self) {
+    if (!self) {
```
tcard: `self` should never be null here. When is it?

tcard:
I'm merging this since it'll improve stability, but I'd like to understand why this happens at some point.

Source/ARTRest.m
ricardopereira: @tcard Sorry, I can't recall when it was null but it wasn't in a normal state.

ricardopereira: @tcard Another thing. I was facing a deadlock as well. I spent some time at my lunch time to figure it out but the time wasn't enough. I will open a new PR.

PR #610, opened 2017-05-31
Deadlock

ricardopereira:
screen shot 2017-05-31 at 22 17 41

### DIFF
--- a/Source/ARTSentry.m
+++ b/Source/ARTSentry.m
@@ -347,6 +347,10 @@ NSString* ART_uuid() {
     [request setHTTPMethod:@"POST"];
     [request setHTTPBody:bodyData];
 
+    if ([[NSProcessInfo processInfo].environment valueForKey:@"ARTUnitTests"]) {
+        return;
+    }
+
     ARTURLSessionServerTrust *session = [[ARTURLSessionServerTrust alloc] init];
     [session get:request completion:^(NSHTTPURLResponse *response, NSData *data, NSError *error) {
         if (error || !response) {
--- a/Source/ARTWebSocketTransport.m
+++ b/Source/ARTWebSocketTransport.m
@@ -272,7 +272,6 @@ enum {
                 break;
             case ARTWsBuggyClose:
             case ARTWsGoingAway:
-            case ARTWsAbnormalClose:
                 // Connectivity issue
                 [s.delegate realtimeTransportDisconnected:s withError:nil];
                 break;
--- a/Source/ARTWebSocketTransport.m
+++ b/Source/ARTWebSocketTransport.m
@@ -39,10 +39,6 @@ enum {
 @implementation ARTWebSocketTransport {
     id<ARTRealtimeTransportDelegate> _delegate;
     ARTRealtimeTransportState _state;
-    /**
-      A dispatch queue for firing the events.
-     */
-    _Nonnull dispatch_queue_t _workQueue;
 }
 
 @synthesize delegate = _delegate;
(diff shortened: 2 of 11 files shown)


Source/ARTSentry.m
```diff
@@ -347,6 +347,10 @@ + (void)report:(NSMutableDictionary *)body to:(NSString *)dns callback:(void (^_
     [request setHTTPMethod:@"POST"];
     [request setHTTPBody:bodyData];
 
+    if ([[NSProcessInfo processInfo].environment valueForKey:@"ARTUnitTests"]) {
```
ricardopereira: BTW, is this enough to ignore Sentry requests?

Source/ARTWebSocketTransport.m
```diff
@@ -272,7 +272,6 @@ - (void)webSocket:(SRWebSocket *)webSocket didCloseWithCode:(NSInteger)code reas
                 break;
             case ARTWsBuggyClose:
             case ARTWsGoingAway:
-            case ARTWsAbnormalClose:
```
ricardopereira: The current code wasn't compiling. Sorry, my fault 😓

ricardopereira:
I think the deadlock happens when:
```
channel attached (channel sync) -> channel on Attached event (listener) -> realtime transport to Failed state (realtime sync) -> transition channels to failed -> channel failed (channel sync <--- Deadlock!)
```

mattheworiordan:
@ricardopereira do you have any ideas how to resolve the deadlocks?

mattheworiordan:
@tcard would thread-safety potentially avoid this deadlock situation because the operations are performed serially, or is this unrelated?

paddybyers:
> would thread-safety potentially avoid this deadlock situation because the operations are performed serially, or is this unrelated?

It depends on what steps we take to ensure thread-safety. If we add locks around accesses to specific data structures then there's every chance we make it worse. This really needs some thought and a systematic approach to which locks exist, and order of acquisition, or we're going to make things worse.

ricardopereira:
We need to ensure that the lock is unique when a `dispatch_sync` is done. It's difficult to guarantee that right now because we have different channel and connection transitions occurring.

ricardopereira:
@tcard BTW, do you know why the test suite is yelling _[!] Unable to find a specification for `KSCrashAblyFork (= 1.15.8-ably-1)`_? Did you push the pod `KSCrashAblyFork`?

tcard:
@ricardopereira Hmm, that sounds right, I didn't think that was possible. I use `dispatch_get_current_queue` (https://github.com/ably/ably-ios/blob/eb3a8f6d16ea74072eb6f39a0a238688e3db950c/Source/ARTRealtimeChannel.m#L433) to avoid this situation on a single queue, but didn't account for several sync dispatches active on top of each other in the same thread. That's a tricky one. We _should_ use `dispatch_async`, but I'm not sure how much of our code relies on those operations being synchronous. Can you do that, see if it fixes things?

So @mattheworiordan, that's not the problem really, but yes, the refactor should fix this.

@paddybyers I'm just going to put every operation on a serial queue. I don't plan to use any locks, really, except if completely necessary (e. g. to hold a lock to something in between several async tasks in the queue, if I need to).

Source/ARTSentry.m
tcard: I guess, yes.

tcard:
@ricardopereira Just to make clear, I'm working now on making the whole library thread-safe, and after those changes this PR would no longer apply. So let's wait for that until doing any new work, since it will likely conflict.

Issue #612, opened 2017-06-07
Connection resume failure detaches all channels


tcard:
From https://support.ably.io/helpdesk/tickets/1915:

> “I’ve been hitting occasional errors which say this "2017-06-06 15:16:02.768 Thred[48904:32755281] WARN: R:0x618000196e70 ARTRealtime: connection has reconnected, but resume failed. Detaching all channels”
>
>According to their docs when you lose connection it should reconnect and re-attach to any channels you’re subscribed to, so why this error is occurring I’m not sure.  I haven’t been able to consistently repro it either so I’m honestly not sure why its happening.”

So I've checked the code and: when after a resume attempt (which happens [if CONNECTING after FAILED, DISCONNECTED and SUSPENDED](https://github.com/ably/ably-ios/blob/a0dbdddcde74b53ff4b8cfcb6417490ca3c622af/Source/ARTRealtime.m#L378-L383)) the received connection ID is a new one, then [we detach all channels](https://github.com/ably/ably-ios/blob/a0dbdddcde74b53ff4b8cfcb6417490ca3c622af/Source/ARTRealtime.m#L556-L562).

At this point, in theory, the channel [got suspended](https://github.com/ably/ably-ios/blob/a0dbdddcde74b53ff4b8cfcb6417490ca3c622af/Source/ARTRealtime.m#L515-L517), so after the detach above [the channel will reattach](https://github.com/ably/ably-ios/blob/a0dbdddcde74b53ff4b8cfcb6417490ca3c622af/Source/ARTRealtime.m#L496-L500) as expected, and the reported log message is just misleading. But if the channel wasn't suspended, e. g. because the library was in the DISCONNECTED state (and thus the channels were still ATTACHED) while the server already forgot its connection state, then nothing will happen and the channels won't reattach.

I think that corner case should be explicitly added to the spec, something like "if, after a CONNECTED message is received when trying to resume, the new connection ID doesn't match the previous one, act as if the connection transitioned from SUSPENDED to CONNECTING to CONNECTED".

mattheworiordan:
> I think that corner case should be explicitly added to the spec, something like "if, after a CONNECTED message is received when trying to resume, the new connection ID doesn't match the previous one, act as if the connection transitioned from SUSPENDED to CONNECTING to CONNECTED".

@tcard sure, that makes sense to me. @paddybyers @SimonWoolf do you agree?

SimonWoolf:
[...]

PR #616, opened 2017-06-24
Fix RTN15c3

ricardopereira:
Discussion #612

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -554,10 +554,10 @@ ART_TRY_OR_MOVE_TO_FAILED_START(self) {
     // Resuming
     if (_resuming) {
         if (![message.connectionId isEqualToString:self.connection.id]) {
-            [self.logger warn:@"R:%p ARTRealtime: connection has reconnected, but resume failed. Detaching all channels", self];
-            // Fatal error, detach all channels
+            [self.logger warn:@"R:%p ARTRealtime: connection has reconnected, but resume failed. Reattaching any attached channels", self];
+            // Reattach all channels
             for (ARTRealtimeChannel *channel in self.channels) {
-                [channel detachChannel:[ARTStatus state:ARTStateConnectionDisconnected info:message.error]];
+                [channel reattachWithReason:message.error callback:nil];
             }
             _resuming = false;
         }
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -2370,7 +2370,12 @@ class RealtimeClientConnection: QuickSpec {
                         defer { client.dispose(); client.close() }
                         let channel = client.channels.get("test")
 
-                        expect(client.connection.state).toEventually(equal(ARTRealtimeConnectionState.connected), timeout: testTimeout)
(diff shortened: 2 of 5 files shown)


Spec/RealtimeClientConnection.swift
```diff
@@ -2390,13 +2395,14 @@ class RealtimeClientConnection: QuickSpec {
                                 done()
                             }
                         }
-                        expect(channel.state).to(equal(ARTRealtimeChannelState.detached))
+                        expect(client.msgSerial).to(equal(0))
+                        expect(channel.state).to(equal(ARTRealtimeChannelState.attaching))
```
ricardopereira: @tcard I changed the test regarding the [last comment](https://github.com/ably/ably-ios/issues/612#issuecomment-306760888) of @SimonWoolf. /cc @mattheworiordan

Source/ARTRealtime.m
```diff
@@ -554,10 +554,10 @@ - (void)onConnected:(ARTProtocolMessage *)message {
     // Resuming
     if (_resuming) {
         if (![message.connectionId isEqualToString:self.connection.id]) {
-            [self.logger warn:@"R:%p ARTRealtime: connection has reconnected, but resume failed. Detaching all channels", self];
-            // Fatal error, detach all channels
+            [self.logger warn:@"R:%p ARTRealtime: connection has reconnected, but resume failed. Reattaching any attached channels", self];
+            // Reattach all channels
             for (ARTRealtimeChannel *channel in self.channels) {
-                [channel detachChannel:[ARTStatus state:ARTStateConnectionDisconnected info:message.error]];
+                [channel reattachWithReason:message.error callback:nil];
```
ricardopereira: @tcard Maybe this should be done after the state transition. WDYT?

Spec/RealtimeClientConnection.swift
mattheworiordan: @ricardopereira this looks good to me. @SimonWoolf do you agre?

SimonWoolf: :+1:
(very minor comment: maybe use `oldConnectionId` or something rather than `expectedConnectionId`, `expect(...).toNot(equal(expectedConnectionId))` reads a bit weirdly :) )

mattheworiordan:
@ricardopereira could you address the minor feedback and release a fix?  It would be good if this fix is applied to 0.8.* if easy.

ricardopereira:
@mattheworiordan Ok but isn't this an issue only for 1.0? It's about the `reattach` feat.

mattheworiordan:
> @mattheworiordan Ok but isn't this an issue only for 1.0? It's about the reattach feat.

Yes, apologies, good point.

Spec/RealtimeClientConnection.swift
ricardopereira: Done fcfef9f. Thanks

Source/ARTRealtime.m
tcard: It would be cleaner, probably, as it's a connection state change side effect so it's weird the side effect happens before the state change. But since there's no dependency on the state change from the reattach it doesn't matter that much.

Issue #607, opened 2017-05-29
Consider not holding references to children objects


tcard:
Right now, this causes a crash:

```swift
var channel: ARTChannel
{
    var rest = ARTRest(key: "...")
    channel = rest.channels.get("foo")
    // rest is deallocated here, since channel only holds a weak ref to it,
    // so its refcount = 0 when the rest var goes out of scope
}
channel.publish(...) // crash here; channel's internal reference to rest is nil!
```

Or just:

```swift
ARTRest(key: "...").channels/* rest deallocated at this point */.get("foo")/* crash here */.publish(...)
```

This is because we're following the pattern of parents creating instances of children, usually at startup (e. g. `rest.channels`, `rest.auth`, `realtime.connection`...), and those children having the parent as a dependency, so they refernce the parent too. To avoid a reference-count cycle that would memory leak, we (correctly) reference the parent objects weakly, ie. not incrementing their reference counts. The problem is that this weak reference can in fact outlive the strong one it copies from.

Instead of using weak refs, we can just not have references from parents to children, and instead allocate new instances each time they're asked for, in the getter. The children would then keep strong references to their parents, and then follow the behavior you'd expect from a GCd language.

Any state children "singleton" objects may have can be stored in the parent instead. For a non-trivial example, `realtime.channels` needs to store its channels somewhere so that you get the same ones, with the same `state` and everything, in each `realtime.channels` instance. So it should store them in `realtime`.

mattheworiordan:
> To avoid a reference-count cycle that would memory leak, we (correctly) reference the parent objects weakly, ie. not incrementing their reference counts. 

I don't follow.  Why do we have a weak reference from a Channel to the Rest/Realtime Client when as far as I can tell, it depends on it?

tcard:
@mattheworiordan They depend on each other. The channel depends on REST/Realtime and REST/Realtime depends on the channel (indirectly, through the `.channels` collection).

[...]

PR #620, opened 2017-07-21
[WIP] Make the library thread-safe.

tcard:
**Work in progress!** There are still some deadlocks around. The test suite is pretty broken right now.

This change makes the library thread-safe. Users can call all methods and access
all properties from whatever thread they want.

To achieve this, this is the general strategy used:

* We have a serial queue where all our code runs.
* To ensure all our code runs in the serial queue, we need to dispatch to it
  from all entry points, namely:
  - User code: all public methods use
  - HTTP requests: all HTTP responses are handled at the same place and dispatched there.
  - Reachability callback from the OS.
  - Events dispatches/timeouts from ARTEventEmitter, which use NSNotificationCenter under the hood: all are dispatched to the queue.
  - WebSocket incoming messages: we set the queue as delegate using SocketRocket's `setDelegateQueue`.
* By default, the queue's delegate is the global queue with QOS_CLASS_BACKGROUND. So all our operations will be very low-priority, which I think is the right choice for Ably. The user can change this by specifying `ARTClientOptions.internalDispatchQueue`.
* Additionally, callbacks provided from the user are dispatched to another queue, which by default is the main queue. This is a convenience, since they probably want to react to Ably's callbacks by updating the UI, which only can be done from the main queue. This can be overriden too, with `ARTClientOptions.dispatchQueue`.
* Public, mutable properties (e. g. connection and channel state) need always to be read and written to from the queue. This is pretty annoying since those reads need to be synchronous, since we're offering a synchronous interface. A synchronous dispatch to a queue is annoying since you can't do it if you're in the queue already, so, from within the queue (ie. from our code) we need to read those properties without the sync dispatch. So those get an additional `<property name>_nosync` getter. Pretty clumsy but, well, this is Objective-C.
* Some public methods don't go async right away, since they have some effects that need to be done right after the call, or need to return something. For instance, `attach` leaves the channel in the ATTACHING state synchronously. Similarly to public, mutable properties as described above, those method get an additional method without the sync dispatch, to use from inside our code.

### DIFF
--- a/Ably.xcodeproj/project.pbxproj
+++ b/Ably.xcodeproj/project.pbxproj
@@ -168,6 +168,7 @@
 		EB2D84F71CD75CCE00F23CDA /* ARTReachability.h in Headers */ = {isa = PBXBuildFile; fileRef = EB2D84F61CD75CCE00F23CDA /* ARTReachability.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		EB2D84FD1CD769B800F23CDA /* ARTOSReachability.m in Sources */ = {isa = PBXBuildFile; fileRef = EB2D84FC1CD769B700F23CDA /* ARTOSReachability.m */; };
 		EB2D85011CD769C800F23CDA /* ARTOSReachability.h in Headers */ = {isa = PBXBuildFile; fileRef = EB2D85001CD769C800F23CDA /* ARTOSReachability.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		EB4B1A0C1F2190BB00467F07 /* ARTRestChannels+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = EB4B1A0B1F2190BB00467F07 /* ARTRestChannels+Private.h */; };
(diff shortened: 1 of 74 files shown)


tcard:
A side effect I didn't see coming from this is that, since more operations are done asynchronously and the library holds to weak references even less time, #607 is even worse now. Something like this doesn't work:

```
{
    let rest = ARTRest(...)
    rest.auth.requestToken(...) {
        // Will never be called!
    }
}
```

Now users will really need to make sure their `rest` and `realtime` objects live until after callbacks are called back.

tcard:
The test suite is passing now. ThreadSanitizer is still coughing out issues, though.

mattheworiordan:
> The test suite is passing now.

🎆 

> ThreadSanitizer is still coughing out issues, though.

Oh, well at least we're getting the warnings I suppose.

tcard:
I believe this is finally ready.

mattheworiordan:
> I believe this is finally ready.

Woop woop.  Thanks @tcard

Issue #625, opened 2017-08-03
Memory leak when publishing via realtime


tcard:
See screencast: https://www.screencast.com/t/IE5uVcUj

The user bisected this as a regression from 1.0.3 to 1.0.4, when Sentry was introduced.

tcard:
I can't reproduce this. I've used this code:

```swift
func publish() {
    let channel = client.channels.get("test")
    channel.publish("event", data: "ASDFASDLPFG,ASDKLPFGMPASDKMFPKOASKMDFPIOADSPIFJ OIMADSOIFCMOIASJDFPIOAMSÒIFPCOIAWJERPMOICJAWEPOCMPEOIRJCMPAOIWECFRPOAIWEJFCPMOIAWEJFMCPOAWIJEFPOCMIOPJAMWIOEFJCMPOAIWEJFPCMO ASDOFKAI,ACSIOEJRSAPOF")
    delay(0.2) {
        publish()
    }
}
```

But the memory kept constant the whole time...

ricardopereira:
@tcard Please ask the user to try this branch and see if it helps: `pod 'Ably', :git => 'https://github.com/ably/ably-ios.git', :branch => 'fix-625'`. I turned off the additional log data for Sentry.

ricardopereira:
Enabling address sanitizer could also help us getting more information: https://developer.apple.com/documentation/code_diagnostics/address_sanitizer/enabling_the_address_sanitizer

The client should enable it and look something like ⬇️ in the log:

```
==7829==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 7 byte(s) in 1 object(s) allocated from:
    #0 0x42c0c5 in __interceptor_malloc /usr/home/hacker/llvm/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:74
    #1 0x43ef81 in main /usr/home/hacker/memory-leak.c:6
    #2 0x7fef044b876c in __libc_start_main /build/buildd/eglibc-2.15/csu/libc-start.c:226

SUMMARY: AddressSanitizer: 7 byte(s) leaked in 1 allocation(s).
```

ricardopereira:
Nevermind. I reproduce it successfully 🙌

ricardopereira:
Code to reproduce the problem:

```
// Using Ably version 1.0.7

import UIKit
import Ably

class ViewController: UIViewController {

    @IBOutlet weak var button: UIButton!

[...]

PR #626, opened 2017-08-05
Fix: memory leak when publishing via realtime

ricardopereira:
#625


ricardopereira:
An attempt to mislead the problem.

mattheworiordan:
@ricardopereira unfortunately not sure I can comment on this.  Will have to rely on @tcard here

tcard:
I'm not really sure what is this. Is this WIP?

ricardopereira:
@tcard The client should try this with `pod 'Ably', :git => 'https://github.com/ably/ably-ios.git', :branch => 'fix-625'` and see if it helps.. I turned off the additional log data for Sentry.

Issue #627, opened 2017-08-08
Objective-C tests review


mattheworiordan:
Our old tests were written in Obj-C and we use the new Swift tests for most of our test coverage now.  We should consider whether we can dispose of the old Obj-C tests completely or not by working out:

* [ ] How much of our coverage is provided by Obj-C tests
* [ ] How easy is it to simply get the Obj-C tests running
* [ ] How easy is it to port the test to Swift

tcard:
There are only four spec items covered only by Objective-C:

- RSC14a
  * https://github.com/ably/ably-ios/blob/96720f7c03fa60c75f69b1d0384648248be499b2/ably-iosTests/ARTRestInitTest.m#L48
- RSL2a
  * https://github.com/ably/ably-ios/blob/96720f7c03fa60c75f69b1d0384648248be499b2/ably-iosTests/ARTRestChannelHistoryTest.m#L198
  * https://github.com/ably/ably-ios/blob/c56fffd1f8e0b2cc7ca0209eaabc19c084030b2f/ably-iosTests/ARTRestChannelHistoryTest.m#L150
  * https://github.com/ably/ably-ios/blob/c56fffd1f8e0b2cc7ca0209eaabc19c084030b2f/ably-iosTests/ARTRestChannelHistoryTest.m#L383
- RTP10d
  * https://github.com/ably/ably-ios/blob/5e4def84ad28da10840d521dbb65424c6b31d531/ably-iosTests/ARTRealtimePresenceTest.m#L1006
- RTP14d
  * https://github.com/ably/ably-ios/blob/96720f7c03fa60c75f69b1d0384648248be499b2/ably-iosTests/ARTRealtimePresenceTest.m#L754

They all look easy enough to port to Swift. I'd recommend that.

mattheworiordan:
Thanks @tcard for reviewing this.

ricardopereira:
@mattheworiordan Will do. Should I then deactivate those legacy tests?

mattheworiordan:
> Should I then deactivate those legacy tests?

Yes, I think so personally so that we have a passing CI again.  @ricardopereira think you'll have time to look at this?

ricardopereira:
Yes, maybe tomorrow or wednesday.

mattheworiordan:
Thanks @ricardopereira

PR #629, opened 2017-08-22
Port legacy tests to Swift

ricardopereira:
#627


ricardopereira:
BTW, why is `RSC14a` missing from the [v1.0 docs](https://docs.ably.io/client-lib-development-guide/features/)?

mattheworiordan:
> BTW, why is RSC14a missing from the v1.0 docs?

@ricardopereira I think because it's covered elsewhere, i.e. https://docs.ably.io/client-lib-development-guide/features/#RSA2 and https://docs.ably.io/client-lib-development-guide/features/#RSA11

Ably.xcodeproj/xcshareddata/xcschemes/Ably.xcscheme
```diff
@@ -20,6 +20,20 @@
                ReferencedContainer = "container:Ably.xcodeproj">
             </BuildableReference>
          </BuildActionEntry>
+         <BuildActionEntry
```
tcard: We might as well just delete them, no?

Spec/RealtimeClientPresence.swift
```diff
@@ -3814,6 +3844,47 @@ class RealtimeClientPresence: QuickSpec {
...
+                        let partialDone = AblyTests.splitDone(4, done: done)
+                        channel.presence.enterClient(john, data: nil) { error in
+                            expect(error).to(beNil())
+                            partialDone()
+                        }
+                        channel.presence.enterClient(max, data: nil) { error in
+                            expect(error).to(beNil())
+                            partialDone()
+                        }
+                        channel.presence.subscribe { message in
```
tcard: Shouldn't this subscription be done before the `enterClient`s, so that we make sure we don't miss them?

mattheworiordan:
@tcard @ricardopereira I was hoping porting these tests would result in a green build.  What's the issue that is preventing a green build then?

PR #636, opened 2017-09-14
Implement RTP11d.

### LINKED ISSUE #630: RTP11d (Presence.get when SUSPENDED) not implemented; throws on DISCONNECTED
> (RTP11d) If the Channel is in the SUSPENDED state then the get function will by default, or if waitForSync is set to true, result in an error with code 91005 and a message stating that the presence state is out of sync due to the channel being in a SUSPENDED state. If however the get function is called with waitForSync set to false, then it immediately returns the members currently stored in the PresenceMap giving developers access to the members that were present at the time the channel became SUSPENDED

Plus, on connection DISCONNECTED, we [throw](https://github.com/ably/ably-ios/blob/master/Source/ARTRealtimePresence.m#L72), while we should follow [RTP16b](https://docs.ably.io/client-lib-development-guide/features/#RTP16b) (although that talks about presence _messages_, so we probably should clarify what happens with `get`).

tcard:
Fixes #630.

### DIFF
--- a/Source/ARTPresenceMap.h
+++ b/Source/ARTPresenceMap.h
@@ -55,6 +55,8 @@ ART_ASSUME_NONNULL_BEGIN
 - (void)onceSyncEnds:(void (^)(__GENERIC(NSArray, ARTPresenceMessage *) *))callback;
 - (void)onceSyncFails:(void (^)(ARTErrorInfo *))callback;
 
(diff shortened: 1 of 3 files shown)


mattheworiordan:
PS. I believe this is a bit of a blocker for a customer, so if we could do an iOS release I think they'd be pretty grateful :)

PR #642, opened 2017-10-26
Push Admin tests

ricardopereira:
It doesn't compile:

screen shot 2017-10-27 at 00 20 42

The Push implementation is missing some methods.

### DIFF
--- a/Spec/PushAdmin.swift
+++ b/Spec/PushAdmin.swift
@@ -1,0 +1,621 @@
+//
+//  PushAdmin.swift
+//  Ably
+//
+//  Created by Ricardo Pereira on 25/10/2017.
+//  Copyright © 2017 Ably. All rights reserved.
+//
+
+import Ably
+import Nimble
+import Quick
+
+class PushAdmin : QuickSpec {
+
+    private static var deviceDetails: ARTDeviceDetails = {
+        let deviceDetails = ARTDeviceDetails(id: "testDeviceDetails")
+        deviceDetails.platform = "ios"
+        deviceDetails.formFactor = "phone"
+        deviceDetails.metadata = NSMutableDictionary()
+        deviceDetails.push.recipient = [
+            "transportType": "apns",
+            "deviceToken": "foo"
+        ]
+        return deviceDetails
+    }()
+
+    private static var deviceDetails1ClientA: ARTDeviceDetails = {
+        let deviceDetails = ARTDeviceDetails(id: "deviceDetails1ClientA")
+        deviceDetails.platform = "android"
+        deviceDetails.formFactor = "tablet"
+        deviceDetails.clientId = "clientA"
+        deviceDetails.metadata = NSMutableDictionary()
+        deviceDetails.push.recipient = [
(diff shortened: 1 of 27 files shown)


ricardopereira:
@tcard The special test-only ablyChannel recipient is reset on each sandbox POST /apps request?

tcard:
@ricardopereira I'm not sure what you mean. Reset in which way? It posts the notification to a channel in the current Ably app, which is set anew in each POST /apps, if that's what you mean.

ricardopereira:
@tcard For example, if the sandbox resets the registered devices to 0 in the second run when I register 2 devices in the first run?

tcard:
@ricardopereira The old registrations won't receive push notifications from the new app created when the test suite runs again, if that's the question. Everything in Ably is encapsulated by app.

Spec/PushAdmin.swift
```diff
@@ -542,13 +542,30 @@ class PushAdmin : QuickSpec {
...
+
+                    waitUntil(timeout: testTimeout) { done in
                         realtime.push.admin.channelSubscriptions.removeWhere(params) { error in
                             expect(error).to(beNil())
                             done()
                         }
                     }
 
                     waitUntil(timeout: testTimeout) { done in
+                        delay(3) {
```
ricardopereira: @tcard This test only passes if I put a delay between the `removeWhere:params` call and the `list:params` call. Without a delay the `list:params` returns 2 subscriptions instead of 0. It seems the backend needs some time to "commit" the changes.

tcard: @ricardopereira Are you setting `fullWait=true`? Sorry, I should've mentioned and it's not in the spec. I set up [a `pushFullWait: bool` client option](https://github.com/ably/ably-java/pull/355/files#diff-f111f958256bd5375e3c22f2b1fed74bR182) and then, if it's on, [add a `fullWait=true` query parameter to every push request](https://github.com/ably/ably-java/pull/355/files#diff-7006bd95b380da0538f1ca6a1971e2b2R76), so that it waits for the full operation to complete. (We don't do that by default because it can be potentially a quite expensive operation.)

ricardopereira: @tcard Thanks. It works now as expected 😄

ricardopereira:
😒

screen shot 2017-11-09 at 10 28 52

ricardopereira:
But we're good:

```
Test Suite 'PushAdmin' passed at 2017-11-09 10:26:12.962.
	 Executed 22 tests, with 0 failures (0 unexpected) in 28.301 (33.053) seconds
Test Suite 'AblySpec.xctest' passed at 2017-11-09 10:26:12.963.
	 Executed 22 tests, with 0 failures (0 unexpected) in 28.301 (33.054) seconds
Test Suite 'All tests' passed at 2017-11-09 10:26:12.964.
	 Executed 22 tests, with 0 failures (0 unexpected) in 28.301 (33.058) seconds
```

screen shot 2017-11-09 at 10 26 19

PR #644, opened 2017-11-09
Fix #640

### LINKED ISSUE #640: Exception thrown when invoking [_channel unsubscribe]
We sometimes experience an exception being thrown for [_channel unsubscribe];

The full stacktrace can be found below. This happens intermittently for us and we can't reliably reproduce the issue.

for now we just added try catch around [unsubscribe] hopefully it doesn't crash our users anymore.

```
Fatal Exception: NSGenericException
*** Collection <__NSArrayM: 0x173c0540> was mutated while being enumerated.
-[ARTEventEmitter resetListeners]
 Raw Text

Fatal Exception: NSGenericException
0  CoreFoundation                 0x1b4bfb3d __exceptionPreprocess
1  libobjc.A.dylib                0x1a747067 objc_exception_throw
2  CoreFoundation                 0x1b4bf5e1 -[NSException name]
3  MY TEST CI                 0x1f7805 -[ARTEventEmitter resetListeners] (ARTEventEmitter.m:258)
4  MY TEST CI                 0x20cb15 -[ARTRealtimeChannel unsubscribe] (ARTRealtimeChannel.m:301)
5  MY TEST CI                 0x16d815 -[AblyManager unsubscribeFromChannelNamed:] (AblyManager.m:25)
6  MY TEST CI                 0x153001 -[BroadcastDisplayerViewController dismissController] (BroadcastDisplayerViewController.m:353)
7  UIKit                          0x207263b5 -[UIApplication sendAction:to:from:forEvent:]
8  UIKit                          0x20726349 -[UIControl sendAction:to:forEvent:]
9  UIKit                          0x20710979 -[UIControl _sendActionsForEvents:withEvent:]
10 UIKit                          0x20725c85 -[UIControl touchesEnded:withEvent:]

```

ricardopereira:
#640

### DIFF
--- a/Spec/RealtimeClient.swift
+++ b/Spec/RealtimeClient.swift
@@ -1328,6 +1328,28 @@ class RealtimeClient: QuickSpec {
                 expect(result).to(equal(expectedOrder))
             }
 
+            // Possible issue of https://github.com/ably/ably-ios/issues/640
+            fit("should unlock queue when the timer ends") {
+                let options = ARTClientOptions(key: "xxxx:xxxx")
+                options.autoConnect = false
+                let client = ARTRealtime(options: options)
+                let channel = client.channels.get("foo")
+
+                waitUntil(timeout: testTimeout) { done in
+                    // Schedule a timer in the `internalDispatchQueue`
+                    let listener = client.internalEventEmitter.once { _ in }
+                    listener.setTimer(0.2, onTimeout: {
+                        // another lock
+                        channel.unsubscribe()
+                        done()
+                    })
+                    listener.startTimer()
+
+                    // lock
+                    channel.unsubscribe()
+                }
+            }
+
             it("should never register any connection listeners for internal use with the public EventEmitter") {
                 let options = AblyTests.commonAppSetup()
                 options.autoConnect = false
--- a/Spec/RealtimeClient.swift
+++ b/Spec/RealtimeClient.swift
@@ -1328,54 +1328,29 @@ class RealtimeClient: QuickSpec {
                 expect(result).to(equal(expectedOrder))
             }
 
-            // Possible issue of https://github.com/ably/ably-ios/issues/640
-            xit("should unlock queue when the timer ends") {
-                let options = ARTClientOptions(key: "xxxx:xxxx")
-                options.autoConnect = false
-                let client = ARTRealtime(options: options)
-                let channel = client.channels.get("foo")
-
-                waitUntil(timeout: testTimeout) { done in
-                    // Schedule a timer in the `internalDispatchQueue`
-                    let listener = client.internalEventEmitter.once { _ in }
-                    listener.setTimer(0.2, onTimeout: {
-                        // another lock
-                        channel.unsubscribe()
-                        done()
-                    })
-                    listener.startTimer()
-
-                    // lock
-                    channel.unsubscribe()
-                }
-            }
-
             class AblyManager {
                 static let sharedClient = ARTRealtime(options: { $0.autoConnect = false; return $0 }(ARTClientOptions(key: "xxxx:xxxx")))
             }
 
-            // Possible issue of https://github.com/ably/ably-ios/issues/640
-            xit("should dispatch in user queue") {
+            // Issue https://github.com/ably/ably-ios/issues/640
+            it("should dispatch in user queue when removing an observer") {
                 class Foo {
-                    let channel = "foo"
                     init() {
-                        AblyManager.sharedClient.channels.get(channel).subscribe { _ in
+                        AblyManager.sharedClient.channels.get("foo").subscribe { _ in
                             // keep reference
                             self.update()
                         }
                     }
-                    deinit {
-                        close()
-                    }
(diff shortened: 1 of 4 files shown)


ricardopereira:
d6353e1 test raises a EXC_BAD_INSTRUCTION in `[ARTRealtimeChannel unsubscribe]` because of a `_dispatch_barrier_sync_f_slow `:

screen shot 2017-11-09 at 19 27 31

tcard:
@ricardopereira This is expected. You can't call `unsubscribe` from an `internalEventEmitter`, since a `internalEventEmitter` works exclusively within the internal queue, already holding the lock.

I'll comment on the issue what I think the problem is.

ricardopereira:
(3e1bd48) Another try to reproduce the issue but no success.

ricardopereira:
@tcard I think we just need to change the `ARTInternalEventEmitter` to `ARTPublicEventEmitter`. 

https://github.com/ably/ably-ios/blob/2027ee3690f5fa2af3c1f629d8471934d6e79348/Source/ARTRealtimeChannel.m#L418-L420

https://github.com/ably/ably-ios/blob/2027ee3690f5fa2af3c1f629d8471934d6e79348/Source/ARTRealtimeChannel.m#L66

tcard:
@ricardopereira No, that's for dispatching to the internal queue when the user is in the user (main) queue. This is the other way around, from the internal queue we need to dispatch to the user queue. And this needs to happen for internal event emitters too, at least those that potentially hold references to user objects.

ricardopereira:
> This is the other way around, from the internal queue we need to dispatch to the user queue.

@tcard But we are already doing that by encapsulating each user callback with our implementation which dispatches the callback to the user queue:

```
    if (cb) {
        void (^userCallback)(ARTMessage *__art_nullable m) = cb;
        cb = ^(ARTMessage *__art_nullable m) {
            ART_EXITING_ABLY_CODE(_realtime.rest);
            dispatch_async(_userQueue, ^{
                userCallback(m);
            });
        };
    }
```

tcard:
@ricardopereira Right, and that's perfect, but what I didn't anticipate is that there's another way of executing user code from our internal queue: by causing some of their object's reference counts to drop to zero, hence automatically calling their `dealloc` method. Which is exactly what happened at #640.

When `off` is called, we drop the reference to a callback, and that callback may be holding references to other objects, and those objects may have a `dealloc` method. Hence, dropping references to user callbacks must be executed in the user queue, not in our internal queue.

Spec/RealtimeClient.swift
```diff
@@ -1355,7 +1355,7 @@ class RealtimeClient: QuickSpec {
             }
 
             // Possible issue of https://github.com/ably/ably-ios/issues/640
-            fit("should dispatch in user queue") {
+            xit("should dispatch in user queue") {
```
tcard: This should be fixed.

Spec/RealtimeClient.swift
```diff
@@ -1328,6 +1328,56 @@ class RealtimeClient: QuickSpec {
                 expect(result).to(equal(expectedOrder))
             }
 
+            // Possible issue of https://github.com/ably/ably-ios/issues/640
+            xit("should unlock queue when the timer ends") {
```
tcard: I think we can remove this one, the other reproduces the issue more clearly from a user persoective without meddling with the internal emitter.

Spec/RealtimeClient.swift
```diff
@@ -1328,6 +1328,56 @@ class RealtimeClient: QuickSpec {
...
+                    }
+                    func update() {
+                    }
+                    func close() {
+                        AblyManager.sharedClient.channels.get(channel).unsubscribe()
+                    }
+                }
+
+                let foo = Foo()
+                foo.close()
```
tcard: Hmm, does this really reproduce the issue? I think if you call `close` here, then it will remove the listener you add at `init` but since you still have a reference to the created object its `deinit` won't be called, and when you drop the reference just after, since you already removed the listener, its `deinit`'s `unsubscribe` will just do nothing.

I think `deinit` should just call `unsubscribe` directly; we don't need `close`.

Please confirm this reproduces the issue by commenting out the `dispatch_async` in `removeListener` and making sure this test then fails, and stops failing if you put it back.

ricardopereira: > Hmm, does this really reproduce the issue?

I think so but it always pass. I'll change the code with your points.

ricardopereira: It always passes d0b8aeb

ricardopereira: Should we ask the client to test using the branch?

tcard: @ricardopereira Yes, but we also need to have a test that actually reproduces the problem. Maybe instead of `self` we can reference a different object, then make sure we drop the other reference to it before dropping the reference to `foo`?

ricardopereira:
Success!

screen shot 2017-11-16 at 17 10 13

ricardopereira:
Squashed.

Issue #641, opened 2017-10-26
Presence map is using clientId as its key, rather than memberKey


SimonWoolf:
The lib solemnly set a memberKey for presence messages... and then ignores it and just uses the clientId as the key in the presence map.

https://github.com/ably/ably-ios/blob/7d18a06e190e6bd1fef7b879a258af521d6fcabe/Source/ARTPresenceMap.m#L109 https://github.com/ably/ably-ios/blob/7d18a06e190e6bd1fef7b879a258af521d6fcabe/Source/ARTPresenceMap.m#L128

So presence behaviour for clients which share a connectionId is currently completely broken.

See e.g. customer reporting an issue caused by this [here](https://app.intercom.io/a/apps/ua39m1ld/respond/inbox/all/conversations/12510079752) .

Relevant spec items: RTP2, RTP2a

SimonWoolf:
Also:

- could we please have a test which enters multiple clients on one connection, leaves one, etc. No such test currently or this wouldn't been caught
- and perhaps some more tests of presence syncing? current test coverage looks a little thin. See e.g. https://github.com/ably/ably-js/blob/master/spec/realtime/sync.test.js for inspiration
- It looks like presence.get doesn't currently support filtering by clientId/connectionId, so also fails RTP11c2 and RTP11c3

mattheworiordan:
@ricardopereira would you mind looking at this ASAP as this sounds like a pretty major bug?  Also, any idea why the issue was closed?

PR #647, opened 2017-11-15
Fix: presence map key and get with query (clientId and connectionId)

ricardopereira:
#641

### DIFF
--- a/Spec/RealtimeClientPresence.swift
+++ b/Spec/RealtimeClientPresence.swift
@@ -3590,6 +3590,174 @@ class RealtimeClientPresence: QuickSpec {
                         }
                     }
 
+                    // RTP11c2
+                    fit("should return members filtered by clientId") {
+                        let options = AblyTests.commonAppSetup()
+                        let now = NSDate()
+
+                        var clientMembers: ARTRealtime?
+                        defer { clientMembers?.dispose(); clientMembers?.close() }
+                        waitUntil(timeout: testTimeout) { done in
+                            clientMembers = AblyTests.addMembersSequentiallyToChannel("foo", members: 101, options: options) {
+                                done()
+                            }
+                        }
+
+                        let clientSubscribed = AblyTests.newRealtime(options)
+                        defer { clientSubscribed.dispose(); clientSubscribed.close() }
+                        let channelSubscribed = clientSubscribed.channels.get("foo")
+
+                        let presenceData: [ARTPresenceMessage] = [
+                            ARTPresenceMessage(clientId: "a", action: .enter, connectionId: "one", id: "one:0:0", timestamp: now as Date),
+                            ARTPresenceMessage(clientId: "a", action: .leave, connectionId: "one", id: "fabricated:0:1", timestamp: (now as Date) + 1),
+                            ARTPresenceMessage(clientId: "b", action: .enter, connectionId: "one", id: "one:0:2", timestamp: now as Date),
+                            ARTPresenceMessage(clientId: "b", action: .leave, connectionId: "one", id: "fabricated:0:3", timestamp: (now as Date) - 1),
+                            ARTPresenceMessage(clientId: "c", action: .enter, connectionId: "one", id: "fabricated:0:4", timestamp: now as Date),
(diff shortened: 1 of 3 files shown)


Spec/RealtimeClientPresence.swift
```diff
@@ -3590,6 +3590,174 @@ class RealtimeClientPresence: QuickSpec {
                         }
                     }
 
+                    // RTP11c2
+                    fit("should return members filtered by clientId") {
```
tcard: `fit`?

Spec/RealtimeClientPresence.swift
```diff
@@ -3590,6 +3590,174 @@ class RealtimeClientPresence: QuickSpec {
...
+                        let options = AblyTests.commonAppSetup()
+                        let now = NSDate()
+
+                        var clientMembers: ARTRealtime?
```
tcard: Is this used for anything? It doesn't seem to, as then you use `presenceData` as mock?

ricardopereira: @tcard Just to be sure it filters different connectionId's from 101 members.

tcard: Sorry, I don't follow. Where are you checking that? I only see that you check against the `presenceData` dataset, not against the members you actually push.

Spec/RealtimeClientPresence.swift
```diff
@@ -3590,6 +3590,174 @@ class RealtimeClientPresence: QuickSpec {
...
+
+                        waitUntil(timeout: testTimeout) { done in
+                            channelSubscribed.presence.get(query) { members, error in
+                                expect(error).to(beNil())
+                                guard let members = members else {
+                                    fail("Members is nil"); done(); return
+                                }
+                                expect(members).to(haveCount(2))
+                                expect(members).to(allPass({ (member: ARTPresenceMessage?) in member!.action != .absent }))
+                                expect(members.filter{ $0.clientId == "a" }).to(beEmpty())
```
tcard: How will this be empty, given that, in `presenceData`, messages with client ID `a` also have connection ID `one`?

ricardopereira: @tcard client ID `a` has a `leave` action.

Sorry, I copy&pasted the mock data from another test and maybe that's confusing you.

Spec/RealtimeClientPresence.swift
ricardopereira: The query to filter members by `connectionId` https://github.com/ably/ably-ios/blob/fa2cf962304fd1d968813b808b62f483cb046d05/Spec/RealtimeClientPresence.swift#L3743

Spec/RealtimeClientPresence.swift
tcard: Ah, I see the timestamps. It may be a bit clearer if the order in the array was the chronological order, but never mind.

Spec/RealtimeClientPresence.swift
tcard: But 1. that's not even in this test, and 2. in that query you're also checking against the `presenceData` dataset by mocking the SYNC process, so `clientMembers` are never retrieved unless I'm missing something.

ricardopereira: Yes, you're right. I was looking at the other test. I'll change it.

ricardopereira: Done d157cb2.

Spec/RealtimeClientPresence.swift
ricardopereira: Done d157cb2. Thanks.

ricardopereira:
Both are passing https://travis-ci.org/ably/ably-ios/builds/302897096#L3032-L3033 but maybe I should look at the other tests. I don't know if it was me that broke the tests or they already were broken.

ricardopereira:
WDYT?

tcard:
@ricardopereira I think this PR can be merged and that should be addressed separately (https://github.com/ably/ably-ios/issues/652).

PR #648, opened 2017-11-15
Fix: ambiguous reference to member 'dataTask(with:completionHandler:)'

ricardopereira:


.travis.yml
```diff
@@ -6,14 +6,16 @@ env:
   - LANG=en_US.UTF-8
 before_install:
 install:
-  # Install Scan
-  - gem install scan -v 0.14.2
+  # Install Fastlane
+  - gem install fastlane -v 2.64.0
```
tcard: Why?

ricardopereira: `scan` is no more a stand alone app.

ricardopereira: it's now bundled in `fastlane`. `scan` always was a part of fastlane.

Issue #614, opened 2017-06-09
Reported crash when an ATTACH is responded to with a DETACHED after a while


SimonWoolf:
User reports crash with ably-ios, see debugger screenshot in https://app.intercom.io/a/apps/ua39m1ld/respond/inbox/all/conversations/10249232111 . Further debugging revealed that this happens when Ably fails to instance the channel, so crashes when an ATTACH request is responded to (after 30s) with a DETACHED (e.g. `{"action":13,"error":{"message":"Timeout waiting for master to become ready","code":50000,"statusCode":500},"channel":"channelname"}`)

tcard:
The screenshot:

<img width="1392" alt="screen shot 2017-06-09 at 16 36 56" src="https://user-images.githubusercontent.com/727422/32372289-8439f55e-c094-11e7-81a0-6cc3fed5dcbc.png">

tcard:
A test should be added that simulates this scenario and checks that it's handled properly. I think it's just calling the `attach` callback with the error and emitting a ChannelEvent; it doesn't need to move from INITIALIZED to DETACHED.

ricardopereira:
I created a test and seems to be working fine: https://github.com/ably/ably-ios/blob/16cfdaa1ba0b4452698cb67116df1cd39683b2b5/Spec/RealtimeClientChannel.swift#L3441-L3457

ricardopereira:
@SimonWoolf @tcard Am I missing something? Thanks.

tcard:
@ricardopereira This was before the thread-safety stuff and everything, so maybe it's fixed. Nothing we can do anyway.

PR #651, opened 2017-11-15
Fix: when an ATTACH request is responded with a DETACHED

ricardopereira:
#614

### DIFF
--- a/Spec/RealtimeClientChannel.swift
+++ b/Spec/RealtimeClientChannel.swift
@@ -3423,6 +3423,40 @@ class RealtimeClientChannel: QuickSpec {
                 }
             }
 
+            // https://github.com/ably/ably-ios/issues/614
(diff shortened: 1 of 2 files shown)


Spec/RealtimeClientChannel.swift
```diff
@@ -3423,6 +3423,40 @@ class RealtimeClientChannel: QuickSpec {
                 }
             }
 
+            // https://github.com/ably/ably-ios/issues/614
+            fit("should not crash when an ATTACH request is responded with a DETACHED") {
```
ricardopereira: `fit` just to show that the test is passing.

tcard: OK, just learned about f* and x*... Kind of ugly in a framework which is all about descriptive test suites.

Issue #638, opened 2017-10-18
Implement RTN23a idle timeout


SimonWoolf:
Part of the 1.0 spec changes is [RTN23a](https://docs.ably.io/client-lib-development-guide/features/#RTN23a) . AFAICT this is not currently implemented in ably-ios, and I can't see any issues referring to `maxIdleInterval` or that spec item -- perhaps an oversight. In any case, this should to be done; ios and android devices can get into periods of dodgy network connectivity, and the idle timer is very useful for detecting that clientside, when otherwise there wouldn't be any.

See https://docs.ably.io/client-lib-development-guide/features/#RTN23a for spec, and https://github.com/ably/ably-js/blob/master/common/lib/transport/transport.js#L187-L213 for an example implementation in ably-js. (If ably-ios can detect websocket pings, don't forget to trigger onActivity [when you get a ping](https://github.com/ably/ably-js/blob/master/common/lib/transport/websockettransport.js#L73)).

PR #660, opened 2017-12-09
RTN23

ricardopereira:
#638

### DIFF
--- a/Source/ARTConnectionDetails+Private.h
+++ b/Source/ARTConnectionDetails+Private.h
@@ -15,6 +15,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (readwrite, strong, nonatomic, nullable) NSString *clientId;
 @property (readwrite, strong, nonatomic, nullable) NSString *connectionKey;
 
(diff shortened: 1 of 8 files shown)


SimonWoolf:
:+1: Looks good. (One thought: could maybe have a quick test that `client.maxIdleInterval` is `15` for the simulated connection which sets it to 15000 in the connectionDetails, to check that it's being converted from milliseconds to seconds).

FWIW: The approach you've gone with here is what I did in ably-js at first. I later changed it to one that [just records the last activity time](https://github.com/ably/ably-js/blob/master/common/lib/transport/transport.js#L187-L213) (and only sets a new timer at most every 15s) after [Paddy noted that the initial way was quite expensive](https://github.com/ably/ably-js/issues/358). The new way gives the additional advantage that you can use the lastActivity time for the connection state freshness check in https://github.com/ably/ably-ios/issues/645 . Not saying you need to change approach to that now -- for the moment, better to get this out as-is -- but something to keep in mind as a possibility when you implement https://github.com/ably/ably-ios/issues/645 .

Issue #654, opened 2017-11-24
Deadlock in [rest device] when called from activation state machine


tcard:
A customer has reported a crash due to this. The call stack shows that `[rest device]` was called from the state machine while already holding the lock. We need to ensure this doesn't happen; we probably need a `[rest device_nosync]` method like the others we have around, and use that from the activation machine, and put an `dispatch_async` on entry points of the state machine.

tcard:
Here's a screenshot of the crash:

<img width="1440" alt="screen shot 2017-11-24 at 10 23 32" src="https://user-images.githubusercontent.com/727422/33219198-53dbe738-d141-11e7-8130-2e2679780fae.png">

itolisto:
Hello @tcard and @ricardopereira

If I can help you, please let me know, I would gladly collaborate to solve this issue.

itolisto:
Hello @tcard and @ricardopereira,

Just wondering if you have been able to take a look on this? And again, if there's anything I can do, I'll be happy to do it.

ricardopereira:
@itolisto Not yet, sorry. But I'll look at it today. You're using the `push` branch, right?

itolisto:
@ricardopereira Yes, I’m using the push branch.

itolisto:
Hi @ricardopereira,

Just checking again if you have had the chance to take a look on this. Sorry if I'm being annoying (: And again, if I can help please tell me. 👍

ricardopereira:
@itolisto Sorry. I had to deal with an other issue the other day. I'll reach you later today to test this out please 😊

ricardopereira:
I reproduced it successfully:

![screen shot 2017-12-13 at 11 27 53](https://user-images.githubusercontent.com/3541185/33937068-e0ec3c60-dff9-11e7-83ed-5eef559c92d5.png)

Basically, I'm calling `activate()` in a `deactivate` event.

ricardopereira:
A test should be added with this issue at #662

ricardopereira:
@itolisto Can you test the fix using the `push-fix-654` branch please? Thanks 👍

itolisto:
@ricardopereira Testing this right now.

itolisto:
@ricardopereira I can't reproduce the issue anymore, I believe this is fixed. 👍 

Thank you for your help

ricardopereira:
Good to hear! Thank you

PR #662, opened 2017-12-13
Push ActivationStateMachine tests

ricardopereira:
This is an old branch that needs to be updated with recent changes from `push`.


ricardopereira:
Add a test for #654

PR #671, opened 2018-01-19
Swift 4: upgrade settings and dependencies

funkyboy:
@ricardopereira We need an estimate of the complexity of migrating to the newest version Xcode, Swift 4 and possibly also upgrading Nimble and Quick.

I created this branch from master just to have a playground and see which are the obstacles.

cc @paddybyers


funkyboy:
@ricardopereira Take a look at the failures. Might help to come up with an estimate for the upgrade.

funkyboy:
@ricardopereira ready for review once everything is green.

PR #683, opened 2018-02-02
Pull specific version of SocketRocket

funkyboy:


funkyboy:
@ricardopereira current declaration could lead to non-deterministic builds. Agree?

ricardopereira:
@funkyboy Agreed. Thanks.

PR #687, opened 2018-02-08
Run tests in parallel

funkyboy:


funkyboy:
Some stats: each job takes roughly 15m.
But we can run only 2 in parallel. 
Thus we need ~30m to have results for iOS9, 10 and 11.

PR #693, opened 2018-02-19
Add supported platform to README

### LINKED ISSUE #692: Document supported platforms
In the README

funkyboy:
Fix #692

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # [Ably](https://www.ably.io) iOS client library
 
 An iOS client library for [ably.io](https://www.ably.io), the realtime messaging service, written in Objective-C.
+The following platform are currently supported: iOS9, iOS10 and iOS11.
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


funkyboy:
@paddybyers Technically I didn't drop the compatibility with iOS8 yet in the configuration. So it will "work", as in compile and not throw runtime errors.

I am afraid the message in the README can be quite long: 

- we are not breaking compatibility with iOS8 on purpose
- but we are not running tests on iOS8
- so we don't know how it behaves on that platform
- and if you spot a bug specific to iOS8 it's likely that we are not going to fix it, or give it low priority.

I'd keep the positive version: "we support x, y and z". Or even "this SDK is compatible with iOS9 and above".
[Firebase for example]("https://firebase.google.com/docs/ios/setup) says "An Xcode project targeting iOS 8 or above".

paddybyers:
> I'd keep the positive version: "we support x, y and z". Or even "this SDK is compatible with iOS9 and above".

OK, fine. If we later we or customers find things that are broken on earlier versions we can decide whether we explicitly state that it is not compatible with iOS 8, or we have a statement that says something like "some features are not available on earlier iOS versions; see <issue>".

funkyboy:
@paddybyers Improved again the description.

PR #697, opened 2018-02-21
Improve supported platforms description

### LINKED ISSUE #696: Have a clearer Supported Platforms section in the README

funkyboy:
Fix #696

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,18 @@
 # [Ably](https://www.ably.io) iOS client library
 
 An iOS client library for [ably.io](https://www.ably.io), the realtime messaging service, written in Objective-C.
-This SDK is compatible with projects that target iOS9, iOS10 and iOS11.
+
+## Supported platforms
+
+This SDK is compatible with projects that target:
+
+- iOS9
+- iOS10 
+- iOS11
+
+We actively run tests on all these platforms during development and before making a new release.
+
+Note: the SDK is still compatible also with iOS8. We are not actively running test on that target, but we are not aware of any incompatility or bugs at the moment. We are planning to drop support to iOS8 in an upcoming release. 
 
 ## Documentation
 


paddybyers:
My suggestion on the call was:

We maintain compatibility and explicitly support these platform versions, including performing CI testing on all library revisions.

We do not explicitly maintain compatibility with older platform versions; we no longer perform CI testing on iOS8 as of version 1.0.11. Any known incompatibilities with older versions can be found at https://github.com/ably/ably-ios/issues?q=is%3Aissue+is%3Aopen+label%3A%22compatibility%22.

Please raise an issue or contact Ably customer support for advice if you find a problem with, or wish to have support for, unsupported platform versions.

README.md
```diff
@@ -1,7 +1,18 @@
...
+
+This SDK is compatible with projects that target:
+
+- iOS9
+- iOS10 
+- iOS11
+
+We actively run tests on all these platforms during development and before making a new release.
+
+Note: the SDK is still compatible also with iOS8. We are not actively running test on that target, but we are not aware of any incompatility or bugs at the moment. We are planning to drop support to iOS8 in an upcoming release.
```
mattheworiordan: `g/incompatility/incompatibility`

> "We are planning to drop support to iOS8 in an upcoming release. "

Not sure about that.  I think we should instead say something like "Note: This SDK should still be compatible with iOS version 8, however we are no longer supporting iOS version 8 and below as of [Date]. If you find any issues with iOS 8, please raise an issue in this repository and we may consider a fix, but cannot guarantee this. We recommend our customers drop support for iOS version 8 given less than [0.5% of users are still using 8.x as of Feb 2017](https://david-smith.org/iosversionstats/)"

funkyboy:
@paddybyers @mattheworiordan WDYT about latest version I pushed?

paddybyers:
> We recommend our customers drop support for iOS8 given that [less than 0.5% of users are still using it as of Feb 2017](https://david-smith.org/iosversionstats/).

Really I don't think it's our place to tell customers what versions they should support in their app.

My suggestion was:

> Please raise an issue or contact Ably customer support for advice if you find a problem with, or wish to have support for, unsupported platform versions.

I think that says what we want to say.

funkyboy:
@paddybyers better now?

paddybyers:
> better now?

LGTM

funkyboy:
@mattheworiordan one last check before merging?

PR #702, opened 2018-03-13
Fix memory consumption bug

### LINKED ISSUE #691: High Memory usage with my trackee channels
We are experiencing high memory usage currently with our application channels and we are not sure about the reason and sometimes it takes more than a GB which is not good for a mobile device . I have simulated this issue with our own channels in a separate project and it has been shared with @funkyboy as project contains our channels so we are not keeping it public on GitHub repository.
We suspect that https://github.com/ably/ably-ios/issues/673  could be because of this issue.
When memory goes high often we get the message "Count response is greater then the total of pending message."
Feel free to further communicate for this project. 
Following is the project url https://drive.google.com/drive/folders/1oYMa0wakNKXfSH7M8gAzPX_2EnShcdTH

message

screen shot 2018-02-15 at 3 19 23 am

Screen Shot 2018-02-15 at 1.56.50 AM

funkyboy:
Fix #691

### DIFF
--- a/Source/ARTLog.m
+++ b/Source/ARTLog.m
@@ -186,7 +186,7 @@ static const char *logLevelName(ARTLogLevel level) {
 }
 
 - (void)verbose:(NSString *)format, ... {
-    if (self.logLevel == ARTLogLevelVerbose) {
+    if (self.logLevel <= ARTLogLevelVerbose) {
         va_list args;
         va_start(args, format);
         [self log:[[NSString alloc] initWithFormat:format arguments:args] level:ARTLogLevelVerbose];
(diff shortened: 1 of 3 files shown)


Source/ARTLog.m
```diff
@@ -109,7 +109,7 @@ - (instancetype)init {
 }
 
 - (instancetype)initCapturingOutput:(BOOL)capturing {
-    return [self initCapturingOutput:true historyLines:100];
+    return [self initCapturingOutput:true historyLines:50];
```
ricardopereira: The history is sent to Sentry when an exception occurs. I think 50 is too low. 100 lines would give us more elements to diagnose an issue. Does this change improve significantly the performance?

ricardopereira:
Tests aren't passing:
https://travis-ci.org/ably/ably-ios/jobs/352945302#L4620-L4621

Source/ARTLog.m
funkyboy: Will measure and let you know.

ricardopereira:
@funkyboy Ok, thanks for checking!

PR #703, opened 2018-03-16
Update README file

funkyboy:


paddybyers:
Ironically this is a PR against `master`. I think `default` should become the default branch, and this PR should be against that. Also, if we're updating the README to reflect the new flow, we should have something that states how things are managed in this repo in addition to that 1 line in the contributing section.

funkyboy:
@paddybyers LOL. All of this just because I forgot to click 
screen shot 2018-03-19 at 10 59 02

paddybyers:
:)

PR #704, opened 2018-03-19
Update contributing instructions

funkyboy:


funkyboy:
@paddybyers better now?

PR #708, opened 2018-03-23
Push Channels

ricardopereira:


Spec/PushChannel.swift
```diff
@@ -0,0 +1,330 @@
...
+                            guard let error = error else {
+                                fail("Error is nil"); done(); return
+                            }
+                            expect(error.message).to(contain("null client ID"))
+                            done()
+                        }
+                    }
+                }
+
+                // RSH7a2
```
funkyboy: I think this is `RSH7b2`

Spec/PushChannel.swift
```diff
@@ -0,0 +1,330 @@
...
+
+                    expect(query).to(haveParam("deviceId", withValue: params["deviceId"]))
+                    expect(query).toNot(haveParam("clientId", withValue: rest.device.clientId))
+                    expect(query).to(haveParam("channel", withValue: params["channel"]))
+                    expect(query).to(haveParam("concatFilters", withValue: "true"))
+                }
+
+                it("should return a paginated result with PushChannelSubscription filtered by channel and client") {
+                    let params = [
+                        "clientId": "tester",
```
funkyboy: probably better to reuse `rest.options.clientId`

Spec/PushChannel.swift
```diff
@@ -0,0 +1,330 @@
...
+                it("should not accept both deviceId and clientId params at the same time") {
+                    let params = [
+                        "deviceId": "x",
+                        "clientId": "y"
+                    ]
+                    let channel = rest.channels.get("foo")
+                    expect { try channel.push.listSubscriptions(params) { _, _ in } }.to(throwError { (error: NSError) in
+                        expect(error.code).to(equal(ARTDataQueryError.invalidParameters.rawValue))
+                    })
+                }
```
funkyboy: Please add a test in which you actually subscribe and then you check that the result of `listSubscriptions` actually includes the result of the subscription.

Spec/PushChannel.swift
ricardopereira: Fixed 2dba0b0 👍

Source/ARTPushChannel.m
```diff
@@ -174,23 +182,24 @@ - (void)listSubscriptions:(NSDictionary<NSString *, NSString *> *)params callbac
...
+    return YES;
 }
 
-- (ARTDeviceDetails *)getDevice:(void(^_Nullable)(ARTErrorInfo *_Nullable))callback {
-    ARTDeviceDetails *device = [_channel device];
-    if (!device) {
+- (ARTLocalDevice *)getDevice:(void(^_Nullable)(ARTErrorInfo *_Nullable))callback {
+    ARTLocalDevice *device = [_channel device];
+    if (![device isRegistered]) {
         if (callback) callback([ARTErrorInfo createWithCode:0 message:@"cannot use device before ARTRest.push.activate has finished"]);
```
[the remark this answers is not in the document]
ricardopereira: Accepted acd4100 👍

Source/ARTPushChannel.m
```diff
@@ -146,26 +150,30 @@ - (void)unsubscribeClient:(void(^_Nullable)(ARTErrorInfo *_Nullable))callback {
     }];
 }
 
-- (void)listSubscriptions:(void(^)(ARTPaginatedResult<ARTPushChannelSubscription *> *_Nullable, ARTErrorInfo *_Nullable))callback {
-    [self listSubscriptions:nil callback:callback];
-}
+- (BOOL)listSubscriptions:(NSDictionary<NSString *, NSString *> *)params callback:(void(^)(ARTPaginatedResult<ARTPushChannelSubscription *> *_Nullable, ARTErrorInfo *_Nullable))callback error:(NSError *_Nullable *_Nullable)errorPtr {
```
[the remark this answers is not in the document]
ricardopereira: It has a `BOOL` because it's how ObjC handle errors. We should respect that for good Swift/ObjC interoperability. An ObjC method with an `NSError **` as last parameter and a BOOL return type will bridge to a Swift method that throw an error.

Spec/PushChannel.swift
ricardopereira: Done 720d602 but it doesn't pass. I already reported that issue in Slack.

ricardopereira:
`should return a paginated result with PushChannelSubscription` still fails with "_40100: push not enabled for this channel_"

ricardopereira:
Rebased with `push` branch.

funkyboy:
@ricardopereira when you pull down this branch to run it locally it throws a bunch of compilation errors
screen shot 2018-06-12 at 16 51 04

ricardopereira:
@funkyboy You're working with Xcode 9.3?

funkyboy:
@ricardopereira no, 9.2. Afair you said you didn’t upgrade any project settings.

ricardopereira:
@funkyboy I did not update the project settings, I'm only using Xcode 9.3 https://github.com/ably/ably-ios/pull/709#discussion_r186029445

funkyboy:
@ricardopereira Yup, the point is that it doesn’t compile on 9.2 :(
Anyway an upgrade is on the way, so this will soon be a non problem.

ricardopereira:
@funkyboy I think it's not a big deal. This won't affect users because it only affects tests and even so this is still in beta, from what I know.

Spec/PushChannel.swift
funkyboy: If you put a breakpoint on [this line](https://github.com/ably/ably-ios/commit/720d602766f712a8ace8bb4d7d75ac667fe4b2ea#diff-d254ef992ef12123539ed0708d7375d1R337) you'll get

```
(lldb) po channel.name
"test-2-pushenabled:foo"

(lldb) 
```

funkyboy: @ricardopereira quite sure it's due to this https://github.com/ably/ably-ios/blob/push-channels/Spec/TestUtilities.swift#L100

funkyboy:
@ricardopereira FYI I cherry-picked this commit https://github.com/ably/ably-ios/commit/1845218b8ea0f3aa3ca204fd6cdef499d0a8b159 to both `push` and `push-channels`. Now Travis kicks in.

funkyboy:
@ricardopereira also, there's a consistent failure on this line https://github.com/ably/ably-ios/blob/push-channels/Spec/PushActivationStateMachine.swift#L584

Likely due to the fact that the previous test doesn't have a 
`defer { rest.device.setAndPersistIdentityTokenDetails(nil) }`

funkyboy:
Rebased on `push` now.

Spec/PushChannel.swift
funkyboy: Fixed in https://github.com/ably/ably-ios/pull/708/commits/02b4d05740d7edf1340f051e1bc69853db8a79cb

PR #714, opened 2018-04-16
Add jwt tests

### LINKED ISSUE #713: Add test for JWT token

funkyboy:
Fix #713 

- [x] Remove old test
- [x] Add tests that use `authURL`
- [x] Add tests that use `ClientOptions`
- [x] Add tests that use `authCallback`
- [x] Add tests for automatic renewal
- [x] Add tests with JWT wrapping an Ably native token `x-ably-token`
- [x] Add tests with encrypted token
- [x] Add tests with JWT that includes TokenDetails in `x-ably-capability`
- [x] Add test for token returned with `application/jwt` content type
- [x] Add refs to spec items
- [x] Investigate RSA4b issue. See https://github.com/ably/ably-ios/pull/714#discussion_r192783902
- [x] Change URL of `auth_url` when echo server is deployed. See TODOs

### DIFF
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -3718,20 +3718,14 @@ class Auth : QuickSpec {
                     it ("receives a 40142 error from the server") {
                         waitUntil(timeout: testTimeout) { done in
                             client.connection.once(.connected) { stateChange in
-                                delay(tokenDuration + 1) {
+                                client.connection.once(.disconnected) { stateChange in
+                                    expect(stateChange!.reason!.code).to(equal(40142))
+                                    expect(stateChange!.reason!.description).to(contain("Key/token status changed (expire)"))
                                     done()
                                 }
                             }
                             client.connect()
                         }
-                        waitUntil(timeout: testTimeout) { done in
-                            client.connection.once(.disconnected) { stateChange in
-                                expect(stateChange!.reason!.code).to(equal(40142))
-                                expect(stateChange!.reason!.description).to(contain("Key/token status changed (expire)"))
-                                done()
-                            }
-                        }
-
                     }
                 }
                 
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -3849,5 +3849,38 @@ class Auth : QuickSpec {
                 }
             }
         }
+        
+        describe("JWT and rest") {
+            let options = AblyTests.clientOptions()
+            
+            context("when the JWT token embeds an Ably token") {
+                options.tokenDetails = ARTTokenDetails(token: getJWTToken(jwtType: "embedded")!)
+                let client = ARTRest(options: options)
+                
+                it ("pulls stats successfully") {
+                    waitUntil(timeout: testTimeout) { done in
+                        client.stats{ stats, error in
+                            expect(error).to(beNil())
+                            done()
+                        }
+                    }
+                }
+            }
+            
+            context("when the JWT token embeds an Ably token and it is requested as encrypted") {
+                options.tokenDetails = ARTTokenDetails(token: getJWTToken(jwtType: "embedded", encrypted: 1)!)
+                let client = ARTRest(options: options)
+                
+                it ("pulls stats successfully") {
+                    waitUntil(timeout: testTimeout) { done in
+                        client.stats{ stats, error in
+                            expect(error).to(beNil())
+                            done()
+                        }
+                    }
+                }
+            }
+            
+        }
     }
 }
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -3906,6 +3906,26 @@ class Auth : QuickSpec {
                 }
             }
             
+            context("when the JWT token is returned with application/jwt content tye") {
+                
+                it("the client successfully connects and pulls stats") {
+                    let options = AblyTests.clientOptions()
+                    let keys = getKeys()
+                    options.authUrl = NSURL(string: echoServerAddress)! as URL
+                    options.authParams = [URLQueryItem]() as [URLQueryItem]?
+                    options.authParams?.append(URLQueryItem(name: "keyName", value: keys["keyName"]) as URLQueryItem)
+                    options.authParams?.append(URLQueryItem(name: "keySecret", value: keys["keySecret"]) as URLQueryItem)
+                    options.authParams?.append(URLQueryItem(name: "returnType", value: "jwt") as URLQueryItem)
+                    let client = ARTRest(options: options)
+                    
+                    waitUntil(timeout: testTimeout) { done in
+                        client.stats{ stats, error in
+                            expect(error).to(beNil())
+                            done()
+                        }
+                    }
+                }
+            }
         }
     }
 }
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -3907,22 +3907,35 @@ class Auth : QuickSpec {
             }
             
             context("when the JWT token is returned with application/jwt content tye") {
+                let options = AblyTests.clientOptions()
+                let keys = getKeys()
+                options.authUrl = NSURL(string: echoServerAddress)! as URL
+                options.authParams = [URLQueryItem]() as [URLQueryItem]?
+                options.authParams?.append(URLQueryItem(name: "keyName", value: keys["keyName"]) as URLQueryItem)
+                options.authParams?.append(URLQueryItem(name: "keySecret", value: keys["keySecret"]) as URLQueryItem)
+                options.authParams?.append(URLQueryItem(name: "returnType", value: "jwt") as URLQueryItem)
+                let client = ARTRest(options: options)
                 
                 it("the client successfully connects and pulls stats") {
-                    let options = AblyTests.clientOptions()
-                    let keys = getKeys()
-                    options.authUrl = NSURL(string: echoServerAddress)! as URL
-                    options.authParams = [URLQueryItem]() as [URLQueryItem]?
-                    options.authParams?.append(URLQueryItem(name: "keyName", value: keys["keyName"]) as URLQueryItem)
-                    options.authParams?.append(URLQueryItem(name: "keySecret", value: keys["keySecret"]) as URLQueryItem)
-                    options.authParams?.append(URLQueryItem(name: "returnType", value: "jwt") as URLQueryItem)
-                    let client = ARTRest(options: options)
-                    
                     waitUntil(timeout: testTimeout) { done in
                         client.stats{ stats, error in
                             expect(error).to(beNil())
                             done()
                         }
+                    }
+                }
+                
+                it("the client can request a new token to initilize another client that connects and pulls stats") {
+                    waitUntil(timeout: testTimeout) { done in
+                        client.auth.requestToken(nil, with: nil, callback: { tokenDetails, error in
+                            let newClientOptions = AblyTests.clientOptions()
+                            newClientOptions.token = tokenDetails!.token
+                            let newClient = ARTRest(options: newClientOptions)
+                            newClient.stats{ stats, error in
+                                expect(error).to(beNil())
+                                done()
+                            }
+                        })
                     }
                 }
             }
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -3662,7 +3662,7 @@ class Auth : QuickSpec {
             }
 
             // RSA8g RSA8c
-            context("when using auth_url") {
+            context("when using authUrl") {
                 let options = AblyTests.clientOptions()
                 let keys = getKeys()
                 options.autoConnect = false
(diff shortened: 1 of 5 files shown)


funkyboy:
Rebased on master.

Spec/Auth.swift
```diff
@@ -3619,5 +3619,7 @@ class Auth : QuickSpec {
                 }
             }
         }
+            }
```
mattheworiordan: What is going on here with the two extra closing squigglies?

Spec/TestUtilities.swift
```diff
@@ -43,6 +43,7 @@ let appSetupJson = JSON(parseJSON: try! String(contentsOfFile: pathForTestResour
 
 let testTimeout: TimeInterval = 10.0
 let testResourcesPath = "ably-common/test-resources/"
+let echoServerAddress = "https://shrouded-plains-50367.herokuapp.com/createJWT" // TODO: change this
```
mattheworiordan: You can just deploy your changes to the Heroku echoserver, even if not fully merged, to avoid having to go back and do this.

Spec/TestUtilities.swift
```diff
@@ -494,6 +495,52 @@ func getTestTokenDetails(key: String? = nil, clientId: String? = nil, capability
     return tokenDetails
 }
 
+func encodeString(_ input: String) -> String? {
+    let allowedQueryParamAndKey =  CharacterSet(charactersIn: "{}[]*: ")
+   if let encodedString_ = input.addingPercentEncoding(withAllowedCharacters: allowedQueryParamAndKey) {
```
mattheworiordan: Indentation issues again

Spec/TestUtilities.swift
```diff
@@ -494,6 +495,52 @@ func getTestTokenDetails(key: String? = nil, clientId: String? = nil, capability
...
+    let options = AblyTests.commonAppSetup()
+    guard let components = options.key?.components(separatedBy: ":"), let keyName = components.first, var keySecret = components.last else {
+        fail("Invalid API key: \(options.key ?? "nil")")
+        return nil
+    }
+    if (invalid) {
+        keySecret = "invalid"
+    }
+
+    var URLCcomponents = URLComponents(string: echoServerAddress)
```
mattheworiordan: Typo I think, `URLCcomponents`?

Spec/Auth.swift
```diff
@@ -3619,5 +3619,331 @@ class Auth : QuickSpec {
...
+                let options = AblyTests.clientOptions()
+
+                context("with valid credentials") {
+                    options.token = getJWTToken()
+                    let client = AblyTests.newRealtime(options)
+                    defer { client.dispose(); client.close() }
+
+                    it("pulls stats successfully") {
+                        waitUntil(timeout: testTimeout) { done in
+                            client.stats{ stats, error in
```
mattheworiordan: Missing whitespace after `stats` I believe?

Spec/Auth.swift
```diff
@@ -3619,5 +3619,331 @@ class Auth : QuickSpec {
...
+                                done()
+                            }
+                            client.connect()
+                        }
+                    }
+                }
+            }
+
+            // RSA8g RSA8c
+            context("when using auth_url") {
```
mattheworiordan: Minor mostly unimportant comment, but this is `authUrl` in iOS, not `auth_url`

Spec/Auth.swift
```diff
@@ -3619,5 +3619,331 @@ class Auth : QuickSpec {
...
+                let options = AblyTests.clientOptions()
+                let keys = getKeys()
+                options.autoConnect = false
```
mattheworiordan: Why are you setting `autoConnect` to false?  You did something similar in Ruby, is this needed in iOS for some reason?

Spec/Auth.swift
```diff
@@ -3619,5 +3619,331 @@ class Auth : QuickSpec {
...
+                context("with valid credentials") {
+                    options.authParams = [URLQueryItem]() as [URLQueryItem]?
+                    options.authParams?.append(URLQueryItem(name: "keyName", value: keys["keyName"]) as URLQueryItem)
+                    options.authParams?.append(URLQueryItem(name: "keySecret", value: keys["keySecret"]) as URLQueryItem)
+                    let client = ARTRealtime(options: options)
+                    defer { client.dispose(); client.close() }
+
+                    it("fetches a channels and posts a message") {
+                        client.channels.get(channelName).publish(messageName, data: nil, callback: { error in
+                            expect(error).to(beNil())
```
mattheworiordan: In the other tests test are wrapped in a `waitUntil` with a call to `done()`.  How does the test suite know when this test is done?

Spec/Auth.swift
```diff
@@ -3619,5 +3619,331 @@ class Auth : QuickSpec {
...
+                    options.authParams?.append(URLQueryItem(name: "keyName", value: keys["keyName"]) as URLQueryItem)
+                    options.authParams?.append(URLQueryItem(name: "keySecret", value: keys["keySecret"]) as URLQueryItem)
+                    options.authParams?.append(URLQueryItem(name: "expiresIn", value: String(UInt(tokenDuration))) as URLQueryItem)
+                    let client = ARTRealtime(options: options)
+                    defer { client.dispose(); client.close() }
+
+                    it ("receives a 40142 error from the server") {
+                        waitUntil(timeout: testTimeout) { done in
+                            client.connection.once(.connected) { stateChange in
+                                delay(tokenDuration + 1) {
```
mattheworiordan: Why do you need this delay?  Should the connection `once(.disconnected)` fire below anyway once the token has expired?  In fact, if waits `tokenDuration`, what's probably happening with this test is that it's being disconnected once here, and once again below.

Spec/Auth.swift
```diff
@@ -3619,5 +3619,331 @@ class Auth : QuickSpec {
...
+                    options.authCallback = { tokenParams, completion in
+                        let token = ARTTokenDetails(token: getJWTToken(invalid: true)!)
+                        completion(token, nil)
+                    }
+                    let client = ARTRealtime(options: options)
+                    defer { client.dispose(); client.close() }
+
+                    it("fails to connect") {
+                        waitUntil(timeout: testTimeout) { done in
+                            client.connection.once(.failed) { stateChange in
```
mattheworiordan: I thought we had agreed that https://docs.ably.io/client-lib-development-guide/features/#RSA4b stated that the connection should not become failed, but instead should become disconnected and retry. I believe @paddybyers updated the spec after chatting with you about this at some point, although oddly I can't find a matching commit.

Spec/Auth.swift
```diff
@@ -3619,5 +3619,331 @@ class Auth : QuickSpec {
...
+                        client.connect()
+                    }
+                }
+            }
+            
+            context("when the token request includes a clientId") {
+                let clientId = "JWTClientId"
+                let options = AblyTests.clientOptions()
+                options.tokenDetails = ARTTokenDetails(token: getJWTToken(clientId: clientId)!)
+                options.autoConnect = false
```
mattheworiordan: More `autoConnect` false although not sure why.

Spec/Auth.swift
```diff
@@ -3619,5 +3619,331 @@ class Auth : QuickSpec {
...
+                        client.stats{ stats, error in
+                            expect(error).to(beNil())
+                            done()
+                        }
+                    }
+                }
+            }
+            
+            // RSA4f, RSA8c
+            context("when the JWT token is returned with application/jwt content tye") {
```
mattheworiordan: Small typo, `type` not `tye`

Spec/TestUtilities.swift
funkyboy: Just didn't want to disrupt existing stuff that currently depends on the echo server. Happy to change this when the echo server PR is merged.

mattheworiordan: But your changes aren’t breaking changes? You only added a new endpoint so I don’t see the risk personally

Spec/TestUtilities.swift
funkyboy: Actually it was a leftover :) https://github.com/ably/ably-ios/pull/714/commits/54c84dea35bb680fbd7785571044c5385f71a4bd

Spec/TestUtilities.swift
funkyboy: Good catch, thanks. https://github.com/ably/ably-ios/pull/714/commits/903bc3e41c45a82e5d3ed858dce095d6cc9ce011

Spec/Auth.swift
funkyboy: Yes. See https://github.com/ably/ably-ios/pull/714/commits/a2100f271b15127f75aa131472ec8831b737f650

Spec/Auth.swift
funkyboy: Done https://github.com/ably/ably-ios/pull/714/commits/7351d465fc5690dfad10464ad6f539223e3ecbb7

Spec/Auth.swift
funkyboy: No needed. Done https://github.com/ably/ably-ios/pull/714/commits/5ec07607a40ef592fab887fb9b8fa5d02ec0be9a

Spec/Auth.swift
funkyboy: Good observation. https://github.com/ably/ably-ios/pull/714/commits/c7efe08867f966f1a3771074d1c636c19a4343fa

Spec/Auth.swift
funkyboy: Done https://github.com/ably/ably-ios/pull/714/commits/936955b51ed438eb74ed0f0f5036e95f9cbc0a22

Spec/Auth.swift
funkyboy: Add this as a TODO above

Spec/Auth.swift
funkyboy: done https://github.com/ably/ably-ios/pull/714/commits/4b5ada45f280b71afac5baf90f1e2ec638ef34f4

Spec/Auth.swift
funkyboy: See https://github.com/ably/ably-ios/pull/714/commits/7351d465fc5690dfad10464ad6f539223e3ecbb7#diff-b5d84fe9bce26902a7481c2ffec84fffR3911

Spec/TestUtilities.swift
funkyboy: Only potential issue could be https://github.com/ably/echoserver/pull/4/commits/49a3a0f8af59a02a18f0c36451c958cfcf6224d0

mattheworiordan: @funkyboy why would that be an issue?  I appreciate it's an upgrade, but you have run a full test suite against the echoServer running the new Node without any issues?

funkyboy: No. Will do that with the Ruby SDK. If everything passes can I deploy then to echo.ably.io ?

Spec/Auth.swift
paddybyers: @mattheworiordan RSA4b says that a REST request fails if a token error is returned after a retry, which is correct, and is unrelated to this test.

https://docs.ably.io/client-lib-development-guide/features/#RTN14b relates to token failures on connection creation - it says that a connection should not fail in that case, but become disconnected. This test is therefore wrong and, because it passes, there is also a bug in the library (which will have existed before these JWT changes). @funkyboy pls raise a separate issue for that (and find the other existing, non-JWT, invalid test).

The discussion you remembered related was https://github.com/ably/docs/issues/429, relating to RTN15h, and that's to do with failures to reconnect (ie not initial connection failures). I'm guessing that there's going to be an issue for iOS there as well. I will do a docs PR for that.

funkyboy: @paddybyers @mattheworiordan The non-JWT corresponding test is https://github.com/ably/ably-ios/blob/develop/Spec/Auth.swift#L305 and again it tests for `FAILED`. 

Issue raised here: https://github.com/ably/ably-ios/issues/730

funkyboy: @paddybyers [This](https://github.com/ably/ably-ios/blob/develop/Spec/Auth.swift#L216) seems to test the new `RTN15h1` that you wrote. 
[These](https://github.com/ably/ably-ios/blob/develop/Spec/Auth.swift#L190-L332) all riff around the same "token expired/wrong" conditions but none seems to tests for the "single attempt to renew the token" clause.

Should I open a new issue about this?

cc @mattheworiordan

mattheworiordan: > @mattheworiordan RSA4b says that a REST request fails if a token error is returned after a retry, which is correct, and is unrelated to this test.

Oops, I should have scrolled up to see the context of the test I found.  That explains why I could not find the record of the change we made in the spec!

Spec/TestUtilities.swift
funkyboy: I ran the full test suite of Ruby against the echo server in staging. All green: https://travis-ci.org/ably/ably-ruby/builds/388329095

Will promote it to production if there are no objections.

mattheworiordan: 👍

Spec/Auth.swift
funkyboy: @paddybyers 
> ably/docs#439. @funkyboy can you check that behaviour please in this library and see if complies with the DISCONNECTED or FAILED behaviour?

Opened issue here #731

funkyboy:
@paddybyers @mattheworiordan @ricardopereira ready for another look

PR #719, opened 2018-04-26
Enforce new connection when last activity > than (idle interval + TTL)

funkyboy:
Enforces a new connection when the last activity is > than `(_maxIdleInterval + _connectionStateTtl)`

Mimics [this behavior in JS](https://github.com/ably/ably-js/blob/c55b4518d6f096e4bde90afcd04c5ba35d466f3c/common/lib/transport/connectionmanager.js#L702-L712), called when [a connection gets started](https://github.com/ably/ably-js/blob/master/common/lib/transport/connectionmanager.js#L1039).

cc @paddybyers

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -388,7 +388,6 @@ ART_TRY_OR_MOVE_TO_FAILED_START(self) {
                     [self.connection setId:nil];
                     [self.connection setKey:nil];
                     [self.connection setSerial:0];
-                    _transport = nil;
                 }
             }
             
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -227,7 +227,10 @@ ART_TRY_OR_MOVE_TO_FAILED_START(self) {
 }
 
 - (void)_connect {
-    if(self.connection.state_nosync == ARTRealtimeClosing) {
+    NSTimeInterval intervalSinceLast = [[NSDate date] timeIntervalSinceDate:_lastActivity];
+    
+    // We want to enforce a new connection also when there hasn't been activity for longer than (idle interval + TTL)
+    if(self.connection.state_nosync == ARTRealtimeClosing || intervalSinceLast > (_maxIdleInterval + _connectionStateTtl)) {
         // New connection
         _transport = nil;
     }
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -231,14 +231,6 @@ ART_TRY_OR_MOVE_TO_FAILED_START(self) {
         // New connection
         _transport = nil;
     }
-    
-    // We want to enforce a new connection also when there hasn't been activity for longer than (idle interval + TTL)
-    NSTimeInterval intervalSinceLast = [[NSDate date] timeIntervalSinceDate:_lastActivity];
-    if (intervalSinceLast > (_maxIdleInterval + _connectionStateTtl)) {
-        [self.connection setId:nil];
-        [self.connection setKey:nil];
-        [self.connection setSerial:0];
-    }
     [self transition:ARTRealtimeConnecting];
 }
 
@@ -388,6 +380,18 @@ ART_TRY_OR_MOVE_TO_FAILED_START(self) {
 
     switch (stateChange.current) {
         case ARTRealtimeConnecting: {
+            
+            // RTN15g We want to enforce a new connection also when there hasn't been activity for longer than (idle interval + TTL)
+            if (stateChange.previous == ARTRealtimeDisconnected || stateChange.previous == ARTRealtimeSuspended) {
+                NSTimeInterval intervalSinceLast = [[NSDate date] timeIntervalSinceDate:_lastActivity];
+                if (intervalSinceLast > (_maxIdleInterval + _connectionStateTtl)) {
+                    [self.connection setId:nil];
+                    [self.connection setKey:nil];
+                    [self.connection setSerial:0];
+                    _transport = nil;
+                }
+            }
+            
             stateChangeEventListener = [self unlessStateChangesBefore:[ARTDefault realtimeRequestTimeout] do:^{
                 [weakSelf onConnectionTimeOut];
             }];
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -227,12 +227,17 @@ ART_TRY_OR_MOVE_TO_FAILED_START(self) {
 }
 
 - (void)_connect {
-    NSTimeInterval intervalSinceLast = [[NSDate date] timeIntervalSinceDate:_lastActivity];
-    
-    // We want to enforce a new connection also when there hasn't been activity for longer than (idle interval + TTL)
-    if(self.connection.state_nosync == ARTRealtimeClosing || intervalSinceLast > (_maxIdleInterval + _connectionStateTtl)) {
+    if(self.connection.state_nosync == ARTRealtimeClosing) {
         // New connection
         _transport = nil;
+    }
+    
+    // We want to enforce a new connection also when there hasn't been activity for longer than (idle interval + TTL)
+    NSTimeInterval intervalSinceLast = [[NSDate date] timeIntervalSinceDate:_lastActivity];
+    if (intervalSinceLast > (_maxIdleInterval + _connectionStateTtl)) {
+        [self.connection setId:nil];
+        [self.connection setKey:nil];
+        [self.connection setSerial:0];
     }
     [self transition:ARTRealtimeConnecting];
 }
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -2712,8 +2712,7 @@ class RealtimeClientConnection: QuickSpec {
                                     client.connection.once(.connecting) { _ in
                                         client.connection.once(.connected) { _ in
                                             expect(client.connection.id).toNot(equal(connectionId))
-                                            let newChannel = client.channels.get(channelName)
-                                            newChannel.once(.attached) { _ in
+                                            channel.once(.attached) { _ in
                                                 done()
                                             }
                                         }
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -2653,12 +2653,12 @@ class RealtimeClientConnection: QuickSpec {
                 // RTN15g RTN15g1
                 context("when connection (ttl + idle interval) period has passed since last activity") {
                     let options = AblyTests.commonAppSetup()
+                    // We want this to be > than the sum of customTtlInterval and customIdleInterval
+                    options.disconnectedRetryTimeout = 5.0
                     var client: ARTRealtime!
                     var connectionId = ""
                     let customTtlInterval: TimeInterval = 1.0
                     let customIdleInterval: TimeInterval = 1.0
-                    // We want this to be > than the sum of the two above
-                    let reconnectAfterInterval: TimeInterval = 5.0
                     
                     it("uses a new connection") {
                         client = AblyTests.newRealtime(options)
@@ -2682,9 +2682,6 @@ class RealtimeClientConnection: QuickSpec {
                                             expect(client.connection.id).toNot(equal(connectionId))
                                             done()
                                         }
-                                    }
-                                    delay(reconnectAfterInterval) {
-                                        client.connect()
                                     }
                                 }
                                 client.onDisconnected()
@@ -2719,9 +2716,6 @@ class RealtimeClientConnection: QuickSpec {
                                             }
                                         }
                                     }
-                                    delay(reconnectAfterInterval) {
-                                        client.connect()
-                                    }
                                 }
                                 client.onDisconnected()
                             }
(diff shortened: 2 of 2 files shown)


Source/ARTRealtime.m
```diff
@@ -227,7 +227,10 @@ - (void)connect {
 }
 
 - (void)_connect {
-    if(self.connection.state_nosync == ARTRealtimeClosing) {
+    NSTimeInterval intervalSinceLast = [[NSDate date] timeIntervalSinceDate:_lastActivity];
+    
+    // We want to enforce a new connection also when there hasn't been activity for longer than (idle interval + TTL)
+    if(self.connection.state_nosync == ARTRealtimeClosing || intervalSinceLast > (_maxIdleInterval + _connectionStateTtl)) {
```
ricardopereira: Hmm, I think that could easily create too many connections. I'll check that in more detail tomorrow.

ricardopereira: Confirmed, tests are failing and from the tests... 

```
[15:26:38]: ▸ ✗ Connection__connection_request_fails__should_transition_to_Failed_state_because_the_token_is_invalid_and_not_renewable, expected to equal <15>, got <50>
[15:26:38]: ▸ ✗ Connection__connection_request_fails__should_transition_to_Failed_state_because_the_token_is_invalid_and_not_renewable, Should have only one connection request fail
```

this is not good.

funkyboy:
Took another stab https://github.com/ably/ably-ios/pull/719/commits/deb79a77d896ecead2bf222a2e033bb205626c70

paddybyers:
Have a look at the tests in https://github.com/ably/ably-ruby/pull/146 for the cases we need to test

funkyboy:
@paddybyers Will do

paddybyers:
> ricardopereira approved these changes 32 minutes ago

I do think we should have test cases included before approving the PR.

mattheworiordan:
@funkyboy I think we should get into the habit with spec changes of writing tests that fail, then making them pass.  This approach IMHO is in most cases wrong.  If you look back at how we worked with @ricardopereira in the past, he wrote tests that failed but met the spec, then Toni wrote the code.  I am not sure anyone can review this code in any meaningful way without tests.

funkyboy:
@mattheworiordan will do. Won't merge this without having tests, which I am working on.

funkyboy:
@mattheworiordan you can read 

```
delay(reconnectAfterInterval) {
  client.connect()
}
```

as the counterpart of the following Ruby

```
sleep(reconnectAfterInterval)
client.connect
```

mattheworiordan:
> you can read
> delay(reconnectAfterInterval) {
>  client.connect()
> }
> as the counterpart of the following Ruby
> sleep(reconnectAfterInterval)
> client.connect

Ok, so then if the client is running in it's own thread, whilst the test is sleeping for X time, what's stopping it from reconnecting itself automatically?

funkyboy:
It puts to sleep the whole test context, on the main thread: https://github.com/ably/ably-ios/blob/develop/Spec/TestUtilities.swift#L498. The client also belongs to that thread.

mattheworiordan:
> Ready for another look.

@funkyboy I don't think that is sufficient.  See https://github.com/ably/ably-ruby/pull/146/files#diff-6994e351b062efe17b0b5c88ac75991f for comparison.

I test both positive & negative conditions, as well as the channel attach requirements in the specification.  I don't believe this is comprehensive enough.

funkyboy:
@mattheworiordan @paddybyers @ricardopereira ready for another look

Spec/RealtimeClientConnection.swift
```diff
@@ -2691,6 +2691,49 @@ class RealtimeClientConnection: QuickSpec {
...
+                    
+                    it("uses the same connection") {
+                        client = AblyTests.newRealtime(options)
+                        client.connect()
+                        
+                        waitUntil(timeout: testTimeout) { done in
+                            client.connection.once(.connected) { _ in
+                                expect(client.connection.id).toNot(beNil())
+                                connectionId = client.connection.id!
+                                client.connectionStateTtl = customTtlInterval
```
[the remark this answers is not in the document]
funkyboy: For our customers `connectionStateTtl` is a private readonly property.

We have a bunch of methods to expose and set private variables for whitebox testing.
`connectionStateTtl` is one of those https://github.com/ably/ably-ios/blob/develop/Source/ARTRealtime%2BPrivate.h#L51

Spec/RealtimeClientConnection.swift
```diff
@@ -2690,6 +2690,42 @@ class RealtimeClientConnection: QuickSpec {
...
+                                    if let error = error {
+                                        fail(error.message)
+                                    }
+                                    expect(channel.state).to(equal(ARTRealtimeChannelState.attached))
+                                }
+                                client.connection.once(.disconnected) { _ in
+                                    client.connection.once(.connecting) { _ in
+                                        client.connection.once(.connected) { _ in
+                                            expect(client.connection.id).toNot(equal(connectionId))
+                                            let newChannel = client.channels.get(channelName)
```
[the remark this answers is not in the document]
funkyboy: @mattheworiordan done https://github.com/ably/ably-ios/pull/719/commits/d1c7a3c048cb99e6c63301b9cb9a1ddec1963482

Spec/RealtimeClientConnection.swift
```diff
@@ -2690,6 +2690,42 @@ class RealtimeClientConnection: QuickSpec {
                             }
                         }
                     }
+                    
+                    it("reattaches to the same channels after a new connection has been established") {
```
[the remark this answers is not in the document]
funkyboy: @mattheworiordan done https://github.com/ably/ably-ios/pull/719/commits/3bb7493e01280f58e6abeda58d7885b93126b87a

Spec/RealtimeClientConnection.swift
```diff
@@ -2649,6 +2649,91 @@ class RealtimeClientConnection: QuickSpec {
...
+                    it("uses a new connection") {
+                        let options = AblyTests.commonAppSetup()
+                        var client: ARTRealtime!
+                        var connectionId = ""
+                        let customTtlInterval: TimeInterval = 1.0
+                        let customIdleInterval: TimeInterval = 1.0
+                        // We want this to be > than the sum of the two above
+                        let reconnectAfterInterval: TimeInterval = 5.0
+                        client = AblyTests.newRealtime(options)
+                        client.connect()
```
ricardopereira: @funkyboy I usually add a `defer` closing the current connection. The connection will remain active. Something like `defer { client.close() }`.

funkyboy: @ricardopereira done https://github.com/ably/ably-ios/pull/719/commits/8565908636c1c3b7c123abbe7520870eb5661c15

funkyboy:
@mattheworiordan If you don't have any further observations, I'd merge this.

funkyboy:
After talking to @paddybyers and after working on the [Java version of this feature](https://github.com/ably/ably-java/pull/390) I realized that these tests are wrong, in that they should wait for `suspended` state (after disconnecting) and then perform checks. Here's a new todo list:

- [ ] Change tests to follow this sequence:
  1. connect
  2. extract data / mock values
  3. disconnect
  4. wait for the connection to go into `suspended` mode
  5. reconnect and wait for the connection to reconnect
  6. verify assertions
- [ ] Change implementation accordingly

funkyboy:
After another discussion about RTN15g it turns out that this (fresh connection after connectionTtl + idle timer has passed) doesn't _need_ to go through the suspended state. 

I tweaked the tests so that the reconnection happens via the state machine instead of calling `connect()`.
That surfaced a bug because the freshness check was performed only when `connect()` was called directly.

mattheworiordan:
Ok, logic seems good

Spec/RealtimeClientConnection.swift
```diff
@@ -2649,6 +2649,118 @@ class RealtimeClientConnection: QuickSpec {
...
+                                connectionId = client.connection.id!
+                                client.connectionStateTtl = customTtlInterval
+                                client.maxIdleInterval = customIdleInterval
+                                client.connection.once(.disconnected) { _ in
+                                    let disconnectedAt = Date()
+                                    expect(client.connectionStateTtl).to(equal(customTtlInterval))
+                                    expect(client.maxIdleInterval).to(equal(customIdleInterval))
+                                    client.connection.once(.connecting) { _ in
+                                        let reconnectionInterval = Date().timeIntervalSince(disconnectedAt)
+                                        expect(reconnectionInterval).to(beGreaterThan(client.connectionStateTtl + client.maxIdleInterval))
```
mattheworiordan: Why is this the case?  I don't see any explicit request to pause / delay the reconnecting.  What has caused this delay then?

Spec/RealtimeClientConnection.swift
```diff
@@ -2649,6 +2649,118 @@ class RealtimeClientConnection: QuickSpec {
...
+                                    client.connection.once(.connecting) { _ in
+                                        client.connection.once(.connected) { _ in
+                                            expect(client.connection.id).toNot(equal(connectionId))
+                                            channel.once(.attached) { _ in
+                                                done()
+                                            }
+                                        }
+                                    }
+                                }
+                                client.onDisconnected()
```
mattheworiordan: Why is this done before the attach is completed?  The sequence of events for this test is wrong i.e. the initial `attach` may only happen once the connection becomes connected again, in fact that is very likely.

Spec/RealtimeClientConnection.swift
```diff
@@ -2649,6 +2649,118 @@ class RealtimeClientConnection: QuickSpec {
...
+                                channel.attach { error in
+                                    if let error = error {
+                                        fail(error.message)
+                                    }
+                                    expect(channel.state).to(equal(ARTRealtimeChannelState.attached))
+                                }
+                                client.connection.once(.disconnected) { _ in
+                                    client.connection.once(.connecting) { _ in
+                                        client.connection.once(.connected) { _ in
+                                            expect(client.connection.id).toNot(equal(connectionId))
```
mattheworiordan: Why is the connection ID changing?  There is no pause as far as I can see.

Spec/RealtimeClientConnection.swift
```diff
@@ -2649,6 +2649,118 @@ class RealtimeClientConnection: QuickSpec {
...
+                                            channel.once(.attached) { _ in
```
mattheworiordan: You are not checking that the attached ChannelStateChange has a flag indicating loss of continuity on the channel.

Spec/RealtimeClientConnection.swift
```diff
@@ -2649,6 +2649,118 @@ class RealtimeClientConnection: QuickSpec {
...
+                    }
+                }
+                
+                // RTN15g2
+                context("when connection (ttl + idle interval) period has NOT passed since last activity") {
+                    let options = AblyTests.commonAppSetup()
+                    options.disconnectedRetryTimeout = 5.0
+                    var client: ARTRealtime!
+                    var connectionId = ""
+                    let customTtlInterval: TimeInterval = 20.0
```
mattheworiordan: Why even set this, surely a standard connection should reconnect automatically without any custom settings?

Spec/RealtimeClientConnection.swift
funkyboy: @mattheworiordan It was this option, which tries the reconnection after 5s the client has been in a `DISCONNECTED` state https://github.com/ably/ably-ios/blob/develop/Spec/RealtimeClientConnection.swift#L2657

Spec/RealtimeClientConnection.swift
funkyboy: @mattheworiordan I don't understand. Would you prefer to see this within the `attach` block and right after this assertion? https://github.com/ably/ably-ios/blob/develop/Spec/RealtimeClientConnection.swift#L2708

Spec/RealtimeClientConnection.swift
funkyboy: This `it` block uses the same settings declared up here https://github.com/ably/ably-ios/blob/develop/Spec/RealtimeClientConnection.swift#L2655

Spec/RealtimeClientConnection.swift
funkyboy: @mattheworiordan you mean I should check if `resumed` is true?

Spec/RealtimeClientConnection.swift
funkyboy: @mattheworiordan good spot. This is not needed.

Spec/RealtimeClientConnection.swift
mattheworiordan: But this is not according to the spec.  The client will retry to reconnect immediately after disconnect, see https://docs.ably.io/client-lib-development-guide/features/#RTN15a, not wait `disconnectedRetryTimeout`.  It only waits `disconnectedRetryTimeout` on subsequent requests.

Spec/RealtimeClientConnection.swift
mattheworiordan: As above, there should be no pause according to https://docs.ably.io/client-lib-development-guide/features/#RTN15a

Spec/RealtimeClientConnection.swift
mattheworiordan: Yes exactly https://docs.ably.io/client-lib-development-guide/features/#TH4

Spec/RealtimeClientConnection.swift
mattheworiordan: Yes, exactly, otherwise how do we know that the channel attached first and thus gets reattached?

Spec/RealtimeClientConnection.swift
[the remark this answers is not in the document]
mattheworiordan: Ok, thanks for clarifying

funkyboy:
@mattheworiordan Made [an issue](https://github.com/ably/ably-ios/issues/725) with your feedback on this PR after merge.

PR #726, opened 2018-05-16
Improve RTN15G tests

### LINKED ISSUE #725: Improve RTN15G tests
- [x] Improve usage of reconnection delay See https://github.com/ably/ably-ios/pull/719#discussion_r188201480
- [x] Disconnect after attaching See https://github.com/ably/ably-ios/pull/719#discussion_r188201665
- [x] Add check for `resume` See https://github.com/ably/ably-ios/pull/719#discussion_r188202070
- [x] Use standard connection values when possible See https://github.com/ably/ably-ios/pull/719#discussion_r188202285

funkyboy:
Fix #725

### DIFF
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -2711,7 +2711,8 @@ class RealtimeClientConnection: QuickSpec {
                                     client.connection.once(.connecting) { _ in
                                         client.connection.once(.connected) { _ in
                                             expect(client.connection.id).toNot(equal(connectionId))
-                                            channel.once(.attached) { _ in
+                                            channel.once(.attached) { stateChange in
+                                                expect(stateChange?.resumed).to(beFalse())
                                                 done()
                                             }
                                         }
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -2657,8 +2657,8 @@ class RealtimeClientConnection: QuickSpec {
                     options.disconnectedRetryTimeout = 5.0
                     var client: ARTRealtime!
                     var connectionId = ""
-                    let customTtlInterval: TimeInterval = 1.0
-                    let customIdleInterval: TimeInterval = 1.0
+                    let customTtlInterval: TimeInterval = 0.1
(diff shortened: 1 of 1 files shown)


mattheworiordan:
@funkyboy what about https://github.com/ably/ably-ios/pull/719#discussion_r188307367?  Do you have any thoughts on why the client is not automatically reconnecting, or am I reading the test incorrectly?

funkyboy:
@mattheworiordan [Looking at this](https://github.com/ably/ably-ios/blob/67d1d943ceba3181a035df8e599fb9b10a5fd618/Source/ARTRealtime.m#L489) I am quite sure it only uses the timer and does not honor the "will immediately attempt to reconnect and restore the connection state" part of [RTN15a](https://docs.ably.io/client-lib-development-guide/features/#RTN15a)

Will check with @ricardopereira.

mattheworiordan:
Yes, looking at https://github.com/ably/ably-ios/blob/ec182697eb4842f67eb28a30bf80eb038f8545ad/Spec/RealtimeClientConnection.swift#L2212-L2263, which is supposedly testing `RTN15a`, yet it's not, there is no test for this.

funkyboy:
filed an issue about RTN15a here https://github.com/ably/ably-ios/issues/727 
Can we merge this?

@mattheworiordan @paddybyers

mattheworiordan:
@funkyboy yes I think so 👍

Spec/RealtimeClientConnection.swift
```diff
@@ -2706,18 +2706,19 @@ class RealtimeClientConnection: QuickSpec {
...
+                                            channel.once(.attached) { stateChange in
+                                                expect(stateChange?.resumed).to(beFalse())
                                                 done()
                                             }
                                         }
                                     }
                                 }
-                                client.onDisconnected()
                             }
                         }
```
ricardopereira: I think the `client.connect()` should be called in the `waitUntil` closure. Just to be sure that all observers are set.

funkyboy: @ricardopereira good point :) See https://github.com/ably/ably-ios/pull/726/commits/f0e2314b37160152a6a1af690fbc3167b0970831

PR #742, opened 2018-07-24
Add cd2c

funkyboy:


Spec/RealtimeClientConnection.swift
```diff
@@ -23,6 +23,35 @@ class RealtimeClientConnection: QuickSpec {
...
+                        client.dispose()
+                        client.close()
+                    }
+                    ARTDefault.setMaxMessageSize(1)
+                    
+                    waitUntil(timeout: testTimeout) { done in
+                        client.connection.once(.connected) { _ in
+                            let transport = client.transport as! TestProxyTransport
+                            let firstConnectionDetails = transport.protocolMessagesReceived.filter{ $0.action == .connected }[0].connectionDetails
+                            expect(firstConnectionDetails!.maxMessageSize).to(equal(16384)) // Sandbox apps have a 16384 limit
```
ricardopereira: This is currently failing, right? Because I don't see any change where the connection is sending the current _MaxMessageSize_.

funkyboy: No that's passing. And once [this](https://github.com/ably/ably-common/pull/29/files#diff-35f9a23050dc9d9017554ed78cb4e77eR6) is merged, we can be sure of that value.

funkyboy: We are already loading that value https://github.com/ably/ably-ios/blob/develop/Source/ARTConnectionDetails.m#L24 It's just not exposed yet in realtime.

PR #745, opened 2018-07-24
Use latest version of fastlane on Travis

funkyboy:
Maybe it's gonna solve those sporadic cases where all the tests pass but we get an `Exit status: 65`.


funkyboy:
Ironically it happens also in this build :( https://travis-ci.org/ably/ably-ios/jobs/407661006#L5363

PR #747, opened 2018-07-24
Add tp3a test

funkyboy:
`(TP3a) id string – unique ID for this presence message. This attribute is always populated for presence messages received over REST. For presence messages received over Realtime, if the presence message does not contain an id, it should be set to protocolMsgId:index, where protocolMsgId is the id of the ProtocolMessage encapsulating it, and index is the index of the presence message inside the presence array of the ProtocolMessage`


ricardopereira:
It's a bit confusing to see changes that has nothing to do with the current issue. Will you rebase this with the base branch?

funkyboy:
@ricardopereira Yes, I'll rebase. Sorry about that.

PR #752, opened 2018-07-27
Add rtn15h2

funkyboy:
`(RTN15h2) If the DISCONNECTED message contains a token error (statusCode value of 401 and error code value in the range 40140 <= code < 40150) and the library has the means to renew the token, a single attempt to create a new token should be made and a new connection attempt initiated using the new token. If the token creation fails or the next connection attempt fails due to a token error, the connection will transition to the DISCONNECTED state and the Connection#errorReason will be set.`

While working on this I got a bunch of failures related to presence. Addressed here: https://github.com/ably/ably-ios/pull/751

### DIFF
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -2842,7 +2842,8 @@ class RealtimeClientConnection: QuickSpec {
                         }                        
                     }
 
-                    it("should transition to Failed when the token renewal fails and the error should be emitted") {
+                    // RTN15h2
+                    it("should transition to disconnected when the token renewal fails and the error should be emitted") {
                         let options = AblyTests.commonAppSetup()
                         options.disconnectedRetryTimeout = 1.0
                         options.autoConnect = false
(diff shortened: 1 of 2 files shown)


Spec/RealtimeClientConnection.swift
```diff
@@ -2878,8 +2878,8 @@ class RealtimeClientConnection: QuickSpec {
                         }
 
                         waitUntil(timeout: testTimeout) { done in
-                            // Renewal will fail
-                            client.connection.once(.failed) { stateChange in
+                            // Renewal will lead to a disconnect (again)
+                            client.connection.once(.disconnected) { stateChange in
```
ricardopereira: Should this test wait and see if a new token comes?

funkyboy: Well, a new valid token will never come in, given the way this test is designed. Or am I misunderstanding your question?

ricardopereira: Sorry, I misunderstood the test.
It looks good.

PR #754, opened 2018-07-27
RSC19a

ricardopereira:
`(RSC19a) Method signature is @request(string method, string path, Dict<String, String> params?, JsonObject | JsonArray body?, Dict<String, String> headers?) -> HttpPaginatedResponse@ with arguments: @method@ is a valid "HTTP verb":https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html (must support @"GET"@, @"POST"@, and @"PUT"@, should support @"PATCH"@, may support others); @path@ is the path component of the URL such as @"/channels"@; @params@ and @headers@ are optional arguments containing pairs of key value strings (multi-valued headers are not supported) that will result in query params and HTTP headers being added respectively in the request (the argument types can be idiomatic for the language such as @Object@ in the case of Javascript); @body@ is an optional @JsonObject@ or @JsonArray@ like object argument that can be easily serialized to MsgPack or JSON`

### DIFF
--- a/Source/ARTProtocolMessage+Private.h
+++ b/Source/ARTProtocolMessage+Private.h
@@ -8,9 +8,15 @@
 
 /// ARTProtocolMessageFlag bitmask
 typedef NS_OPTIONS(NSUInteger, ARTProtocolMessageFlag) {
-    ARTProtocolMessageFlagHasPresence = (1UL << 0), //1
-    ARTProtocolMessageFlagHasBacklog = (1UL << 1), //2
-    ARTProtocolMessageFlagResumed = (1UL << 2) //4
+    ARTProtocolMessageFlagHasPresence = (1UL << 0),
+    ARTProtocolMessageFlagHasBacklog = (1UL << 1),
+    ARTProtocolMessageFlagResumed = (1UL << 2),
+    ARTProtocolMessageFlagHasLocalPresence = (1UL << 3),
+    ARTProtocolMessageFlagTransient = (1UL << 4),
+    ARTProtocolMessageFlagPresence = (1UL << 16),
+    ARTProtocolMessageFlagPublish = (1UL << 17),
+    ARTProtocolMessageFlagSubscribe = (1UL << 18),
+    ARTProtocolMessageFlagPresenceSubscribe = (1UL << 19)
 };
 
 NS_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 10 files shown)


funkyboy:
@ricardopereira the essence is there 👍. 
I'd enrich it by testing for headers, statusCodes and such. You can take some inspiration from this [.NET PR](https://github.com/ably/ably-dotnet/pull/238/files#diff-8623e76647b61ac8a886d9b822983920R61)

ricardopereira:
Rebased with `develop`.

Spec/PushActivationStateMachine.swift
```diff
@@ -293,7 +293,14 @@ class PushActivationStateMachine : QuickSpec {
                         guard let rawBody = request.httpBody else {
                             fail("should have a body"); return
                         }
-                        guard let body = stateMachine.rest.defaultEncoder.decode(rawBody, error: nil) as? NSDictionary else {
+                        let decodedBody: Any
```
ricardopereira: This started failing because I changed the `- (id)decode:(NSData *)data error:(NSError *_Nullable *_Nullable)error;` header to the right form for Swift interoperability (`- (nullable id)decode:(NSData *)data error:(NSError *_Nullable *_Nullable)error;`). The `error` argument should be translated to a a `@throw` (Swift Error Handling).

ricardopereira:
@funkyboy about the `Accept` header 6cd9230

@paddybyers please review 2e211bf, 9a23fd0 and cc762ee.

ricardopereira:
@paddybyers The iOS lib doesn't have `HttpPaginatedResponse.statusCode` & `HttpPaginatedResponse.errorCode` https://github.com/ably/ably-ios/blob/develop/Source/ARTPaginatedResult.h#L16-L27. Is that a recent change?

paddybyers:
> The iOS lib doesn't have HttpPaginatedResponse.statusCode & HttpPaginatedResponse.errorCode https://github.com/ably/ably-ios/blob/develop/Source/ARTPaginatedResult.h#L16-L27. Is that a recent change?

No. I think you're confusing `PaginatedResult` (https://docs.ably.io/client-lib-development-guide/features/#TG1) which is used for the existing paginated queries such as stats and history; and `HttpPaginatedResponse` (https://docs.ably.io/client-lib-development-guide/features/#HP1) which is a different type, and exists only for RTN19. This API is a generic HTTP API which can handle arbitrary requests and responses, and doesn't treat error responses in the same way as the normal APIs.

ricardopereira:
Oh, I see. Then this implementation is wrong. Thanks. I'll change it.

PR #757, opened 2018-08-02
Fix failing tests

funkyboy:
This: 

- fixes the [implementation of RTN15h2](https://github.com/ably/ably-ios/commit/9f984a2e1063c9f0489e61ead13a7cb2d11fef01)
- fixed test regressions related to the addition of RTN15h2
- [fixes](https://github.com/ably/ably-ios/commit/4ff2639bf9434b021cc1a8467d6b4fc2ed272396) a bug introduced in a [previous PR](https://github.com/ably/ably-ios/pull/743) about the values passed in the `Accept` header

### DIFF
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -3691,7 +3691,7 @@ class Auth : QuickSpec {
 
                     it("fails to connect with reason 'invalid signature'") {
                         waitUntil(timeout: testTimeout) { done in
-                            client.connection.once(.failed) { stateChange in
+                            client.connection.once(.disconnected) { stateChange in
                                 expect(stateChange!.reason!.code).to(equal(40144))
                                 expect(stateChange!.reason!.description).to(contain("invalid signature"))
                                 done()
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -213,7 +213,7 @@ class Auth : QuickSpec {
                 }
 
                 // RSA4a
-                it("should transition the connection to the disconnected state when the server responds with a token error and there is no way to renew the token") {
+                it("should transition the connection to the FAILED state when the server responds with a token error and there is no way to renew the token") {
                     let options = AblyTests.clientOptions()
                     options.tokenDetails = getTestTokenDetails(ttl: 0.1)
                     options.autoConnect = false
@@ -234,10 +234,11 @@ class Auth : QuickSpec {
                     let channel = realtime.channels.get("test")
 
                     waitUntil(timeout: testTimeout) { done in
-                        realtime.connection.on(.disconnected) {stateChange in
+                        realtime.connection.on(.failed) {stateChange in
                             guard let error = stateChange?.reason else {
                                 fail("Error is nil"); done(); return
                             }
+                            expect(error.statusCode).to(equal(401))
                             expect(error.code).to(equal(40142))
                             done()
                         }
@@ -3690,7 +3691,7 @@ class Auth : QuickSpec {
 
                     it("fails to connect with reason 'invalid signature'") {
                         waitUntil(timeout: testTimeout) { done in
-                            client.connection.once(.disconnected) { stateChange in
+                            client.connection.once(.failed) { stateChange in
                                 expect(stateChange!.reason!.code).to(equal(40144))
                                 expect(stateChange!.reason!.description).to(contain("invalid signature"))
                                 done()
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -213,10 +213,11 @@ class Auth : QuickSpec {
                 }
 
                 // RSA4a
-                it("should transition the connection to the FAILED state when the server responds with a token error and there is no way to renew the token") {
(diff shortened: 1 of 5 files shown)


Spec/RealtimeClientConnection.swift
```diff
@@ -4247,6 +4247,7 @@ class RealtimeClientConnection: QuickSpec {
                 // https://github.com/ably/wiki/issues/22
                 it("should encode and decode fixture messages as expected") {
                     let options = AblyTests.commonAppSetup()
+                    options.useBinaryProtocol = false
```
ricardopereira: This started failing because of 4ff2639, right?

funkyboy: Correct, same for https://github.com/ably/ably-ios/pull/757/files#diff-ebbc11fa8ab1c043fe6ed321bbbe72bbR999

Spec/Auth.swift
```diff
@@ -213,10 +213,11 @@ class Auth : QuickSpec {
                 }
 
                 // RSA4a
-                it("should transition the connection to the FAILED state when the server responds with a token error and there is no way to renew the token") {
+                it("should transition the connection to the disconnected state when the server responds with a token error and there is no way to renew the token") {
                     let options = AblyTests.clientOptions()
                     options.tokenDetails = getTestTokenDetails(ttl: 0.1)
                     options.autoConnect = false
+                    options.logLevel = .verbose
```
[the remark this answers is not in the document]
funkyboy: Ooops :) https://github.com/ably/ably-ios/pull/757/commits/a2d09e06184562ea4ef89d708a3a9dccf9400432

Spec/Auth.swift
```diff
@@ -213,10 +213,11 @@ class Auth : QuickSpec {
                 }
 
                 // RSA4a
-                it("should transition the connection to the FAILED state when the server responds with a token error and there is no way to renew the token") {
+                it("should transition the connection to the disconnected state when the server responds with a token error and there is no way to renew the token") {
```
ricardopereira: This spec still says that it should transition to Failed state:

> (RSA4a) When a token or tokenDetails is used to instance the library, and no means to renew the token is provided (either an API key, authCallback or authUrl), if the server responds with a token error (401 HTTP status code and an Ably error value 40140 <= code < 40150), then the client library should indicate an error, not retry the request and in the case of the realtime library, transition the connection to the FAILED state

Is the spreadsheet of 1.1 outdated?

paddybyers: The spec is correct, and in the RSA4a case you have to transition to `FAILED` because you have no option - there is no means to renew the token, and the token you have doesn't work. The same is true in RTN15h1.

RTN15h2 describes the case where you do have the ability to renew the token, and in this case you go to `DISCONNECTED`. Remember the `DISCONNECTED` state means "I'm currently not connected, but I think I can be connected imminently".

We think this is the definitive text: https://github.com/ably/docs/blob/integration-spec-1-1/content/client-lib-development-guide/features.textile

If the spreadsheet disagrees, then the spreadsheet is wrong.

funkyboy:
See https://github.com/ably/ably-ios/commit/abba1492e8a929221d55550471a6b80554f08d06

Spec/RealtimeClientConnection.swift
```diff
@@ -1962,7 +1962,7 @@ class RealtimeClientConnection: QuickSpec {
                     }
                 }
 
-                it("should transition to Failed when the token renewal fails") {
+                it("should transition to disconnected when the token renewal fails") {
```
ricardopereira: Maybe we should add a test for the Failed case because the spec says:

> **RTN14b** ... connection will transition to the DISCONNECTED state, and the Connection#errorReason should be set. **(If no means to renew the token is provided, RSA4a applies)**

Spec/Auth.swift
```diff
@@ -234,15 +234,15 @@ class Auth : QuickSpec {
                     let channel = realtime.channels.get("test")
 
                     waitUntil(timeout: testTimeout) { done in
-                        realtime.connect()
-                        channel.publish("message", data: nil) { error in
-                            guard let error = error else {
+                        realtime.connection.on(.failed) {stateChange in
```
ricardopereira: This test doesn't pass like it was before? Just to be sure that we aren't changing tests that were validated (working the right way) and are important to give us confidence.

funkyboy: done 
https://github.com/ably/ably-ios/pull/757/commits/b64024f8a23ed5524fc171cb4f63101f8fcb9f3c

Spec/RealtimeClientConnection.swift
funkyboy: @ricardopereira Is this what you mean? https://github.com/ably/ably-ios/blob/develop/Spec/RealtimeClientConnection.swift#L2008

ricardopereira: Yes, that's it 👍

PR #759, opened 2018-08-08
Add max message size

funkyboy:
This addresses spec `TO3l8*` and `RSL1i` items, aka throw an error before publishing messages when the size exceeds `maxMessageSize`.

NOTE: queued messages exceeding `maxMessageSize` (`RTL6d1`) will be addressed in a separate PR

### DIFF
--- a/Source/ARTBaseMessage.h
+++ b/Source/ARTBaseMessage.h
@@ -36,6 +36,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSString *)description;
 
+- (NSInteger)messageSize;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Source/ARTBaseMessage.h
+++ b/Source/ARTBaseMessage.h
@@ -7,6 +7,7 @@
 //
 
 #import <Foundation/Foundation.h>
+#import <Ably/ARTTypes.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -28,6 +29,10 @@ NS_ASSUME_NONNULL_BEGIN
 @property (strong, nonatomic, nullable) NSString *encoding;
 
 @property (strong, nonatomic, nullable) id data;
+
+/// The event name, if available
+@property (nullable, readwrite, strong, nonatomic) NSString *name;
+@property (nullable, nonatomic) id<ARTJsonCompatible> extras;
 
 - (NSString *)description;
 
--- a/Source/ARTChannel.m
+++ b/Source/ARTChannel.m
@@ -15,6 +15,7 @@
 #import "ARTBaseMessage+Private.h"
 #import "ARTDataQuery.h"
 #import "ARTRest+Private.h"
+#import "ARTDefault.h"
 
 @implementation ARTChannel {
     __weak ARTRest *_rest;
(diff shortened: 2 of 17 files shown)


Source/ARTChannel.m
```diff
@@ -73,6 +74,17 @@ - (void)publish:(art_nullable NSString *)name data:(art_nullable id)data extras:
         if (callback) callback([ARTErrorInfo createFromNSError:error]);
         return;
     }
+    
+    // Checked after encoding, so that the client can receive callback with encoding errors
+    if ([self exceedMaxSize:@[message]]) {
+        ARTErrorInfo *sizeError = [ARTErrorInfo createWithCode:40009
```
ricardopereira: Is this a common error between all libs?

ricardopereira: I think we should add those error codes in an enum or at least have it in a global&static var, even though, I know there are a bunch of hard coded codes in the source. Maybe raising an issue is appropriated. What do you think?

funkyboy: Make sense. Issue here: https://github.com/ably/ably-ios/issues/760

Source/ARTBaseMessage.h
```diff
@@ -29,6 +30,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (strong, nonatomic, nullable) id data;
 
+/// The event name, if available
+@property (nullable, readwrite, strong, nonatomic) NSString *name;
```
paddybyers: I don't understand why `name` is moved. Is the idea to include `name` in `PresenceMessage` ? That shouldn't happen.

Spec/RestClient.swift
```diff
@@ -1378,8 +1378,42 @@ class RestClient: QuickSpec {
...
+                let channel = client.channels.get(channelName)
+                let name = buildStringThatExceedMaxMessageSize()
                 
+                waitUntil(timeout: testTimeout, action: { done in
+                    channel.publish(name, data: nil, extras: nil, callback: {err in
+                        expect(err?.code).to(equal(40009))
+                        expect(err?.message).to(contain("maximum message length exceeded"))
+                        done()
+                    })
+
```
paddybyers: Extra whitespace?

Source/ARTBaseMessage.h
paddybyers: So is it not easy to do:

```
ARTMessage.messageSize() {
  return super.messageSize() + name.length;
}
```

?

Source/ARTTypes.m
```diff
@@ -186,6 +190,16 @@ - (NSDictionary *)toJSON:(NSError *_Nullable *_Nullable)error {
     return self;
 }
 
+- (NSString *)toJSONString {
```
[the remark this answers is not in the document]
funkyboy: We needed a way to have a JSON-stringified representation of a dictionary, which is not a built-in function.

Source/ARTBaseMessage.h
funkyboy: That's one way to do it. I was caught by other aspects of this PR and I didn't think of it :)
Issue here: https://github.com/ably/ably-ios/issues/762

PR #765, opened 2018-08-09
Use latest version of Fastlane

funkyboy:
Let's use the latest, without specifying a version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
   # Carthage
   - carthage bootstrap --platform iOS
   # Update fastlane
-  - gem install fastlane -v '2.100.1'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,7 +19,7 @@ install:
   # Carthage
   - carthage bootstrap --platform iOS
   # Update fastlane
-  - gem install fastlane -v '2.100.1'
+  - gem update fastlane
```
ricardopereira: Not a good thing. I had this once. The CI can brake quite easily.

funkyboy: My idea was to avoid making a PR like this multiple times a year :)

But you are right, consistency is better.

PR #774, opened 2018-08-10
Turn on and fix RSL1h RSL6a2 tests

funkyboy:


Spec/RestClientChannel.swift
```diff
@@ -327,15 +327,17 @@ class RestClientChannel: QuickSpec {
...
-            pending("should provide an optional argument that allows the extras value to be specified") {
-                // TODO: pushenabled doesn't appear to be working.
+            it("should provide an optional argument that allows the extras value to be specified") {
                 let client = ARTRest(options: AblyTests.commonAppSetup())
+                let originalARTChannels_getChannelNamePrefix = ARTChannels_getChannelNamePrefix
+                defer { ARTChannels_getChannelNamePrefix = originalARTChannels_getChannelNamePrefix }
+                ARTChannels_getChannelNamePrefix = nil // Force that channel name is not changed.
                 let channel = client.channels.get("pushenabled:test")
-                let extras = ["push": ["key": "value"]] as ARTJsonCompatible
+                let extras = ["hello": ["key": "value"]] as ARTJsonCompatible
```
funkyboy: Note that `push` is a reserved key in this case: https://docs.ably.io/client-lib-development-guide/features/#TM2i

paddybyers: But `"hello"` isn't guaranteed to work by the spec. The test should use `extras.push`, but include data that succeeds validation (eg it could be a valid set of push notification data).

funkyboy: @paddybyers ooops https://github.com/ably/ably-ios/pull/774/commits/c189e92875169bba4222811ca4e5bbbad4eabef2

PR #783, opened 2018-08-23
HTTP Paginated Response

ricardopereira:


Source/ARTHttp.m
```diff
@@ -51,7 +52,7 @@ - (void)executeRequest:(NSMutableURLRequest *)request completion:(void (^)(NSHTT
         } else {
             [self.logger debug:@"%@ %@: statusCode %ld", request.HTTPMethod, request.URL.absoluteString, (long)httpResponse.statusCode];
             [self.logger verbose:@"Headers %@", httpResponse.allHeaderFields];
-            NSString *headerErrorMessage = httpResponse.allHeaderFields[@"X-Ably-ErrorMessage"];
+            NSString *headerErrorMessage = httpResponse.allHeaderFields[ARTHttpHeaderFieldErrorMessageKey];
```
ricardopereira: BTW, I hope this won't break anything. The header is definitely `X-Ably-Errormessage` but it was using `"-Ably-ErrorMessage` 🤷‍♂️

Source/ARTConstants.m
```diff
@@ -0,0 +1,13 @@
...
+//  Ably
+//
+//  Created by Ricardo Pereira on 23/08/2018.
+//  Copyright © 2018 Ably. All rights reserved.
+//
+
+#import "ARTConstants.h"
+
+NSString *const ARTHttpHeaderFieldErrorCodeKey = @"X-Ably-Errorcode";
+NSString *const ARTHttpHeaderFieldErrorMessageKey = @"X-Ably-Errormessage";
```
funkyboy: I am guessing we can move other constants in here once this is merged.

Source/ARTHTTPPaginatedResponse.m
```diff
@@ -0,0 +1,139 @@
...
+ART_TRY_OR_REPORT_CRASH_START(self.rest) {
+    if (callback) {
+        void (^userCallback)(ARTHTTPPaginatedResponse *_Nullable result, ARTErrorInfo *_Nullable error) = callback;
+        callback = ^(ARTHTTPPaginatedResponse *_Nullable result, ARTErrorInfo *_Nullable error) {
+            ART_EXITING_ABLY_CODE(self.rest);
+            dispatch_async(self.userQueue, ^{
+                userCallback(result, error);
+            });
+        };
+    }
```
funkyboy: This `if` looks the same as implemented in `first:`. Worth extracting it in separate method?

Source/ARTHTTPPaginatedResponse.m
```diff
@@ -0,0 +1,139 @@
...
+
+            NSDictionary *links = [response extractLinks];
+
+            NSMutableURLRequest *firstRel = [NSMutableURLRequest requestWithPath:links[@"first"] relativeTo:request];
+            NSMutableURLRequest *currentRel = [NSMutableURLRequest requestWithPath:links[@"current"] relativeTo:request];
+            NSMutableURLRequest *nextRel = [NSMutableURLRequest requestWithPath:links[@"next"] relativeTo:request];
+
+            ARTHTTPPaginatedResponse *result = [[ARTHTTPPaginatedResponse alloc] initWithResponse:response items:items rest:rest relFirst:firstRel relCurrent:currentRel relNext:nextRel responseProcessor:responseProcessor];
+
+            callback(result, [ARTErrorInfo createFromNSError:error]);
```
funkyboy: I don't understand why the error is passed along here. Shouldn't it be nil?

Source/ARTStatus.m
```diff
@@ -35,6 +35,9 @@ + (ARTErrorInfo *)createWithCode:(NSInteger)code status:(NSInteger)status messag
 }
 
 + (ARTErrorInfo *)createFromNSError:(NSError *)error {
+    if (!error) {
+        return nil;
+    }
```
funkyboy: Now I see why you passed `[ARTStatus createFromNSError: ... ]` in the successful callback above.
I think it's confusing to see a "I successfully got results and here they are" callback that takes a (at first sight) non-null error value. Did you do it so that the syntax looks nicer in Swift?

Source/Ably.modulemap
```diff
@@ -5,6 +5,7 @@ framework module Ably {
     module * { export * }
 
     explicit module Private {
+        header "ARTConstants.h"
```
funkyboy: Why is this listed as private?

Spec/RestPaginated.swift
```diff
@@ -0,0 +1,55 @@
...
+import Quick
+import SwiftyJSON
+
+class RestPaginated : QuickSpec {
+    override func spec() {
+        describe("RestPaginated") {
+
+            let links = "<./messages?start=0&end=1535035746063&limit=100&direction=backwards&format=msgpack&firstEnd=1535035746063&fromDate=1535035746063&mode=all>; rel=\"first\", <./messages?start=0&end=1535035746063&limit=100&direction=backwards&format=msgpack&firstEnd=1535035746063&fromDate=1535035746063&mode=all>; rel=\"current\""
+
+            let url = URL(string: "https://sandbox-rest.ably.io:443/channels/foo/messages?limit=100&direction=backwards")!
```
funkyboy: This won't work as soon as we change the value of `ABLY_ENV`

Spec/RestPaginated.swift
```diff
@@ -0,0 +1,55 @@
...
+
+                guard let firstLink = extractedLinks["first"] as? String else {
+                    fail("First link is missing from extracted links"); return
+                }
+
+                guard let firstRequest = NSMutableURLRequest(path: firstLink, relativeTo: request) else {
+                    fail("First link isn't a valid URL"); return
+                }
+
+                expect(firstRequest.url?.absoluteString).to(equal("https://sandbox-rest.ably.io:443/channels/foo/messages?start=0&end=1535035746063&limit=100&direction=backwards&format=msgpack&firstEnd=1535035746063&fromDate=1535035746063&mode=all"))
```
funkyboy: Same here.

Spec/RestPaginated.swift
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+    }
+}
```
funkyboy: Can we have also tests for next and last?

Source/ARTConstants.m
ricardopereira: Yes, exactly. I think it makes more sense to have a `ARTConstants.h` file instead putting everything into `ARTTypes.h`.

Spec/RestClient.swift
```diff
@@ -1623,13 +1631,24 @@ class RestClient: QuickSpec {
 
                         waitUntil(timeout: testTimeout) { done in
                             do {
-                                try rest.request("get", path: "feature", params: nil, body: nil, headers: nil) { paginatedResult, error in
+                                try rest.request("get", path: "feature", params: nil, body: nil, headers: nil) { paginatedResponse, error in
                                     guard let error = error else {
-                                        fail("Error is nil"); done(); return
+                                        fail("Error is empty"); done(); return
```
paddybyers: I don't understand why you expect there to be an error in this case. The semantics of this method is that the request was made, and there was a response, and that's that. There was no error.

The response includes a statusCode that is an "error" result, but the request itself didn't fail. It succeeded, and returned a result.

Also, pls can we have tests for requests that do indeed fail - for example it times out, fails to resolve the host, returns a response body that cannot be parsed, etc.

funkyboy: @paddybyers this is making a `GET /feature` which doesn't exist, so the request fails and the error should not be nil. This `guard` simply makes the test fail if the error is nil.

paddybyers: No, that's not right. If the server responds to the request, and that response is well-formed (even though it is a 404), then that counts as success *in this API*. There is no `error`, but the response, encapsulated in the `HttpPaginatedResponse`, will indicate that the *content of the response* was an error, appropriately populating `success`, `statusCode`, `errorCode` and `errorMessage`.

RSC19e states that this method does result in an error if the server did not respond, or its response was not well-formed in some way, so it was not possible to return a `HttpPaginatedResponse`.

funkyboy: Oh I see now. So the error in nested in the pagination block of the response, not on the first level of the returned object.

ricardopereira: > No, that's not right. If the server responds to the request, and that response is well-formed (even though it is a 404), then that counts as success *in this API*. There is no `error`, but the response, encapsulated in the `HttpPaginatedResponse`, will indicate that the *content of the response* was an error, appropriately populating `success`, `statusCode`, `errorCode` and `errorMessage`.

That makes it clear, yes. Thanks. I'll fix it.

Spec/RestPaginated.swift
ricardopereira: Hmm, that's a good point but we need something to respond with the ref next and last links.

Spec/RestPaginated.swift
ricardopereira: Those requests will fail so I think they doesn't need look at ABLY_ENV. I just use `sandbox` to not overload a production server.

Source/Ably.modulemap
ricardopereira: Because the lib user doesn't need to have access to those constants.

Source/ARTStatus.m
ricardopereira: I change it with 861f32b. Thanks  👍

Source/ARTHTTPPaginatedResponse.m
ricardopereira: I changed it 861f32b. Thanks

Spec/RestClient.swift
ricardopereira: Fixed 861f32b.

ricardopereira: Fixed 63ebbf5.

Spec/RestClient.swift
```diff
@@ -1632,14 +1632,10 @@ class RestClient: QuickSpec {
...
                             do {
                                 try rest.request("get", path: "feature", params: nil, body: nil, headers: nil) { paginatedResponse, error in
-                                    guard let error = error else {
-                                        fail("Error is empty"); done(); return
-                                    }
+                                    expect(error).to(beNil())
                                     guard let paginatedResponse = paginatedResponse else {
                                         fail("PaginatedResult is empty"); done(); return
                                     }
-                                    expect(error.statusCode) == 404
```
paddybyers: Why are there no corresponding assertions for the properties of `paginatedResponse` ?

ricardopereira: @paddybyers There already exist at the bottom. Did you miss them? https://github.com/ably/ably-ios/blob/63ebbf5a24d5807a19e07ddbfb2de0e9e65def24/Spec/RestClient.swift#L1642-L1647

paddybyers: Yes :(

ricardopereira: @paddybyers 😄 So we're good?

PR #789, opened 2018-09-05
Add RTN15a

funkyboy:
`(RTN15a) If a Connection transport is disconnected unexpectedly or if a token expires, then the Connection manager will immediately attempt to reconnect and restore the connection state. Connection state recovery is provided by the Ably service and ensures that whilst the client is disconnected, all events are queued and channel state is retained on the Ably servers. When a new connection is made with the correct connection recovery key, the client is able to catch up by receiving the queued ProtocolMessages from Ably.`

- Includes cleanup of tests that were tweaking the value of `disconnectedRetryTimeout`
- adds a private `suspendImmediateReconnection` property to prevent an immediate reconnection after a disconnection happens. Handy for tests.

### DIFF
--- a/Source/ARTRealtime+Private.h
+++ b/Source/ARTRealtime+Private.h
@@ -68,6 +68,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (readonly, getter=getClientOptions) ARTClientOptions *options;
 
+/// Suspend the behavior defined in RTN15a, that is trying to immediately reconnect after a disconnection
+@property (readwrite, assign, nonatomic) BOOL suspendImmediateReconnection;
+
 @end
 
 @interface ARTRealtime (Private)
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -506,7 +506,12 @@ ART_TRY_OR_MOVE_TO_FAILED_START(self) {
 
             [self.transport close];
             _transport = nil;
-            [stateChange setRetryIn:self.options.disconnectedRetryTimeout];
+            NSTimeInterval retryInterval = self.options.disconnectedRetryTimeout;
+            // RTN15a - retry immediately if client was connected
+            if (stateChange.previous == ARTRealtimeConnected && !_suspendImmediateReconnection) {
+                retryInterval = 0.1;
+            }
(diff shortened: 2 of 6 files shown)


Source/ARTRealtime+Private.h
```diff
@@ -68,6 +68,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (readonly, getter=getClientOptions) ARTClientOptions *options;
 
+/// Suspend the behavior defined in RTN15a, that is trying to immediately reconnect after a disconnection
+@property (readwrite, assign, nonatomic) BOOL suspendImmediateReconnection;
```
ricardopereira: Why this flag? If the client isn't respecting the spec then it should not be a flag to be activated only for the test suite, right? Maybe I'm missing something.

funkyboy: @ricardopereira correct, now it waits 15s.

The flag is needed because sometimes you need a "disconnect and stay in the disconnected for x seconds" case. For example when you are testing that there's a new connection id if you connect after (ttl + idle interval). Without this flag you'd have an immediate connection retry.
We had to do [something similar in Java](https://github.com/ably/ably-java/commit/79dc37d6d2926a424311b96a715dcdf76e694f0e#diff-4cfebe2518c34c9f3975cde19f3ced08R469).

ricardopereira: Ok, understood.

paddybyers:
Am I missing it or is there a test that verifies there is an immediate reconnection?

funkyboy:
@paddybyers been so focused on avoiding immediate reconnection in tests that I forgot to test if it actually happens :) 
Here you go https://github.com/ably/ably-ios/pull/789/commits/8bab329f3f96471273b830643c20cf0528261f7b

paddybyers:
LGTM, although I worry that anything with only a 2s margin of error will break in Travis. I think t's safe to check the reconnect time is anything in the interval 0 .. 10s, because at least then you know it's not waiting the usual 15s.

funkyboy:
Travis tests were passing but let's make sure we have more leeway https://github.com/ably/ably-ios/pull/789/commits/e16c7a6d73c2bf1cfa1672b981e3adb873de9945

PR #791, opened 2018-09-05
Improve RTN14b test

funkyboy:


Spec/RealtimeClientConnection.swift
```diff
@@ -1991,7 +1991,9 @@ class RealtimeClientConnection: QuickSpec {
                             guard let reason = stateChange?.reason else {
                                 fail("Reason is nil"); done(); return;
                             }
-                            expect(reason.code) == 40142
+                            expect(reason.code).to(equal(40142))
+                            expect(reason.statusCode).to(equal(401))
+                            expect(stateChange!.reason!.message).to(contain("Key/token status changed (expire)"))
```
ricardopereira: Can we replace `stateChange!.reason!.message` with `reason.message`?

Spec/RestClient.swift
```diff
@@ -1384,71 +1384,35 @@ class RestClient: QuickSpec {
...
-                    channel.publish("test", data: "test-data", callback: { error in
-                        guard let error = error else {
-                            fail("Error should not be empty")
+            context("if in the course of a REST request an attempt to authenticate using authUrl fails due to a timeout") {
+                // RSA4e
+                it("the request should result in an error with code 40170, statusCode 401, and a suitable error message") {
+                    let options = AblyTests.commonAppSetup()
+                    let token = getTestToken()
+                    options.httpRequestTimeout = 3 // short timeout to make it fail faster
+                    options.authUrl = NSURL(string: "http://10.255.255.1")! as URL
```
ricardopereira: You can use directly the `URL` object, so `URL(string: "http://10.255.255.1")!` instead of we you have now.

ricardopereira: Did you moved the code to another place, is that it?

funkyboy: @ricardopereira No, as part of a rebase this PR includes one of your commits https://github.com/ably/ably-ios/pull/791/commits/09c428bd4c95afde03556e4caa8641f6d1b2780a

funkyboy: @ricardopereira 
> You can use directly the URL object, so URL(string: "http://10.255.255.1")! instead of we you have now.

There's many occurrences of `NSURL(...)` in the swift code. I propose to address this in a separate PR. [Issue here](https://github.com/ably/ably-ios/issues/794). Ok?

Issue #787, opened 2018-08-30
Add Transient publishing


ricardopereira:

ricardopereira:
@mattheworiordan Could you please add some information about this?

mattheworiordan:
You created the issue, so shouldn't you? 😄 

See https://github.com/ably/docs/pull/323

ricardopereira:
@mattheworiordan Didn't know where to look 🙄 You're right. I was lazy. Thanks man 😄

ricardopereira:
@funkyboy I saw your #793 but I was starting to implement this. Are you doing the same?

funkyboy:
@ricardopereira no. That PR was just meant to add the new values to the enum.

PR #795, opened 2018-09-06
Transient Publishing

ricardopereira:
#787


ricardopereira:
@funkyboy Yap. I'm looking at them now 😄

ricardopereira:
@paddybyers @mattheworiordan So, I'm facing a lot of failing tests related with ACK's/NACK's because some of the tests were created with the original intention to queue an array of messages and wait for the ATTACHED state. With this, it resulted in one only ACK message with multiple acknowledgments.

Now, with the Transient Publishing implementation that I did, the behaviour is different because the array of messages is sent right away because the connection is ready, resulting in multiple ACK messages.

Should I revisit all those failing tests and update them with the new behaviour? Or the existing tests should work as usual?

paddybyers:
@ricardopereira the existing tests that now fail are testing ... what? If they're not related to transient publishing, can't those tests continue to work in the same way by calling `attach()` before the first publish?

ricardopereira:
@paddybyers They're not related to transient publishing. Ok, attaching the channel before the publish should work, I'll try. Thanks.

ricardopereira:
@paddybyers The test "should reset msgSerial serially if the connection does not resume" is failing. It receives 2 or 3 ACKs and sometimes only 1 but it's not consistent like it was before the new code https://github.com/ably/ably-ios/blob/e6e0c6dcf76f0ede6ea630a92d9f72cece25e90a/Spec/RealtimeClientConnection.swift#L1046

In Transient Publishing, in that case, how can I get only one ACK?

Source/ARTRealtimeChannel.m
```diff
@@ -276,21 +268,84 @@ - (void)publishProtocolMessage:(ARTProtocolMessage *)pm callback:(void (^)(ARTSt
...
+        case ARTRealtimeChannelDetaching:
+        case ARTRealtimeChannelDetached:
+        case ARTRealtimeChannelAttaching:
+        case ARTRealtimeChannelAttached: {
+            if (_realtime.connection.state_nosync == ARTRealtimeConnected) {
+                [self sendMessage:pm callback:cb];
+            }
+            else {
+                [self addToQueue:pm callback:queuedCallback];
+                [self.realtime.internalEventEmitter once:[ARTEvent newWithConnectionEvent:ARTRealtimeConnectionEventConnected] callback:^(ARTConnectionStateChange *_Nullable change) {
```
SimonWoolf: This seems unnecessary. Why have a channel-level queue for messages waiting for the connection to become connected (and manually listen for connection events) when the ARTRealtime class has its own queue that will do that automatically?

(Not to mention, seems like having a channel-level queue that mixes Messages waiting for connection CONNECTED with PresenceMessages waiting for channel ATTACHED is error-prone -- the event added here will drain the whole queue when the conneciton connects, including presence messages that actually need to wait for an attached, no?)

Spec/RealtimeClientChannel.swift
```diff
@@ -1725,167 +1725,228 @@ class RealtimeClientChannel: QuickSpec {
...
-                        it("INITIALIZED") {
                             waitUntil(timeout: testTimeout) { done in
-                                expect(client.connection.state).to(equal(ARTRealtimeConnectionState.initialized))
-                                publish(done)
-                                client.connect()
-                                expect(channel.queuedMessages).to(haveCount(1))
+                                channel.publish(nil, data: "message") { error in
+                                    expect(error).to(beNil())
+                                    done()
+                                }
```
SimonWoolf: This test doesn't seem to test that the message is actually published immediately (ie without waiting for the channel to attach) -- it just checks that the publish completes eventually, you're not checking channel state after the publish completed. (similarly with the other tests)

SimonWoolf:
> It receives 2 or 3 ACKs and sometimes only 1 but it's not consistent like it was before the new code https://github.com/ably/ably-ios/blob/e6e0c6dcf76f0ede6ea630a92d9f72cece25e90a/Spec/RealtimeClientConnection.swift#L1046

> In Transient Publishing, in that case, how can I get only one ACK?

You should stop making assumptions about the number of ACKs you receive. The server is free to roll up multiple acks into one (using a `count` > 1), or not, at its discretion, to achieve some balance of latency minimization vs efficiency. You can not assume any particular way it will do this, other than that if the transport remains active every message will be acked/nacked eventually.

mattheworiordan:
@ricardopereira when do you expect this PR to be able to land?

ricardopereira:
@mattheworiordan I can work on this next week.

mattheworiordan:
Ok, please do, it's worrying to have a PR open for this long

mattheworiordan:
I realise you've done a lot of work on iOS recently @ricardopereira, but hopefully once the push stuff is out the way, we can rebase this and get it merged.

Source/ARTRealtimeChannel.m
ricardopereira: Fixed a4466b7

Spec/RealtimeClientChannel.swift
ricardopereira: Fixed a93a589

Source/ARTRealtimeChannel.m
```diff
@@ -331,23 +323,79 @@ - (void)publishProtocolMessage:(ARTProtocolMessage *)pm callback:(void (^)(ARTSt
...
+} ART_TRY_OR_MOVE_TO_FAILED_END
+}
 
-                [self.realtime.internalEventEmitter once:[ARTEvent newWithConnectionEvent:ARTRealtimeConnectionEventConnected] callback:^(ARTConnectionStateChange *__art_nullable change) {
-                    [weakSelf sendQueuedMessages];
-                }];
+- (void)publishProtocolMessage:(ARTProtocolMessage *)pm callback:(void (^)(ARTStatus *))cb {
+ART_TRY_OR_MOVE_TO_FAILED_START(_realtime) {
+    ARTStatus *statusInvalidConnectionState = [ARTStatus state:ARTStateError info:[ARTErrorInfo createWithCode:90001 message:@"channel operation failed (invalid connection state)"]];
+    ARTStatus *statusInvalidChannelState = [ARTStatus state:ARTStateError info:[ARTErrorInfo createWithCode:90001 message:@"channel operation failed (invalid channel state)"]];
```
SimonWoolf: Could these not be a bit more specific? (cf https://github.com/ably/ably-cocoa/issues/843)

Source/ARTRealtimeChannel.m
```diff
@@ -331,23 +323,79 @@ - (void)publishProtocolMessage:(ARTProtocolMessage *)pm callback:(void (^)(ARTSt
...
+        case ARTRealtimeChannelInitialized:
+        case ARTRealtimeChannelDetaching:
+        case ARTRealtimeChannelDetached:
+        case ARTRealtimeChannelAttaching:
+        case ARTRealtimeChannelAttached: {
+            if (_realtime.connection.state_nosync == ARTRealtimeConnected) {
+                [self sendMessage:pm callback:cb];
+            }
+            else {
+                [self addToQueue:pm callback:queuedCallback];
```
SimonWoolf: I still don't understand why you're queueing Messages in a channel-level queue here. What's the reasoning behind doing that rather than just sending them to the connection queue? (cf previous comment https://github.com/ably/ably-cocoa/pull/795#discussion_r278043050 )

Source/ARTRealtimeChannel.m
ricardopereira: PR https://github.com/ably/ably-cocoa/pull/895 created.

PR #802, opened 2018-09-25
Add rtl6d1

funkyboy:
Adds `RTL6d1`
```
(RTL6d) Messages for a single channel that have been queued may be sent in a single ProtocolMessage by bundling them into the ProtocolMessage#messages array, subject to the following constraints:
    (RTL6d1) The resulting ProtocolMesssage must not exceed the maxMessageSize
```

### DIFF
--- a/Spec/RealtimeClientChannel.swift
+++ b/Spec/RealtimeClientChannel.swift
@@ -2066,13 +2066,13 @@ class RealtimeClientChannel: QuickSpec {
                     defer { client.dispose(); client.close() }
                     let channel = client.channels.get("test-maxMessageSize")
                     // This amount of messages would be beyond maxMessageSize, if bundled together
-                    let messagesToBeSent = 5000
+                    let messagesToBeSent = 1000
                     
                     // call publish before connecting, so messages are queued
                     waitUntil(timeout: testTimeout*2) { done in
                         let partialDone = AblyTests.splitDone(messagesToBeSent, done: done)
                         for i in 1...messagesToBeSent {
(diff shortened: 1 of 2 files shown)


Spec/RealtimeClientChannel.swift
```diff
@@ -2057,6 +2057,37 @@ class RealtimeClientChannel: QuickSpec {
...
+                        client.connect()
+                    }
+                    
+                    let transport = client.transport as! TestProxyTransport
+                    let protocolMessages = transport.protocolMessagesSent.filter{ $0.action == .message }
+                    // verify that messages are not bundled in a single protocol message
+                    expect(protocolMessages.count).to(beGreaterThan(1))
+                    // verify that all the messages have been sent
+                    let messagesSent = protocolMessages.compactMap{$0.messages?.count}.reduce(0, +)
+                    expect(messagesSent).to(equal(messagesToBeSent))
```
ricardopereira: I can't detect where the resulting `ProtocolMesssage` is not exceeding the `maxMessageSize`. I think you should check if the messages are being blocked if the total size is exceeding the `maxMessageSize`, right?

funkyboy: I had exactly the same thought at first :)
But this is queued messages, so the iOS client should bundle them in a way that the resulting  protocol message does not exceed max size.

ricardopereira: Oh I see 😄 That's why the `beGreaterThan(1)`.

Issue #460, opened 2016-08-05
Cocoa Mac development support


mattheworiordan:
When trying to use this library for Mac development, the following error is shown:

```
"The platform of the target `iris` (OS X 10.11) is not compatible with `AblyRealtime (0.8.2)`, which does not support `osx`."
```

We need to ensure the library is compatible with OSX

hemezh:
Thanks @mattheworiordan for the quick response.

ricardopereira:
**For macOS compability, please have in mind:**
 - Reachability
 - Locale changes
 - Date and Time changes

mattheworiordan:
@ricardopereira not something to look at now, but any idea how much work it may be to add Mac OS support?

ricardopereira:
Adding a new framework target for the other platform and fix incompatibilities should be enough. I would say 2 days of work.

mattheworiordan:
Ok, let's review once 0.9 is done

rynosoft:
Did anything come of this?

funkyboy:
@rynosoft Not yet. It shouldn't be a complicated task. Here's the plan I'd follow:

- import `AppKit` instead of `UIKit` if the target platform is macOS
- configure the test suite to run also on the `macOS` platform (Sierra and High Sierra).
- build a sample macOS app to make sure that our dependencies (KSCrash, msgpack and SocketRocket) work correctly also on macOS with both Carthage and Cocoapods.

@rynosoft Feel free to take a stab at it if you like :) 
Next week I plan to work on this for a few hours to see how far I get.

rynosoft:
Hi Cesare:

I am taking a stab at it today. The first problem I’ve run into is that the Carthage update fails on msgpack. I’m researching for a solution. I’ll keep you apprised of my progress.

/Mick

rynosoft:
The carthage update fails for msgpack. I tweeted at the author about Mac support.

rynosoft:
I have successfully added Mac targets for the framework and tests. The tests appear to run successfully. I'm going to build a sample app next per your suggestion. Once I get it all working, I'll fork and push it.

funkyboy:
@rynosoft Cool. So in the end `msgpack` works with macOS as a target?

rynosoft:
I had to do a little work with msgpack which I'll also need to commit. Right now it's all cobbled together and won't work with carthage or cocoapods. I'll try to get that worked out in the next couple of days.

funkyboy:
[...]

PR #817, opened 2018-11-29
Mac support

ricardopereira:
### Installation Guide

⚠️ Push Notifications are unavailable in **macOS**.

#### CocoaPods

`pod 'Ably', :git => 'https://github.com/ably/ably-ios.git', :branch => 'mac-support'`

#### Carthage

`github "ably/ably-ios" "mac-support"`

#### Demo

screenshot 2018-11-30 at 10 04 55

Demo available [here](https://github.com/ably/demo-macos).

### Most important changes
Mostly project configurations has been changed.

Travis file has been updated:
 - iOS 9, 10, 11, 12 is being tested
 - macOS 10.14 is being tested

New targets and scheme:
 - `Ably` has been renamed to `Ably-iOS`
 - `Ably-macOS` has been added (minimum deployment target is 10.10 Yosemite)
 - `Ably-tvOS` has been added (but not ready because of one dependency that hasn't support for `tvOS`).

Dependencies:
 - New `ULID` version because the macOS support was broken.
 - New `msgpack-objective-C` from ably-forks/msgpack-objective-C

Reference #460


paddybyers:
I've not looked through in detail.

What needs to happen before we can merge the mac support branch?

> Maybe we need to consider changing the name of the repo to `ably-cocoa` instead of `ably-ios` like other library devs companies do

I'd be ok with that in principle if we set up a repo redirect so it's transparent to users of the existing git url.
I think iOS is still the primary use-case and is probably the most natural name. But if we do rename it,

ricardopereira:
> What needs to happen before we can merge the mac support branch?

@paddybyers Nothing in particular. This is ready to land in `develop` branch.

> I'd be ok with that in principle if we set up a repo redirect so it's transparent to users of the existing git url.

Yes indeed. I never done that. Most likely things will brake for users that are using old versions if we rename the repository to `ably-cocoa`. Need to do some investigation.

mattheworiordan:
Thanks Ricardo.

I think we should not only land it on develop, but bump the minor version and document tvOS and macOS support in the readme & changeling.  We also need to update the www.ably.io/download <http://www.ably.io/download> page to include details on those OS’ being supported.

In regards to renaming, I think that is sensible.  Github offers redirects now, so that will be done seemlessly.  In regards to the package though and where it’s published, I don’t know how that works.  Ricardo would you mind looking into whether there’s a way to rename libraries?  

We should probably defer the rename until the 1.1 release, which is around the corner (aiming for 2nd week of Dec).

PR #824, opened 2019-01-10
Release 1.1.3

ricardopereira:


paddybyers:
I did have a look at the two demo projects referenced, and I think now if we're linking to them then it would be good to improve the READMEs so it's more obvious what they do

PR #826, opened 2019-01-18
Push State Machine: main thread sometimes gets stuck when accessing Local Device

ricardopereira:


ricardopereira:
@paddybyers  Yes, if you try to access the local device by, i.e., trying to subscribe your device, and at the same time call activate (which syncs the main thread) then it could do a deadlock.

barrier queue ably stacktrace

PR #828, opened 2019-01-20
Stop notifying Slack for every platform and every iOS build

mattheworiordan:
Now that we've disabled the Slack webhooks for every iOS build on every platform, the builds need to also stop attempting to notify Slack.

@ricardopereira I assume we still see the results of Fastlane in the build output itself so this change is fine?


ricardopereira:
Thank you Matt! That's it.

PR #831, opened 2019-02-07
Timestamp should not be generated in the client

ricardopereira:
#576


Spec/Auth.swift
```diff
@@ -1470,6 +1470,7 @@ class Auth : QuickSpec {
                         clientOptions.authMethod = "POST"
                         clientOptions.authHeaders = ["X-Header-1": "foo", "X-Header-2": "bar"]
                         let tokenParams = ARTTokenParams()
+                        tokenParams.timestamp = Date()
```
SimonWoolf: You shouldn't need this. The timestamp should be generated in the auth server, when creating the signed token request

ricardopereira: @SimonWoolf done e4a4591. Thanks.

ricardopereira:
@SimonWoolf Please review. Thanks

Issue #737, opened 2018-06-21
my-members presenceMap requirement change for 1.1


SimonWoolf:
See https://github.com/ably/docs/pull/446

SimonWoolf:
@ricardopereira Just want to check in on the status of this? Bit worried that this issue seems to be left open after the 1.1 release -- as noted in the linked docs pr, the server assumes that any client library that sends `v=1.1` implements the local presence map up to spec

ricardopereira:
@SimonWoolf Ok, I will look at it and after this I'll review all the remaining issues and see if something else is missing. Thanks

PR #835, opened 2019-03-01
RTP17b

ricardopereira:
> (RTP17b) The events that should be applied to the RTP17 presence map are:
> * any ENTER, PRESENT or UPDATE event with a connectionId that matches the current client’s connectionId;
> * any LEAVE event with a connectionId that matches the current client’s connectionId and is NOT a ‘synthesized leave’ (an event that has a connectionId which is not an initial substring of its id, per RTP2b1 )

Related #737

### DIFF
--- a/Source/ARTPresenceMap.m
+++ b/Source/ARTPresenceMap.m
@@ -170,18 +170,21 @@ NSString *ARTPresenceSyncStateToStr(ARTPresenceSyncState state) {
 }
 
 - (void)startSync {
+    [_logger debug:__FILE__ line:__LINE__ message:@"%p PresenceMap sync started", self];
     _syncSessionId++;
     _syncState = ARTPresenceSyncStarted;
     [_syncEventEmitter emit:[ARTEvent newWithPresenceSyncState:_syncState] with:nil];
 }
 
 - (void)endSync {
+    [_logger verbose:__FILE__ line:__LINE__ message:@"%p PresenceMap sync ending", self];
     [self cleanUpAbsentMembers];
     [self leaveMembersNotPresentInSync];
     _syncState = ARTPresenceSyncEnded;
     [self reenterLocalMembersMissingFromSync];
     [_syncEventEmitter emit:[ARTEvent newWithPresenceSyncState:ARTPresenceSyncEnded] with:[_members allValues]];
     [_syncEventEmitter off];
+    [_logger debug:__FILE__ line:__LINE__ message:@"%p PresenceMap sync ended", self];
 }
 
 - (void)failsSync:(ARTErrorInfo *)error {
@@ -192,18 +195,11 @@ NSString *ARTPresenceSyncStateToStr(ARTPresenceSyncState state) {
 }
 
 - (void)onceSyncEnds:(void (^)(NSArray<ARTPresenceMessage *> *))callback {
-    if (self.syncInProgress) {
-        [_syncEventEmitter once:[ARTEvent newWithPresenceSyncState:ARTPresenceSyncEnded] callback:callback];
-    }
-    else {
-        callback([_members allValues]);
-    }
+    [_syncEventEmitter once:[ARTEvent newWithPresenceSyncState:ARTPresenceSyncEnded] callback:callback];
 }
 
 - (void)onceSyncFails:(void (^)(ARTErrorInfo *))callback {
-    if (self.syncInProgress) {
-        [_syncEventEmitter once:[ARTEvent newWithPresenceSyncState:ARTPresenceSyncFailed] callback:callback];
-    }
+    [_syncEventEmitter once:[ARTEvent newWithPresenceSyncState:ARTPresenceSyncFailed] callback:callback];
 }
 
 - (BOOL)syncComplete {
--- a/Source/ARTPresenceMap.m
+++ b/Source/ARTPresenceMap.m
@@ -111,7 +111,7 @@ NSString *ARTPresenceSyncStateToStr(ARTPresenceSyncState state) {
     // Local member
     if ([message.connectionId isEqualToString:self.delegate.connectionId]) {
         [_localMembers addObject:message];
-        [_logger debug:__FILE__ line:__LINE__ message:@"local member %@ added", message.memberKey];
+        [_logger debug:__FILE__ line:__LINE__ message:@"local member %@ with action %@ has been added", message.memberKey, ARTPresenceActionToStr(message.action).uppercaseString];
     }
 }
 
--- a/Source/ARTPresenceMap.m
+++ b/Source/ARTPresenceMap.m
@@ -78,7 +78,7 @@ NSString *ARTPresenceSyncStateToStr(ARTPresenceSyncState state) {
 
 - (BOOL)add:(ARTPresenceMessage *)message {
     ARTPresenceMessage *latest = [_members objectForKey:message.memberKey];
-    if ([self isNewestPresence:message comparingWith:latest]) {
+    if ([message isNewerThan:latest]) {
         ARTPresenceMessage *messageCopy = [message copy];
         switch (message.action) {
             case ARTPresenceEnter:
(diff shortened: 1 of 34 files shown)


Source/ARTRealtime.m
```diff
@@ -550,12 +550,6 @@ - (ARTEventListener *)transitionSideEffects:(ARTConnectionStateChange *)stateCha
 
     if ([self shouldSendEvents]) {
         [self sendQueuedMessages];
-        // For every Channel
-        for (ARTRealtimeChannel* channel in self.channels.nosyncIterable) {
-            if (channel.state_nosync == ARTRealtimeChannelSuspended) {
-                [channel _attach:nil];
-            }
```
ricardopereira: I did the fix in this PR because the RTP17b test was failing because of this. Discussion [here](https://ably-real-time.slack.com/archives/C030C5YLY/p1550849880011900).

Source/ARTPresenceMap.m
```diff
@@ -119,6 +120,10 @@ - (void)internalRemove:(ARTPresenceMessage *)message {
 }
 
 - (void)internalRemove:(ARTPresenceMessage *)message force:(BOOL)force {
+    if ([message.connectionId isEqualToString:self.delegate.connectionId] && !message.isSynthesized) {
+        [_localMembers removeObject:message];
```
SimonWoolf: Now you've added this, you need to remove line 134, or it'll do the remove unconditionally

Source/ARTRealtime.m
SimonWoolf: @ricardopereira With this removal, does the library comply with RTN15c3? "when a resume request results in a CONNECTED with a new connectionId ... The client library should initiate an attach for channels that are in the SUSPENDED state. For all channels in the ATTACHING or ATTACHED state, the client library should fail any previously queued messages for that channel and initiate a new attach..."

cf my comment in slack "you shouldn't be sending an attached after a successful resume, only for an unsuccessful one" -- you still need to do the latter.

(From that test log it looks like the connection moved to `suspended`, but it's not obvious why from the log, the resume was successful so I can't see any reason for it to do so)

Source/ARTRealtimeChannel.m
```diff
@@ -164,6 +164,59 @@ - (void)internalPostMessages:(id)data callback:(void (^)(ARTErrorInfo *__art_nul
...
+    [self.logger verbose:__FILE__ line:__LINE__ message:@"R:%p C:%p (%@) requesting a sync operation", _realtime, self, self.name];
+
+    ARTProtocolMessage * msg = [[ARTProtocolMessage alloc] init];
+    msg.action = ARTProtocolMessageSync;
+    msg.channel = self.name;
+
+    if (self.presenceMap.syncMsgSerial) {
+        msg.channelSerial = self.presenceMap.syncChannelSerial;
+        msg.msgSerial = [NSNumber numberWithLongLong:self.presenceMap.syncMsgSerial];
+    }
```
SimonWoolf: I don't understand this. What is a syncMsgSerial? It looks like it's being set from a SYNC msgSerial, but that isn't a thing (no incoming messages ever have msgSerials, and SYNCs don't have them in either direction). Nothing in RTP3 tells you to do this, it only asks for the channelSerial. 

Also, AFAICS syncChannelSerial and syncMsgSerial are never unset in endSync, so any sync() call will result in the lib trying to resume an actually-successfully-completed sync, no?

Spec/RealtimeClientPresence.swift
```diff
@@ -2990,6 +2998,93 @@ class RealtimeClientPresence: QuickSpec {
...
+                                fail("Presences is nil"); done(); return
+                            }
+                            expect(channel.presenceMap.syncComplete).to(beTrue())
+                            expect(presences).to(haveCount(2))
+                            expect(presences.map({$0.clientId})).to(contain(["one", "two"]))
+                            done()
+                        }
+                    }
+                }
+
```
SimonWoolf: I can't see the bit of this test that tests synthesized leave's effect on the own-members set, as the test name implies? (If it is there than it's odd that it passed, since https://github.com/ably/ably-cocoa/pull/835/files#r278056029 means that the logic wasn't actually taking effect)

Source/ARTRealtime.m
ricardopereira: > With this removal, does the library comply with RTN15c3?

@SimonWoolf Yes it does. When I implemented `RTN15c3`, I forgot to remove that bit because the new implementation already handle that case. You can see that the channel will reattach https://github.com/ably/ably-cocoa/blob/8dbc03d9e1ec9776e4fbeb58a2d85469edc9182f/Source/ARTRealtime.m#L608-L615 and it does check the SUSPENDED channel state https://github.com/ably/ably-cocoa/blob/8dbc03d9e1ec9776e4fbeb58a2d85469edc9182f/Source/ARTRealtimeChannel.m#L924-L934

ricardopereira: > (From that test log it looks like the connection moved to suspended, but it's not obvious why from the log, the resume was successful so I can't see any reason for it to do so)

@SimonWoolf That's because the test forces to suspend:

https://github.com/ably/ably-cocoa/blob/8dbc03d9e1ec9776e4fbeb58a2d85469edc9182f/Spec/RealtimeClientPresence.swift#L3034-L3039

SimonWoolf: > I forgot to remove that bit because the new implementation already handle that case

:+1:

Though looking at `reattachWithReason`, it looks to me like it's reattaching channels in every state _except_ attaching. So (1) it's starting an attach for channels in `initialized`, `detached`, or `failed`, which it really shouldn't be; and (2) it isn't initiating a new attach for a channel in the `attaching` state (we've just started a completely new connection, so any previously-in-progress attach needs to be started again). Or am I misreading it?

Spec/RealtimeClientPresence.swift
ricardopereira: @SimonWoolf You're right, that part is missing. What's the best way to test the synthesized leave's effect? I'm thinking in forcing a synthesized leave of one existing presence and force a sync. Then after the sync, the user should be updated since the user is re-entered and the final members map should remain the same, no leave has taken effect.

SimonWoolf: A proper end-to-end test that forces realtime to do a synthesized leave is a bit annoying if only because it means the test will take at least 15 seconds. But if you want to do that, something like: enter, force a disconnect, wait 20s, reconnect, wait to see the synthesized leave (& sync showing the member not present), wait for lib to automatically re-enter.

Source/ARTRealtimeChannel.m
ricardopereira: @SimonWoolf Thanks for pointing this out. Fixed dc84181.

Source/ARTPresenceMap.m
ricardopereira: Fixed ca32ef2. Thanks

Spec/RealtimeClientPresence.swift
ricardopereira: @SimonWoolf Thanks, updated the test case 46dd5e7.

Source/ARTRealtime.m
ricardopereira: @SimonWoolf You're not misreading. It seems so, it's not respecting `RTN15c3`. Issue has been created: #847

ricardopereira:
It needs #845 for the test suite pass.

Source/ARTRealtimeChannel.m
SimonWoolf: And the second issue (second paragraph of my comment)? `presenceMap.syncChannelSerial` is only ever set, never unset, and you're setting it on the sync unconditionally, requesting a resume of a sync that actually completed successfully?

ricardopereira: @SimonWoolf Right, fixed 0d7cdc0. Set `syncChannelSerial` to nil when the sync ends.

PR #839, opened 2019-04-06
Test suite: Swift `allHeaderFields` property is case sensitive

ricardopereira:
Swift Regression https://bugs.swift.org/browse/SR-2429

### DIFF
--- a/Source/ARTConstants.m
+++ b/Source/ARTConstants.m
@@ -8,5 +8,5 @@
 
 #import "ARTConstants.h"
 
-NSString *const ARTHttpHeaderFieldErrorCodeKey = @"X-Ably-Errorcode";
-NSString *const ARTHttpHeaderFieldErrorMessageKey = @"X-Ably-Errormessage";
+NSString *const ARTHttpHeaderFieldErrorCodeKey = @"x-ably-errorcode";
+NSString *const ARTHttpHeaderFieldErrorMessageKey = @"x-ably-errormessage";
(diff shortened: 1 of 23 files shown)


ricardopereira:
Had to cherry pick f3933f7a0b282dbf110fa90e3eb7496157f74f46, 068b0ae984a8c6432e19ad4fb42d6759ccef98ac and b2165d2adb38f42dfcad4c4f94947fd8820694eb

Source/ARTConstants.m
```diff
@@ -8,5 +8,5 @@
 
 #import "ARTConstants.h"
 
-NSString *const ARTHttpHeaderFieldErrorCodeKey = @"X-Ably-Errorcode";
-NSString *const ARTHttpHeaderFieldErrorMessageKey = @"X-Ably-Errormessage";
+NSString *const ARTHttpHeaderFieldErrorCodeKey = @"x-ably-errorcode";
```
mattheworiordan: I am a little confused:

- If these are lower cased, why are all the tests upper cased?
- If you've corrected `X-Ably-Errorcode` to `X-Ably-ErrorCode` in https://github.com/ably/ably-cocoa/pull/839/commits/e9e5c22fd2f003eabe88251c69617f1a94e7ea23#diff-401a259ee49c8a3a0b9f14bd723f4430R23, why is that not reflected in the tests?

ricardopereira: @mattheworiordan I changed it initially but then I discovered what was really wrong:

> Swift Regression: HTTPURLResponse allHeaderFields is now case-sensitive https://bugs.swift.org/browse/SR-2429

That's why sometimes some tests were failing because the server responded differently: X-Ably-Errorcode or X-Ably-errorcode, etc.

So the case should not matter, the access to these header values should be case-insensitive as described in RFC 2616.

ricardopereira:
@mattheworiordan Please could you approve it (if it's ok for you)? I need it for other PR's and cherry-picking is becoming more tedious. Thanks

PR #845, opened 2019-04-20
Fix RTN16f (old RTN10c)

ricardopereira:


ricardopereira:
Had to cherry pick b2165d2adb38f42dfcad4c4f94947fd8820694eb and e9e5c22fd2f003eabe88251c69617f1a94e7ea23

mattheworiordan:
@ricardopereira I am a bit confused, `RTN10c` does not appear to exist at https://docs.ably.io/client-lib-development-guide/features?  Which spec item is this addressing.  What are the cherry-picked commits for?

Spec/RealtimeClientConnection.swift
```diff
@@ -1429,6 +1429,7 @@ class RealtimeClientConnection: QuickSpec {
                     expect(lastSerial).to(equal(4))
 
                     options.recover = client.connection.recoveryKey
+                    client.onError(AblyTests.newErrorProtocolMessage())
```
mattheworiordan: @SimonWoolf @ricardopereira I assume you two have discussed this.  Shouldn't a connection that fails to resume just start a new connection?  The comment above indicates that we should put the connection into the Failed state, but that seems wrong given failed is terminal.

Source/ARTRealtime.m
```diff
@@ -641,8 +641,10 @@ - (void)onConnected:(ARTProtocolMessage *)message {
             [self.connection setMaxMessageSize:message.connectionDetails.maxMessageSize];
             if (!_resuming) {
                 [self.connection setSerial:message.connectionSerial];
-                [self.logger debug:@"RT:%p msgSerial of connection \"%@\" has been reset", self, self.connection.id_nosync];
-                self.msgSerial = 0;
+                if (message.error != nil || self.options.recover == nil || [[self.options.recover stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]] isEqualToString:@""]) {
+                    [self.logger debug:@"RT:%p msgSerial of connection \"%@\" has been reset", self, self.connection.id_nosync];
+                    self.msgSerial = 0;
```
mattheworiordan: I get the recover bit, but why the error part i.e. if there is an error, we don't reset the message serial.  I don't really know what that is to be honest.  @ricardopereira @SimonWoolf?

SimonWoolf: I get that you're trying to enumerate the scenarios in which you believe there should be a new connection per RTN15c3. But it's overcomplicated and possibly error-prone compared to just checking whether the `connectionId` is new (ie not the same as a previous connectionId, or there was no previous connectionId). See [the logic in ably-js](https://github.com/ably/ably-js/blob/33f4b69a50ca7b1dc25241ac7780908fdd2ed78d/common/lib/transport/connectionmanager.js#L736-L740).

Source/ARTRealtime.m
```diff
@@ -641,8 +641,10 @@ - (void)onConnected:(ARTProtocolMessage *)message {
             [self.connection setMaxMessageSize:message.connectionDetails.maxMessageSize];
             if (!_resuming) {
                 [self.connection setSerial:message.connectionSerial];
```
SimonWoolf: Why are you only setting the connectionSerial if not resuming? RTN10a isn't conditional

Spec/RealtimeClientConnection.swift
SimonWoolf: @mattheworiordan the test constructs a new, second client library instance, which recovers from the first. It is not valid to recover from a connected, active client library instance -- when the new one connects the old gets disconnected as it's been superseded, so it tries to reconnect and steals the connection back from the new, and so on. The two client lib instances fight over the connection, trading it back and forth.

SimonWoolf:
> RTN10c does not appear to exist at https://docs.ably.io/client-lib-development-guide/features

@mattheworiordan RTN10c was removed by https://github.com/ably/docs/pull/585 as a duplicate of RTN15b2. 

@ricardopereira are you working from an old copy of the spec or something?

Spec/RealtimeClientConnection.swift
mattheworiordan: Sure, I get that.  But if a client has its connection disconnected because it's been taken over, it shouldn't become `FAILED`, it should lose continuity and start a new connection.  I don't agree that's a terminal condition.

mattheworiordan: Also, if this behaviour is expected in other libs, shouldn't this be in the spec?

SimonWoolf: > if a client has its connection disconnected because it's been taken over, it shouldn't become FAILED, it should lose continuity and start a new connection. I don't agree that's a terminal condition.

I don't follow your objection. Being superseded doesn't trigger it to become failed, it's becoming failed because Ricardo is explicitly calling onError to make it do so, as an easy way of making it disconnect all transports and stop talking to realtime. The point is to simulate the real-life situation when recover would be used (where the previous client lib instance no longer exists because iOS has killed it or whatever).

You [do the same thing in ably-ruby tests](https://github.com/ably/ably-ruby/blob/f7a8756a27e85e1a43c92a4d4632344a83fadaef/spec/acceptance/realtime/connection_spec.rb#L1223)
I [do a similar thing in ably-js tests (but force suspended rather than failed)](https://github.com/ably/ably-js/blob/33f4b69a50ca7b1dc25241ac7780908fdd2ed78d/spec/realtime/resume.test.js#L415).

> Also, if this behaviour is expected in other libs, shouldn't this be in the spec?

I'm not clear on what you're saying should be in the spec?

Source/ARTRealtime.m
ricardopereira: @SimonWoolf Yes, it's way simpler to just check the `connectionId`. Thanks, I will change the code.

Source/ARTRealtime.m
```diff
@@ -636,14 +636,18 @@ - (void)onConnected:(ARTProtocolMessage *)message {
 
     switch (self.connection.state_nosync) {
         case ARTRealtimeConnecting:
+            if (self.connection.id_nosync && ![message.connectionId isEqualToString:self.connection.id_nosync]) {
+                [self.logger debug:@"RT:%p msgSerial of connection \"%@\" has been reset", self, self.connection.id_nosync];
+                self.msgSerial = 0;
```
ricardopereira: @SimonWoolf Did the same as JS lib but [this](https://github.com/ably/ably-cocoa/blob/3b8c581fa4312a93ed3eae2c86aabc7223fa8eca/Spec/RealtimeClientConnection.swift#L3106) test is failing, so I suspect that I should have another condition to reset the msgSerial when the client is new (`this.connectionId` is `nil`) and the recover fails i.e. with _80008 - "Unable to recover connection"_, right?

Source/ARTRealtime.m
ricardopereira: Right, fixed 5b61635.

Source/ARTRealtime.m
ricardopereira: Fixed with 6334568 and 8cd4d50.

Source/ARTRealtime.m
ricardopereira: Fixed with 8bb7467.

SimonWoolf: lgtm

ricardopereira:
Had to cherry pick 09ef1fd.

Issue #852, opened 2019-05-22
Expose Auth.tokenDetails


paddybyers:
We're planning to expose a public way to discover the current token in use by the library - see https://github.com/ably/docs/pull/673.

How easy is it to implement this now for iOS?

ricardopereira:
@paddybyers Quite easy, that property already exist but it's hidden.

ricardopereira:
Sure, I'll add this in the next release. I'll prepare it today.

PR #853, opened 2019-05-23
RSA11

ricardopereira:
#852

### DIFF
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -4141,6 +4141,39 @@ class Auth : QuickSpec {
                 }
             }
         }
+
+        // RSA11
+        context("currentTokenDetails") {
+
+            // RSA11b
+            it("should hold a @TokenDetails@ instance in which only the @token@ attribute is populated with that token string") {
+                let token = getTestToken()
+                let rest = ARTRest(token: token)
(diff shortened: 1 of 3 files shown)


Spec/Auth.swift
```diff
@@ -4141,6 +4141,39 @@ class Auth : QuickSpec {
                 }
             }
         }
+
+        // RSA11
+        context("currentTokenDetails") {
```
mattheworiordan: this should be named `tokenDetails`

Spec/Auth.swift
```diff
@@ -4141,6 +4141,39 @@ class Auth : QuickSpec {
...
+                        done()
+                    }
+                }
+                expect(rest.auth.tokenDetails).to(equal(authenticatedTokenDetails))
+            }
+
+            // RSA11d
+            it("should be empty if there is no current token") {
+                let rest = ARTRest(options: AblyTests.commonAppSetup())
+                expect(rest.auth.tokenDetails).to(beNil())
```
mattheworiordan: @paddybyers do you not think we should encourage everyone to have a test that confirms a token that becomes invalid results in `auth.tokenDetails` being nil as well?

PR #861, opened 2019-06-13
Update Example

ricardopereira:


paddybyers:
So why are we using a fork of msgpack?

ricardopereira:
@paddybyers Because the original msgpack is missing the macOS target so I forked it. I created a PR but the owner ignored it https://github.com/rvi/msgpack-objective-C/pull/5.

ricardopereira:
Currently, this lib is using 3 forks:
 - [KSCrash](): crash/exceptions handler with our custom needs.
 - [SocketRocket](https://github.com/ably-forks/SocketRocket): removed warnings and added Carthage support.
 - [MsgPack](https://github.com/ably-forks/msgpack-objective-C): added macOS support.

ricardopereira:
@paddybyers Are you clarified?

PR #867, opened 2019-07-01
Push Device Update Registration: fix request authentication

ricardopereira:
- `deviceUpdateRegistration` method was setting a wrong Authorization header and was trying to read an updated `deviceIdentityToken` from the response body. Since that field doesn't exist, the local `deviceIdentityTokenDetails` was updated with a `token=nil`.

### DIFF
--- a/Spec/PushActivationStateMachine.swift
+++ b/Spec/PushActivationStateMachine.swift
@@ -638,7 +638,7 @@ class PushActivationStateMachine : QuickSpec {
                         expect(url.query).to(contain(rest.device.id))
                         expect(rest.device.identityTokenDetails).to(beNil())
                         expect(request.allHTTPHeaderFields?["Authorization"]).toNot(beNil())
-                        let deviceAuthorization = request.allHTTPHeaderFields?["X-Ably-DeviceIdentityToken"]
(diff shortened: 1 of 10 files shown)


ricardopereira:
I'll clean this mess and rebase.

Spec/PushActivationStateMachine.swift
```diff
@@ -634,8 +637,9 @@ class PushActivationStateMachine : QuickSpec {
                         expect(request.httpMethod) == "DELETE"
                         expect(url.query).to(contain(rest.device.id))
                         expect(rest.device.identityTokenDetails).to(beNil())
-                        let authorization = request.allHTTPHeaderFields?["X-Ably-DeviceIdentityToken"]
-                        expect(authorization).to(equal(testIdentityTokenDetails.token.base64Encoded()))
+                        expect(request.allHTTPHeaderFields?["Authorization"]).toNot(beNil())
+                        let deviceAuthorization = request.allHTTPHeaderFields?["X-Ably-DeviceIdentityToken"]
```
paddybyers: Still lots of references to this header. These should all be updated to `X-Ably-DeviceToken`. (That's not what the spec says, but that's what's implemented in the system, so we will update those as a separate piece of work.)

ricardopereira: Done 256c1e5. Thanks

Issue #866, opened 2019-07-01
ably-ios apparently interpreting the connectionStateTtl with incorrect units


SimonWoolf:
Customer reported that ably-ios wasn't going into the suspended state as expected.

Looking at ably-ios, noticed that:
1. It is passing the connectionDetails connectionStateTtl an NSTimeInterval constructor -- which [takes a number in seconds](https://developer.apple.com/documentation/foundation/nstimeinterval) -- to https://github.com/ably/ably-cocoa/blob/develop/Source/ARTConnectionDetails.m#L18 without doing any unit conversion
2. The tests are mocking a connected message with a connectionStateTtl of `60`: https://github.com/ably/ably-cocoa/blob/develop/Spec/TestUtilities.swift#L1023

Which makes it look like ably-ios is intrepreting that fields in seconds. Which is wrong, it's in milliseconds: The features spec IDL says that fields of type `Duration` are in milliseconds unless otherwise specified, not to mention that the value of the field is generally `120000`.

This is a bit worrying -- times and durations from the server are generally in milliseconds; are there any other places where ably-ios is wrongly interpreting them as fields in seconds?

mattheworiordan:
>  The features spec IDL says that fields of type Duration are in milliseconds unless otherwise specified, not to mention that the value of the field is generally 120000.

Something to bear in mind is that because times and dates are quite inconsistent across languages, we have where possible avoided these issues by using interval/time types when the language supports it natively and idiomatically.  So the attribute in the constructor being an NSTimeInterval is not an issue necessarily, however [iOS docs state](https://developer.apple.com/documentation/foundation/nstimeinterval) a "NSTimeInterval value is always specified in seconds; it yields sub-millisecond precision".

What makes things definitely wrong though is that the Ably documentation is misleading, see below:

![screenshot_2019-07-01_08-43-51_pm](https://user-images.githubusercontent.com/43789/60462530-74ece400-9c41-11e9-874b-fd163c818689.png)

Based on those docs, consumers of the API couldn't be blamed for thinking durations are in milliseconds.  

@ricardopereira can you check that we are consistent specifying all durations in seconds (well NSTimeInterval technically), and all READMEs etc. refer to durations in seconds too?  We'll then get an issue raised in the docs repo to avoid confusion around iOS durations.

mattheworiordan:
Oops, seems I missed what @SimonWoolf was saying, that this is about time/durations from Ably being misinterpreted...

[...]

PR #869, opened 2019-07-03
Fix milliseconds conversions

ricardopereira:
Related with #866

### DIFF
--- a/Spec/RealtimeClient.swift
+++ b/Spec/RealtimeClient.swift
@@ -361,6 +361,8 @@ class RealtimeClient: QuickSpec {
                 if let secs = endInterval {
                     expect(secs).to(beLessThanOrEqualTo(UInt(options.suspendedRetryTimeout)))
                 }
+
+                expect(client.connectionStateTtl).to(equal(120))
             }
 
             // RTC8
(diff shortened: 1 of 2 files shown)


Spec/RealtimeClient.swift
```diff
@@ -361,6 +361,8 @@ class RealtimeClient: QuickSpec {
                 if let secs = endInterval {
                     expect(secs).to(beLessThanOrEqualTo(UInt(options.suspendedRetryTimeout)))
                 }
+
+                expect(client.connectionStateTtl).to(equal(120))
```
paddybyers: Is there not a way to express this interval `120` with units?

ricardopereira: Yes, done ab79140

Spec/RealtimeClient.swift
```diff
@@ -361,6 +361,8 @@ class RealtimeClient: QuickSpec {
                 if let secs = endInterval {
                     expect(secs).to(beLessThanOrEqualTo(UInt(options.suspendedRetryTimeout)))
                 }
+
+                expect(client.connectionStateTtl).to(equal(120 as TimeInterval /*seconds*/))
```
mattheworiordan: Personally I think this test is a little brittle.  If we change our TTL to anything other than 120s, this test will fail.  Why not intercept the CONNECTED protocol message and make sure the `client.connectionStateTtl = connectionStateTtlFromProtocolMessage / 1000)`

PR #874, opened 2019-07-12
Update LocalDevice.clientId

ricardopereira:
#877


Source/ARTPushActivationStateMachine.m
```diff
@@ -226,6 +226,7 @@ - (void)deviceUpdateRegistration:(ARTErrorInfo *)error {
     NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[[NSURL URLWithString:@"/push/deviceRegistrations"] URLByAppendingPathComponent:local.id]];
     request.HTTPMethod = @"PATCH";
     request.HTTPBody = [[_rest defaultEncoder] encode:@{
+        @"clientId": local.clientId,
```
ricardopereira: @paddybyers Is this necessary for the new release?

ricardopereira: Removed that line because ClientId should be immutable after a successful device registration.

ricardopereira:
@paddybyers updated for the wildcard case.

paddybyers:
> updated for the wildcard case

LGTM, thanks - lets see what happens with the tests

PR #883, opened 2019-07-27
[WIP] Replace weak refs by explicit cycle breaking where possible.

tcard:
Throughout the codebase, we were liberally using weak references to
break all possible ref cycles, even those that are temporary and should
be broken explicitly. Other weak refs are there just for no good reason,
just to signify that we don't expect an object to lose all other
references to it before our reference is lost.

This by itself results in brittle code. The default should be strong,
which ensures a reference won't disappear before you expected. In other
cases, a more explicit handling of the reference lifecycle (e. g.
explicitly calling a "close" or "cancel" method) makes things clearer.

This PR removes all weak references except those where there's a stable
cycle and the ownership relationship between the objects in the cycle
is clear; for example, ARTRealtime clearly "owns" ARTRest, insofar their
lifetimes are entangled: ARTRealtime initializes ARTRest, and when the
ARTRealtime is released, so should be its ARTRest, which should have no
other references to it. All those cases have been documented.

In some places, we relied on weak references to implicitly invalidate
timers, enqueued operations and things like that. The pattern looks like
this:

    __weak MyClass *weakSelf = self;
    scheduleBlockToBeCalledLater(^{
        [weakSelf someMethod];
    });

The intent here is to cancel scheduled operations on the object if we
stop being interested in the object. And this "interest" is expressed
as holding a reference to the object somewhere.

This way we're forcing the user to hold on to their reference to (say)
ARTRealtime even if they're not going to use it anymore. That's not
what one would expect in a garbage-collected language, in which other
client libraries are written. In those, you would explicitly call
realtime.close if you want it to stop operating.

So this change makes this behavior uniform with other client libraries,
but now we're forced to explicitly unschedule the operations on close.

For this, a little helper class called ARTFlusher to register things
that should be "flushed" (cancelled, emptied, etc.) once we call "close"
or similar on the object they operate on. We flush this way event
listeners, timers, enqueued messages, etc. related to a realtime or
realtime channel instance.

### DIFF
--- a/Ably.xcodeproj/project.pbxproj
+++ b/Ably.xcodeproj/project.pbxproj
@@ -649,6 +649,12 @@
 		EB0505FC1C5BD7C4006BA7E2 /* ARTBaseMessage+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = EB0505FB1C5BD7C4006BA7E2 /* ARTBaseMessage+Private.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		EB1AE0CC1C5C1EB200D62250 /* ARTEventEmitter+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = EB1AE0CB1C5C1EB200D62250 /* ARTEventEmitter+Private.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		EB1AE0CE1C5C3A4900D62250 /* Utilities.swift in Sources */ = {isa = PBXBuildFile; fileRef = EB1AE0CD1C5C3A4900D62250 /* Utilities.swift */; };
+		EB1B53F122EC5354006A59AC /* ARTFlusher.h in Headers */ = {isa = PBXBuildFile; fileRef = EB1B53F022EC5354006A59AC /* ARTFlusher.h */; settings = {ATTRIBUTES = (Private, ); }; };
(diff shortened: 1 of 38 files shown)


tcard:
I'm closing this because `Realtime.close` isn't really the "cleanup everything" method I assumed it is. Event emitters, channels, etc. shouldn't be "flushed" on a close, they should be flushed once the Realtime object is really done with, but a Realtime can be reused after a close.

So if we shouldn't use `Realtime.close`, what should we use instead?

After #882, there will be a hard barrier between user-referenced objects and internally-referenced objects. In particular, if user code drop all references to a (Public)Realtime object, since internal code won't hold any references to it, it will be deallocated. This is in contrast to now, when internal code may be referencing the same object as user code, preventing a deallocation.

So we should leverage this by breaking the cycles when the (Public)Realtime object is deallocated, ie. call the equivalent of `[_flushOnClose flush]` on the `(Public)Realtime.dealloc` method, since the user dropping all references to Ably objects _is_ what means "cleanup everything" (except timers, which still need to be released with `Realtime.close`), as it does in client libraries for garbage-collected languages.

So this PR should be revived, if at all, after #882 is done.

paddybyers:
Yes, if this was C++ I'd be expecting to release resources (other than those associated with the connection itself) in `delete`.

tcard:
Actually, the whole flush thing is misguided. The flushed things that may hold reference cycles are either timed operations that will eventually execute or be cancelled, so it's OK for them to be strong, or indefinitely-lived event listeners that after #882 should go back to weak references, since either the user or the transport will prevent referenced objects (Rest, Realtime, or an owned child) to be released before the event listener that reference them is executed.

So this PR should then be adapted to just remove unnecessary weak refs, ie. weak refs that resolve no cycle at all or that resolve cycles that are solved in some other way anyway.

Issue #876, opened 2019-07-22
Implement push spec update https://github.com/ably/docs/pull/710


paddybyers:
There has been a spec update with a few push-related changes: https://github.com/ably/docs/pull/710

PR #886, opened 2019-08-02
Push spec update

ricardopereira:
#876

### DIFF
--- a/Spec/Push.swift
+++ b/Spec/Push.swift
@@ -76,6 +76,35 @@ class Push : QuickSpec {
                 }
             }
 
+            // RSH2d
+            it("should occur a GettingPushDeviceDetailsFailed event when the APNS registration fails") {
+                let stateMachine = rest.push.activationMachine()
+                let stateMachineDelegate = StateMachineDelegate()
+                stateMachine.delegate = stateMachineDelegate
+                defer {
+                    stateMachine.transitions = nil
+                    stateMachine.delegate = nil
+                }
+
+                let expectedError = ARTErrorInfo(domain: ARTAblyErrorDomain, code: 1234, userInfo: nil)
+
+                let hook = stateMachine.testSuite_injectIntoMethod(after: NSSelectorFromString("registerForRemoteNotifications")) {
+                    AblyTests.extraQueue.async(execute: {
+                        ARTPush.didFailToRegisterForRemoteNotificationsWithError(expectedError, rest: rest)
+                    })
+                }
+                defer { hook.remove() }
+
+                waitUntil(timeout: testTimeout) { done in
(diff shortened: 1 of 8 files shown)


Spec/Push.swift
```diff
@@ -76,6 +79,70 @@ class Push : QuickSpec {
...
+                            expect(event.error) == expectedError
+                            done()
+                        }
+                    }
+                    rest.push.activate()
+                }
+            }
+
+            // RSH2e
+            it("should verify the validity of the push device details") {
```
ricardopereira: @paddybyers I noticed that this isn't well implemented. The spec says **_Each time the library is instanced_** and I was doing the check when `activate` is called. So I did that change and it's not recommended because the instance needs to  access the `device` property in the initialiser. Some of the `device` properties (like the device secret) will be loaded from disk, more precisely from the iOS Keychain and that could be a problem. Since we can create an instance right before the app is running, the Keychain is not ready to be used and errors like `ERROR: Device Secret couldn't be updated (SecBase.h I/O error)` can happen. Maybe the best would be adding a `[rest/realtime].push.setup()` or `[rest/realtime].push.configure()` that should be called in the AppDelegate like other libraries do.

I.e. Firebase:

Screenshot 2019-08-07 at 11 30 32

tcard: @ricardopereira But wasn't `activate` the equivalent of that `configure` method you propose?

ricardopereira: @tcard Not quite because the user can call `activate()` whatever he/she wants and the `configure()` method was meant only for the initialisation of the app.

> paddy 2:11 PM
> There will be a further push spec change where we re-do the registration request each time activate() is called, even if already registered

Even so, this spec will change so this is in stand by.

ricardopereira:
New spec is being implemented in #974.

ricardopereira:
Actually, part of this PR is still valid. I just need to remove/update these specs:
 - (RSH2c) (Moved to (RSH8g)).
 - (RSH2d) (Moved to (RSH8h)).
 - (RSH2e) (Moved to (RSH8i)).

ricardopereira:
@tcard Could you update this PR regarding your changes on #974?

ricardopereira:
@QuintinWillison do you know if this still relevant?

tcard:
@ricardopereira Sure, will do. Should've done this before, really. I just forgot this PR existed... Also it claims to implement https://github.com/ably/docs/pull/710, which is different than what #974 implements (https://github.com/ably/docs/pull/786).

ricardopereira:
@tcard Ok, thanks a lot, since I'm a little confuse with what's still relevant or not because of the different changes that the Push spec suffered.

PR #890, opened 2019-09-12
Expose Push State Machine current, last and pending events

ricardopereira:


ricardopereira:
@paddybyers Sure, this was something mentioned by @Srushtika and you when we were working on the [iOS Push Example](https://github.com/ably/push-example-ios). The chat we had: https://ably-real-time.slack.com/archives/C030C5YLY/p1562948709355600

PR #891, opened 2019-09-12
Push should use internal logger and avoid system NSLog

### LINKED ISSUE #896: Push is using the system NSLog directly instead of the ARTLooger
I.e., currently users always see Push Activate State Machine logs because it's using NSLog directly instead of ARTLogger so ARTPush classes doesn't respect the log level when the user changes it, for example, to only see errors.

ricardopereira:
Fixes #896


ricardopereira:
Done, thanks!

ricardopereira:
@paddybyers Sorry, I can't merge since I merged with changes from `develop` because of some conflicts.

ricardopereira:
@paddybyers You missed this one. Thanks.

PR #893, opened 2019-09-12
Replace NSData.description to stringify device tokens correctly

### LINKED ISSUE #889: Push token: replace NSData.description usage
`-[NSData description]` is not a good way to stringify push tokens and the follow code will break in iOS 13:

https://github.com/ably/ably-cocoa/blob/15d3ca2a3acbf74f39952eea454477b7a4ef6eb1/Source/ARTPush.m#L121

Replace with something like:

```objc
const unsigned char *dataBuffer = data.bytes;
NSMutableString *hexString  = [NSMutableString stringWithCapacity:(dataLength * 2)];
for (int i = 0; i < dataLength; ++i) {
    [hexString appendFormat:@"%02x", dataBuffer[i]];
}

NSLog(@"%@", [hexString copy]);
```

More info about this: https://stackoverflow.com/a/57842886/3641812

ricardopereira:
Fix #889.

### DIFF
--- a/Source/ARTPush.m
+++ b/Source/ARTPush.m
@@ -116,20 +116,27 @@ static dispatch_once_t *activationMachine_once_token;
 }
 
 + (void)didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceTokenData restInternal:(ARTRestInternal *)rest {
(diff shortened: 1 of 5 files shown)


ricardopereira:
@paddybyers Yes, it's compatible with older versions.

Issue #690, opened 2018-02-12
when using authUrl with authMethod=POST, authParams are sent JSON-encoded, not form-encoded


SimonWoolf:
Customer [reports](https://app.intercom.io/a/apps/ua39m1ld/inbox/inbox/conversation/14584344238) that ably-ios doesn't conform to https://docs.ably.io/client-lib-development-guide/features/#RSA8c1b -- i.e. if using authUrl with some authParams and authMethod = POST it sends the post body JSON-encoded, not form-encoded. (It's a breaking change, so we'll have to consider whether to put off the fix for the next major version).

Incorrect test: https://github.com/ably/ably-ios/blob/4a22afe3da2af00d687748330d2ac45c23aaf5f1/Spec/Auth.swift#L1387

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-371) by [Unito](https://www.unito.io)

PR #899, opened 2019-09-27
Fix RSA8c1b

ricardopereira:
#690 

⚠️ Add an extra notice in the release description that includes this change because it's a breaking change for the lib users.


Source/ARTFormEncode.m
```diff
@@ -0,0 +1,66 @@
...
+    [allowedSet formUnionWithCharacterSet:[NSCharacterSet URLQueryAllowedCharacterSet]];
+    [allowedSet removeCharactersInString:@":/?&=;+!@#$()',*"];
+
+    return [[in description] stringByAddingPercentEncodingWithAllowedCharacters:allowedSet];
+}
+
+#define enckey(in) enc(in, @"[]")
+#define encval(in) enc(in, @"")
+
+static NSArray *DoQueryMagic(NSString *key, id value) {
```
paddybyers: Can you suggest a more descriptive name?

SimonWoolf:
> warning Add an extra notice in the release description that includes this change because it's a breaking change for the lib users.

Did we decide whether we wanted to make this change now or put it off till the 2.0 release, given it's a breaking change? If so, link to discussion?

ricardopereira:
@SimonWoolf We didn't decided yet if it should be part of the 2.0 or in the next 1.1 as a breaking change. What do you prefer?

Source/ARTFormEncode.m
ricardopereira: @paddybyers This code has been taken from another project, that's why I decided to leave it has it is in the original:

https://github.com/ably/ably-cocoa/blob/8efef941d36566207bf68a7ddd5f5f8b57781e86/Source/ARTFormEncode.m#L8-L9

Should I still change the name of the internal function?

SimonWoolf:
Semver would dictate that it should wait until 2.0. On the other hand we're trying to tie library versions to spec versions and 2.0 could be years away, during which time lots more people might start using the library, and run into issues where it posts params to their auth server as json when other libs use formencoding. So I'm conflicted. @paddybyers wdyt?

tcard:
Historically we've broken the API on minor releases when the API wasn't conforming to the spec in the first place. The library doesn't conform to semver; only the spec does.

(Not saying this is ideal.)

SimonWoolf:
As a compromise, we could merge as part of the upcoming [1.2 release](https://github.com/ably/docs/pull/766)? Still not semver compliant, but at least there's maybe a bit more chance people will read the release notes for 1.2 than for a 1.1 patch. @tcard / @paddybyers wdyt?

SimonWoolf:
> As a compromise, we could merge as part of the upcoming [1.2 release](https://github.com/ably/docs/pull/766)? Still not semver compliant, but at least there's maybe a bit more chance people will read the release notes for 1.2 than for a 1.1 patch.

@ricardopereira fyi I brought this up during the tech meeting and we agreed we should do this ^

cc @QuintinWillison

tcard:
Given how breaking of a change this is (clients will have to change their _server code_ because they upgraded _their iOS library_) for basically no gain at all for users, I'd say this should stay broken in all 1.x and wait until 2.0, where such a change would be more justified (and still, I'd say it will be worth keeping a "legacy compatibility" option so that you can upgrade your iOS app without changing your server code).

I, for one, would be pretty mad if a library I use did this.

As a temporary fix, we can introduce a library-specific `authBodyEncoding` option that lets new users opt-in to the form encoding and clearly document the situation in the README.

SimonWoolf:
> for basically no gain at all for users

I don't think that's true. Point is that it's rare for a customer to _only_ use ably-cocoa -- generally they'll have an android version of their app as well, and usually a web version, all using similar auth methods against their auth server. The reason this bug is annoying for customers is that if they use authmethod=POST (which to be clear is not the default), they have different libraries sending them auth requests with different content types.

Which means:

1. any customer using authMethod = POST probably already has an authServer that can cope with url-encoded auth requests, because that's what other client libraries are sending them

2. my guess is the number of customers actually doing that is small if not zero, because anyone who's complained we've told to just use authMethod = GET -- which is the default anyway -- and no-one's indicated they have a problem with that.

So the gain to customers of the change is that it becomes practical for them to actually start using authMethod = POST.

If you want to add a `uselegacyAuthBodyEncoding` option to ease compatibility then sure, I don't object to that. But I do think the default behaviour for 1.2 should be the one that matches the other client libraries, so that new customers can join and use authMethod=POST without running into the bug, so they're not weighed down by backwards-compatibility constraints. Having the sense be that way round is less good for existing customers using authMethod=POST, sure, but there aren't going to be many of those precisely because of this bug...

cc @QuintinWillison

tcard:
Good points.

Another thought to consider: why not do _both_? Try one of the two, and if we get a 4xx, try the other.

Maybe even issue an OPTIONS first and decide based on the `Accept` header.

SimonWoolf:
> Another thought to consider: why not do both? Try one of the two, and if we get a 4xx, try the other.

I see where you're coming from, but I really don't think that's a good idea. As a user, a library trying again with a different content-type on error would strike me as bizarre, surprising, hard-to-debug behaviour. The client library is.. occasionally sending me an auth request with a different content type than normal? What does that mean? Do I need to be able to parse and respond to both?

(re the other, OPTIONS responses aren't required to include an accept header, and afaict generally don't)

At least if auth breaks when you try ably-cocoa 1.2, it'll do so consistently, and then you can fix it, and it will consistently work again. (And as I said, I strongly suspect that there are almost no people for whom it will break).

ricardopereira:
I'm a little confuse.
@QuintinWillison What should I do with this? Any suggestion?

QuintinWillison:
@ricardopereira I was just noticing that this was an outstanding PR so hoping we can progress or close if it's no longer relevant. Or had you transferred it to @tcard, in which case assign to him please.

tcard:
As [per above](https://github.com/ably/ably-cocoa/pull/899#issuecomment-557496622), this is on hold until 1.2.

tcard:
Oh wait, spec 1.2 seems to be released already. Well, then I guess we should create a branch to adapt to 1.2 or something?

@QuintinWillison WDYT?

ricardopereira:
> As per above, this is on hold until 1.2.

Ok, then this is ready to be merged to 1.2 branch.

Source/ARTAuth.m
```diff
@@ -271,11 +272,11 @@ - (NSMutableURLRequest *)buildRequest:(ARTAuthOptions *)options withParams:(ARTT
     if ([options isMethodPOST]) {
         // TokenParams take precedence over any configured authParams when a name conflict occurs
         NSDictionary *unitedParams = [params toDictionaryWithUnion:options.authParams];
```
QuintinWillison: It seems that `toDictionaryWithUnion:` returns `<NSString *, NSString *> *` so could you consider:
1. Using that specific type for `unitedParams`
2. Replacing use of `description` on `id` pointers elsewhere (e.g. `enc` function) with `NSString *`

QuintinWillison:
Oops, @ricardopereira. Somehow I didn't spot that this had already been merged before I reviewed (was just working through notifications and hadn't noticed or scrolled to the bottom). Perhaps raise a new PR with these fixes when you get a chance. Or I do them if you are swamped.

ricardopereira:
@QuintinWillison Sorry, I thought that you did have a look at this. Since I saw the comment about merging this only to v1.2, I decided to merge it.

Could you handle your comments please? Thanks

QuintinWillison:
No problem, @ricardopereira - I will sort them. Thanks!

Issue #913, opened 2019-10-27
Realtime suspended connection retrial using wrong timeout


ricardopereira:
### Which version of the Ably SDK are you using? 
`v1.1.11 - Beta 1`

### On which platform does the issue happen? 
`iOS 13`

### Which version of Xcode are you using? 
Xcode 11.1

### What did you do?
I set my iPhone to flight mode and the connection moved to `DISCONNECTED` as expected but the connection retrial seems to not respect [RTN14e](https://docs.ably.io/client-lib-development-guide/features/#RTN14e).

> ([RTN14e](https://docs.ably.io/client-lib-development-guide/features/#RTN14e)) Once the connection state has been in the DISCONNECTED state for more than the default connectionStateTtl, the state will change to SUSPENDED and be emitted with the reason, and the Connection#errorReason will be updated. In this state, a new connection attempt will be made periodically as specified within suspendedRetryTimeout of ClientOptions.

The network connection is down: moves to `DISCONNECTED` and retries after 15s again and again until the `connectionStateTtl` is reached and then it moves to `SUSPENDED`. The connection retries after 30s (`SUSPENDED` -> `CONNECTING`) but fails and moves to `DISCONNECTED`. Then the library starts again with retrials of 15s (DISCONNECTED to CONNECTING).

### What did you expect to happen?

Moving from `DISCONNECTED` (15s) -> `CONNECTING` (fails) -> `DISCONNECTED` (15s) -> `CONNECTING` (fails) -> `DISCONNECTED` ... after TTL --> `SUSPENDED` (30s) -> `CONNECTING` (fails) -> `SUSPENDED` (30s) -> `CONNECTING` (fails) -> `SUSPENDED` (30s) ...

### What happened instead?

Moving from `DISCONNECTED` (15s) -> `CONNECTING` (fails) -> `DISCONNECTED` (15s) -> `CONNECTING` (fails) -> `DISCONNECTED` ... after TTL --> `SUSPENDED` (30s) -> `CONNECTING` (fails) -> `DISCONNECTED` (15s) -> `CONNECTING` (fails) -> `DISCONNECTED` (15s) -> `CONNECTING` (fails) -> `DISCONNECTED` ... after TTL --> `SUSPENDED` (30s).

PR #917, opened 2019-11-11
Fix connection SUSPENDED timeout

ricardopereira:
#913


Source/ARTRealtime.m
```diff
@@ -640,48 +643,48 @@ - (ARTEventListener *)transitionSideEffects:(ARTConnectionStateChange *)stateCha
...
-                }];
-            }
-
-            if ([[NSDate date] timeIntervalSinceDate:_startedReconnection] >= _connectionStateTtl) {
-                artDispatchScheduled(0, _rest.queue, ^{
+            if ([self isSuspendMode]) {
+                [self.logger verbose:@"RT:%p realtime connection has staled", self];
+                [_connectionRetryFromDisconnectedListener stopTimer];
+                _connectionRetryFromDisconnectedListener = nil;
+                stateChangeEventListener = [self unlessStateChangesBefore:0 do:^{
```
tcard: `0`? Why?

Source/ARTRealtime.m
```diff
@@ -404,6 +405,7 @@ - (void)_connect {
         // New connection
         _transport = nil;
     }
+    [self setSuspendTime];
```
tcard: Maybe it's because I have a pretty bad headache but I can't follow the logic here at all.

What's `suspendTime`? Why are we setting it to the moment the connection is established + `connectionStateTtl`?

Spec/RealtimeClientConnection.swift
```diff
@@ -2176,6 +2173,75 @@ class RealtimeClientConnection: QuickSpec {
...
+                        }
+                        client.connection.on(.suspended) { stateChange in
+                            guard let error = stateChange?.reason else {
+                                fail("SUSPENDED reason should not be nil"); done(); return
+                            }
+                            expect(error.message).to(contain("network is down"))
+                            expect(client.connection.errorReason).to(beIdenticalTo(error))
+                            partialDone()
+                        }
+                        client.simulateRestoreInternetConnection(after: 7.0)
```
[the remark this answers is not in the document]
ricardopereira: @tcard But that's something we already do on other tests and we missed the real issue. I caught it only when I was testing the library on my device.

Spec/RealtimeClientConnection.swift
```diff
@@ -2176,6 +2173,75 @@ class RealtimeClientConnection: QuickSpec {
...
+                            expect(stateChange?.reason).to(beNil())
+                            done()
+                        }
+                        client.connect()
+                    }
+
+                    waitUntil(timeout: testTimeout) { done in
+                        // Should move 6 times to the DISCONNECTED state
+                        //   3.0 connectionStateTtl / 0.5 disconnectedRetryTimeout = 6 times + 1 (SUSPENDED)
+                        let partialDone = AblyTests.splitDone(7, done: done)
```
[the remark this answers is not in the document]
ricardopereira: @tcard Ok, I like that.

Source/ARTRealtime.m
ricardopereira: The suspend time it's the current time + TTL, the time to considerate the connection is suspended. I assigned the suspend time in the moment that the connection starts.

Spec/RealtimeClientConnection.swift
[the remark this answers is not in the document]
tcard: OK, I guess it's worth the wait then.

Spec/TestUtilities.swift
```diff
@@ -1301,6 +1301,12 @@ extension ARTRealtime {
         }
     }
 
+    func overrideConnectionStateTTL(_ ttl: TimeInterval) {
+        self.internal.testSuite_injectIntoMethod(before: NSSelectorFromString("setSuspendTime")) {
```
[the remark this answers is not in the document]
ricardopereira: Done ae95fa5.

Source/ARTRealtime.m
```diff
@@ -640,48 +643,48 @@ - (ARTEventListener *)transitionSideEffects:(ARTConnectionStateChange *)stateCha
...
-                    if (change.current != ARTRealtimeDisconnected && change.current != ARTRealtimeConnecting) {
-                        self->_startedReconnection = nil;
-                    }
-                }];
-            }
-
-            if ([[NSDate date] timeIntervalSinceDate:_startedReconnection] >= _connectionStateTtl) {
-                artDispatchScheduled(0, _rest.queue, ^{
+            if ([self isSuspendMode]) {
+                [self.logger verbose:@"RT:%p realtime connection has staled", self];
```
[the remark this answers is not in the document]
ricardopereira: Agree  ae95fa5. Thanks.

Spec/RealtimeClientConnection.swift
```diff
@@ -2176,6 +2173,75 @@ class RealtimeClientConnection: QuickSpec {
...
+                        client.connection.on(.disconnected) { stateChange in
+                            partialDone()
+                        }
+                        client.connection.once(.suspended) { stateChange in
+                            client.connection.off()
+                            guard let error = stateChange?.reason else {
+                                fail("SUSPENDED reason should not be nil"); done(); return
```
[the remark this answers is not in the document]
ricardopereira: Just playing safe but I understand what you mean. I'll simplify it. Thanks

Source/ARTRealtime.m
ricardopereira: It wasn't a nice solution. Changed on 928051b.

Source/ARTRealtime.m
```diff
@@ -507,14 +509,15 @@ - (void)transition:(ARTRealtimeConnectionState)state {
 
 - (void)transition:(ARTRealtimeConnectionState)state withErrorInfo:(ARTErrorInfo *)errorInfo {
 ART_TRY_OR_MOVE_TO_FAILED_START(self) {
+    if (state == ARTRealtimeDisconnected && [self isSuspendMode]) {
+        state = ARTRealtimeSuspended;
```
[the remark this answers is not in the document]
ricardopereira: Agree 928051b. The suspend mode logic is simpler now. Thanks!

Spec/RealtimeClientConnection.swift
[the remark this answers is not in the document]
ricardopereira: Done 52fbc49. Thanks!

ricardopereira:
@tcard Can you please check my changes? Thanks

Source/ARTRealtime+Private.h
```diff
@@ -95,7 +95,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (readonly, getter=getClientOptions) ARTClientOptions *options;
 
 /// Suspend the behavior defined in RTN15a, that is trying to immediately reconnect after a disconnection
-@property (readwrite, assign, nonatomic) BOOL suspendImmediateReconnection;
+@property (readwrite, assign, nonatomic) BOOL disableImmediateReconnection;
```
tcard: Extreme nitpick: a verb makes me think of a method; a noun makes me think of a variable. So, `immediateReconnectionDisabled` (or `Enabled`).

Source/ARTRealtime.m
tcard: May I suggest then `connectedAt`, and then do `if (now > connectedAt + TTL)` to check if we're suspended? That's easier to understand for me anyway. `suspendTime` just tells you it's a time and has to do with suspend.

But anyway... This isn't how it works at all, is it? We're suspended if the time since the last _disconnection_ is greater than the connection state TTL.

I'd suggest an alternative solution: before moving to CONNECTING, store whether we are coming from DISCONNECTED or SUSPENDED and, on connection failure, move back to that state (the move from DISCONNECTED to SUSPENDED already works). As simple as that, we don't need to deal with more deadlines if I'm understanding things correctly.

tcard: On second thought, the new way of moving from DISCONNECTED to SUSPENDED is probably clearer that what we had before. So, maybe just fix this:

> But anyway... This isn't how it works at all, is it? We're suspended if the time since the last _disconnection_ is greater than the connection state TTL.

(So instead of `connectedAt` it could be `disconnectedAt` or `connectionLostAt` or something like that.)

Source/ARTRealtime+Private.h
QuintinWillison: I would prefer `Enabled` to `Disabled` in the name, as I think it makes it easier to understand the use.

Or, perhaps, use a prefix like `shouldImmediatelyReconnect`

Source/ARTRealtime.m
ricardopereira: @tcard I understand your point. I think `connectionLostAt` is a better name because I should only set the time on the first disconnection. I need some sort of logic to identify the connection lost, something like:

```
if (stateChange.previous == ARTRealtimeConnected || stateChange.current == ARTRealtimeDisconnected) {
    [self setConnectionLostTime];
}
```

Even so, I think it's simpler setting the expected suspension time right in the beginning like it's done right now. Maybe the naming is not the best, maybe `setExpectedSuspensionTime` instead of `setSuspendTime` would be better?

Source/ARTRealtime+Private.h
ricardopereira: I like that more: `shouldImmediatelyReconnect`. Thanks folks.

ricardopereira: Done 69e5cf0.

Source/ARTRealtime.m
ricardopereira: @tcard Does b600118 turn things more clear?

tcard: @ricardopereira 

> Even so, I think it's simpler setting the expected suspension time right in the beginning like it's done right now. 

My point is that this isn't how it should work. The suspension timer starts when the connection is _broken_, not when it's _established_. You're calling `setConnectionLostTime` when the connection is established, while it should be called when it's broken.

What you're doing now implies that, if we connect, then let more than 2 minutes (the connection state TTL) pass, then the connection is broken, we go straight to SUSPEND. What should happen is that we go to DISCONNECTED for 2 minutes, then to SUSPEND, now matter how long we were connected.

tcard: (But I'm OK with `setExpectedSuspensionTime` as a name otherwise. I still think storing `connectionLostAt` or similar is more clear, but that's a subjective thing.)

ricardopereira: @tcard Got it, please look at 26baa96. Thanks for your explanation and patience.

Source/ARTRealtime.m
```diff
@@ -405,7 +405,7 @@ - (void)_connect {
         // New connection
         _transport = nil;
     }
-    [self setExpectedSuspensionTime];
+    _connectionLostAt = nil;
```
tcard: This should be done after the connection is established, not before we try to connect. Otherwise, we're going to be DISCONNECTED for more than the connection state TTL.

Source/ARTRealtime.m
```diff
@@ -855,7 +859,7 @@ - (void)onConnected:(ARTProtocolMessage *)message {
     }
 
     _resuming = false;
-    [self setExpectedSuspensionTime];
+    _connectionLostAt = nil;
```
tcard: Why do this both here and on the CONNECTED transition side effects?

Source/ARTRealtime.m
ricardopereira: Ups this was a mistake. I did "Replace all". Fixed 53a2867.

Source/ARTRealtime.m
ricardopereira: Same as https://github.com/ably/ably-cocoa/pull/917#discussion_r369057510. Thanks!

PR #923, opened 2019-12-05
Revert "Should not force token auth when clientId is set"

tcard:
Build has possibly been broken by this, so reverting.

Reverts ably/ably-cocoa#898

PR #924, opened 2019-12-05
Revert "Fix Network Reachability: instance not found for target"

tcard:
Build has possibly been broken by this.

Reverts ably/ably-cocoa#910


tcard:
This seems not to be necessary, as #923's build, which includes the commit this PR reverts, passes.

PR #903, opened 2019-10-03
Release 1.1.12

ricardopereira:
# Main Changes

**Fixed bugs:**

 - Push token: replace `NSData.description` usage [\#889](https://github.com/ably/ably-cocoa/issues/889)

PR #942, opened 2019-12-09
Bring 1.1.11 and 1.1.12 changes in

QuintinWillison:
This should have been done when they were released, though it's not entirely clear to me why they were not. This needs to be done before we can work on 1.1.13 release.


tcard:
This is essentially the same as #903. I probably shouldn't have closed it.

Note that 3bc4b8d108893c20cc5638e4b8ad74c4c825ad1e was also included at https://github.com/ably/ably-cocoa/pull/893 which is already on `develop`. A PR like this shouldn't include a commit like that. I think we should remove it from here.

Issue #945, opened 2019-12-10
Agent should be ably-cocoa by default


mattheworiordan:
I have not checked the code for this lib, but I think we send `ably-ios` as the agent, yet this library is now `ably-cocoa`.  We should I believe do the following:

* [x] Change default agent to `ably-cocoa`
* [x] Include a variant based on platform such as `macos`, `ios`, `watchos`, `ipados`.  Not sure what the official variants are, but assuming it was ios it would be `ably-cocoa.ios`
* [x] Check for legacy references in code to `ably-ios` and replace with `ably-cocoa`
* [ ] Make sure we have no references elsewhere in docs or the website to `ably-ios`

tcard:
#954 does the parts of this that concern this repo.

PR #954, opened 2019-12-22
Update ably-ios references to ably-cocoa

tcard:
From #945:

> * [x] Change default agent to `ably-cocoa`
> * [x] Include a variant based on platform such as `macos`, `ios`, `watchos`, `ipados`.  Not sure what the official variants are, but assuming it was ios it would be `ably-cocoa.ios`
> * [x] Check for legacy references in code to `ably-ios` and replace with `ably-cocoa`


tcard:
@mattheworiordan Please confirm we can release this without breaking any internal machinery.

mattheworiordan:
@tcard I am not aware of any places this will break internal machinery.  As far as I am aware, we have no "hard coded" dependencies on the iOS agent string.  Probably worth @SimonWoolf and/or @paddybyers confirming for sure as I don't code any of the "internal machinery" as such.

tcard:
[Bitrise: ✅](https://app.bitrise.io/build/ed764139dc91a099#?tab=log)

PR #955, opened 2019-12-23
KSCrash fork has been renamed to KSCrashAblyFork

ricardopereira:


paddybyers:
I don't understand why we're reverting the version number change.

PR #957, opened 2019-12-23
Alleviate RTN5 flakiness

### LINKED ISSUE #934: Flaky test: RTN5 (basic operations should work simultaneously)
Pretty often, in Travis, we get a timeout:

```
[10:42:01]: ▸     ✗ Connection__basic_operations_should_work_simultaneously, Waited more than 10.0 seconds
```

This doesn't happen on my machine, and given this is a timeout on stressful test, we're probably just asking for too much. But we should double-check this isn't a deadlock or something.

tcard:
Now it receives 20% of the messages it used to receive and gets more time to do so.

Fixes #934, for now.

### DIFF
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -575,7 +575,8 @@ class RealtimeClientConnection: QuickSpec {
                 let options = AblyTests.commonAppSetup()
                 options.echoMessages = false
                 var disposable = [ARTRealtime]()
-                let max = 25
+                let numClients = 25
+                let numMessages = 50
                 let channelName = "chat"
 
                 defer {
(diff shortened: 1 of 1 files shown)


Spec/RealtimeClientConnection.swift
```diff
@@ -587,47 +588,47 @@ class RealtimeClientConnection: QuickSpec {
...
+                            messagesReceived += 1
+                            if messagesReceived == messagesExpected {
                                 done()
                             }
-                            sync.unlock()
                         }
-
-                        channel.publish(nil, data: "message_string", callback: nil)
+                        
+                        if messagesSent < numMessages {
```
ricardopereira: The `messagesSent` should be reset to 0 for each client, right?

ricardopereira: Never mind, I misread the code.

PR #959, opened 2019-12-23
Make RTN17 tests resilient to concurrent requests

### LINKED ISSUE #931: Flaky test: RTN17* (fallback hosts)
```
[10:56:12]: ▸     ✗ Connection__Host_Fallback__should_retry_hosts_in_random_order_after_checkin_if_an_internet_connection_is_available, expected to have Array<NSURL> with count 6, got 7
[10:56:12]: ▸     ✗ Connection__Host_Fallback__should_retry_hosts_in_random_order_after_checkin_if_an_internet_connection_is_available, expected to equal <[d.ably-realtime.com, e.ably-realtime.com, a.ably-realtime.com, c.ably-realtime.com, b.ably-realtime.com]>, got <[d.ably-realtime.com, a.ably-realtime.com, d.ably-realtime.com, c.ably-realtime.com, b.ably-realtime.com, e.ably-realtime.com]>
```

```
[09:38:12]: ▸     ✗ Connection__Host_Fallback__should_retry_custom_fallback_hosts_in_random_order_after_checkin_if_an_internet_connection_is_available, expected to have Array<NSURL> with count 6, got 7
[09:38:12]: ▸     ✗ Connection__Host_Fallback__should_retry_custom_fallback_hosts_in_random_order_after_checkin_if_an_internet_connection_is_available, expected to equal <[i.ably-realtime.com, j.ably-realtime.com, f.ably-realtime.com, h.ably-realtime.com, g.ably-realtime.com]>, got <[i.ably-realtime.com, f.ably-realtime.com, i.ably-realtime.com, h.ably-realtime.com, g.ably-realtime.com, j.ably-realtime.com]>
```

The test suite tries to arrange things so that the order is always [a given](https://github.com/ably/ably-cocoa/blob/a89a864cd9c6ec0c881eaa034ba7de2d947dadf0/Spec/RealtimeClientConnection.swift#L3143) 3, 4, 0, 2, 1, but then we see tests where it's sometimes 3, 0, 3, 2, 1, 4 (notice there's one more than expected).

I can't reproduce this on my machine, so the flakiness probably has to do with Travis limited system resources.

tcard:
In the test, we were checking that each fallback host got _exactly one_ request. But that's more than we need to check; we just need to check that all fallback hosts are used, in the fake-random order we set.

That extra strictness was causing the test to fail sometimes, as some request would sneak in during the test and duplicate a fallback host. This is because we first set each fallback host as `rest.prioritizedHost`, _then_ check if it works, so in between that, all requests are sent to that host.

I still don't know _why_ did some extra requests happened since I couldn't reproduce this on my machine and we have limited visibility into CI, but the point is, there's no reason why that should affect the test.

Fixes #931.

### DIFF
--- a/Spec/RealtimeClientConnection.swift
+++ b/Spec/RealtimeClientConnection.swift
@@ -3582,12 +3582,13 @@ class RealtimeClientConnection: QuickSpec {
                     }
 
                     expect(NSRegularExpression.match(testHttpExecutor.requests[0].url!.absoluteString, pattern: "//internet-up.ably-realtime.com/is-the-internet-up.txt")).to(beTrue())
-                    expect(urlConnections).to(haveCount(6)) // default + 5 fallbacks
 
                     let extractHostname = { (url: NSURL) in
                         NSRegularExpression.extract(url.absoluteString, pattern: "[a-e].ably-realtime.com")
                     }
-                    let resultFallbackHosts = urlConnections.compactMap(extractHostname)
+                    // While connecting to a fallback host, concurrent requests will be sent to that
+                    // host too, so deduplicate those.
+                    let resultFallbackHosts = urlConnections.compactMap(extractHostname).deduplicateContiguous()
                     let expectedFallbackHosts = Array(expectedHostOrder.map({ ARTDefault.fallbackHosts()[$0] }))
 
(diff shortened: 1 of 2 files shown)


tcard:
Superseded by #961.

PR #963, opened 2020-01-02
Simplify hooks on RTP2f test

### LINKED ISSUE #938: Flaky test: RTP2f (incoming LEAVE while SYNCing)
Happening sometimes in Travis:

```
[09:52:28]: ▸     ✗ Presence__PresenceMap__if_a_SYNC_is_in_progress__then_when_a_presence_message_with_an_action_of_LEAVE_arrives__it_should_be_stored_in_the_presence_map_with_the_action_set_to_ABSENT, expected to have Dictionary<String, ARTPresenceMessage> with count 1, got 0
```

tcard:
Now we have a single hook, directly on the presence map add method, so
that we can't possibly miss an ABSENT message.

Also, attach has been moved to after the hook and event subscription are
set, so that we can't miss those events.

May fix #938.

### DIFF
--- a/Source/ARTPresenceMap.h
+++ b/Source/ARTPresenceMap.h
@@ -55,6 +55,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)onceSyncFails:(void (^)(ARTErrorInfo *))callback;
 
 - (void)internalAdd:(ARTPresenceMessage *)message;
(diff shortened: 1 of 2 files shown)


tcard:
No occurrences of #938 on 8 CI runs; I'm restarting the builds to double-check.

Issue #966, opened 2020-01-11
Non-persistent push state machine states conflict with RSH3


tcard:
See https://github.com/ably/ably-java/issues/546 for details.

From there:

> (This happens on Cocoa already.)

What happens here is that, when we get a device token, we 1. persist it and 2. emit a GotPushDeviceDetails event that provokes a transition from WaitingForPushDeviceDetails to WaitingForDeviceRegistration.

If, while at that state, the app is killed, we're back at WaitingForPushDeviceDetails, but even if we get a device token again, we have code that checks that it isn't the same token as the one persisted so that we don't emit a GotPushDeviceDetails again for the same token.

This would be correct if, once on WaitingForDeviceRegistration, we didn't "irregularly" go back to WaitingForPushDeviceDetails, but we do, because WaitingForDeviceRegistration is not persistent.

maratal:
Correcting invalid state during initialisation seems pretty legit to me, WDYT @QuintinWillison @ben-xD @lukasz-szyszkowski   Shouldn't #967 close this issue?

ben-xD:
> Correcting invalid state during initialisation seems pretty legit to me, WDYT @QuintinWillison @ben-xD @lukasz-szyszkowski Shouldn't #967 close this issue?

The comment in https://github.com/ably/ably-cocoa/pull/967 says
> This also serves as a workaround, so that the "proper" fix for #966 is
not urgent.

I don't if additional work has been done, but just seeing https://github.com/ably/ably-cocoa/pull/967 and https://github.com/ably/ably-cocoa/issues/966, I don't think we can say its finished.

QuintinWillison:
The opening comment on this issue doesn't describe the problem well enough for me, nor does it suggest a fix. I'm finding this issue very difficult to parse and understand.

lukasz-szyszkowski:
same here, it needs deeper diving into code - can't tell you if #967 should close it since it's marked as a workaround

jamienewcomb:
@maratal can you make a decision if we can close this issue please

maratal:
To better understand what Tony was trying to say you should replace event names to something more obvious:

So, his

> What happens here is that, when we get a device token, we 1. persist it and 2. emit a `GotPushDeviceDetails` event that provokes a transition from `WaitingForPushDeviceDetails` to `WaitingForDeviceRegistration`.
> 
[...]

PR #972, opened 2020-01-13
Don't have push state machine singleton

tcard:
Instead, just cache it in the Rest instance.

Otherwise, if the user stops using the first instance and activates push
through a new instance, push activation will still use the first
instance an things won't work as expected.

Unfortunately, this now means that it's possible to have more than one
state machines per process, which is bad; RSH3 mandates that the state
machine must be a single app-wide concurrent process.

Once #966 is fixed, this drawback may be lifted. For now, is has been
documented in the README.

### DIFF
--- a/Source/ARTPush.m
+++ b/Source/ARTPush.m
@@ -81,6 +81,7 @@ NSString *const ARTDeviceTokenKey = @"ARTDeviceToken";
 @implementation ARTPushInternal {
     ARTRestInternal *_rest;
     ARTLog *_logger;
+    ARTPushActivationStateMachine *_activationMachine;
 }
 
 - (instancetype)init:(ARTRestInternal *)rest {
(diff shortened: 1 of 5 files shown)


Source/ARTPush.m
```diff
@@ -98,14 +100,8 @@ - (dispatch_queue_t)queue {
...
 
-// Store address of once_token to access it in debug function.
-static dispatch_once_t *activationMachine_once_token;
-
 - (ARTPushActivationStateMachine *)activationMachine {
-    static dispatch_once_t once;
-    activationMachine_once_token = &once;
-    static id activationMachineInstance;
-    dispatch_once(&once, ^{
+    if (_activationMachine == nil) {
```
QuintinWillison: Should we concerned about the potential that this method might be called from multiple threads?

tcard: This is an internal method which is only called from the library's GCD queue.

Generally, the library is architected as a public facade with an "internal" counterpart. The public facade wraps calls to the internal counterpart in GCD dispatches to a single, serial queue.

tcard:
PR kept open pending feedback from client.

PR #975, opened 2020-01-19
Remove redundant direct calls to push delegate callbacks

tcard:
We already do those properly through the state machine transitions' side effects, as per the spec.

For example, on `GotDeviceRegistration`:

https://github.com/ably/ably-cocoa/blob/6a7ef16e5be084e6ea9155a7cee6603aff035fa2/Source/ARTPushActivationState.m#L147

That calls:

https://github.com/ably/ably-cocoa/blob/6a7ef16e5be084e6ea9155a7cee6603aff035fa2/Source/ARTPushActivationStateMachine.m#L309-L321

Also, this was calling `didActivateAblyPush` after a registration update, which isn't spec-compliant. That's what the `updateFailedCallback` is for.

Also, we weren't calling those methods from the user queue.


tcard:
Before merging this, I want to make sure this works in a more end-to-end activation test.

PR #976, opened 2020-01-24
Channel options (per 1.2 spec)

ricardopereira:
- (RSL7) Channel#setOptions takes a ChannelOptions object and sets or updates the stored channel options, then indicates success

 - (RTL16) Channel#setOptions takes a ChannelOptions object and sets or updates the stored channel options.
    - (RTL16a) If the user has provided either ChannelOptions.params or ChannelOptions.modes and the channel is in either the attached or attaching state, Channel#setOptions sends an ATTACH message to the server with the params & modes encoded per RTL4, and indicates success once the server has replied with an ATTACHED (or indicates failure if the channel becomes detached or failed before that happens, as with Channel#attach); else it indicates success immediately

### DIFF
--- a/Source/ARTRealtimeChannels.m
+++ b/Source/ARTRealtimeChannels.m
@@ -34,8 +34,8 @@
     return [[ARTRealtimeChannel alloc] initWithInternal:[_internal get:(NSString *)name] queuedDealloc:_dealloc];
 }
 
-- (ARTRealtimeChannel *)get:(NSString *)name options:(ARTChannelOptions *)options {
-    return [[ARTRealtimeChannel alloc] initWithInternal:[_internal get:(NSString *)name options:(ARTChannelOptions *)options] queuedDealloc:_dealloc];
+- (ARTRealtimeChannel *)get:(NSString *)name options:(ARTRealtimeChannelOptions *)options {
+    return [[ARTRealtimeChannel alloc] initWithInternal:[_internal get:(NSString *)name options:options] queuedDealloc:_dealloc];
 }
 
 - (void)release:(NSString *)name callback:(nullable void (^)(ARTErrorInfo *_Nullable))errorInfo {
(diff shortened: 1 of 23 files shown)


ricardopereira:
This isn't ready! The idea was to create this PR as a Draft but I hit the wrong button 🤦‍♂️

ricardopereira:
Rebased: moved the commits since `develop` to `integration/1.2`.

ricardopereira:
@QuintinWillison Ready to be reviewed. Please assign the review to the right and available person. Ty

QuintinWillison:
@SimonWoolf, if there's any chance we can have your input on this one then it would be greatly appreciated. I'll check it over from an Objective-C correctness standpoint so could you please just let @ricardopereira know if there's anything obviously wrong in terms of the overarching story. Thanks!

Source/ARTRealtimeChannels.m
```diff
@@ -34,8 +34,8 @@ - (ARTRealtimeChannel *)get:(NSString *)name {
     return [[ARTRealtimeChannel alloc] initWithInternal:[_internal get:(NSString *)name] queuedDealloc:_dealloc];
 }
 
-- (ARTRealtimeChannel *)get:(NSString *)name options:(ARTChannelOptions *)options {
-    return [[ARTRealtimeChannel alloc] initWithInternal:[_internal get:(NSString *)name options:(ARTChannelOptions *)options] queuedDealloc:_dealloc];
+- (ARTRealtimeChannel *)get:(NSString *)name options:(ARTRealtimeChannelOptions *)options {
+    return [[ARTRealtimeChannel alloc] initWithInternal:[_internal get:(NSString *)name options:options] queuedDealloc:_dealloc];
```
SimonWoolf: Can't see any attempt at RTS3c1, but might be missing something

ricardopereira: I'll do another PR for that. Thanks

PR #986, opened 2020-02-13
Fix handling of token error when first connecting.

### LINKED ISSUE #984: Auth token refresh misses when macOS is sleeping
`1.1.18`

`macOS 10.14.6`

### Are you using Carthage? 
`0.34.0`

```
Xcode 11.3
Build version 11C29
```

### What did you do?

`authCallback` is not properly re-fetching TokenRequests when the computer is sleeping during the TTL expiration. Reproduction steps:

1. Create TokenRequests returned from servers with TTL of ~2 minutes
2. Set up client with auth options using the callback to get new TokenRequests

```swift
        let options = ARTClientOptions()
        options.authCallback = { tokenParams, complete in
            self.apiService.authenticateAbly { (response, error) in
                if let response = response {
                    do {
                        let parsed = try ARTTokenRequest.fromJson(response.tokenRequest as ARTJsonCompatible)
                        complete(parsed, nil)
                    } catch let error as NSError {
                        complete(nil, error)
                    }
                } else {
                    complete(nil, error)
                }
            }
        }
        options.clientId = "\(authInfo.id)"

        let ably = ARTRealtime(options: options)
```

3. Put your computer to sleep and wait for >2min
4. Watch as the client enters an infinite disconnect -> suspend loop and never tries to refetch TokenRequests properly.

**Note**: if you don't sleep the computer it refetches the tokens fine!

Client to properly reconnect once computer wakes from sleep.

### What happened instead?

It never fetches the new token and enters this infinite loop of being unable to reconnect.

Edit: My guess is this issue is very related to [this one](https://github.com/ably/ably-cocoa/issues/881)

tcard:
Removed complex and apparently unnecessary logic on error protocol message handler. No tests broken by this.

I first thought this was an issue with RSA4b, so I also added a missing test for it and some documentation.

Fixes #984.

### DIFF
--- a/Source/ARTAuth.m
+++ b/Source/ARTAuth.m
@@ -306,7 +306,13 @@ ART_TRY_OR_REPORT_CRASH_START(_rest) {
         if (self.tokenDetails.expires == nil) {
             return YES;
         }
-        else if (![self hasTimeOffset] || [self.tokenDetails.expires timeIntervalSinceDate:[self currentDate]] > 0) {
+        
+        // RSA4b1: Only check expiry client-side if local clock has been adjusted.
+        // If it hasn't, assume the token remains valid.
+        if (![self hasTimeOffset]) {
+            return YES;
+        }
+        if ([self.tokenDetails.expires timeIntervalSinceDate:[self currentDate]] > 0) {
             return YES;
         }
     }
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -920,14 +920,6 @@ ART_TRY_OR_MOVE_TO_FAILED_START(self) {
         [self onChannelMessage:message];
     } else {
         ARTErrorInfo *error = message.error;
-        if (error.code >= 40140 && error.code < 40150) {
-            if (![self shouldRenewToken:&error]) {
-                [self transition:ARTRealtimeFailed withErrorInfo:message.error];
-                return;
-            }
-            [self transitionToDisconnectedOrSuspendedWithError:message.error];
-            return;
-        }
         if ([self shouldRenewToken:&error]) {
             [self transportReconnectWithRenewedToken];
             return;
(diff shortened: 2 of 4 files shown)


tcard:
[Bitrise: ❌](https://app.bitrise.io/build/1f26ab5038ad073c#?tab=log)

Well, it looks like one does not simply remove a bunch of lines.

> No tests broken by this.

This is so false I don't know how I reached that conclusion originally. I've fixed a bunch of them, but for some reason there are tests for unrelated things such as presence, etc. that do seem broken by this (since they pass on master), I don't know how yet.

SpencerCDixon:
Ouch :( Is this still on yalls radar to resolve 🤞?

tcard:
@SpencerCDixon Yes, of course.

SpencerCDixon:
Hey, @tcard sorry for being annoying but is there any chance you have a timeline on when this might get fixed and merged in? 

I have a few weeks of work porting over to Ably that is being blocked by this. We'd like to start beta testing the integration but if Ably clients can't get new tokens when folks computers are asleep that's a nonstarter. Most of our users keep our app running 24/7 and their computers are likely asleep overnight and would miss refresh token windows.

Assuming our initial testing goes well we're going to be using your service with thousands of users and paying yall a lot of money 😄

Again, sorry to be annoying, I know you probably have a LOT going on just trying to get a handle on what I'll be working on next week and if I should be considering forking.

tcard:
@SpencerCDixon I didn't realize this was such a blocker for your development, since it's _almost_ ready to be released. We'll prioritize getting this released as soon as possible.

In any case, note that a workaround is possible by making a new connection each time the computer wakes up, passing the `recoveryKey` from the old connection in the new connection's `ARTClientOptions.recover`. That will recover the old connection's state if less that 2 minutes passed. But I think you can expect this to be released by tomorrow, so that won't be necessary. (But maybe it's good to have that trick up your sleeve anyway, in case you want to dispose of your ARTRealtime instance for some reason without losing state!)

tcard:
After deeper review, it seems this change is fine after all. Test failures were coincidentally caused by the flakiness of the test suite.

SpencerCDixon:
Awesome, thanks so much!

PR #987, opened 2020-02-16
Revert "Update KSCrash dependency"

tcard:
Reverts ably/ably-cocoa#981

@QuintinWillison This broke the build:

```
❌  /Users/vagrant/git/Source/ARTSentry.m:19:9: 'KSCrashAblyFork/NSData+GZip.h' file not found
#import <KSCrashAblyFork/NSData+GZip.h>
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I'm reverting it pending a fix.

PR #988, opened 2020-02-20
Second Attempt to Update KSCrash dependency

QuintinWillison:
I didn't test #981 as I naively assumed it would be a simple fix. I was wrong.

If this fixes then we can close @tcard's #987.


QuintinWillison:
I am leaving [the branch](https://github.com/ably/ably-cocoa/tree/feature/rename-kscrash-categories) in place for the moment as we may have a customer testing from it.

PR #1000, opened 2020-03-05
Avoid leak from user incorrectly holding to authCallback's callback.

tcard:
This is a corner case, but users may encounter it.

### DIFF
--- a/Source/ARTAuth.m
+++ b/Source/ARTAuth.m
@@ -402,11 +402,14 @@ ART_TRY_OR_REPORT_CRASH_START(self->_rest) {
             // and release any references it may hold to Rest/Realtime. Even if the user holds on to
             // the callback we give to them, they would be then holding to a now-nil reference to the
             // original callback.
-            __block void (^cancellableCallback)(ARTTokenDetails *, NSError *) = callback;
-
-            task = artCancellableFromCallback(^{
-                cancellableCallback = nil;
-            });
+            __weak void (^safeCallback)(id<ARTTokenDetailsCompatible> tokenDetailsCompat, NSError * error);
+            task = artCancellableFromCallback(^(const id<ARTTokenDetailsCompatible> tokenDetailsCompat, NSError *const error) {
+                if (error) {
+                    callback(nil, error);
+                } else {
+                    [tokenDetailsCompat toTokenDetails:[self toAuth] callback:callback];
+                }
+            }, &safeCallback);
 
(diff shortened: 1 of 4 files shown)


QuintinWillison:
I'm having a look locally and may just end up pushing a couple of tweaks to save time.

Source/ARTAuth.m
```diff
@@ -391,6 +391,26 @@ - (void)requestToken:(ARTTokenParams *)tokenParams withOptions:(ARTAuthOptions *
...
+            // to Rest/Realtime.
+            //
+            // So we're going to apply the Fundamental Theorem of Software Engineering here and wrap the
+            // callback in another reference. Then, we pass a callback that calls this reference,
+            // and not the original callback. Then, we return a task that, if cancelled, sets that
+            // intermediate reference to nil, so that we lose our reference to the original callback
+            // and release any references it may hold to Rest/Realtime. Even if the user holds on to
+            // the callback we give to them, they would be then holding to a now-nil reference to the
+            // original callback.
+            __weak void (^safeCallback)(id<ARTTokenDetailsCompatible> tokenDetailsCompat, NSError * error);
```
tcard: Why is this weak? Which reference cycle is this breaking?

Source/ARTTypes.m
```diff
@@ -308,3 +308,74 @@ + (NSDate *)dateWithMillisecondsSince1970:(uint64_t)msecs {
...
+    }
+    self = [super init];
+    if (!self) {
+        return nil;
+    }
+    
+    _lock = [NSObject new];
+    _callback = callback;
+    
+    __weak ARTCancellableFromCallback * _cancellable = self;
```
tcard: Same, why is this weak?

Source/ARTAuth.m
QuintinWillison: Just being defensive as I don't think the current scope has any business in 'owning' that callback. On other words, erring on the side of caution, in which case my question in return would be, why _should_ it be strong? It feels like a delegation pattern.

Source/ARTTypes.m
QuintinWillison: This one is more important and I believe justified. We don't want `_wrapper` retaining a strong reference to `self`, given it's a strong ivar on `self` itself. Unless I missed something?

QuintinWillison:
@tcard is there any chance my responses satisfied you and you're happy to resolve the two conversations above so I can get this merged?

tcard:
@QuintinWillison Sorry, I got drowned in other things. LGTM

Issue #1001, opened 2020-03-11
request_id in REST requests


QuintinWillison:
See [1.2 spec](https://github.com/ably/docs/pull/796) `RSC7c`/`TO3p` (`addRequestIds` on `ClientOptions` generating `request_id` query param in REST calls)  and `TI1` (`requestId` on `ErrorInfo`).

This is helpful for debugging certain types of customer issue.

PR #1002, opened 2020-03-12
REST Request IDs

ricardopereira:
#1001

### DIFF
--- a/Source/ARTRestChannel.m
+++ b/Source/ARTRestChannel.m
@@ -327,7 +327,7 @@ ART_TRY_OR_REPORT_CRASH_START(self->_rest) {
         [request setValue:self.rest.defaultEncoding forHTTPHeaderField:@"Content-Type"];
     }
 
-    [self.logger debug:__FILE__ line:__LINE__ message:@"RS:%p C:%p (%@) post message %@", self->_rest, self, self.name, [[NSString alloc] initWithData:encodedMessage encoding:NSUTF8StringEncoding]];
+    [self.logger debug:__FILE__ line:__LINE__ message:@"RS:%p C:%p (ID:%@) (%@) post message %@", self->_rest, self, requestId, self.name, [[NSString alloc] initWithData:encodedMessage encoding:NSUTF8StringEncoding]];
     [self->_rest executeRequest:request withAuthOption:ARTAuthenticationOn completion:^(NSHTTPURLResponse *response, NSData *data, NSError *error) {
         if (callback) {
             ARTErrorInfo *errorInfo = error ? [ARTErrorInfo createFromNSError:error] : nil;
--- a/Source/ARTRestChannel.m
+++ b/Source/ARTRestChannel.m
@@ -21,6 +21,7 @@
 #import "ARTNSArray+ARTFunctional.h"
 #import "ARTPushChannel+Private.h"
 #import "ARTCrypto+Private.h"
+#import "ARTClientOptions.h"
 
 @implementation ARTRestChannel {
     ARTQueuedDealloc *_dealloc;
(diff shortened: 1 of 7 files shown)


jamienewcomb:
@ricardopereira is this ready for review

ricardopereira:
@jamienewcomb not yet. I'm need run the test suite on my machine to see what's failing.

jamienewcomb:
@ricardopereira Okay, are you looking to complete this over the next few days?

jamienewcomb:
Deleting my above comment. I will link to this PR in Jira

Spec/RestClient.swift
```diff
@@ -1698,6 +1698,85 @@ class RestClient: QuickSpec {
...
+                            guard let firstRequestId = extractURLQueryValue(testHTTPExecutor.requests.first?.url, key: "request_id") else {
+                                fail("First request attempt doesn't have the 'request_id'."); return
+                            }
+                            requestId = firstRequestId
+                            expect(error.message).to(contain(requestId))
+                            done()
+                        }
+                    }
+
+                    expect(fallbackRequests).to(allPass { extractURLQueryValue($0?.url, key: "request_id") == requestId })
```
QuintinWillison: Do you need to also ensure that `fallbackRequests` is not empty at this point? If it were I think this expectation would still succeed, which is misleading. Perhaps even go as far as to assert the count.

Spec/RestClient.swift
```diff
@@ -1698,6 +1698,85 @@ class RestClient: QuickSpec {
...
+
+                    mockHttpExecutor.reset()
+
+                    options.addRequestIds = false
+                    let restB = ARTRest(options: options)
+                    restB.internal.httpExecutor = mockHttpExecutor
+                    waitUntil(timeout: testTimeout) { done in
+                        restB.channels.get("foo").publish(nil, data: "something") { error in
+                            expect(error).to(beNil())
+                            guard let url = mockHttpExecutor.requests.first?.url else {
```
QuintinWillison: Given you're reliant on the implementation of the `mockHttpExecutor.reset()` call above, I would perhaps assert at this point that there is only one request.

Source/ARTRestChannel.m
```diff
@@ -305,19 +306,37 @@ - (void)internalPostMessages:(id)data callback:(void (^)(ARTErrorInfo *__art_nul
             return;
         }
     }
-    
-    NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:[self->_basePath stringByAppendingPathComponent:@"messages"]]];
+
+    NSString *randomId = [NSUUID new].UUIDString;
```
QuintinWillison: Could you avoid generating here, only assigning `randomId` within the check for `options.addRequestIds` below? For those that don't want request ids, which for now will be most users, this may have a slight performance hit. Also, less 'new code' run for every customer upgrading their library.

Spec/RestClient.swift
ricardopereira: Nice catch! Fixed 2ecd0dc. Thanks

Spec/RestClient.swift
ricardopereira: Yap, makes sense. Done 33b1356.

Source/ARTRestChannel.m
ricardopereira: Ups, I forgot to remove that. Fixed 2f87540.

ricardopereira:
Arrg, I forgot to squash the fixes 😩

sacOO7:
@ricardopereira , correct me if I am wrong, request-id is different for different request right ( ofc if not retried to fallback host )

ricardopereira:
Hello @sacOO7. Yes, that's correct.

Issue #894, opened 2019-09-12
Remove queueing messages in a channel-level queue


ricardopereira:
Regarding https://github.com/ably/ably-cocoa/pull/795#discussion_r320167019

The `ARTRealtimeChannel` should be refactored. It's doing checks that are meant to be done by the connection, for example: https://github.com/ably/ably-cocoa/blob/f6f722badb6455f6b90d840c24305b556f32800d/Source/ARTRealtimeChannel.m#L332-L341

Other example, it keeps a list of queued messages that should be handled by the connection: https://github.com/ably/ably-cocoa/blob/f6f722badb6455f6b90d840c24305b556f32800d/Source/ARTRealtimeChannel.m#L66

Issue #1004, opened 2020-03-24
lib fails all the user's channels on transition to connecting/disconnected if queueMessages is disabled?


SimonWoolf:
https://github.com/ably/ably-cocoa/blob/f0e4018040bf0c4b014e2fb6907fb5d75016d8df/Source/ARTRealtime.m#L701-L739

This is in transitionSideEffects. The implemented logic appears to be that if sendEvents and queueEvents are both false, and it's not in the closing, closed, or suspended states, then it should set all the channels to failed. Perhaps it's assuming that the only remaining state that it can be in in that branch is the `failed` state; if so this is not a correct assumption because `queueMessages` can be disabled, which will make `shouldSendEvents` and `shouldQueueEvents` both return false in states like `connecting` and `disconnected`.

If I'm reading the code right, the effect of this is to make the queueMessages client option completely unusable, all subsequent operations on any channels that had already been accessed will then fail even once the lib has reconnected after the first time it goes into the connecting or disconnected states

QuintinWillison:
Internal Breadcrumbs: [Slack](https://ably-real-time.slack.com/archives/C030C5YLY/p1584973153034800), [Intercom](https://app.intercom.com/a/apps/ua39m1ld/inbox/inbox/conversation/26069109023)

Expected behaviours, as supplied by @SimonWoolf via Slack: [RTL3](https://docs.ably.io/client-lib-development-guide/features/#RTL3) and [TO3g](https://docs.ably.io/client-lib-development-guide/features/#TO3g).

tomczoink:
<div>Agent Tom Camp linked Freshdesk ticket <a href="https://ably.freshdesk.com/a/tickets/106205">106205</a> for this issue.</div>

PR #1005, opened 2020-03-26
Fix 'queueMessages' expected behaviour

ricardopereira:
#1004


Spec/RealtimeClientChannel.swift
```diff
@@ -496,6 +496,82 @@ class RealtimeClientChannel: QuickSpec {
...
+                        options.queueMessages = false
+                        options.autoConnect = false
+                        let client = ARTRealtime(options: options)
+                        defer { client.dispose(); client.close() }
+                        let channel = client.channels.get("foo")
+                        waitUntil(timeout: testTimeout) { done in
+                            expect(client.connection.state).to(equal(.initialized))
+                            channel.publish(nil, data: "message") { error in
+                                expect(error).toNot(beNil())
+                                done()
```
ricardopereira: @paddybyers The publish should fail if the connection is in the INITIALIZED state, right? [`TO3g`](https://docs.ably.io/client-lib-development-guide/features/#TO3g) only mentions about connecting but it contradicts with [`RTL6c2`](https://docs.ably.io/client-lib-development-guide/features/#RTL6c2).

ricardopereira: cc @QuintinWillison

QuintinWillison:
**Breadcrumbs**: [internal discussion](https://ably-real-time.slack.com/archives/C030C5YLY/p1585300078019400) around failing tests which may be blocking @ricardopereira from fully testing this, which may be why this PR is still in Draft state.

ricardopereira:
Yes, exactly. Thanks @QuintinWillison.

ricardopereira:
The CI is getting better but even so I had to restart some of them.

ricardopereira:
@QuintinWillison BTW, this fixes https://github.com/ably/ably-cocoa/issues/894 too 🙌

Source/ARTRealtime.m
```diff
@@ -699,42 +701,53 @@ - (ARTEventListener *)transitionSideEffects:(ARTConnectionStateChange *)stateCha
...
+                    [channel detachChannel:[ARTStatus state:ARTStateOk]];
+                    break;
+                case ARTRealtimeSuspended:
+                    [channel setSuspended:channelStatus];
+                    break;
+                case ARTRealtimeFailed:
+                    [channel setFailed:channelStatus];
+                    break;
+                default:
+                    break;
```
SimonWoolf, line 750: Why is all of this still gated under a condition that fails if `options.queueMessages` is false? This is [RTL3](https://sdk.ably.com/builds/ably/specification/main/features/#RTL3), right? That spec item doesn't mention anything about queueMessages, what do they have to do with each other?

maratal: See https://github.com/ably/ably-cocoa/pull/1941

PR #1009, opened 2020-04-05
Test suite: keep channel name prefix for current ClientOptions while calling 'setupOptions' and other improvements

ricardopereira:


Source/ARTTypes.m
```diff
@@ -363,7 +363,7 @@ -(void)invokeWithResult:(const id)result error:(NSError *const)error {
     }
     
     if (callback) {
-        _callback(result, error);
+        callback(result, error);
```
ricardopereira, line 366: Small but important fix.

QuintinWillison: 😬 .... 😌

PR #1018, opened 2020-05-16
Update message bundling regarding RTL6d

ricardopereira:
#979

### DIFF
--- a/Source/ARTBaseMessage.m
+++ b/Source/ARTBaseMessage.m
@@ -79,10 +79,18 @@
         }
         else if ([self.data isKindOfClass:[NSData class]]) {
             finalResult += [self.data length];
-        } else {
+        }
+        else {
             NSError *error = nil;
+            NSJSONWritingOptions options;
+            if (@available(iOS 13.0, *)) {
+                options = NSJSONWritingWithoutEscapingSlashes;
+            }
+            else {
+                options = 0; //no specific format
+            }
             NSData *jsonData = [NSJSONSerialization dataWithJSONObject:self.data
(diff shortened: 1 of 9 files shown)


Source/ARTBaseMessage.m
```diff
@@ -79,10 +79,18 @@ - (NSInteger)messageSize {
         }
         else if ([self.data isKindOfClass:[NSData class]]) {
             finalResult += [self.data length];
-        } else {
+        }
+        else {
             NSError *error = nil;
+            NSJSONWritingOptions options;
+            if (@available(iOS 13.0, *)) {
```
QuintinWillison, line 86: I thought I would load this up quickly in Xcode as I was curious as to whether it was acceptable to just make this runtime switch based on iOS platform detection alone, given Apple's [sparse documentation](https://developer.apple.com/documentation/foundation/nsjsonwritingoptions/nsjsonwritingwithoutescapingslashes?language=objc) on the new enumeration value you are exercising.

But what I then discovered, which perhaps I could have spotted just by looking at the code here, is that you appear to have some scope issues with your local variables. I'm seeing this in Xcode 11.3.1:

Screenshot 2020-05-18 at 09 41 32

Can you take a look, please.

ricardopereira: @QuintinWillison Nice catch! Thanks. Wondering why I didn't saw the warnings 🤔

ricardopereira: Now I know why I didn't see those warning: #1020 😄

PR #1026, opened 2020-06-09
Test suite: fix lib version

ricardopereira:
I noticed that all tests failed in master so saw the problem and fixed it.


QuintinWillison:
Thanks for jumping on that so quickly, @ricardopereira. 😁

Issue #939, opened 2019-12-08
Flaky test: EventEmitter timeout - Utilities__EventEmitter__the_timed_method__should_call_onTimeout_and_off_the_listener_if_the_deadline_is_reached


tcard:
Happening sometimes in Travis:

```
[09:52:28]: ▸     ✗ Utilities__EventEmitter__the_timed_method__should_call_onTimeout_and_off_the_listener_if_the_deadline_is_reached, expected to be true, got <false>
[09:52:28]: ▸     ✗ Utilities__EventEmitter__the_timed_method__should_call_onTimeout_and_off_the_listener_if_the_deadline_is_reached, expected to be nil, got <123>
```

PR #1027, opened 2020-06-19
Fix 939: improve EventEmitter tests

ricardopereira:
#939

### DIFF
--- a/Spec/Utilities.swift
+++ b/Spec/Utilities.swift
@@ -201,7 +201,6 @@ class Utilities: QuickSpec {
             }
 
             context("EventEmitter") {
-                
 
                 var eventEmitter = ARTInternalEventEmitter<NSString, AnyObject>(queue: AblyTests.queue)
                 var receivedFoo1: Int?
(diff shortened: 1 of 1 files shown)


Spec/Utilities.swift
```diff
@@ -201,7 +201,6 @@ class Utilities: QuickSpec {
             }
 
             context("EventEmitter") {
-                
 
                 var eventEmitter = ARTInternalEventEmitter<NSString, AnyObject>(queue: AblyTests.queue)
```
ricardopereira, line 205: The `EventEmitter` instance is using the `AblyTests.queue` so `delay` is not ideal because it uses the main queue.

QuintinWillison:
@ricardopereira do, please, merge this at your leisure. Thanks. 😁

ricardopereira:
@QuintinWillison I didn't saw your initial comment, sorry. There are maaaany things to improve and to refactor. For example, the TestUtilities file is a mess with a lot of small classes and types that could land to tis own file. Or 3rd party libs like the SwiftyJSON that could be replaced with the standard `JSONDecoder`, etc. I'll have those closely in mind for sure. Thanks

PR #1028, opened 2020-06-19
Fix some `waitUntil` error "expects its completion closure to be only called once"

ricardopereira:


Spec/RestClient.swift
```diff
@@ -1173,12 +1173,15 @@ class RestClient: QuickSpec {
                             testHTTPExecutor.http = MockHTTP(network: caseTest, logger: options.logHandler)
                             let channel = client.channels.get("test")
 
+                            let lock = NSLock()
                             testHTTPExecutor.afterRequest = { _, callback in
+                                lock.lock()
```
QuintinWillison, line 1178: I would prefer to see this acquisition constrained, so it would perhaps be healthier to consider a sensible timeout and fail the test if that's exceeded via [`lock(before:)`](https://developer.apple.com/documentation/foundation/nslock/1411133-lock).

ricardopereira:
I don't understand. After Paddy fixes I can have easily have a green in my machine 🤷‍♂️  

Screenshot 2020-07-24 at 11 30 33

Spec/RestClient.swift
ricardopereira: Done f6196d7. Thanks

QuintinWillison (approved):

PR #1056, opened 2020-07-01
Improve unhandled exception handling code

### LINKED ISSUE #1030: Data race in ARTstartHandlingUncaughtExceptions
`1.2.0`

iOS 13.5

### Are you using Cocoapods? 
`1.9.2`

Xcode 11.5
Build version 11E608c

### What did you do?
Running our app’s test target.

The test succeeds.

### What happened instead?
Xcode’s Thread Sanitizer trips on `ARTstartHandlingUncaughtExceptions()`. It seems that the read of `_handlingUncaughtExceptions` followed by the write to it can happen simultaneously on multiple threads.

image

QuintinWillison:
Fixes #1030.

I've not been able to reproduce the thread sanitizer data race that the customer has experienced at runtime, so I have taken the approach of inspecting this area of the codebase and improving what I see while trying to keep the scope of change as minimal as possible.

The diff's a bit messier than I would have liked as I've also fixed some indentation and thrown in some `const`s. There's a little bit more detail in my commit messages, in particular https://github.com/ably/ably-cocoa/commit/471cce18a4b4bee35184f4ae0e498d9f5c1ec6cc.

Meanwhile I would like to invite @SlaunchaMan to try directly from this branch to see if it solves the issue, assuming it is regularly reproducible! Jeff, could you please rebuild using this Cocoapods directive instead of the one you currently have for bringing Ably in to your project and let us know if that passes the gaze of the thread sanitizer in your context?:

```
pod 'Ably', :git => 'https://github.com/ably/ably-cocoa.git', :branch => 'feature/1030-data-race'
```

Thanks all! 😁

### DIFF
--- a/Source/ARTRealtime+Private.h
+++ b/Source/ARTRealtime+Private.h
@@ -125,29 +125,30 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
-NS_ASSUME_NONNULL_END
-
-#define ART_TRY_OR_MOVE_TO_FAILED_START(realtime) \
+#define ART_TRY_OR_MOVE_TO_FAILED_START(REALTIME) \
 	do {\
-	ARTRealtimeInternal *__realtime = realtime;\
-    BOOL __started = ARTstartHandlingUncaughtExceptions(__realtime.rest);\
-    BOOL __caught = false;\
-	@try {\
-		do {\
+        ARTRealtimeInternal *const __realtime = REALTIME; \
+        ARTRestInternal *const __rest = __realtime.rest; \
+        const BOOL __started = [__rest startHandlingUncaughtExceptions]; \
+        BOOL __caught = false; \
+        @try { \
+            do {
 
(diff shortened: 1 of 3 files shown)


SlaunchaMan:
@QuintinWillison Using this branch, I was able to run our entire test suite without triggering any data races (except for our own, which I fixed) with the thread sanitizer enabled. 👍

QuintinWillison:
Thanks for the prompt feedback, @SlaunchaMan! That's excellent news. 😁 

Sounds like a good time to do a patch release. I'll get that sorted.

QuintinWillison:
(sorry forgot about the auto-branch delete on merge ... @SlaunchaMan I'll keep this branch alive until the release is out and you've told me you've migrated back to that)

SlaunchaMan:
@QuintinWillison No worries, I never committed any changes with that branch, just tested locally, so you’re good to delete it.

PR #1063, opened 2020-07-24
GitHub action for the test suite (experiment)

ricardopereira:


ricardopereira:
@QuintinWillison FYI, I [have successfully be able to run the test suite on Github actions](https://github.com/whitesmith/ably-cocoa/runs/1037661568?check_suite_focus=true) but it is the same behaviour as Travis. So, we have no other choice, we need to fix the failing tests.

PR #1080, opened 2020-10-16
Update README with new push registration changes

ricardopereira:


README.md
```diff
@@ -146,6 +116,55 @@ All internal operations are dispatched to a single serial GCD queue. You can spe
...
 
+## Push Notifications
+
+If you haven’t yet, you should first check the detailed [documentation](https://www.ably.io/documentation/general/push).
+
+### Activation and device registration
+
+**`ARTPushRegistererDelegate`** is the interface for handling Push activation/deactivation-related actions. The activation process, by default, will check if the `UIApplication.sharedApplication.delegate` has the `ARTPushRegistererDelegate` implementation.
+
+Since version 1.2.2, there's a new preferable way to pass the `ARTPushRegistererDelegate` implementation into the library. You can assign the delegate directly in the client options using the `pushRegistererDelegate` property, without relying on `UIApplicationDelegate`.
```
QuintinWillison, line 127: I think the wording needs to be stronger. It's not "preferable" if the app is created with defaults in Xcode 12 because it will be using the SwiftUI App Lifecycle. Perhaps this paragraph should start with something like "For new apps created with Xcode 12 and later, using the SwiftUI App Lifecycle, ..."?

QuintinWillison (changes requested):

README.md
ricardopereira: @QuintinWillison got it but people that doesn't use the new App lifecycle can also use the new property and it's safer to use it instead of the `UIApplicationDelegate`.

What about this:
```
Since version 1.2.2, there's a new way to pass the `ARTPushRegistererDelegate` implementation into the library. It should be used for new apps created with Xcode 12 and later, using the SwiftUI App Lifecycle. You can assign the delegate directly...
```

QuintinWillison: Better... though "should" is still not strong enough. I think this is a "must". 😁

ricardopereira: 😄 got it! Thanks. 5f054059d55c4e2adcc47bdf1182eadef4f0a829

QuintinWillison (approved):

Issue #1074, opened 2020-10-06
Publishing messages with authCallback is broken after upgrading from 1.1.16 to 1.2.1


tiholic:
Ably SDK Version: `1.2.1`
iOS: `iOS13.6`, simulator
Cocoapods: `1.9.3`
Xcode: `11.6, Build version 11E708`

Unable to publish messages with authCallback enabled. Shared logs from 2 scenarios for better understanding.

1. This scenario is based on authCallback, for publishing one message(same as above) by acquiring TokenRequest json form https://www.ably.io/ably-auth/token-request/demos
> As far as I remember, this used to work well before upgrade

```py
2020-10-03 12:11:01.025098+0530 Runner[92073:745806] flutter: Sending rest message
2020-10-03 12:11:01.026661+0530 Runner[92073:745566] AblyFlutterPlugin.m:327 publish(AblyFlutterMessage)
2020-10-03 12:11:01.027102+0530 Runner[92073:751706] VERBOSE: RS:0x600001cd8d80 ARTJsonLikeEncoder<msgpack>: messageToDictionary {
    clientId = "*";
    data = "Flutter 1";
    name = Hello;
}
2020-10-03 12:11:01.028955+0530 Runner[92073:751706] DEBUG: RS:0x600001cd8d80 ARTJsonLikeEncoder<msgpack> encoding '{
    clientId = "*";
    data = "Flutter 1";
    name = Hello;
}'; got: {length = 38, bytes = 0x83a46e61 6d65a548 656c6c6f a8636c69 ... 6c757474 65722031 }
2020-10-03 12:11:01.030651+0530 Runner[92073:751706] DEBUG: (ARTRestChannel.m:332) RS:0x600001cd8d80 C:0x600000aef1e0 (test) post message (null)
2020-10-03 12:11:01.033253+0530 Runner[92073:751706] DEBUG: (ARTRest.m:286) RS:0x600001cd8d80 calculating authorization 1
[...]

PR #1082, opened 2020-10-23
Fixes 40012 "Malformed message; invalid clientId" when message has no clientId and credentials can assume any clientId

ricardopereira:
#1074

### DIFF
--- a/Source/ARTHttp.m
+++ b/Source/ARTHttp.m
@@ -47,22 +47,14 @@ Class configuredUrlSessionClass = nil;
 }
 
 - (NSObject<ARTCancellable> *)executeRequest:(NSMutableURLRequest *)request completion:(void (^)(NSHTTPURLResponse *_Nullable, NSData *_Nullable, NSError *_Nullable))callback {
-    NSString *requestBodyStr = [[NSString alloc] initWithData:request.HTTPBody encoding:NSUTF8StringEncoding];
-    if (!requestBodyStr) {
-        requestBodyStr = [request.HTTPBody base64EncodedStringWithOptions:NSDataBase64EncodingEndLineWithCarriageReturn];
-    }
-    [self.logger debug:@"--> %@ %@\n  Body: %@\n  Headers: %@", request.HTTPMethod, request.URL.absoluteString, requestBodyStr, request.allHTTPHeaderFields];
+    [self.logger debug:@"--> %@ %@\n  Body: %@\n  Headers: %@", request.HTTPMethod, request.URL.absoluteString, [self debugDescriptionOfBodyWithData:request.HTTPBody], request.allHTTPHeaderFields];
 
     return [_urlSession get:request completion:^(NSHTTPURLResponse *response, NSData *data, NSError *error) {
         NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *)response;
         if (error) {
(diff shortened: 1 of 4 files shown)


ricardopereira:
Screenshot 2020-10-23 at 09 26 56

Source/ARTHttp.m
```diff
@@ -47,14 +47,22 @@ - (void)dealloc {
 }
 
 - (NSObject<ARTCancellable> *)executeRequest:(NSMutableURLRequest *)request completion:(void (^)(NSHTTPURLResponse *_Nullable, NSData *_Nullable, NSError *_Nullable))callback {
-    [self.logger debug:@"--> %@ %@\n  Body: %@\n  Headers: %@", request.HTTPMethod, request.URL.absoluteString, [[NSString alloc] initWithData:request.HTTPBody encoding:NSUTF8StringEncoding], request.allHTTPHeaderFields];
+    NSString *requestBodyStr = [[NSString alloc] initWithData:request.HTTPBody encoding:NSUTF8StringEncoding];
+    if (!requestBodyStr) {
+        requestBodyStr = [request.HTTPBody base64EncodedStringWithOptions:NSDataBase64EncodingEndLineWithCarriageReturn];
+    }
+    [self.logger debug:@"--> %@ %@\n  Body: %@\n  Headers: %@", request.HTTPMethod, request.URL.absoluteString, requestBodyStr, request.allHTTPHeaderFields];
```
QuintinWillison, line 54: This new code adds runtime overhead that's required for normal function. Do we have the capability to detect at runtime if debug logs are enabled and, therefore, only construct this string in that case?

Also, why can the call to `initWithData:encoding:` fail? And if it can fail, would it not be better to encapsulate this logic (used twice here) in to a function to perhaps document why it can fail in the one place?

Source/ARTRestChannel.m
```diff
@@ -247,9 +247,6 @@ - (void)internalPostMessages:(id)data callback:(void (^)(ARTErrorInfo *__art_nul
             callback([ARTErrorInfo createWithCode:ARTStateMismatchedClientId message:@"attempted to publish message with an invalid clientId"]);
             return;
         }
-        else {
```
QuintinWillison, line 250: Could you explain somewhere in this PR, or against the original issue even, why this change was necessary and why it fixes the issue?

QuintinWillison (changes requested):
I've not looked in detail at the tests as I'm short for time, but I've made some comment on the implementation.

Source/ARTHttp.m
ricardopereira: @QuintinWillison I agree, it's necessary overhead for the normal use case. I'll find out a solution. `initWithData:encoding:` can fail when the data is encoded with MsgPack and that's why it should at least show a B64 string instead of null. When I was debugging I thought: `Oh, here's the problem, the body is empty` and that was not the case 😄

Source/ARTRestChannel.m
ricardopereira: I think the removed code existed because of a misinterpretation of the spec. The library should not change the clientId. The clientId is explicitly configured with the `ClientOptions` or implicitly through Token Auth.

tiholic (approved):
`Source/ARTRestChannel.m` looks good to me.
Haven't looked at other changes.

QuintinWillison (approved):

Issue #949, opened 2019-12-17
Reduce library dependencies


QuintinWillison:
### Which version of the Ably SDK are you using? 
`1.1.14`

### On which platform does the issue happen? 
`iOS13`

### Are you using Cocoapods? 
`1.8.4`

### Which version of Xcode are you using? 
`11.3`

### Problem
We have a customer who is integrating our SDK within their own SDK which is then offered out to their customers. The problem they have is that the five dependencies of [our CocoaPod](https://cocoapods.org/pods/Ably) are proving to be a big concern to their customers. They would like to make their own SDK as lightweight as possible. Our transitive dependencies don't help!

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-390) by [Unito](https://www.unito.io)

QuintinWillison:
I'm awaiting confirmation from the customer as to their build and distribution process for their SDK.

QuintinWillison:
Some more information from the customer in respect of how they are distributing their SDK to their customers:

> Our SDK can be distributed by both Cocoapods and Carthage, and will support Swift Package Manager in the future. Since Carthage doesn't support prebuilt framework with dependencies, our customers have to add all of our dependencies in their carthfile. After that, they also have to copy the dependencies path (including the dependencies of each dependency) into the build phase section. It's pretty inconvenient. Cocoapods is the most recommended way as they only need to add our SDK into the podfile. However, seeing a lot of dependencies are installed into their app is scary. That's why we have to make our dependencies as few as possible.

In terms of linking choices (e.g. static vs dynamic) the customer has put it very succinctly for us:

> In the beginning, we tried to archive our SDK as an umbrella framework, but we faced the code-sign problem, and Apple also doesn't recommend the umbrella framework. Now we still include some static libraries in our SDK, that's ok because code-sign problem affects the dynamic library only. FYI.

Furthermore, when they build an app against their SDK (via CocoaPods), they are getting the following on launching the app:

```
error: module importing failed: invalid pathname
dyld: Library not loaded: @rpath/KSCrash.framework/KSCrash
  Referenced from:
      [REDACTED path to customer SDK within customer test app]
  Reason: image not found
Message from debugger: Terminated due to signal 6
```

With their succinct analysis being:

[...]

PR #1090, opened 2020-11-12
Compile SocketRocket directly into the SDK

lawrence-forooghian:
# What’s new

This removes SocketRocket as a dependency from the `Cartfile` and podspec, and compiles SocketRocket directly into the SDK. It uses a submodule to pull the SocketRocket code in.

We're doing this as part of a wider effort to reduce the number of transitive dependencies that we impose on users of the SDK. See #949 for more details.

The commit messages explain what I’ve done in lot more detail, and I'd recommend reviewing this commit-by-commit.

# Questions for reviewers

I've done a fair bit of testing myself - mainly to check that the Carthage and CocoaPods builds are compiling correctly, and that the LICENCE is correctly included in the zip file that we generate for upload to GitHub. However, I'm not too sure how to test that everything is working functionally. I have a handful of test failures when running the unit tests, but my understanding is that this is to be expected as we do have some intermittent failures. What can we do to increase our confidence in merging these changes?


ricardopereira (approved):
🙌 Noice! So with this we can use websocket using URLSession in iOS 13+ and maintain SocketRocket for older versions, right?

lawrence-forooghian:
Yep, we could certainly do that if we wanted to, but that would be a future enhancement. This piece of work is just to hide the SocketRocket dependency from consumers of the SDK.

QuintinWillison:
I had a quick run at making this pull request viable as it had been left to languish for far too long, however it's not immediately obvious to me why the build would be failing now (in something seemingly unrelated).

QuintinWillison (changes requested):
Thanks for removing the superfluous files, @maratal. 👍 

However (and apologies for only just noticing this) but can you please remove the unnecessary layer of folder indirection for socket rocket. From root I would expect to be seeing source files at `SocketRocket/` rather than `SocketRocket/SocketRocket/` as is currently the case.

maratal:
> Thanks for removing the superfluous files, @maratal. 👍
> 
> However (and apologies for only just noticing this) but can you please remove the unnecessary layer of folder indirection for socket rocket. From root I would expect to be seeing source files at `SocketRocket/` rather than `SocketRocket/SocketRocket/` as is currently the case.

It was done not to face possible “include” issues since the whole thing is to be replaced soon. The less we touch it, the better. But probably I shouldn’t cut corner there… 🤪

QuintinWillison:
Oh, I see. That's a fair point, @maratal.

Also, we need to get checks passing. FYI, @lukasz-szyszkowski

QuintinWillison (approved):

lukasz-szyszkowski (approved):
LGTM, but we have to be sure that everything works fine ... so, all tests have to be "green." 🤷‍♂️

maratal:
Since this is a small subset of what @lukasz-szyszkowski has done in https://github.com/ably/ably-cocoa/pull/1136
(and directly included in it via commit https://github.com/ably/ably-cocoa/pull/1136/commits/a4c4aa97e5d917c789c8ca4e6b4a88bab1e13606)
I would prefer to just close this PR bc time cost of trying to merge it with main is too high and thus makes no sense.

lukasz-szyszkowski:
> Since this is a small subset of what @lukasz-szyszkowski has done in #1136
> (and directly included in it via commit [a4c4aa9](https://github.com/ably/ably-cocoa/commit/a4c4aa97e5d917c789c8ca4e6b4a88bab1e13606))
> I would prefer to just close this PR bc time cost of trying to merge it with main is too high and thus makes no sense.

I'm closing this one, changes are in #1136

PR #1098, opened 2021-01-12
Fixed crash by adding a nil-pointer check.

### LINKED ISSUE #1083: Common crash related to websocket failing with error
`1d0d4df4ff4fd816a1e92d3c3b8407b76e5e9114`

`macOS`

### Are you using Cocoapods? 
We've added the dep as a submodule 

Xcode 12.2
Build version 12B5025f

### What did you do?
Have a regular crash related to the library and was wondering if other folks also experienced it.

### What happened instead?

Here is a stack trace of where the crash is occurring. 

```
Crashed: io.ably.main
0  libdispatch.dylib              0x7fff7299517c _dispatch_continuation_init_slow + 64
1  libdispatch.dylib              0x7fff729a46b0 dispatch_after + 521
2  Ably                           0x10ea88712 artDispatchScheduled + 324
3  Ably                           0x10ea9196a -[ARTEventListener startTimer] + 131
4  Ably                           0x10eab79ee -[ARTRealtimeInternal transition:withErrorInfo:] + 481
5  Ably                           0x10eab7a71 -[ARTRealtimeInternal transitionToDisconnectedOrSuspendedWithError:] + 76
6  Ably                           0x10eabeaa0 -[ARTRealtimeInternal realtimeTransportFailed:withError:] + 807
7  Ably                           0x10ea8f044 -[ARTWebSocketTransport webSocket:didFailWithError:] + 180
8  libdispatch.dylib              0x7fff72990d53 _dispatch_call_block_and_release + 12
9  libdispatch.dylib              0x7fff72991dcf _dispatch_client_callout + 8
10 libdispatch.dylib              0x7fff72998124 _dispatch_lane_serial_drain + 618
11 libdispatch.dylib              0x7fff72998bdc _dispatch_lane_invoke + 388
12 libdispatch.dylib              0x7fff729a1090 _dispatch_workloop_worker_thread + 603
13 libsystem_pthread.dylib        0x7fff72bd260b _pthread_wqthread + 409
14 libsystem_pthread.dylib        0x7fff72bd2405 start_wqthread + 13
```

Has anything like this been reported before? Is there any other info I could try and get which would be useful in debugging?

mikepulaski:
Calling `[ARTRealtime transitionSideEffects:]` while in a failed connection state returns a nil reference to `ARTEventListener*`. 

This PR adds a nil-pointer check to avoid a crash, which fixes #1083.

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -483,7 +483,10 @@ dispatch_async(_queue, ^{
 
     [_internalEventEmitter emit:[ARTEvent newWithConnectionEvent:(ARTRealtimeConnectionEvent)state] with:stateChange];
 
-    [stateChangeEventListener startTimer];
+    // stateChangeEventListener may be nil if we're in a failed state
+    if (stateChangeEventListener != nil) {
+        [stateChangeEventListener startTimer];
+    }
 }
 
 - (void)transitionToDisconnectedOrSuspended {
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -514,7 +514,10 @@ dispatch_async(_queue, ^{
 
     ARTEventListener *stateChangeEventListener = [self transitionSideEffects:stateChange];
 
-    [stateChangeEventListener startTimer];
+    // stateChangeEventListener may be nil if we're in a failed state
+    if (stateChangeEventListener != nil) {
+        [stateChangeEventListener startTimer];
+    }
 }
 
 - (ARTEventListener *)transitionSideEffects:(ARTConnectionStateChange *)stateChange {


ricardopereira:
Hi @mikepulaski, thank you for your contribution. About your changes, in my opinion, it doesn't need a nil-pointer check because Objective-C won't produce a crash when the listener reference is `nil`.

mikepulaski:
> Hi @mikepulaski, thank you for your contribution. About your changes, in my opinion, it doesn't need a nil-pointer check because Objective-C won't produce a crash when the listener reference is `nil`.

@ricardopereira  Yeah, I think you're right because we still get the crash I was trying to fix with this :)

I suspect a similar fix is probably needed, though. I suspect the `__weak _eventHandler` reference in `ARTEventListener` is nil when trying to start the timer, as the timer depends on its queue. It might make sense to make a strong reference to it and checking for nil wherever it's used so that it doesn't unexpectedly become nil after passing a nil check.

Something like:

```objc
- (void)startTimer {
    ARTEventEmitter *eventHandler = _eventHandler;
    if (eventHandler == nil) {
        return;
    }
    
    // ...
}
```

ricardopereira:
@mikepulaski Did you tried the tentative fix from #1089?

mikepulaski:
> @mikepulaski Did you tried the tentative fix from #1089?

@ricardopereira Nope! It's not something we've been able to reproduce ourselves, unfortunately. We have quite a few people running our app daily so we see some weird edge cases pop up in the wild. This one is our most frequent, unfortunately.

It looks like your fix is more or less what I was suggesting, but a potential problem is that a nil-check alone wouldn't fix the crash if it happened to dealloc on another queue/thread while this code block is running, which is possible since it's a weak reference. That's why I thought making a local strong reference before checking if it's nil check might do the trick.

QuintinWillison:
I'm hoping to attend to this soon. See [my comment on your other pull request](https://github.com/ably/ably-cocoa/pull/1100#issuecomment-779007590).

QuintinWillison:
@mikepulaski Sorry for the delay. Is there any chance you could rebase atop our updated `main` branch?
Now that https://github.com/ably/ably-cocoa/pull/1103 has landed I'm hoping that this pull request will pass green in CI.
FYI, @SpencerCDixon.

QuintinWillison (approved):

Issue #1099, opened 2021-01-12
Memory leak(s) when using event emitters


mikepulaski:
### Which version of the Ably SDK are you using? 
1.2.3

### On which platform does the issue happen? 
All

### Are you using Carthage? 
No

### Are you using Cocoapods? 
No

### Which version of Xcode are you using? 
Xcode 12.2
Build version 12B45b

### What did you do?
Subscribed to a channel, which caused memory leaks to appear in Instruments.

### What did you expect to happen?
Its resources to be cleaned up when torn down.

### What happened instead?
Memory leaked.

### Link to sample project
I've created a PR to demonstrate the root cause of the issue (`ARTScheduledBlockHandle`) and added a fix (with a test).

PR #1100, opened 2021-01-12
Fixed memory leaks within real time client

mikepulaski:
In short, `ARTScheduledBlockHandle` creates a reference cycle to itself which is never broken if the handler is not cancelled. 

Real time event handlers also create strong references to themselves and this cycle was not broken when the timers were stopped.

I added a `XCTest` to assert the behavior, which could probably be converted to use your testing stack pretty easily -- I just wasn't familiar enough to do to it justice (and it seems like it had its own effect on reference counting) 🙃 

This fixes issue #1099.

### DIFF
--- a/Source/Private/ARTGCD.h
+++ b/Source/Private/ARTGCD.h
@@ -8,8 +8,34 @@
 
 #import <Foundation/Foundation.h>
 
+/// ARTScheduledBlockHandle wraps a block for delayed invocation within a queue. If/when the handle deallocates, the scheduled invocation will be cancelled if it has not already been executed.
 @interface ARTScheduledBlockHandle : NSObject
+- (instancetype)initWithDelay:(NSTimeInterval)delay queue:(dispatch_queue_t)queue block:(dispatch_block_t)block;
+- (void)cancel;
 @end
 
-ARTScheduledBlockHandle *artDispatchScheduled(NSTimeInterval seconds, dispatch_queue_t queue, dispatch_block_t block);
-void artDispatchCancel(ARTScheduledBlockHandle *handle);
+
+static inline ARTScheduledBlockHandle *artDispatchScheduled(NSTimeInterval seconds, dispatch_queue_t queue, dispatch_block_t block) {
+    // We don't pass the block directly; instead, we put it in a property, and
+    // read it back from the property once the timer fires. This gives us the
+    // chance to set the property to nil when cancelling the timer, thus
+    // releasing our retain on the block early. dispatch_block_cancel doesn't do
+    // this, it retains the block even if you cancel the dispatch until the
+    // dispatch time passes. (How this is a good idea escapes me.)
+    //
+    // From Apple's documentation [1]:
+    //
+    // > Release of any resources associated with the block object is delayed
+    // > until execution of the block object is next attempted (or any execution
+    // > already in progress completes).
+    //
+    // https://developer.apple.com/documentation/dispatch/1431058-dispatch_block_cancel
+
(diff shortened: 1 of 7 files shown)


Source/Private/ARTGCD.h
```diff
@@ -8,8 +8,34 @@
...
+    // this, it retains the block even if you cancel the dispatch until the
+    // dispatch time passes. (How this is a good idea escapes me.)
+    //
+    // From Apple's documentation [1]:
+    //
+    // > Release of any resources associated with the block object is delayed
+    // > until execution of the block object is next attempted (or any execution
+    // > already in progress completes).
+    //
+    // https://developer.apple.com/documentation/dispatch/1431058-dispatch_block_cancel
```
mikepulaski, line 32: I wrote this to preserve the existing behavior / intention, but I'm not sure this is actually needed. 

When using blocks with GCD, they're copied (not retained). Variables referenced outside of the scope of the block have different semantics, but typically the thing we care about is how it affects an objects life cycle and whether or not we can mutate its value from within the block. That said, the resources that "aren't released" are either primitive values (bools, ints, etc) or a _pointer_, all of which have very little overhead.

It seems to me that capturing weak variables is likely the solution to whatever problem this was trying to solve, but I could be wrong!

SpencerCDixon:
@QuintinWillison Any updates on possibly getting some of @mikepulaski 's PRs merged? They fix some mem leaks with your client that others would benefit from.

Looks like the Swift compiler is adding a feature to properly detect bridged Objc -> Swift nullable issues and we noticed that it might affect your codebase. The ARTBaseMessage.connectionId can become nil if the connection enters a failed state when there are messages still in the queue. Simple fix might be to make some objects nullable (that are become nil) so that Swift code can handle it properly by returning early or whatever. 

Let us know if you want more details!

QuintinWillison:
Hi @SpencerCDixon and @mikepulaski - sorry for the delay on this. My time has been soaked up by another project recently and nobody else has picked this up either. I'm going to try to set aside some time in the next 48 hours or so to take a look. Thanks for your patience. 😁

ricardopereira:
Nice one! Thank you. Having a semaphore coordinating the scheduled block seems reasonable to me. This will also solve the issue with `ARTEventEmitter` sometimes not canceling the scheduled work.

BTW, I ran the whole test suite (including the Soak tests) a couple of times and I had no issues.

ricardopereira (approved):

QuintinWillison:
@mikepulaski Sorry for the delay. Is there any chance you could rebase atop our updated `main` branch?
Now that https://github.com/ably/ably-cocoa/pull/1103 has landed I'm hoping that this pull request will pass green in CI.
FYI, @SpencerCDixon.

mikepulaski:
@QuintinWillison I'm not entirely sure what you want me to do here, to be honest. It looks like the conflict is just related to the Xcode project (because I added a new test class to verify the fix). It should be pretty easy to fix the conflict on your side, and was hoping this would have been fixed much quicker as we've had to ship a forked version of your client for several releases now.

mikepulaski:
No problem, I get it! Thanks for your help, @QuintinWillison.

lukasz-szyszkowski:
@QuintinWillison Can we close this PR when we have this one https://github.com/ably/ably-cocoa/pull/1130 ?

QuintinWillison:
@lukasz-szyszkowski that would make sense, yes... but I see, currently at least, we're awaiting green CI before we can land #1130.

lukasz-szyszkowski:
I'm closing this PR, changes are [HERE](https://github.com/ably/ably-cocoa/pull/1130)

PR #1107, opened 2021-03-18
Fix of RSH3a2a1 implementation when the clientId doesn't change

ricardopereira:
Related with https://ably-real-time.slack.com/archives/CSQEKCE81/p1615834853011400 cc @cruickshankpg


cruickshankpg (approved):
Great, LGTM

PR #1118, opened 2021-05-13
Fix flaky test: authorization cancellation

ricardopereira:


Source/ARTAuth.m
```diff
@@ -548,30 +548,43 @@ - (void)authorize:(ARTTokenParams *)tokenParams options:(ARTAuthOptions *)authOp
...
+    __block BOOL hasBeenExplicitlyCanceled = false;
     // Request always a new token
     [self.logger verbose:@"RS:%p ARTAuthInternal [authorize.%@, delegate=%@]: requesting new token", _rest, authorizeId, lastDelegate ? @"YES" : @"NO"];
     NSObject<ARTCancellable> *task;
     self->_authorizationsCount += 1;
     task = [self _requestToken:currentTokenParams withOptions:replacedOptions callback:^(ARTTokenDetails *tokenDetails, NSError *error) {
         self->_authorizationsCount -= 1;
 
-        void (^successBlock)(void) = ^{
+        void (^successCallbackBlock)(void) = ^{
```
QuintinWillison, line 559: The block pointers should be `const` (they're never explicitly set to `nil` anywhere, as they just fall out of scope, so we should make that clear).

Source/ARTAuth.m
```diff
@@ -548,30 +548,43 @@ - (void)authorize:(ARTTokenParams *)tokenParams options:(ARTAuthOptions *)authOp
     __weak id<ARTAuthDelegate> lastDelegate = self.delegate;
 
     NSString *authorizeId = [[NSUUID new] UUIDString];
+    __block BOOL hasBeenExplicitlyCanceled = false;
```
QuintinWillison, line 551:
```suggestion
    __block BOOL hasBeenExplicitlyCanceled = NO;
```

Source/ARTAuth.m
```diff
@@ -581,36 +594,39 @@ - (void)authorize:(ARTTokenParams *)tokenParams options:(ARTAuthOptions *)authOp
...
             }];
         }
         else {
-            successBlock();
+            successCallbackBlock();
         }
     }];
 
     [_cancelationEventEmitter once:^(ARTErrorInfo * _Nullable error) {
+        hasBeenExplicitlyCanceled = true;
```
QuintinWillison, line 629:
```suggestion
        hasBeenExplicitlyCanceled = YES;
```

Spec/Auth.swift
```diff
@@ -3725,47 +3725,37 @@ class Auth : QuickSpec {
...
-                    expect(tokenDetailsFirst?.token).toNot(equal(tokenDetailsLast?.token))
                     expect(realtime.auth.tokenDetails).to(beIdenticalTo(tokenDetailsLast))
                     expect(realtime.auth.tokenDetails?.token).to(equal(tokenDetailsLast?.token))
 
                     if let transport = realtime.internal.transport as? TestProxyTransport, let query = transport.lastUrl?.query {
                         expect(query).to(haveParam("accessToken", withValue: realtime.auth.tokenDetails?.token ?? ""))
                     }
                     else {
-                        XCTFail("MockTransport is not working")
+                        fail("MockTransport is not working")
```
QuintinWillison, line 3758: I'm interested as to why `XCTFail` was not suitable here. 🤔

Source/ARTAuth.m
```diff
@@ -548,30 +548,43 @@ - (void)authorize:(ARTTokenParams *)tokenParams options:(ARTAuthOptions *)authOp
...
-        void (^successBlock)(void) = ^{
+        void (^successCallbackBlock)(void) = ^{
             [self.logger verbose:@"RS:%p ARTAuthInternal [authorize.%@]: success callback: %@", self->_rest, authorizeId, tokenDetails];
             if (callback) {
                 callback(tokenDetails, nil);
             }
         };
 
-        void (^failureBlock)(NSError *) = ^(NSError *error) {
+        void (^failureCallbackBlock)(NSError *) = ^(NSError *error) {
```
QuintinWillison, line 566: In this case, not only should the block pointer be `const` but the `error` arg passed to the callback should be too, for strict correctness. I've not run this by the compiler but if my rusty Objective-C memories are correct then that would like this:

```suggestion
        void (^const failureCallbackBlock)(NSError *) = ^(NSError *const error) {
```

QuintinWillison (changes requested):
@lukasz-szyszkowski I'm going to assign this pull request to you so that you can push some commits to address review comments, as well as tweak anything you feel needs change yourself. Thanks.

Spec/Auth.swift
lukasz-szyszkowski: When we're using `Quick/Nimble` framework for tests we should use methods provided by them, the `fail(...)` method is a part of `Nimble` framework. The above change is correct IMO but there is a lot of `XCTFail` usages in this class that should be changed for code cohesion.

Source/ARTAuth.m
lukasz-szyszkowski: changed here https://github.com/ably/ably-cocoa/pull/1118/commits/23b90b26f1002fe137b9013d0a7b7e577e1dc8ca

Source/ARTAuth.m
lukasz-szyszkowski: changed here https://github.com/ably/ably-cocoa/pull/1118/commits/23b90b26f1002fe137b9013d0a7b7e577e1dc8ca

Source/ARTAuth.m
lukasz-szyszkowski: changed here https://github.com/ably/ably-cocoa/pull/1118/commits/23b90b26f1002fe137b9013d0a7b7e577e1dc8ca

Source/ARTAuth.m
lukasz-szyszkowski: changed here https://github.com/ably/ably-cocoa/pull/1118/commits/23b90b26f1002fe137b9013d0a7b7e577e1dc8ca

maratal: Haven't seen that extensive usage of `const` before in objc. Basically it's mimicking swift `let` which is probably right thing to do, but definitely is not common practice. IMHO.

Spec/Auth.swift
```diff
@@ -3725,47 +3725,37 @@ class Auth : QuickSpec {
...
                             partialDone()
                         }
+                        // One of them will be canceled by the connection:
+                        realtime.auth.authorize(callback)
+                        realtime.auth.authorize(callback)
                     }
 
-                    expect(tokenDetailsFirst?.token).toNot(equal(tokenDetailsLast?.token))
                     expect(realtime.auth.tokenDetails).to(beIdenticalTo(tokenDetailsLast))
                     expect(realtime.auth.tokenDetails?.token).to(equal(tokenDetailsLast?.token))
```
maratal, line 3753: Originally `tokenDetailsLast` made sense only in a duet with tokenDetailsFirst, meaning "The first call was cancelled by the second, and the result contained in `realtime.auth.tokenDetails` is from the second call". Without the `tokenDetailsFirst` this check does nothing, except ensuring `realtime.auth.tokenDetails` is the same thing that was thrown in the callback param, which is not the purpose of this check as I can guess.

Source/ARTAuth.m
lukasz-szyszkowski: I agree with @maratal, but it's not a PR merge blocker, and to be honest, I've never seen this before.

Spec/Auth.swift
lukasz-szyszkowski: Ok, I've added one flag to ensure that the previous authorization callback was cancelled by next. The rest of these looks good to me.

lukasz-szyszkowski: changed here 7e23fe5501b4956a5f33ff070f5311a41d912aa2

maratal: Hmm.. that doesn't check whether the second call cancels the first one. This callback potentially can be called by the first call and set `didCancelAuthorization` to `true`. And this is what originally this test was trying to ensure library doesn't do.

lukasz-szyszkowski: As I understand this, in this test, one of these two requests should be cancelled, but it's no matter which one

maratal (approved):

lukasz-szyszkowski (approved):

QuintinWillison (approved):

Issue #1111, opened 2021-04-06
Implement RSC7d (Ably-Agent header)


owenpearson:
see [PR](https://github.com/ably/docs/pull/1034) and [spec point](https://ably-docs.herokuapp.com/client-lib-development-guide/features/#RSC7d) for context.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-685) by [Unito](https://www.unito.io)

QuintinWillison:
For this platform I do wonder whether _how_ the client library was built may also be important, given the customer is often building our library from source to integrate into their app. See [this internal conversation](https://ably-real-time.slack.com/archives/CSQEKCE81/p1619606795034000).

PR #1119, opened 2021-05-14
Ably-Agent Header

maratal:
Issue: https://github.com/ably/ably-cocoa/issues/1111


Spec/RealtimeClientConnection.swift
```diff
@@ -250,7 +252,7 @@ class RealtimeClientConnection: QuickSpec {
                             done()
                         case .connected:
                             if let transport = client.internal.transport as? TestProxyTransport, let query = transport.lastUrl?.query {
-                                expect(query).to(haveParam("lib", withValue: "cocoa\(ARTDefault_variant)-1.2.4"))
+                                expect(query).to(haveParam("lib", hasPrefix: "\(ARTDefault_libName)/1.2."))
```
ben-xD, line 255: This new method is a duplicate of the existing `haveParam` except is uses `hasPrefix` instead of `==`. Im not familiar with Objective-C though. 

I think it does make the test less annoying (you don't have to update the patch version) though by checking for a prefix we are no longer testing this versioning behaviour. 

I think this is not an important detail, feel free to ignore.

ben-xD:
I had a look, it looks good but I only had 1 comment. Thanks :)

Source/ARTDefault.h
```diff
@@ -37,6 +37,8 @@
 
 + (NSString *)libraryVersion;
 
++ (NSString *)userAgent;
```
QuintinWillison, line 40: I would rename this to `ablyAgent`, or even just `agent`.

The `user` suffix suggests too strongly that this is a `User-Agent` as classically understood by most developers (i.e. HTTP context), whereas `User-Agent` was just used as inspiration for [RSC7d2](https://docs.ably.io/client-lib-development-guide/features/#RSC7d2).

QuintinWillison (changes requested):
A few things I think could be improved, please. 😁 

Also, I think there is additional insight we could get from the device at runtime, in particular when running alongside `UIKit` (i.e. iPhone or iPad). In which case could we do something with [model](https://developer.apple.com/documentation/uikit/uidevice/1620044-model?language=objc) and/or [userInterfaceIdiom](https://developer.apple.com/documentation/uikit/uidevice/1620037-userinterfaceidiom?language=objc). @kavalerov or @AndyNicks, do either of you have any thoughts on that from a product perspective, in terms of insight this may bring in future?

maratal:
Thanks @QuintinWillison for comments. I can put there an exact model of the iOS device if it makes sense for statistics.

Source/ARTDefault.m
```diff
@@ -109,4 +110,30 @@ + (NSString *)bundleBuildNumber {
...
+            @"macOS"
+        #else
+            @"Unknown"
+        #endif
+        ;
+}
+
++ (NSString *)osVersionString {
+    NSOperatingSystemVersion version = [[NSProcessInfo processInfo] operatingSystemVersion];
+    NSString *versionString = [NSString stringWithFormat:@"%@.%@.%@", @(version.majorVersion), @(version.minorVersion), @(version.patchVersion)];
```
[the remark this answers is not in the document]
maratal: Not fully understood here, should I replace static property with static function?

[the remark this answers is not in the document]
QuintinWillison: The distinction is writing a `static` function, not a class method (a.k.a. "static method"). Just a nice way of containing a unit of functionality within this compilation unit without polluting namespaces. I'll pick this one up with you in Slack. 😁

kavalerov:
> Also, I think there is additional insight we could get from the device at runtime, in particular when running alongside UIKit (i.e. iPhone or iPad). In which case could we do something with model and/or userInterfaceIdiom.

Collecting `model` would be great, but I don't think `userInterfaceIdiom` will provide us particular value as it will mostly be a duplication of the `model`

kavalerov:
Now thinking about this - @QuintinWillison is there equivalent for `model` in the Android?

Source/ARTDefault.m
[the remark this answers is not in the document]
maratal: fixed

Source/ARTDefault.h
maratal: fixed

maratal:
System gives only a manufacturing model identifier, such as 'iPhone12,5' which is 'iPhone 11 Pro Max'. To get to the latter, all possible names should be hardcoded or plist-ed. https://stackoverflow.com/a/20062141/961037

QuintinWillison:
@maratal I think the identifier is fine (i.e. `iPhone12,5`), because:

1. we want to avoid complex client-side logic, especially not hard-coded maps that will not adapt to future devices being released - these apps could be out in the wild for a while without update
2. we can map at the server-side log analysis stage later on

In terms of those manufacturing model identifiers, we would need to define some logic to remove or replace any characters that would not be compatible with our `Ably-Agent` format. Perhaps a baseline canonicalisation might be something like:

1. convert to lowercase
2. replace all characters not in set 'a' thru 'z', or '0' thru '9' with a hyphen
3. trim any hyphens from either end

thus `iPhone12,5` would become `iphone12-5`

@owenpearson we could consider making this a standard for "wild" / "raw" string conforms going into agent strings. What do you think? worth adding to the spec?

Spec/RealtimeClientConnection.swift
QuintinWillison: @ben-xD I'm going to resolve this conversation but feel free to further comment and/or unresolve if you feel there is work to be done. Thanks.

maratal: @ben-xD @QuintinWillison Sorry for late comment. I've chosen to match "\(ARTDefault_libName)/1.2." because REST request matches this (spec 'RSC7d'), so why match different substrings for the same check?

owenpearson:
@QuintinWillison I think that's a good idea. I do wonder if we could be a little less strict than "only alphanumeric characters and hyphens" - it would be easier to tighten the rules later if we need to than vice versa. But yes, we definitely need to agree on a standard way of handling, for example, whitespace characters and what you described seems like a good way to do that.

owenpearson:
@QuintinWillison To support SemVer we would need to at the very least preserve "+", "-" and "." (That's not to say we can't replace other characters with "-" as well). I can't think of any reason why we'd want to keep commas, for example, but I also don't know if we have a good enough reason to replace them.

Source/ARTDefault.m
```diff
@@ -109,4 +117,37 @@ + (NSString *)bundleBuildNumber {
     return infoDictionary[ARTDefault_bundleBuildNumberKey];
 }
 
++ (NSString *)osName {
```
QuintinWillison, line 120: In light of [my comment](https://github.com/ably/ably-cocoa/pull/1119#issuecomment-843370936) around conforming dynamic agent components (i.e. where the OS vends us a string which we have little control over), what do people think about us emitting strings with mixed case like those presented here?

I like the idea that we're going for "correctness" with these (i.e. how the vendor would represent their product in technical docs or marketing materials), however I don't know if there's any potential impact in terms of how we report on these downstream (i.e. server-side).

Also, see @owenpearson's [recent comment](https://github.com/ably/ably-cocoa/pull/1119#issuecomment-845805830).

Thoughts, @lmars, @SimonWoolf, @paddybyers, @kavalerov, @AndyNicks, Owen, @tomkirbygreen, @ben-xD?

lmars: With respect to the characters allowed, I'd say stick with the same format as User-Agent which is what we based this off, where each product and version is a `token`, and `token` is defined here: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6

As for mixed case, I don't see an issue with that.

QuintinWillison: Thanks, @lmars ... any ideas why [this Stackoverflow answer](https://stackoverflow.com/a/44652895) would omit the `; any VCHAR, except delimiters` from the end of that RFC BNF? 🤔 

Just trying to imagine conformed algorithm / regular expression to validate.

ben-xD: Would it help to see specific examples of the full set of data to be collected (In their worst case) to test the design/ spec, including encoding and decoding? The best (and only) example I see from [RSC7d1](https://ably-docs.herokuapp.com/client-lib-development-guide/features/#RSC7d1) is `ably-flutter/1.2.0 ably-java/1.2.1 android/24` and it now seems a bit simplistic to the case we have now?

lmars: > any ideas why this Stackoverflow answer would omit the ; any VCHAR, except delimiters from the end of that RFC BNF?

No idea TBH 😐

QuintinWillison: @ben-xD what I'm trying to define is the _limits_ of what we'll accept, and it looks like those limits are going to be pretty broad at the end of the day. In the scope of this PR and the Flutter example you quote from the spec, it's primarily the OS products (e.g. `macOS`) and the dynamic model products (e.g. `iPhone12,5`) that we're talking about - and, for me, that feels concrete enough.

I'll have a run at making a docs PR to reflect this.

maratal: While this is still under discussion (https://github.com/ably/docs/pull/1096/files#r636950729), I've pushed the draft version of the possible solution. @QuintinWillison please check if I'm moving in the right direction.

Source/ARTDefault.h
QuintinWillison: https://github.com/ably/ably-cocoa/pull/1119/commits/103f979058b99abad6d497b83697638825f1710e

Source/ARTDefault.m
```diff
@@ -6,9 +6,35 @@
...
 
-#import "ARTDefault+Private.h"
-
 #import "Ably.h"
+#import "ARTDefault+Private.h"
+#import <sys/utsname.h>
+
+// NSOperatingSystemVersion has NSInteger as version components for some reason, so mitigate it here.
+static UInt32 conformVersionComponent(const NSInteger component) {
+    if (component < 0) return 0;
```
QuintinWillison, line 15: Perhaps a little bit too complex. This implementation could perhaps be a one liner:

```objc
return (component < 0) ? 0 : (UInt32)component;
```

Though, if I'm honest, despite my initial direction that should make this a static function, it could be argued that a macro would also suffice.e.g.:

```objc
#define conformVersionComponent(component) ((component < 0) ? 0 : (Uint32)component)
```

Equally, happy to leave it as is - what you've got is still clear.

Source/ARTDefault.m
```diff
@@ -6,9 +6,35 @@
...
+    UInt32 conformedValue = (UInt32)component;
+    return conformedValue;
+}
+
+static NSString *canonizeStringAsUserAgentToken(NSString *_Nonnull inputString) {
```
QuintinWillison, line 20: Per my earlier comment on this review, these are not user agents so please do not use "user" in method or variable names. This could be just:

```suggestion
static NSString *canonizeStringAsAgentToken(NSString *const _Nonnull inputString) {
```

Source/ARTDefault.m
```diff
@@ -6,9 +6,35 @@
...
+    NSMutableString *canonizedString = [NSMutableString string];
```
QuintinWillison, line 21: If `canonizedString` (a pointer) is not intended to be modified then please make it const. i.e.:

```suggestion
    NSMutableString *const canonizedString = [NSMutableString string];
```

Same goes for `allowedSet` and `scanner`.

Source/ARTDefault.m
QuintinWillison, line 20: This method is looking complex enough that it should have some unit tests.

Source/ARTDefault.m
```diff
@@ -109,4 +136,54 @@ + (NSString *)bundleBuildNumber {
...
+                         conformVersionComponent(version.majorVersion),
+                         conformVersionComponent(version.minorVersion),
+                         conformVersionComponent(version.patchVersion)];
+    });
+    return versionString;
+}
+
++ (NSString *)deviceModel {
+    struct utsname systemInfo;
+    uname(&systemInfo);
```
QuintinWillison, line 170: I believe that `uname` has a return value and can fail, so we should guard against that.

Source/ARTDefault.m
```diff
@@ -109,4 +136,54 @@ + (NSString *)bundleBuildNumber {
...
+    static NSString *modelToken;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        modelToken = canonizeStringAsUserAgentToken([self deviceModel]);
+    });
+    return modelToken;
+}
+
++ (NSString *)agent {
+    NSString* agentString = [NSString stringWithFormat:@"%@/%@ %@/%@ hw/%@",
```
QuintinWillison, line 184: I'm not sure where the `hw/` prefix has come from? The second part of that (`deviceModelToken`) is definitely not SemVer compatible! I think you meant this:

```suggestion
    return [NSString stringWithFormat:@"%@/%@ %@/%@ %@",
```

(no need for `agentString` local variable, either 😉)

QuintinWillison (changes requested):
Thanks, @maratal ... there's more to work on here, please...

Source/ARTDefault.m
```diff
@@ -109,4 +136,54 @@ + (NSString *)bundleBuildNumber {
...
+        #if TARGET_OS_IOS
+            @"iOS"
+        #elif TARGET_OS_TV
+            @"tvOS"
+        #elif TARGET_OS_WATCH
+            @"watchOS"
+        #elif TARGET_OS_OSX
+            @"macOS"
+        #else
+            @"Unknown"
```
QuintinWillison, line 150: Hi @maratal - sorry for the late spot on this one but I think this is a problem. We're going to end up with `Unknown/x.y.z` arriving at our servers and that carries little context or value.

I would prefer that this method returns `nil` if the OS cannot be identified, meaning that the `agent` method will need to be refactored so that it doesn't attempt to pack OS name or version in that scenario.

maratal:
@lukasz-szyszkowski  All test are passed
Screenshot 2021-07-22 at 20 20 08
Screenshot 2021-07-22 at 19 51 08

lukasz-szyszkowski:
image
just one of two iOS tests failed, the rest of them passed - it's qualified to merge. But there are still unresolved conversations @QuintinWillison and @maratal

lukasz-szyszkowski (approved):

Source/ARTDefault.m
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/890c069f75096e056d668f9d45f0300248ffc843

Source/ARTDefault.m
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/890c069f75096e056d668f9d45f0300248ffc843

Source/ARTDefault.m
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/890c069f75096e056d668f9d45f0300248ffc843

Source/ARTDefault.m
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/8e0203c2f37cccdee60b1e20ea8082386a1ece02

Source/ARTDefault.m
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/68b1db88db1ce4b54ec9f2715a5761e87c2aa12a

Source/ARTDefault.m
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/68b1db88db1ce4b54ec9f2715a5761e87c2aa12a

Source/ARTDefault.m
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/68b1db88db1ce4b54ec9f2715a5761e87c2aa12a

maratal:
Somehow my previous comments on conversations disappeared 🤔 So I've made new ones. Some commits are combined fixes for multiple conversations, so you'll need to find the fix among others there, sorry 🤷‍♀️ 😅

Source/ARTDefault.m
```diff
@@ -6,8 +6,37 @@
...
 #import "ARTDefault+Private.h"
 #import "ARTNSArray+ARTFunctional.h"
+#import <sys/utsname.h>
+
+// NSOperatingSystemVersion has NSInteger as version components for some reason, so mitigate it here.
+static inline UInt32 conformVersionComponent(const NSInteger component) {
+    return (component < 0) ? 0 : (UInt32)component;
+}
+
+NSString *canonizeStringAsAgentToken(NSString *const inputString) {
```
QuintinWillison, line 19: Thanks for all your work on defining this method, but with our current requirements for `Ably-Agent` it is superfluous to requirements. Please remove it as any code emitting product names to the agents string must be capable of being statically validated at compile time (i.e. product names must not be dynamically sourced or otherwise generated at runtime).

Source/ARTDefault+Private.h
```diff
@@ -9,7 +9,7 @@
 #import <Ably/ARTDefault.h>
 
 extern NSString *const ARTDefaultProduction;
-
+extern NSString *const ARTDefault_libName;
```
QuintinWillison, line 12: Please avoid abbreviation. Perhaps rename to `ARTDefault_libraryName`.

Source/ARTDefault.m
```diff
@@ -135,4 +165,64 @@ + (NSString *)bundleBuildNumber {
...
+
++ (NSString *)agent {
+    NSMutableString *agentString = [NSMutableString stringWithFormat:@"%@/%@", ARTDefault_libName, [self bundleVersion]];
+    NSString *osName = [self osName];
+    if (osName != nil) {
+        [agentString appendFormat:@" %@/%@", osName, [self osVersionString]];
+    }
+    NSString *deviceModelToken = [self deviceModelToken];
+    if (deviceModelToken != nil) {
+        [agentString appendFormat:@" %@", deviceModelToken];
```
QuintinWillison, line 223: For the time being we will drop any attempt to collect device information. The above `osName`/`osVersionString` is adequate for our needs, and matches [the requirement](https://docs.ably.com/client-lib-development-guide/features/#RSC7d5) to have all possible product string emissions databased in [agents.json](https://github.com/ably/ably-common/blob/main/protocol/agents.json).

Spec/RealtimeClientConnection.swift
```diff
@@ -245,7 +247,7 @@ class RealtimeClientConnection: QuickSpec {
...
```
QuintinWillison, line 250: I know you based this on the previous test boilerplate, but I would prefer if you test against a static string (i.e. don't make the test rely on `ARTDefault_libraryName` being correctly assigned).

There is another instance of this that also needs replacing, please.

QuintinWillison (changes requested):
As [mentioned in Slack (internal)](https://ably-real-time.slack.com/archives/CSQEKCE81/p1627561193078800?thread_ts=1627557695.078000&cid=CSQEKCE81), some simplifications, now that we understand the requirements a bit better - especially having landed [the equivalent PR in ably-java](https://github.com/ably/ably-java/pull/671).

Source/ARTDefault+Private.h
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/9eb087a3fb274c10ef78e479154d38278a027228

Source/ARTDefault.m
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/088bec2fc18adf12ef8528fd2f39f3da758c005d

Source/ARTDefault.m
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/088bec2fc18adf12ef8528fd2f39f3da758c005d

Spec/RealtimeClientConnection.swift
maratal: https://github.com/ably/ably-cocoa/pull/1119/commits/abbad0642963b0c8b3f8d11fe78be014fe4a478e

QuintinWillison (changes requested):
Hi @maratal and thank you very much for your recent tranche of changes, in response to my last round of feedback.

Unfortunately, I still managed to overlook something when I reviewed last time. I am very sorry about this! Sometimes I forget to zoom out and consider the full picture. 🙄 

You need to add an API to client options that allows upstream SDKs that depend on this SDK to insert their own components into the agents string.

You can find this prototyped by `ably-java`'s `ClientOptions` class [here](https://github.com/ably/ably-java/blob/4841965193ecabaf6586f53626728d118ea2629f/lib/src/main/java/io/ably/lib/types/ClientOptions.java#L215) (via https://github.com/ably/ably-java/pull/671). Please ensure to copy the API commentary over as it's important to stress that this is only for use by Ably-authored SDKs.

The immediate consumers of this new client options `agents` API will be [ably-asset-tracking-swift](https://github.com/ably/ably-asset-tracking-swift) and [ably-flutter](https://github.com/ably/ably-flutter/issues/100).

QuintinWillison:
I've raised https://github.com/ably/docs/pull/1155 which may help to clarify the `agents` client option requirement.

Source/ARTWebSocketTransport.m
```diff
@@ -184,7 +184,7 @@ - (NSURL *)setupWebSocket:(NSDictionary<NSString *, NSURLQueryItem *> *)params w
     [queryItems addValueAsURLQueryItem:[ARTDefault version] forKey:@"v"];
     
     // Lib
-    [queryItems addValueAsURLQueryItem:[ARTDefault agent] forKey:@"lib"];
+    [queryItems addValueAsURLQueryItem:[options agents] forKey:@"lib"];
```
maratal, line 187: Not sure if it still should be 'lib', in android it's 'agent'. @QuintinWillison

QuintinWillison: Well spotted. I think you're probably right. But please go back to [the spec](https://docs.ably.com/client-lib-development-guide/features/#RSC7d3) and double check.

maratal: It's `agent`

QuintinWillison: https://github.com/ably/ably-cocoa/pull/1119/commits/a09039f4fd081eb71dc408d6c5d52fe1e9ffcae1

QuintinWillison (approved):

QuintinWillison:
> All checks have passed

Quick... LAND IT! 😆

Issue #1127, opened 2021-06-07
Remove `ULID` dependency (replace usage with `NSUUID`)


ben-xD:
I propose we remove `[[ULID new] ulidString]` (last usage of [ULID](https://github.com/whitesmith/ulid)), which is in [`ARTLocalDevice.m`](https://github.com/ably/ably-cocoa/blob/main/Source/ARTLocalDevice.m).

@ricardopereira, I wanted to see what your thoughts are about this. There might be a good reason its being used instead of the built-in `NSString *randomId = [NSUUID new].UUIDString;`?

Benefits:
- Remove uncommonly used dependency (ULID) and just use built-in functionality (small increase in portability, security, usability, readability, maintainability, etc).

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-923) by [Unito](https://www.unito.io)

QuintinWillison:
See also: https://github.com/ably/ably-cocoa/issues/949

ricardopereira:
At that time, `ULID` was the to-go solution defined in the spec. As far as I know, it is safe to use `NSUUID`. Toni can give you more details on that.

QuintinWillison:
Thanks, @ricardopereira.

@tcard has since responded [internally](https://ably-real-time.slack.com/archives/CSQEKCE81/p1623078164135600?thread_ts=1623073211.125700&cid=CSQEKCE81) with:

> I'm the person who originally introduced ULID for push. there's no need for it, UUID4 is fine

PR #1128, opened 2021-06-07
Remove ULID dependency and replace usage with NSUUID

ben-xD:
Closes https://github.com/ably/ably-cocoa/issues/1127

The feature spec has:
> (RSH3a2b) If the local device does not have id or deviceSecret, both are generated locally. The id must be a ulid or similar globally-unique identifier. The deviceSecret must be created using secure random data with sufficient entropy to generate a digest of at least 32 bytes (eg using sha256) and encoding that digest with base64. The local DeviceDetails is updated with the resulting deviceId and deviceSecret. If either the id or the deviceSecret is lost then a new pair must be created.

@paddybyers , it was your commit (`923b7a184839017ea49ee02ad5c901f7aeff7162`) at `3/28/2018, 9:40:38 PM` that added `ulid` explicitly to the feature spec. Do you have any concerns with merging this PR (removing ulid)? I can add a follow up to remove `ULID` from the spec if desired.


ben-xD:
Tests are failing in CI but not failing locally on my machine.

QuintinWillison (approved):

PR #1130, opened 2021-06-08
Real time client memory leaks fixes

lukasz-szyszkowski:
This change was originally made by @mikepulaski in `ably-cocoa` fork (https://github.com/tupleapp/ably-cocoa/tree/gcd-leaks).


ben-xD (approved):

QuintinWillison (approved):
Looks solid. Any ideas why checks aren't passing?

lukasz-szyszkowski:
> Looks solid. Any ideas why checks aren't passing?

well, it looks weird ... because sometimes they pass and sometimes not. I'll re-run all jobs an we will see if they'll pass or not

Source/Private/ARTGCD.h
```diff
@@ -9,7 +9,13 @@
...
 
 @interface ARTScheduledBlockHandle : NSObject
+- (instancetype)initWithDelay:(NSTimeInterval)delay queue:(dispatch_queue_t)queue block:(dispatch_block_t)block;
+- (void)cancel;
 @end
 
 ARTScheduledBlockHandle *artDispatchScheduled(NSTimeInterval seconds, dispatch_queue_t queue, dispatch_block_t block);
-void artDispatchCancel(ARTScheduledBlockHandle *handle);
+static inline void artDispatchCancel(ARTScheduledBlockHandle *handle) {
+    if (handle) {
```
maratal, line 18: Just calling [handle cancel] should be fine. Check on NULL necessary on direct member access with -> operator. Just saying, can be ignored.

ben-xD: True, but some (including myself) may prefer to avoid the "feature" of objective-C where a message to `nil` is _ignored_, and this is more explicit.

Spec/GCDTest.swift
```diff
@@ -0,0 +1,27 @@
...
+            _ = obj
+            invokedExpectation.fulfill()
+        }
+
+        waitForExpectations(timeout: 2, handler: nil)
+
+        _ = scheduledBlock // silence warning
+        scheduledBlock = nil
+
+        XCTAssertEqual(CFGetRetainCount(obj), 1)
```
maratal, line 25: Check this in Instruments (like the original author did) is probably a good idea. Maybe request sample project from him?

maratal (approved):

Source/Private/ARTGCD.h
ben-xD: PS: I started writing a bit of objective-C last week, so my opinion there is under developed 😅

Spec/GCDTest.swift
lukasz-szyszkowski: good idea, I emailed @mikepulaski for demo app.

mikepulaski: Sorry, I don't have an example project. However, if you run this test without the fix, it'll fail: any object(s) captured within the block will have a +1 retain count and will therefore never be deallocated.

lukasz-szyszkowski: I understand. Thanks @mikepulaski you did a great job!

lukasz-szyszkowski:
I double-checked the failing GCD iOS test and there was an issue with this https://forums.swift.org/t/retain-count-set-to-2-and-no-deinit-called/43571/4 
and finally, I changed it to this https://www.swiftbysundell.com/articles/using-unit-tests-to-identify-avoid-memory-leaks-in-swift/#closures

commit here: 0f8e29f9f6089d670a7abf145b754ff6f284d9ea

Source/Private/ARTGCD.h
lukasz-szyszkowski: is this one could be resolved? @maratal

maratal:
> > I would mark each line with retain/release counter, such as "+1 to scheduledBlock, total = 2". Otherwise it's very frustrating to read this with a fresh eye.
> 
> why? In this solution, we don't need to check to retainCounter, this was in the previous solution and it was doubtful.

Just to give the next reader a quick idea on what's going on in this test.

lukasz-szyszkowski:
> > > I would mark each line with retain/release counter, such as "+1 to scheduledBlock, total = 2". Otherwise it's very frustrating to read this with a fresh eye.
> > 
> > 
> > why? In this solution, we don't need to check to retainCounter, this was in the previous solution and it was doubtful.
> 
> Just to give the next reader a quick idea on what's going on in this test.

here 6df222c746bc08e4eda0ca94332cc32865fec592

PR #1131, opened 2021-06-10
Readme improvement: Add install carthage step, and remove `make setup`

ben-xD:
A small change to the readme (2 actual lines of content) because Carthage is required to be installed. I propose we use bullet points going forward to avoid the diff being larger than necessary.

One of the reasons for this PR is to test the CI (checks tab), it was failing.


lukasz-szyszkowski:
> One of the reasons for this PR is to test the CI (checks tab), it was failing.

no luck it fails again ... it requires further investigation globally, not just here in this PR

lukasz-szyszkowski (approved):

PR #1132, opened 2021-06-11
[PREVIEW] Swift Package Manager support

lukasz-szyszkowski:
Closes https://github.com/ably/ably-cocoa/issues/819

PR #1133, opened 2021-06-14
My attempt at SPM support

ben-xD:
This is still in progress and was based on @lukasz-szyszkowski's initial work. Although Ably-cocoa can now be imported into a Swift package or Xcode project, there are lots more to do. I am not doing these immediately, but am instead am using this branch in Ably Asset Tracking SDK, to add support to SPM [there](https://github.com/ably/ably-asset-tracking-cocoa/pull/156).

## Potential Concerns
- **Depending on other open PRs & Broken CI**: I'm depending on a few open PRs right now though, and none have all the other changes, so the diff is very large. CI tests on `ably-cocoa` are failing (even ones with only README changes). I will create a PR and issue to temporarily disable (and reminder to re-enable them in the future). 
- **Many files moved:** I moved a lot of files to keep the `Package.swift` simple and ensured `import Ably` works. There are many files being moved in this PR (moved into the `include` directory). This allows the `Package.swift` to be quite neat, without much configuration. I considered symlinking, but someone reported this was more problems than benefits in a blog post (trying to find it now). Alternatively, we could configure everything in `Package.swift`, but that approach hasn't worked yet (see other PR below). One downside of using symlinks instead (Joe's idea below, is that its not obvious that developers need to add symlinks of public header files in the include directory.). By doing the work upfront now (moving the files), future developers will find it easier. I propose we **do** move many files like I have done.
   > After poo-pooing this idea I wish I had given it more consideration. It would have been worth the up-front cost to have this structured ‘correctly’. The only downside to having the ‘.h’ files separated from the ‘.m’ files is that Xcode has trouble find them for some keyboard shortcuts. A small price to pay to avoid having to mess with symlinks which cause their own issues in Xcode. Whether or not you can do this will depend on the size and complexity of your codebase. [Joe sisnick's experience of adding SPM support to an Objective-C library](https://joesusnick.medium.com/swift-package-manager-with-a-mixed-swift-and-objective-c-project-part-2-2-e71dad234e6)
- **Broken CI code:** CI (`xcodebuild`) might not work now that we are using Swift packages without `.xcodeproj`. This will need rewriting to use e.g. `swift run`
- Ensure **carthage and cocoapods release** still works as expected
  - Or drop Carthage support, like other repos have have done. [Swiftlint PR](https://github.com/realm/SwiftLint/pull/3414)
- **macOS min version:** I bumped the minimum macOS version to 10.11 (El Capitan) instead of 10.10 (Yosemite). Yosemite was released in 2014, and was unsupported by Apple as of August 2017. El Capitan is also not supported by Apple as of December 2019, but we support it, I think this is good enough, LMK if you think 10.10 is important.

## There is another [PR](https://github.com/ably/ably-cocoa/pull/1132) open
@lukasz-szyszkowski and I discussed and debugged some issues together over slack. His [PR](https://github.com/ably/ably-cocoa/pull/1132) is cleaner (much fewer file changes though it has more configuration in Package.swift), but `import Ably` doesn't work yet. ~~Mine though, crashes at runtime instead (when using Ably in a Swift executable, I am trying to build the asset tracking SDK now, but a lot of code changes are needed on that side as well).~~


ben-xD:
Closing in favour of using https://github.com/ably/ably-cocoa/pull/1136. I will create issues to do some of the additional things this PR attempts which 1136 doesn't (e.g. delete Carthage files, delete `xcodeproj`).

Issue #819, opened 2018-11-30
Swift Package Manager support


ricardopereira:
Add support for the [Swift PM](https://swift.org/package-manager/).

Acceptance Criteria: 
- `ably-cocoa` available to customers via Swift Package Manager

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-379) by [Unito](https://www.unito.io)

makoni:
any plans for Swift PM?

QuintinWillison:
Hello @makoni and thanks for commenting. This item is very much alive in our backlog, however we're still gauging the level of interest from developers in order that we can give it appropriate prioritisation.

Can you give us any more detail around your use case?

makoni:
I'm using it in iOS app for events. Right now Swift Package Manager is included in Xcode and it would be nice to have it's support :)

makoni:
any updates?

QuintinWillison:
Sorry, @makoni - we've not done any work on this, nor has it been prioritised to work on in the near future.

This would, clearly, be much easier to "just do" if our codebase were pure Swift. Unfortunately that's not the case so we need to make sure we fully understand what this involves and how to make it happen. I'm going to try to push those conversations forward here at Ably as part of the ongoing, wider process we have around planning work that will increase our developer experience.

Thanks for chasing this. 🙂

makoni:
@QuintinWillison the project doesn't have to be pure Swift. As an example: One Signal SDK is Objective-C project and they've added Swift PM support recently. It might be a good reference :)

https://github.com/OneSignal/OneSignal-iOS-SDK

makoni:
One more example of Objective-C project that added Swift Package Manager support recently:
https://github.com/BranchMetrics/ios-branch-deep-linking-attribution

benjamincombes:
Any news on this? I use 25+ libs, all are on SPM that is now really standard, I was planning to replace Pusher with Ably, but unfortunately I won't start using Cocoapods again just for it... Hope that it could be supported soon so that I can drop Pusher 🙏 !

QuintinWillison:
Hello @makoni and @benjamincombes ... I hear you and thank you for the nudges. I'm prioritising this up our backlog. We need to ramp up the activity in this repository and that is also high on our list.

benjamincombes:
Thanks @QuintinWillison, hope this could be done soon so that we can enjoy all the benefits of Ably over Pusher on our project, can't wait for it!

[...]

PR #1136, opened 2021-06-15
Swift Package Manager support

lukasz-szyszkowski:
[Issue that relates to this PR](https://github.com/ably/ably-cocoa/issues/819)

### This feature requires this PR https://github.com/ably/ably-cocoa/pull/1128 to be merged, instead it won't work.

The base structure of the project has its initial form, and there are no changes in files tree and files content.

### What's new?
- `include` directory in project dir &rarr; `$(SRCROOT)/include`
- `Ably.modulemap` inside `include` directory &rarr;`$(SRCROOT)/include/Ably.modulemap`
- `Ably` directory inside `include` directory which contains symlinked public header files both, the ably SDK and socketrocket SDK  &rarr; `$(SRCROOT)/include/Ably/*.h`
- `PrivateHeaders` directory in project dir that contains symlinked private headers files both, the ably SDK and socketrocket SDK &rarr; `$(SRCROOT)/PrivateHeaders`
- `Package.swift `have defined `headerSearchPaths`


ben-xD:
Have you checked that it installable into a Carthage project, and a Cocoapods project? And if CI is working?

I think my PR has the benefits of this PR, but also:
- reduces the length of the [package.swift](https://github.com/ably/ably-cocoa/blob/7ef47c423ae081c3bbe1186c1120f0761e896182/Package.swift) to 42 lines (this one is 86)
- avoids using Symlinks (because developers need to actually update this is in the future, and there are issues with it mentioned [here](Joe sisnick's experience of adding SPM support to an Objective-C library)). e.g. Xcode is also not good at using Symlinks. Additional CI steps need to be added to check. Symlinks seem to cause long term problems.
- This PR has a `.xcodeproj` and a `Package.swift`, which do not work together, I've asked about this on the swift forums. This would also be confusing, because users might open the `.xcodeproj`, update the settings, but all of it will be overwritten when someone uses the _deprecated_ command `swift package generate-xcodeproj`, which is the only way to get a `xcodeproj` from a `Package.swift`. 
- i've been working on understanding Carthage/ Cocoapods support (mainly investigating Carthage is support is not possible), and trying to get Cocoapods to work.

Unfortunately my PR has a lot more files being moved, but this is the cost of keeping it tidy in the future. I've explained my decisions [here](https://github.com/ably/ably-cocoa/pull/1133). I could also create a PR to move files without adding anything extra, that would make the diff much nicer.

ben-xD:
To elaborate, here is what Joe wrote [here](https://joesusnick.medium.com/swift-package-manager-with-a-mixed-swift-and-objective-c-project-part-2-2-e71dad234e6) 

**Move all of your header files into the include directory.** (this is how I did it).
Umm. no. This is a huge pain and breaks the convention of colocating header and implementation files.
> 
> edit: After poo-pooing this idea I wish I had given it more consideration. It would have been worth the up-front cost to have this structured ‘correctly’. The only downside to having the ‘.h’ files separated from the ‘.m’ files is that Xcode has trouble find them for some keyboard shortcuts. A small price to pay to avoid having to mess with symlinks which cause their own issues in Xcode. Whether or not you can do this will depend on the size and complexity of your codebase.

And about using symlinks:

> edit: A few months in I’ve found that this technique works but it required writing a script to recursively check that public header files were exposed via symlink correctly. I have not had issues with this approach since that was added to the CI pipeline.

Anyway, the challenges we have now are:
- Do we support carthage? (consider dropping it like Swiftlint)
- Do we support Cocoapods?
- Fixing CI to use this `Package.swift`

Package.swift
```diff
@@ -0,0 +1,86 @@
...
+            ],
+            publicHeadersPath: "Source/include",
+            cSettings: [
+                .headerSearchPath("include"),
+                .headerSearchPath("Source"),
+                .headerSearchPath("Source/Private"),
+                .headerSearchPath("SocketRocket/SocketRocket"),
+                .headerSearchPath("SocketRocket/SocketRocket/Internal"),
+                .headerSearchPath("SocketRocket/SocketRocket/Internal/RunLoop"),
+                .headerSearchPath("SocketRocket/SocketRocket/Internal/Security"),
```
ben-xD, line 77: Here we can replace all the search paths including `"SocketRocket...` ones with `.headerSearchPath("SocketRocket/**"),`. This worked for me

Package.swift
```diff
@@ -0,0 +1,86 @@
...
+                "Ably.podspec",
+                "Version.xcconfig",
+                "ably-common/protocol/README.md"
+            ],
+            sources: [
+                "Source",
+                "SocketRocket"
+            ],
+            resources: [
+                .copy("COPYRIGHT"),
```
ben-xD, line 61: What are the benefits of  specifying these `.copy`, I didn't see any warnings or issues when I did not include any `resources`.  Reading the [docs](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#resource), I'm still not sure of the benefits.

All the resources in project root directory seem to copy just fine without specifying anything.

Package.swift
```diff
@@ -0,0 +1,86 @@
+// swift-tools-version:5.4.0
+
+import PackageDescription
+
+let package = Package(
+    name: "Ably",
+    platforms: [
+        .macOS(.v10_11),
```
ben-xD, line 8: I'm happy with bumping minimum version of macOS to 10.11. @QuintinWillison , do you forsee any issues with this?

Package.swift
```diff
@@ -0,0 +1,86 @@
...
+    ],
+    products: [
+        .library(
+            name: "Ably",
+            targets: ["Ably"]
+        ),
+    ],
+    dependencies: [
+        .package(name: "msgpack", url: "https://github.com/rvi/msgpack-objective-C", from: "0.4.0"),
+        .package(name: "AblyDeltaCodec", url: "https://github.com/ably/delta-codec-cocoa", .branch("main"))
```
ben-xD, line 20: Instead of relying on `main`, does it make sense to use:
```.package(name: "AblyDeltaCodec", url: "https://github.com/ably/delta-codec-cocoa/blob/main/Package.swift")```, **without** the version?
or 
```.package(name: "AblyDeltaCodec", url: "https://github.com/ably/delta-codec-cocoa/blob/main/Package.swift", from: "1.3.0")``` or explicitly specifying the version/ git tag

File: Package.swift [hunk 5]
```diff
@@ -0,0 +1,86 @@
+// swift-tools-version:5.4.0
```
ben-xD, line 1: I wanted to point out that we could _potentially_ lower this version to Swift 5.3 (or even lower), but I don't see any clear benefits. So we can leave this here. Does anyone have desire to reduce it?

Source/include/Ably/ARTAuth+Private.h
```diff
@@ -0,0 +1 @@
+../../../Source/ARTAuth+Private.h
```
ben-xD, line 1: As discussed in the meeting, these types of files (ending in `Private.h` should not be in the `include` directory.

ben-xD:
We spoke about removing private headers from `include`. Although some headers end with `Private.h`, some private headers confusingly, don't (though there might be a **good** reason). The list of private headers are in the `Ably.modulemap` file, under the `explicit module Private`. So even though `ARTConstants.h` does not end in `...Private.h`, it should still not be in the `include` folder.

ben-xD (changes requested):

Package.swift
lukasz-szyszkowski: definitely, this should be a specific version

Package.swift
lukasz-szyszkowski: My explanation on video https://s.tape.sh/SQU1HYd8

Package.swift
lukasz-szyszkowski: now it's 5.3 - it cant be lower according to using `.target()` method (in lower version some arguments are unavailable)

Package.swift
lukasz-szyszkowski: done

Source/include/Ably/ARTAuth+Private.h
lukasz-szyszkowski: done, defined in module as Private

lukasz-szyszkowski (commented):
I've added a script to generate symlinks and modulemap. 
You can use it with `make spm_create_module`

ben-xD:
> I've added a script to generate symlinks and modulemap.
> You can use it with `make spm_create_module`

I think this script adds a lot of complexity. Why does a modulemap need to be generated, we had a file already containing the header files? In my PR, i was able to keep headers private by not putting the header in the `include`.

lukasz-szyszkowski:
> > I've added a script to generate symlinks and modulemap.
> 
> > You can use it with `make spm_create_module`
> 
> 
> 
> I think [this script](https://github.com/ably/ably-cocoa/blob/35a38b82f09937daf2c63cb7248f5319ab628b17/Scripts/create-module.sh) adds a lot of complexity. Why does a `modulemap` need to be generated, we had a `modulemap` already containing the header files? In my PR, i was able to keep headers private by not putting the header in the `include`.

As I know it's not possible to make some headers private (not visible public) without Private sub module using modulemap. About "complexity", you don't need to understand what this file do, it's just a helper.
If you have a better idea how to make some headers private, I'd love to watch it.

ben-xD:
> you don't need to understand what this script does, it's just a helper.

I can't agree with that. I need to understand it if its going into the repo. It requires configuration of headers in the script `HEADERS` variable, why not just write it `modulemap` directly? A developer never needs to generate another `modulemap` if it exists.

> If you have a better idea how to make some headers private, I'd love to watch it.
The way I did it in my PR was to move only the public headers into the `include` directory. Like I said in a comment above, private headers should not go into `include`. The private headers remain in `Sources/Ably/Internal/`.

Package.swift
ben-xD: 35a38b82f09937daf2c63cb7248f5319ab628b17

Package.swift
ben-xD: ebb1f08aedafefd03c2800e3e86d4f9b665f5f8b

Package.swift
ben-xD: Okay, I think the warnings you are getting is because you are setting path of your target to the be root of the project (`path: "."`). We can fix this in a separate PR and we can remove these lines in the future.

For example, SwiftLint is Swift Package, and has a lot of resources in the root folder, however, it doesn't not need to declare any resources.

lukasz-szyszkowski:
> > you don't need to understand what this script does, it's just a helper.
> 
> I can't agree with that. I need to understand it if its going into the repo. The script needs configuration of headers in the script's `HEADERS` variable, why not just write headers in the `modulemap` directly? A developer never needs to generate another `modulemap` if it already exists. We would reduce the cognitive overload needed to see how this script interacts with the rest of the developer workflow/ build system. If a developer updates the `modulemap` manually, it will be overwritten when using `make spm_create_module`. I don't see anything good about this.
> 
> > If you have a better idea how to make some headers private, I'd love to watch it.
> 
> The way I did it in my PR was to move only the public headers into the `include` directory. Like I said in a comment above, private headers should not go into `include`. The private headers remain in `Sources/Ably/Internal/`. You should remove the symlinks to private headers you added in `include`.

Ok, I should explain a little bit 🙂

objective-c libs are mainly different than swift ones, they need a modulemaps to connect all headers and expose them (or not) to API. 
I think Realm is a good example of how to do this, please take a look at this directory https://github.com/realm/realm-cocoa, it's almost the same structure as ours. 
I recommend reading about modulemaps for LLVM, [HERE](https://clang.llvm.org/docs/Modules.html) is a really good document.

ben-xD:
I don't think the Clang modules page addresses my main concerns. I understand how Realm does it, but I don't think they have a **script** to generate a `.modulemap`. What are the benefits of having one? Don't users still need to update the headers variable in the script?

> objective-c libs are mainly different than swift ones, they need a modulemaps to connect all headers and expose them (or not) to API.

The main issue I'm raising is you're putting **private** headers (using symlinks) inside the **`include`** directory, which is unusual. 

> By convention, the contents of the include directory are the headers exposed for public consumption. [source](https://stackoverflow.com/questions/48554758/why-create-an-include-directory-in-c-and-c-projects#48554791)

lukasz-szyszkowski:
> I don't think the Clang modules page addresses my main concerns. I understand how Realm does it, but I don't think they have a **script** to generate a `.modulemap`. What are the benefits of having one? Don't users still need to update the headers variable in the script?
> 
> > objective-c libs are mainly different than swift ones, they need a modulemaps to connect all headers and expose them (or not) to API.
> 
> The main issue I'm raising is you're putting **private** headers (using symlinks) inside the **`include`** directory, which is unusual.
> 
> > By convention, the contents of the include directory are the headers exposed for public consumption. [source](https://stackoverflow.com/questions/48554758/why-create-an-include-directory-in-c-and-c-projects#48554791)

private headers moved outside `include` 
"noisy" script removed
changes here: https://github.com/ably/ably-cocoa/pull/1136/commits/8d96bc84c98349f1e86e56e69ad14fa6cfdbf453

SocketRocket/SocketRocket/Internal/Delegate/ARTSRDelegateController.h
```diff
@@ -9,7 +9,7 @@
 
 #import <Foundation/Foundation.h>
 
-#import <SocketRocketAblyFork/ARTSRWebSocket.h>
```
lukasz-szyszkowski, line 12: @maratal I removed `SocketRocketAblyFork` alias and changed import here

SocketRocket/SocketRocket/Internal/Utilities/ARTSRError.h
```diff
@@ -9,7 +9,7 @@
 
 #import <Foundation/Foundation.h>
 
-#import <SocketRocketAblyFork/ARTSRWebSocket.h>
```
lukasz-szyszkowski, line 12: @maratal I removed `SocketRocketAblyFork` alias and changed import here

lukasz-szyszkowski (commented):
info for team devloper

Package.swift
```diff
@@ -0,0 +1,74 @@
...
+
+import PackageDescription
+
+let package = Package(
+    name: "Ably",
```
ben-xD, line 6: Having `name` different to the Github repo name will mean users may face confusion/ unhelpful error messages. This would lead users to ask question on Stack Overflow like [this](https://stackoverflow.com/questions/61837852/swift-package-error-product-dependency-linenoise-in-package-linenoise-swift/67773295#67773295).

On my answer there, I recommend that packages use the same package name as the github repo name. This is done by most Swift Packages. The alternative is to force/ guide users to explicitly use `.package(name: "Ably", url: "https://github.com/ably/ably-cocoa", .branch("feature/SPM-rc"))` instead of `.package(url: "https://github.com/ably/ably-cocoa", .branch("feature/SPM-rc"))`. However, then they have to reference `Ably` in the target dependencies.

TLDR: Users will find it simpler to use the swift package if `ably-cocoa` is used instead of `Ably`.

include/Ably/ARTRest+Private.h
```diff
@@ -0,0 +1 @@
+../../Source/ARTRest+Private.h
```
ben-xD, line 1: Did you forget to remove the private headers from `include` directory? Since you've now added them to `PrivateHeaders/` already.

ben-xD (changes requested):

ben-xD:
After we resolve the Swift package name comment (`Ably` vs `ably-cocoa`), we should add a README about how to use ably-cocoa via SPM. [Here](https://github.com/ably/ably-asset-tracking-swift/blob/d411e6c0db118a71b6ee4f0bd56882a049497325/README.md#installation) is an example of how I've added the SPM usage for ably-asset-tracking-swift

ben-xD:
I tried using the branch, and unfortunately it doesn't work because ULID is still being used, and ULID is not a dependency in `Package.swift`. You can merge in https://github.com/ably/ably-cocoa/pull/1128 to remove the ULID dependency.

Package.swift
ben-xD: Look's like you accidentally reverted it back to 5.4. 5.3 is needed to run in (our) github runners at the moment.

include/Ably/ARTRest+Private.h
lukasz-szyszkowski: I missed `ARTRest+Private.h` in `include/Ably` - it's removed now.
To be clear, there is no `Source/include` directory, now it's moved one level up and it's on the same level  as `Source`
changed id: https://github.com/ably/ably-cocoa/pull/1136/commits/d43ac41d14872a5c85600a24d5c2664591841b03

Package.swift
lukasz-szyszkowski: changed in: https://github.com/ably/ably-cocoa/pull/1136/commits/d43ac41d14872a5c85600a24d5c2664591841b03

lukasz-szyszkowski:
> I tried using the branch, and unfortunately it doesn't work because ULID is still being used, and ULID is not a dependency in `Package.swift`. You can merge in #1128 to remove the ULID dependency (you did this before so I guess you accidentally undid this?).

Ok, I'll merge it. I set this PR as a draft to prevent it from accidentally merge before PR with removed ULID will be merged.

lukasz-szyszkowski:
> After we resolve the Swift package name comment (`Ably` vs `ably-cocoa`), we should add a README about how to use ably-cocoa via SPM. [Here](https://github.com/ably/ably-asset-tracking-swift/blob/d411e6c0db118a71b6ee4f0bd56882a049497325/README.md#installation) is an example of how I've added the SPM usage for ably-asset-tracking-swift

changes https://github.com/ably/ably-cocoa/pull/1136/commits/652e745690e6a05b157cf4ac7bee0615b10a0151 and https://github.com/ably/ably-cocoa/pull/1136/commits/97cf2939be8b0e364cb05597abd6ab1281b5d2a0

Package.swift
```diff
@@ -0,0 +1,74 @@
...
+    name: "ably-cocoa",
+    platforms: [
+        .macOS(.v10_11),
+        .iOS(.v9),
+        .tvOS(.v10)
+    ],
+    products: [
+        .library(
+            name: "ably-cocoa",
+            targets: ["ably-cocoa"]
```
ben-xD, line 15: You shouldn't make the target name `ably-cocoa`, because now users have to `import ably-cocoa`. This should remain the. same as before, so imports are unchanged, `import Ably`. The package name is different to the target name, which is different the library name.

include/Ably/ARTRest+Private.h
ben-xD: There are more that you missed in `include/Ably`: https://github.com/ably/ably-cocoa/tree/feature/SPM-rc/include/Ably: `ARTPush+Private.h`, `ARTAuth+Private.h`

You've also exposed `ARTSRConstants.h`, `ARTSRURLUtilities.h`, `ARTSRSIMDHelpers.h`, `ARTSRRunLoopThread.h`, `ARTSRRandom.h`, `ARTSRProxyConnect.h`, `ARTSRPinningSecurityPolicy.h`, `ARTSRMutex.h`, `ARTSRLog.h`, `ARTSRIOConsumerPool.h`, `ARTSRIOConsumer.h`, `ARTSRHash.h`, `ARTSRHTTPConnectMessage.h`, `ARTSRError.h`, `ARTSRDelegateController.h`, which are **internal** headers of SocketRocket. 

I can also use symbols in this header file in my application (because they are available). Probably not a good idea. 😅

Package.swift
lukasz-szyszkowski: changed: https://github.com/ably/ably-cocoa/pull/1136/commits/18b0b100e953cdcc446272aa403d2a31ceb91cfb

lukasz-szyszkowski:
@ben-xD [HERE](https://github.com/ably/ably-cocoa/pull/1136/commits/18b0b100e953cdcc446272aa403d2a31ceb91cfb) is new commit with Private / Public headers fix ;)

PrivateHeaders/Ably/ARTChannel.h
```diff
@@ -0,0 +1 @@
+../.././Source/ARTChannel.h
```
ben-xD, line 1: Thanks for the header changes. However, ARTChannel, ARTChannels, ARTDataQuery, ARTLog, ARTStatus, ARTTypes, ARTRealtimeChannels are not listed in the private module in the modulemap, so they **should** be in `include/` and exposed to users. All the headers which are in the private module should **not** be in the `include/`, but those that aren't should be in the `include/`. 

Also, you've missed more headers in the `include/` so users can't use features. e.g. `ARTRealtime.h` is not exposed, so users cannot subscribe to a channel. I haven't checked other headers.

You can have a look at my branch, which is based off the existing `Ably.modulemap`:
All the public headers [here](https://github.com/ably/ably-cocoa/tree/feature/ben-butterworth-temporary-spm/Sources/Ably/include)

All the internet headers [here](https://github.com/ably/ably-cocoa/tree/feature/ben-butterworth-temporary-spm/Sources/Ably/Internal)

Package.swift
```diff
@@ -0,0 +1,74 @@
...
+            targets: ["Ably"]
+        ),
+    ],
+    dependencies: [
+        .package(name: "msgpack", url: "https://github.com/rvi/msgpack-objective-C", from: "0.4.0"),
+        .package(name: "AblyDeltaCodec", url: "https://github.com/ably/delta-codec-cocoa", .branch("main"))
+    ],
+    targets: [
+        .target(
+            name: "Ably",
```
ben-xD, line 24: What's the purpose of moving SocketRocket code into Ably? They were previously a module on their own, `SocketRocket`. I've noticed you also now include `SocketRocket` headers in the Ably public header. We're losing separation of concerns/ modularity here.

SocketRocket/SocketRocket/Internal/Delegate/ARTSRDelegateController.h
ben-xD: The main issue is not removing the alias, but actually moving all the code in SocketRocket into Ably. There is no clear reason to me.

Package.swift
lukasz-szyszkowski: here: 5fa3058ed5d53cbbc52e70d0125d74ce35c5170a

Package.swift
ben-xD: SocketRocket is now separate

SocketRocket/SocketRocket/Internal/Delegate/ARTSRDelegateController.h
ben-xD: They were separated into 2 modules again 👍

Package.swift
```diff
@@ -0,0 +1,48 @@
...
+            dependencies: [
+                "SocketRocket",
+                .byName(name: "msgpack"),
+                .byName(name: "AblyDeltaCodec")
+            ],
+            path: "Source",
+            publicHeadersPath: "include",
+            cSettings: [
+                .headerSearchPath("PrivateHeaders/**"),
```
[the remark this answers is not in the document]
ben-xD: Same comment for the `PrivateHeaders/` in SocketRocket.

PrivateHeaders/Ably/ARTChannel.h
ben-xD: Looks like you've fixed it

Package.swift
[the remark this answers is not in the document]
ben-xD: I messaged with @lukasz-szyszkowski  over slack. This was required, because Swift Packages seem to look for the `Ably/` directory within the header search path. This is because the imports are using `#import "Ably/HeaderName.h"`. There were 2 options to fix: change the imports so it doesn't have `"Ably/`, or add symlinks to `PrivateHeaders/ably/` directory. We went with the second option.

Source/include/Ably.modulemap
```diff
@@ -0,0 +1,64 @@
+framework module Ably {
+    umbrella header "Ably/Ably/Ably.h"
```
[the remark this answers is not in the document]
lukasz-szyszkowski: fixed here https://github.com/ably/ably-cocoa/pull/1136/commits/faf6a52757e69927129e562f767d9fb2e1cfa36f

ben-xD (commented):
I think the last thing to do would be to document the development process going forward? e.g. Public headers should be symlinked in `include/` and private headers should be symlinked `Source/PrivateHeaders/Ably`? And also the reasoning why

lukasz-szyszkowski:
> I think the last thing to do would be to document the development process going forward? e.g. Public headers should be symlinked in `include/` and private headers should be symlinked `Source/PrivateHeaders/Ably`? And also the reasoning why. And also make sure the Readme is refreshed/ still makes sense with these changes.

https://github.com/ably/ably-cocoa/issues/1144 I created an issue for this

ben-xD (approved):
@QuintinWillison also might want to have a look though

ben-xD:
Can you also check if it still supports Cocoapods and Carthage (getting it from this branch), so that we are backwards compatible?

maratal (approved):
I have very limited experience with SPM (used it once in server-side project), so I hope you guys sorted this out. @ben-xD expertise here looks solid.

lukasz-szyszkowski:
until there is no tests for this one it shouldn't be merged https://github.com/ably/ably-cocoa/issues/1151

lukasz-szyszkowski:
@maratal there is a lot of commit's because I had a problem with revision and branch name passing to SPM
tl; dr; 
changes are here:
https://github.com/ably/ably-cocoa/blob/c9fdeffd879df57b2c339d1dc396044352ff3845/Examples/SPM/Package.swift
https://github.com/ably/ably-cocoa/blob/3b487dff146c4d10d464c69c1a6ea052281cd9b1/.github/workflows/check.yml

PR #1153, opened 2021-07-16
Move deviceSecret into ARTLocalDevice

ben-xD:
The deviceSecret should be in `ARTLocalDevice`, instead of `ARTDeviceDetails`. ARTLocalDevice inherits from ARTDeviceDetails, so not much is changing.

The feature spec was previously inconsistent, and there is a PR to fix this: https://github.com/ably/docs/pull/1141

### DIFF
--- a/Source/ARTLocalDevice.h
+++ b/Source/ARTLocalDevice.h
@@ -17,6 +17,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nullable, nonatomic, readonly) ARTDeviceIdentityTokenDetails *identityTokenDetails;
 
+/**
+ Device secret generated using random data with sufficient entropy. It's a sha256 digest encoded with base64.
+ */
+@property (nullable, nonatomic) ARTDeviceSecret *secret;
+
 - (instancetype)init NS_UNAVAILABLE;
 
 @end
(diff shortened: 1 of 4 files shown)


lukasz-szyszkowski (approved):

lukasz-szyszkowski:
> The deviceSecret should be in `ARTLocalDevice`, instead of `ARTDeviceDetails`. ARTLocalDevice inherits from ARTDeviceDetails, so not much is changing.
> 
> The feature spec was previously inconsistent, and there is a PR to fix this: [ably/docs#1141](https://github.com/ably/docs/pull/1141)

could you point out tests for this one?

Source/ARTLocalDevice.h
```diff
@@ -17,6 +17,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nullable, nonatomic, readonly) ARTDeviceIdentityTokenDetails *identityTokenDetails;
 
+/**
+ Device secret generated using random data with sufficient entropy. It's a sha256 digest encoded with base64.
+ */
+@property (nullable, nonatomic) ARTDeviceSecret *secret;
```
lukasz-szyszkowski, line 23: are there any tests for `ARTDeviceSecret` in `ARTLocalDevice`? How do you know that everything works properly after this change?

lukasz-szyszkowski:
Some tests are still failing, can merge this only if all tests are green and all conversations are resolved.

Source/ARTLocalDevice.h
ben-xD: I just noticed the inconsistency between libraries and made a change. I'm not sure about the tests, as I don't maintain this library or even read objective-C that well 😅 . Can I hand over this PR to you?

PR #1170, opened 2021-08-20
Documentation: Add link to example app, and clarify `ARTPushRegistererDelegate` usage

ben-xD:


lukasz-szyszkowski (approved):

lukasz-szyszkowski:
Since this is just a README update we don't require that all checks need to be green.

PR #1171, opened 2021-08-22
1003 mac catalyst warnings

### LINKED ISSUE #1003: Warnings about deprecations when used in a Mac Catalyst app
We've had a customer report that they're getting warnings, especially with our SocketRocket fork. They are also reporting that they had to change a line of our code in order to get rid of an error building.

maratal:
Closes #1003

### DIFF
--- a/Source/ARTDeviceIdentityTokenDetails.m
+++ b/Source/ARTDeviceIdentityTokenDetails.m
@@ -6,6 +6,7 @@
 //  Copyright © 2018 Ably. All rights reserved.
 //
 
+#import "ARTTypes.h"
 #import "ARTDeviceIdentityTokenDetails.h"
 
 NSString *const ARTCoderTokenKey = @"token";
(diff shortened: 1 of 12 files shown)


lukasz-szyszkowski (approved):

Source/ARTDeviceIdentityTokenDetails.m
```diff
@@ -72,34 +73,11 @@ + (BOOL)supportsSecureCoding {
...
-        NSData *data = [NSKeyedArchiver archivedDataWithRootObject:self requiringSecureCoding:false error:&error];
-        if (error) {
-            NSLog(@"ARTDeviceIdentityTokenDetails Archive failed: %@", error);
-        }
-        return data;
-    }
-    else {
-        return [NSKeyedArchiver archivedDataWithRootObject:self];
-    }
+    return [self art_archive];
```
lukasz-szyszkowski, line 76: We don't use the "snake_notation" if it's possible please change it to the "camelCase".

Source/ARTDeviceIdentityTokenDetails.m
```diff
@@ -72,34 +73,11 @@ + (BOOL)supportsSecureCoding {
...
-        ARTDeviceIdentityTokenDetails *result = [NSKeyedUnarchiver unarchivedObjectOfClass:[self class] fromData:data error:&error];
-        if (error) {
-            NSLog(@"ARTDeviceIdentityTokenDetails Unarchive failed: %@", error);
-        }
-        return result;
-    }
-    else {
-        return [NSKeyedUnarchiver unarchiveObjectWithData:data];
-    }
+    return [NSObject art_unarchive:data];
```
lukasz-szyszkowski, line 80: Same as above, if it's possible please use the "camelCase" notation.

Source/ARTDeviceIdentityTokenDetails.m
maratal: I was just following notation already presented in this file. And it's "snake". I think we should either rename all at once or just leave it as is (in this file at least).

PR #1178, opened 2021-09-20
Remove `Copyright` and `Created by` from comment header

lukasz-szyszkowski:
There is a lot of changes but all of them is just removing the header comment lines.

### DIFF
--- a/Source/ARTAuth+Private.h
+++ b/Source/ARTAuth+Private.h
@@ -1,9 +1,3 @@
-//
-//  ARTAuth+Private.h
-//  ably
-//
-//
-
 #import <Ably/ARTAuth.h>
 #import <Ably/ARTEventEmitter.h>
 #import "ARTQueuedDealloc.h"
--- a/Source/ARTAuth+Private.h
+++ b/Source/ARTAuth+Private.h
@@ -2,8 +2,6 @@
 //  ARTAuth+Private.h
 //  ably
 //
-//  Created by Ricardo Pereira on 03/10/2015.
-//  Copyright (c) 2015 Ably. All rights reserved.
 //
 
 #import <Ably/ARTAuth.h>
--- a/Source/ARTAuth.h
+++ b/Source/ARTAuth.h
@@ -1,8 +1,6 @@
 //
 //  ARTAuth.h
 //
-//  Created by Jason Choy on 08/12/2014.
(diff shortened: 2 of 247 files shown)


Source/ARTAuth+Private.h
```diff
@@ -2,8 +2,6 @@
 //  ARTAuth+Private.h
```
QuintinWillison, line 2: These are superfluous. They also add maintenance overhead (if this file is renamed) and cognitive load for maintainers, inspectors and reviewers alike.

QuintinWillison (changes requested):
Thanks, but more work needs to be done here. Please remove the ENTIRE header blocks.

Source/ARTAuth+Private.h
lukasz-szyszkowski: here db1308345753fbe2e41eea5d247233ad3afd2934

maratal (approved):

ben-xD (approved):
LGTM

I was wondering though, what tool did you use to autoformat `Source/ARTPushChannelSubscriptions.m`

Edit: Okay Xcode then :P. its more rudimentary compared to AppCode 😂🤓

lukasz-szyszkowski:
> I was wondering though, what tool did you use to autoformat `Source/ARTPushChannelSubscriptions.m`

CMD+A, CTRL+I ;)

QuintinWillison (approved):
All those bytes saved when people `clone` this repo now. 😆 

Thanks for doing this. I know it was tedious but it keeps the house clean.

ben-xD:
Actually because every change causes a new file to be stored in git, it should add bytes, not remove them 😅. A clone will download all files, including the old version and the new one.

In practice, I hope this is not a problem because of compression somewhere. That got me thinking, does git cloning between https vs ssh have different performance? And what the compression is like

Source/ARTAuth+Private.h
lukasz-szyszkowski: @QuintinWillison I can't merge it without resolving this conversation - please resolve it (if it's ok) or add another comment (if it's not). Thanks

QuintinWillison: Sorry for the delay in responding to this. I'm resolving this conversation now but will leave it for you to land the PR.

PR #1179, opened 2021-09-24
Changed ARTPushActivationStateWaitingForDeviceRegistration

maratal:
IDK whether these two lines of code much better than previous two lines, this issue is very difficult to comprehend within short period of time, but it does what @tcard mentioned here https://github.com/ably/ably-java/issues/546#issuecomment-686710130
Lets give it a shot and see what will happen.

### DIFF
--- a/Source/ARTPushActivationState.h
+++ b/Source/ARTPushActivationState.h
@@ -40,7 +40,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface ARTPushActivationStateNotActivated : ARTPushActivationPersistentState
 @end
 
-@interface ARTPushActivationStateWaitingForDeviceRegistration : ARTPushActivationState
+@interface ARTPushActivationStateWaitingForDeviceRegistration : ARTPushActivationPersistentState
 @end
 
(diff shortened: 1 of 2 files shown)


QuintinWillison (approved):
Honestly, I have so little knowledge of how this implementation has been architected or is supposed to work, that I will find it very difficult to make a call either way as to whether this pull request represents a step forwards or backwards. What you say sounds reasonable, so I trust your judgement.

maratal:
These small changes have broken the whole state machine in many places. I'm putting this issue on hold, because fixing such a sophisticated and tangled piece of code requires _unpredictable_ amount of time. What was the purpose of having such a monstrous mechanism for controlling whether you need register on Ably servers or not? It can be degraded to a simple flag IMHO. Or at least make it synchronous (in a thread of course) to keep all this logic in one place. But something that takes  days to figure out for someone looking at it for the first time is just wrong. If this is absolutely required I'll reopen it and continue, but my suggestion is to leave it as-is short term, and re-write it to something more maintainable in the future.

PR #1181, opened 2021-09-26
Replaced bundleVersion to hardcoded ARTDefault_version

maratal:
`[ARTDefault bundleVersion]` always used `CFBundleShortVersionString` for the version string, so when the framework is embed into the main executable, it takes main bundle version which is app version which is wrong. So I've replaced it with hardcoded version string. Bingo.

### DIFF
--- a/Source/ARTDefault.m
+++ b/Source/ARTDefault.m
@@ -181,7 +181,7 @@ static NSInteger _maxMessageSize = 65536;
 }
 
 + (NSString *)libraryAgent {
-    NSMutableString *agent = [NSMutableString stringWithFormat:@"%@/%@", ARTDefault_libraryName, [self bundleVersion]];
+    NSMutableString *agent = [NSMutableString stringWithFormat:@"%@/%@", ARTDefault_libraryName, ARTDefault_version];
     return agent;
 }
 


lukasz-szyszkowski (approved):

Spec/RealtimeClientConnection.swift
```diff
@@ -247,7 +247,7 @@ class RealtimeClientConnection: QuickSpec {
                             done()
                         case .connected:
                             if let transport = client.internal.transport as? TestProxyTransport, let query = transport.lastUrl?.query {
-                                expect(query).to(haveParam("agent", hasPrefix: "ably-cocoa/1.2."))
+                                expect(query).to(haveParam("agent", hasPrefix: "ably-cocoa/1.2"))
```
QuintinWillison, line 250: Why did you drop the `.` suffix on these checks?

Also, I would prefer to be testing against a static string that's specific to this release - i.e. fully qualified SemVer (at the time of writing `1.2.6`). I was going to illustrate for you how we do that in `ably-java`, but it turns out we're no longer doing this there so I've just had to create https://github.com/ably/ably-java/issues/721.

QuintinWillison (changes requested):

Spec/RealtimeClientConnection.swift
maratal: Because `ARTDefault_version` doesn't have patch version, maybe to add it and use `ARTDefault_version` in tests as well?

QuintinWillison: We have [RSC7d4](https://docs.ably.com/client-lib-development-guide/features/#RSC7d4) in the spec because it's important that entire product versions get passed to the server.

Just `1.2` on its own is clearly a protocol / spec / API version number (we're limited by [this](https://docs.ably.com/client-lib-development-guide/versioning/)).

Looking at [the definition of `ARTDefault_version`](https://github.com/ably/ably-cocoa/blob/703df8d4b611424dd56ef233b6f2a31acb77f2ed/Source/ARTDefault.m#L17), it's undocumented (no explanatory comment) and the name doesn't make its purpose any clearer. In `ably-js`, for example, the `Defaults` module has [apiVersion](https://github.com/ably/ably-js/blob/6605d3bdf4bea41ce6d871db8d20c8840f0e82fa/common/lib/util/defaults.js#L39) (`1.2`) and [version](https://github.com/ably/ably-js/blob/6605d3bdf4bea41ce6d871db8d20c8840f0e82fa/common/lib/util/defaults.js#L38) (`1.2.14`, at the time of writing).

I think this pull request is a good opportunity to clean that up in this codebase. 😄 

For what you should be using in the agent, [see `CURRENT_PROJECT_VERSION`](https://github.com/ably/ably-cocoa/blob/6b3c9beb46f9bb373f946d660e6d70d7dbc40661/Version.xcconfig#L1).

maratal: Fixed 8d908ba

PR #1184, opened 2021-10-04
Full reset device details

maratal:
Currently device details loaded from storage once for each run. So after deactivation you will not be able to activate it without relaunching your app. I haven't touched this yet bc push/state machine is not the place for a lot of changes at the same time.


ben-xD:
I've tested it, and there are some issues.

- After deactivation, you cannot re-activate the device. 
- Even though I deactivate the device, and the details are removed. Restarting the device will cause the device details to be populated once again.
- Launching the app from a fresh install, there device details are already populated, when they shouldn't be.

ben-xD:
I'm not sure if it's just me, but it's crashing when I try to get the `clientId` from the DeviceDetails. Sure, its nullable, why doesn't it just print null, instead of crashing? Please see the recording, thanks :)

https://user-images.githubusercontent.com/24711048/140615937-67987250-0852-4136-851b-baf4e5bf0033.mp4

lukasz-szyszkowski:
> I'm not sure if it's just me, but it's crashing when I try to get the `clientId` from the DeviceDetails. Sure, its nullable, why doesn't it just print null, instead of crashing? Please see the recording, thanks :)
> 
>  CleanShot.2021-11-06.at.15.59.45.mp4

`NSNull` is not `nil` you can't perform any selector on its line on `nil` - maybe that's the case https://nshipster.com/nil/

ben-xD:
> `NSNull` is not `nil` you can't perform any selector on its line on `nil` - maybe that's the case https://nshipster.com/nil/

Thanks for spotting that @lukasz-szyszkowski :)

Okay, in that case, why is this value `NSNull`? I didn't really know about the difficulties with `NSNull`, but after [reading](https://developer.apple.com/documentation/foundation/nsnull) it seems like this should only be used in collections. Why not just use `nil` in this case (or if we are using a dictionary internally, translate the `NSNull` to `nil` when decoding the NSDictionary). **Would that be a good idea?**

However, why does `let clientId = device.clientId;` crash the application? Why would `NSNull` be so evil. Perhaps because it shouldn't be `NSNull`, and the application is assuming it is a `NSString`, and calls `_fastCStringContents`. The field is declared as `- (NSString *_Nullable)clientId;`.

lukasz-szyszkowski:
> > `NSNull` is not `nil` you can't perform any selector on its line on `nil` - maybe that's the case https://nshipster.com/nil/
> 
> Thanks for spotting that @lukasz-szyszkowski :)
> 
> Okay, in that case, why is this value `NSNull`? I didn't really know about the difficulties with `NSNull`, but after [reading](https://developer.apple.com/documentation/foundation/nsnull) it seems like this should only be used in collections. Why not just use `nil` in this case (or if we are using a dictionary internally, translate the `NSNull` to `nil` when decoding the NSDictionary, or alternatively, we could just set it to `nil` and delete the key from the dictionary). **Would that be a good idea?**
> 
> However, why does `let clientId = device.clientId;` crash the application? Why would `NSNull` be so evil. Perhaps because it shouldn't be `NSNull`, and the application is assuming it is a `NSString`, and calls `_fastCStringContents`. The field is declared as `- (NSString *_Nullable)clientId;`.

you're trying to push a `nil` object to the collection (Dictionary) that don't accept any `nil` (optional) values, maybe this is the problem. I don't know how the objective-c `nullable` is converted to swift representation.
IMO you can try one of these solutions:
- accept `nil` values `Dictionary<String, Any?>`
- check if the value is `nil` before write it to the Dictioanry

maratal:
Looks like swift converts `nil` to `NSNull` if type is `Any`. Which doesn't include `Any?` which is a different thing (and nullable). So, as @lukasz-szyszkowski pointed out the easiest thing would be to change the type of `Dictionary` to `[String: Any?]`.

ben-xD:
# Problem 1 (solved): `NSNull` is a problem.

> check if the value is nil before writing it into the Dictionary

Simply **checking** the value actually made the application crash. I previously mentioned: 

> why does `let clientId = device.clientId`; crash the application?

Isn't this caused by **mismatch** between the type of `clientId` (`nullable NSString`) and its value (`NSNull`).

@maratal 
>  Looks like swift converts nil to NSNull if type is Any.

Fortunately, it has now fixed itself... It is nil now, without any changes to my code.

# Problem 2: restarting app to reactivate the device?

@maratal said:
> So after deactivation you will not be able to activate it without relaunching your app

I don't think we should merge code that requires users to restart app to access a specific functionality. This would be a step backwards and be very annoying for users.

ben-xD:
> > I don't think we should merge code that requires users to restart app to access a specific functionality.
> 
> I think this is the reason why `reset` was never implemented in the cocoa library - nuances of the device implementation.

Do you also agree we shouldn't merge this? I guess then we have to fix the problem you mentioned. How difficult would it be to fix the problem: 

> It's how it is designed now in ably-cocoa. So there is no simple way to just "stop loading device details". After you press "Activate device" button the only thing that is populated is ARTDeviceIdentityTokenDetails. Which is not even displayed in demo app.

maratal:
@ben-xD I've tested it with flutter example app from `feature/update-ably-cocoa-1.2.6` branch, but here it's already merged with ably-cocoa `main`, so you'll need to update flutter example to test it. Activation/deactivation works now.

ben-xD:
Great, you have fixed the problem where we cannot reactivate the device after deactivation. We needed to restart the app to do so. Now, we don't. 👍

## One observation

When restarting the app, the `ARTLocalDevice`/`ARTDeviceDetails` is populated once again, instead of being null. However, this time with different deviceId, which is good. This is not a big issue then.

## Problem 2: Error log
When launching my app (where I call `ARTRealtime#device`), I get the following errors:

```
DEBUG: (ARTLocalDeviceStorage.m:33) Device Secret not found
ARTDeviceIdentityTokenDetails unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
DEBUG: ARTPush: device token stored
ARTDeviceIdentityTokenDetails unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
DEBUG: ARTPush: device token stored
ARTPushActivationState unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
ARTPushActivationState unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
ARTPushActivationEvent unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
ARTPushActivationEvent unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
```

Should / could this be removed, since I don't think I'm doing anything wrong?

lukasz-szyszkowski:
> When launching my app (where I call `ARTRealtime#device`), I get the following errors:
> 
> ```
> DEBUG: (ARTLocalDeviceStorage.m:33) Device Secret not found
> ARTDeviceIdentityTokenDetails unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
> DEBUG: ARTPush: device token stored
> ARTDeviceIdentityTokenDetails unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
> DEBUG: ARTPush: device token stored
> ARTPushActivationState unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
> ARTPushActivationState unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
> ARTPushActivationEvent unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
> ARTPushActivationEvent unarchive failed: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is NULL}
> ```
> 
> Should / could this be removed, since I don't think I'm doing anything wrong?

It looks like there is no validation if data is nil/NULL/empty before archive/unarchive

maratal:
@ben-xD fixed https://github.com/ably/ably-cocoa/pull/1184/commits/cae5e0eeae55d68ebdb1aa2a5d1581909562ccae

lukasz-szyszkowski (approved):

ben-xD:
I'm still getting some logs being printed out which might not be intentional or ideal. This only happens when I install the app from fresh. It might be a flutter-only issue https://stackoverflow.com/questions/65011405/xcode-ios-project-what-is-the-meaning-of-fopen-failed-and-invalidating-cache. Have you faced these log lines?

When creating a 
```
fopen failed for data file: errno = 2 (No such file or directory)
Errors found! Invalidating cache...
fopen failed for data file: errno = 2 (No such file or directory)
Errors found! Invalidating cache...
```

Source/ARTPushActivationState.m
```diff
@@ -74,7 +74,7 @@ @implementation ARTPushActivationPersistentState
 
 ARTPushActivationState *validateAndSync(ARTPushActivationStateMachine *machine, ARTPushActivationEvent *event) {
     #if TARGET_OS_IOS
-    ARTLocalDevice *const local = machine.rest.device_nosync;
+    ARTLocalDevice *const local = machine.localDevice;
```
ben-xD, line 77: You have replaced some usages of `rest.device_nosync` with `localDevice`, but there are other uses. If we do this change, we should not leave half the codebase using the old way?
Screenshot 2021-11-29 at 08 39 37

ben-xD: Then, you can possibly remove `device_nosync` and `device` properties from `ARTRestInternal`

https://github.com/ably/ably-cocoa/blob/44cbbb0d18af938411d8787e8af9301e80242813/Source/ARTRest%2BPrivate.h#L20-L23

Source/ARTPushActivationStateMachine.m
```diff
@@ -85,6 +85,15 @@ - (ARTPushActivationState *)current_nosync {
...
 }
 
+- (ARTLocalDevice *)localDevice {
+    ARTLocalDevice *device = _rest.device_nosync;
+    if (device.id == nil) {
+        [_rest resetDeviceSingleton];
+        device = _rest.device_nosync;
+    }
+    return device;
+}
```
ben-xD, line 95: Should this logic not live with the rest of the device information (inside `ARTRest`)? It feels out of place here, when the state is inside `_rest`. We use `_rest.device_nosync` elsewhere in the code, and now there are 2 ways (and 2 "properties") for people to use. IMHO we should only have 1

Source/ARTPushActivationState.m
ben-xD: Actually, `localDevice` is a method... Please start with my conversation about the `localDevice` method instead

I will resolve this for now.

ben-xD:
I am able to reset the device details now in Ably Flutter. Now we just need the code reviewed and approved.

Source/ARTPushActivationStateMachine.m
maratal: Good idea, will now move logic from new `localDevice` method to old getter `device_nosync`.

Source/ARTDeviceDetails.h
```diff
@@ -7,7 +7,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface ARTDeviceDetails : NSObject
 
-@property (nonatomic) ARTDeviceId *id;
+@property (nullable, nonatomic) ARTDeviceId *id;
```
lawrence-forooghian, line 10: I'd like to know whether `id` can, in normal use, be `nil`, or whether we're just using `id == nil` to communicate to `-[ARTRest device_nosync]` that the device needs to be regenerated? If the latter, then perhaps we should use a separate flag to represent this state, and keep this non-nullable.

Source/ARTPushActivationStateMachine.m
maratal: Actually not a good idea, there was a reason why they were separated - you can't recreate `ARTLocalDevice.shared_nosync` each time you call `device_nosync` even if `ARTLocalDevice.id` is `nil`.

maratal: Both methods are private, so if you refer to people as ably-cocoa contributors, then one should be aware what method to use - the one that recreates or the one that returns as-is.

maratal: Log messages are not related to ably-cocoa IMHO.

Source/ARTRest.m
```diff
@@ -724,9 +724,6 @@ - (ARTLocalDevice *)device {
     return ret;
 }
 
-// Store address of once_token to access it in debug function.
-static dispatch_once_t *device_once_token;
-
 - (ARTLocalDevice *)device_nosync {
```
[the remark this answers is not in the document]
maratal: This is expected behavior, I vote for returning new instance, since `deviceId` is different, hence the device itself should be different too. Users can keep reference to this object, but the should update it after they reset device.

Source/ARTDeviceDetails.h
maratal: I would avoid flags like "wasReset" or "shouldBeRegenerated" and keep built-in mechanism for that - nullability. That states exactly that -  the property "was reset" or "should be reassigned". But in this particular scenario I should probably went with resetting the whole object with `resetDeviceSingleton` and checking that instead. So yeah, `id` shouldn't be `nil` here.

Source/ARTPushActivationStateMachine.m
ben-xD: Can/shoud this be documented? It's not obvious _to me_, and I'm afraid of the harm and friction this will do for future developers. Having said that, I am not an objective-C developer.

maratal: I've now removed all calls where device can be created except two places where it's actually should be created. Now it's explicitly, except `device` property, which is heavily used in tests, so I left it for now until API settles down. Then I'll fix tests so they don't use this property as well. @ben-xD

maratal: b18bc1ae

Source/ARTRest.m
```diff
@@ -735,18 +732,17 @@ - (ARTLocalDevice *)device_nosync {
...
     // client ID.
-
-    static dispatch_once_t once;
-    device_once_token = &once;
-    static id device;
-    dispatch_once(&once, ^{
-        device = [ARTLocalDevice load:self.auth.clientId_nosync storage:self.storage];
-    });
-    return device;
+    if (ARTLocalDevice.shared_nosync.id == nil) {
```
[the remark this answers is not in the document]
maratal: Because `ARTLocalDevice` should have its own queue for accessing storage, I left it there. `[ARTRest device_nosync]` is just a wrapper accessor. If and when this PR new API accepted, I'll remove usage of this property everywhere and update tests.

Source/ARTDeviceDetails.h
maratal: I've returned nonnull device `id`. Also I don't check on nil anymore, since all creation of a device singlton are explicit in places where it's needed. Also `reset` method nullifies all properties of the device because example app can't understand `nil` value. Once it's updated `reset` will only clear storage and nullify shared device itself.
@ben-xD @lawrence-forooghian

maratal: e3cb20eb

maratal:
After slack discussion it's became clear that this PR should be splitted on two - one for API changes, and the second one for introducing `ARTLocalDevice` internal queue to fix race while device data loading, leading to creation of multiple state machines.

maratal:
There were a lot of chaotic changes in this PR due to me finding solution. Only few of them finally made sense, so I decided for clarity to close this PR in favor of a new one - #1253, which have a clean and easy for review commit history. I've included only the bare minimum of required for this to work changes.

PR #1190, opened 2021-10-18
Add a CODEOWNERS file

lawrence-forooghian:
GitHub will automatically add the users listed here as reviewers on all pull requests in this repo. I think it’ll be helpful for contributors opening pull requests for the first time.

This list is the core development team for the repo; we can still add additional reviewers on an ad-hoc basis.

This is a quick example so we can test its utility. If we want to keep it, we might instead want to use a single codeowner which is a GitHub team.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @maratal @ben-xD @lukasz-szyszkowski @QuintinWillison @lawrence-forooghian


QuintinWillison:
@lawrence-forooghian Regarding GitHub Teams: As far as I'm aware, backed up by my experiences of trying to use teams to organise our outside collaborators (not possible), teams are only visible - and thus useful - to those _within_ our GitHub orgs. See GitHub docs [here](https://docs.github.com/en/organizations/organizing-members-into-teams/about-teams#team-visibility).

QuintinWillison (commented):
Sorry, only just looping back to this one. It's not obvious to me how (or if it's possible) from [the GitHub docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), but is there a way to format this so we have a line per person? That would be more git-diff friendly when we come to adding or removing team members. 🤔

lawrence-forooghian:
I can't see anything from the docs to suggest it's possible, unfortunately.

QuintinWillison:
I think, on further reflection, that our ways of working in the SDK Team - for the time being at least - don't really support the adoption of code owners, and more specifically the implication that we would enable "Require review from Code Owners" on the `main` branch protection rule.

As such, for the moment, this will add more confusion and noise than benefit. As such, I'm going to close this pull request with the caveat that I am completely comfortable with us reopening this discussion in future if we have a focussed need for this to solve issues we're encountering with workflow management. I just don't see that need at the current time.

lawrence-forooghian:
Yep, makes sense. I was coming at it from an ease-of-onboarding angle primarily, but given we seem to bring new developers onto this repo relatively infrequently, agree that it's not worth it given the downsides.

PR #1194, opened 2021-10-22
Fix/1188 Added allowed class for NSKeyedUnarchiver

maratal:
`NSKeyedUnarchiver` becomes more tricky with new version of iOS SDK. So it took a little while to figure out how to use it properly. With iOS 11 it is not happy if you indirectly allow to parse any type within an archive, so now it requires to provide allowed classes list. So I did.

Also I've provided dedicated test for unarchiving function specifically in Objective-C so Swift wouldn't interfere with its under the hood collection types conversions.

### DIFF
--- a/Source/ARTTypes.m
+++ b/Source/ARTTypes.m
@@ -388,7 +388,7 @@ NSString *ARTChannelEventToStr(ARTChannelEvent event) {
 }
 
 + (nullable id)art_unarchiveFromData:(NSData *)data {
-    NSSet* allowedTypes = [NSSet setWithArray:@[ [NSArray class], [NSDictionary class], [self class]]];
+    NSSet* allowedTypes = [NSSet setWithArray:@[ [NSArray class], [NSDictionary class], self]];
 #if TARGET_OS_MACCATALYST
     NSError *error;
     id result = [NSKeyedUnarchiver unarchivedObjectOfClasses:allowedTypes fromData:data error:&error];
@@ -401,7 +401,7 @@ NSString *ARTChannelEventToStr(ARTChannelEvent event) {
         NSError *error;
         id result = [NSKeyedUnarchiver unarchivedObjectOfClasses:allowedTypes fromData:data error:&error];
         if (error) {
-            NSLog(@"%@ unarchive failed: %@", [self class], error);
+            NSLog(@"%@ unarchive failed: %@", self, error);
         }
         return result;
     }
--- a/Source/ARTTypes.m
+++ b/Source/ARTTypes.m
@@ -387,20 +387,25 @@ NSString *ARTChannelEventToStr(ARTChannelEvent event) {
 #endif
 }
 
-+ (nullable id)art_unarchive:(NSData *)data {
++ (nullable id)art_unarchiveWithAllowedClass:(Class)cls fromData:(NSData *)data {
+    if (data == nil) {
+        NSLog(@"%@ unarchive abortion due to data is NULL", cls);
+        return nil;
+    }
+    NSSet* allowedTypes = [NSSet setWithArray:@[ [NSArray class], [NSDictionary class], cls]];
 #if TARGET_OS_MACCATALYST
     NSError *error;
-    id result = [NSKeyedUnarchiver unarchivedObjectOfClass:[self class] fromData:data error:&error];
+    id result = [NSKeyedUnarchiver unarchivedObjectOfClasses:allowedTypes fromData:data error:&error];
     if (error) {
-        NSLog(@"ARTDeviceIdentityTokenDetails Unarchive failed: %@", error);
+        NSLog(@"%@ unarchive failed: %@", cls, error);
     }
     return result;
 #else
     if (@available(macOS 10.13, iOS 11, tvOS 11, *)) {
         NSError *error;
-        id result = [NSKeyedUnarchiver unarchivedObjectOfClass:[self class] fromData:data error:&error];
+        id result = [NSKeyedUnarchiver unarchivedObjectOfClasses:allowedTypes fromData:data error:&error];
         if (error) {
-            NSLog(@"ARTDeviceIdentityTokenDetails Unarchive failed: %@", error);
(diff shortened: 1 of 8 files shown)


Source/ARTTypes.m
```diff
@@ -387,20 +387,25 @@ - (nullable NSData *)art_archive {
 #endif
 }
 
-+ (nullable id)art_unarchive:(NSData *)data {
++ (nullable id)art_unarchiveWithAllowedClass:(Class)cls fromData:(NSData *)data {
+    if (data == nil) {
```
lawrence-forooghian, line 391: I'm wondering why you introduced this check – `data` is not declared as nullable. Did you see something that led you to believe it might be `nil` sometimes?

Source/ARTTypes.m
```diff
@@ -387,20 +387,25 @@ - (nullable NSData *)art_archive {
 #endif
 }
 
-+ (nullable id)art_unarchive:(NSData *)data {
++ (nullable id)art_unarchiveWithAllowedClass:(Class)cls fromData:(NSData *)data {
```
lawrence-forooghian, line 390: I'm a bit confused about why the `cls` parameter is now required. Why was it no longer sufficient to use `[self class]` as the expected class (in addition to the `NSArray` and `NSDictionary` that you've added)?

Source/ARTTypes.m
maratal: It's just migrated from older function, I didn't pay attention to nil checks before recent issues with them. Thanks for pointing out, will fix it.

maratal: @lawrence-forooghian

Source/ARTTypes.m
maratal: Xcode now gives a warning when you use functions without `Class` parameter. See
`+ (nullable id)unarchiveObjectWithData:(NSData *)data API_DEPRECATED`

maratal: @lawrence-forooghian

maratal: Screenshot 2021-10-29 at 23 41 00

Source/ARTTypes.m
lawrence-forooghian: I think it's good that we try and fix this warning about `data` being `nil`. But my worry is that the check you've added here is covering up a larger problem. Because, we have a non-nullable `data` parameter here, which we've observed is actually sometimes `nil` at runtime. So this method's declaration is incorrect and needs to be fixed. And similarly for the declarations in the methods that call this method, e.g. `+[ARTPushActivationState unarchive:]`, whose argument is also declared as non-nullable. And possibly the methods that call that method and so on.

So I'd suggest we keep the scope of this PR limited to fixing the warning about not passing a class to the unarchiver, and create a separate issue for fixing the nullability stuff. What do you think?

Source/ARTTypes.m
lawrence-forooghian: I suspect that the intention of the author of this `art_unarchive` method (i.e. what I think that `[self class]` was trying to achieve) was that it would be called directly on the class that you wish to unarchive. For example, if you want to unarchive a `ARTPushActivationState`, you'd call `[ARTPushActivationState art_unarchive:data]`. If we were to change the call sites of the `art_unarchive` to call the method on the class instead of on `NSObject` then I don't think we'd need the `cls` parameter.

maratal: It was true prior Xcode 13. v13 keeps spitting complaining warnings about that.

Source/ARTTypes.m
maratal: Yes, correct.

Source/ARTTypes.m
maratal: Actually you are right, `[self class]` is pretty enough. I have no idea why I've refused to use <class> instead of `NSObject`, because "the author of this art_unarchive method" is me 😅 Probably because I played around with this method and then forgotten to replace `NSObject` to the actual classes or something like that.

Source/ARTTypes.m
```diff
@@ -387,20 +387,21 @@ - (nullable NSData *)art_archive {
 #endif
 }
 
-+ (nullable id)art_unarchive:(NSData *)data {
++ (nullable id)art_unarchiveFromData:(NSData *)data {
+    NSSet* allowedTypes = [NSSet setWithArray:@[ [NSArray class], [NSDictionary class], [self class]]];
```
lawrence-forooghian, line 391: I think that all the places in this PR where you've written `[self class]` can just be `self`, since they're all inside class methods so `self` is already an instance of `Class`.

Source/ARTTypes.m
```diff
@@ -387,20 +387,21 @@ - (nullable NSData *)art_archive {
...
 #if TARGET_OS_MACCATALYST
     NSError *error;
-    id result = [NSKeyedUnarchiver unarchivedObjectOfClass:[self class] fromData:data error:&error];
+    id result = [NSKeyedUnarchiver unarchivedObjectOfClasses:allowedTypes fromData:data error:&error];
     if (error) {
-        NSLog(@"ARTDeviceIdentityTokenDetails Unarchive failed: %@", error);
+        NSLog(@"%@ unarchive failed: %@", cls, error);
```
lawrence-forooghian, line 396: What's `cls` here?

maratal: Fixed b263103

Source/ARTTypes.m
maratal: Fixed 40b7a64

Source/ARTTypes.m
maratal: Fixed e934771

Source/ARTTypes.m
maratal: Fixed b28c7c2

lawrence-forooghian (approved):

jamienewcomb:
Think this is ready to go? @lukasz-szyszkowski

PR #1198, opened 2021-10-28
Fix/694 static analyser issues

maratal:
Fixed the most obvious static analyser issues

### DIFF
--- a/Source/ARTChannel.m
+++ b/Source/ARTChannel.m
@@ -111,15 +111,18 @@
     [self publish:messages callback:nil];
 }
 
-- (void)publish:(__GENERIC(NSArray, ARTMessage *) *)messages callback:(art_nullable ARTCallback)callback {
+- (void)publish:(__GENERIC(NSArray, ARTMessage *) *)messages callback:(nullable ARTCallback)callback {
     NSError *error = nil;
 
     NSMutableArray<ARTMessage *> *messagesWithDataEncoded = [NSMutableArray new];
     for (ARTMessage *message in messages) {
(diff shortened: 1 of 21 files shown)


lawrence-forooghian:
Hey @maratal – it looks like you're still working on this one (new commits pushed yesterday). Perhaps it would be best to mark it as a draft so that we know when it's ready for review? Because I see that @lukasz-szyszkowski has already approved it, but it's changed since he approved it.

lawrence-forooghian:
I'm also curious where in Xcode you're seeing these static analyser warnings?

maratal:
> Perhaps it would be best to convert it to a draft so that we know when it's truly ready for review?
It certainly would, but I've finished already =) Now it's ready.

> I'm also curious where in Xcode you're seeing these static analyser warnings?
Menu->Product->Analyze

QuintinWillison (commented):
As discussed in yesterday's in-person call, I would like to see evidence that this is being (or will be) checked in the CI environment (GitHub workflows). I believe that would be one of two options:

1. Add to this PR a commit that switches on static analyzer when building (`RUN_CLANG_STATIC_ANALYZER` build setting, I think) - or, perhaps, run `xcodebuild analyze` from workflow(s)
2. Create an issue to add this to our backlog - recognising that when we work on that there may be additional transgressions to fix

lawrence-forooghian (approved):
Looks good – I agree with what @QuintinWillison said about getting this done in CI, too.

maratal:
I agree too, but I think it should be done in a separate issue. Will do.

QuintinWillison (approved):

Issue #1201, opened 2021-10-29
Investigate how we might get rid of the Quick testing framework


lawrence-forooghian:
## What?

I'd like us to consider getting rid of the Quick testing framework.

## Why?

It's a non-standard thing for new developers to learn, but more importantly it's actually making the test suite harder to maintain.

I was recently trying to investigate some tests that are intermittently failing in CI and on my machine. I wanted to see what happens when I try running these tests in isolation instead of in the full suite. Unfortunately, this wasn’t really possible because of Quick’s lack of integration with the Xcode testing UI controls. So, I think removing it would help us towards our short-term goal of identifying and quarantining flaky tests.

## How?

This would be a very time-consuming thing to do manually – our test suite is big. I think it would be possible to automate it, though, using some sort of code transformation tool such as [SwiftSyntax](https://github.com/apple/swift-syntax) with some simple rules, to transform our `QuickSpec` subclasses into `XCTestCase` ones.

I think automating it, other than saving us a heap of time, would save us from having to have disruptive code freezes whilst we migrate a file. It would also allow us to experiment with different approaches for how we’d like the migrated code to look.

I suggest that we prototype something to see whether this automation is viable.

## References

- [Internal Slack conversation](https://ably-real-time.slack.com/archives/CSQEKCE81/p1635456482018900)

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1333) by [Unito](https://www.unito.io)

lawrence-forooghian:
A progress update, and some further thoughts.

## What I've done

I've put together [a prototype](https://github.com/lawrence-forooghian/ably-cocoa-tests-transformer-prototype) using SwiftSyntax, which is capable of:

- changing the superclass of a `QuickSpec` object into `XCTestCase` 
- all `it` invocations contained _directly inside_ (i.e. without an intermediate `context` or `describe`) a `spec()` declaration into top-level `func test*` instance methods

Neither of these are hugely complicated things but they've given me confidence in how to use SwiftSyntax to interrogate the contents of a method and extract things like calls to `context()`, `describe()` etc.

## Viability of the approach

I believe that we will be able to extend this prototype to handle the (more complicated) nested test cases. This is where having access to the AST will truly show its utility over a simple find-and-replace.

## Potential complications for automation

[...]

PR #1206, opened 2021-11-04
[Groundwork for #1201] Remove capture in Quick method call trailing closures

lawrence-forooghian:
## What does this do?

Removes a capture in some trailing closures in the tests. See commit message for more details.

Whilst I was there, I also changed some `var`s to `let`s.

No, it's just a refactor.

## Why are we doing this?

It’s groundwork for removing the Quick testing framework using an automated migrator tool (#1201). The
migrator that I’m writing won’t be able to handle the capture lists in the QuickSpec method call trailing closures, so I’m replacing them.

### DIFF
--- a/Spec/PushAdmin.swift
+++ b/Spec/PushAdmin.swift
@@ -4,7 +4,7 @@ import Quick
 
 class PushAdmin : QuickSpec {
 
-    private static var deviceDetails: ARTDeviceDetails = {
+    private static let deviceDetails: ARTDeviceDetails = {
         let deviceDetails = ARTDeviceDetails(id: "testDeviceDetails")
         deviceDetails.platform = "ios"
         deviceDetails.formFactor = "phone"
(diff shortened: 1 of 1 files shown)


Spec/PushAdmin.swift
```diff
@@ -346,7 +330,7 @@ class PushAdmin : QuickSpec {
                             guard let device = device else {
                                 fail("Device is missing"); done(); return;
                             }
-                            expect(device).to(equal(self.deviceDetails))
+                            expect(device).to(equal(PushAdmin.deviceDetails))
                             expect(error).to(beNil())
```
maratal, line 334: Here and below it can be `Self`?

maratal (approved):

Spec/PushAdmin.swift
lawrence-forooghian: Thanks for teaching me something new! Fixed.

lukasz-szyszkowski (approved):

PR #1208, opened 2021-11-04
Fix `ARTChannel.setOptions` and `ARTRealtimeChannel.setOptions`: Recreate ARTDataEncoder on ARTChannel options update

lukasz-szyszkowski:


Source/ARTChannel.m
```diff
@@ -51,6 +51,17 @@ - (void)setOptions:(ARTChannelOptions *)options {
...
     _options = options;
+    [self recreateDataEncoderWith:options.cipher];
+}
+
+- (void)recreateDataEncoderWith:(ARTCipherParams*)cipher {
+    NSError *error = nil;
+    _dataEncoder = [[ARTDataEncoder alloc] initWithCipherParams:cipher error:&error];
+    
+    if (error != nil) {
+        [_logger warn:@"creating ARTDataEncoder: %@", error];
```
ben-xD, line 62: Should this be a warning? Or a verbose log?

Source/ARTChannel.m
```diff
@@ -51,6 +51,17 @@ - (void)setOptions:(ARTChannelOptions *)options {
...
 - (void)setOptions_nosync:(ARTChannelOptions *)options {
     _options = options;
+    [self recreateDataEncoderWith:options.cipher];
+}
+
+- (void)recreateDataEncoderWith:(ARTCipherParams*)cipher {
+    NSError *error = nil;
+    _dataEncoder = [[ARTDataEncoder alloc] initWithCipherParams:cipher error:&error];
+    
+    if (error != nil) {
```
ben-xD, line 61: Should we handle the case where there is an error?

ben-xD (approved):
Thanks for quickly working on this :)

Unfortunately, I find that the application crashes now when I call `[channel setOptions:channelOptions];`. I've recorded this in a video that I attached. Notice that the application is paused at a breakpoint before calling `[channel setOptions:channelOptions];`. After playing, it immediately crashes at a line of code _related_ to this PR.

https://user-images.githubusercontent.com/24711048/140435813-862453c1-2721-4321-9dea-0fc10886f972.mp4

Source/ARTChannel.m
lukasz-szyszkowski: it's a copy/paste from init
```objc
- (instancetype)initWithName:(NSString *)name andOptions:(ARTChannelOptions *)options rest:(ARTRestInternal *)rest {
    if (self = [super init]) {
        _name = name;
        _logger = rest.logger;
        _queue = rest.queue;
        _options = options;
        NSError *error = nil;
        _dataEncoder = [[ARTDataEncoder alloc] initWithCipherParams:_options.cipher error:&error];
        if (error != nil) {
            [_logger warn:@"creating ARTDataEncoder: %@", error];
            _dataEncoder = [[ARTDataEncoder alloc] initWithCipherParams:nil error:nil];
        }
    }
    return self;
}
```

lukasz-szyszkowski:
> Thanks for quickly working on this :)
> 
> Unfortunately, I find that the application crashes now when I call `[channel setOptions:channelOptions];`. I've recorded this in a video that I attached. Notice that the application is paused at a breakpoint before calling `[channel setOptions:channelOptions];`. After playing, it immediately crashes at a line of code _related_ to this PR.
> 
>  CleanShot.2021-11-04.at.23.49.11.2.mp4

fixed here 0a7fd0c8fa7cf40897a6b671ac27c7dad43c2c5b

ben-xD (approved):
Thanks. I tested these changes and there don't seem to be any crashes anymore. The crash turned out to be a race condition issue, as when I was testing it, it worked fine if using the debugger and stepping through the program more slowly. This led to @lukasz-szyszkowski trying multiple fixes, and we finally settled on `dispatch_barrier_async` as it stopped causing the crashes.

I would appreciate others with more objective-C and especially those with **objective-C + concurrency eyes** here to approve if the race condition is fully fixed, not just a fix for some cases.

maratal:
I would rather assure wether we need `setOptions:` to be called after creation of the channel. It's declared as private and probably shouldn't exist at all. Playing with multithreading like that is not what you would want, not saying making release out of it in a rush. ON FRIDAY.

Source/ARTChannel.m
```diff
@@ -51,6 +51,19 @@ - (void)setOptions:(ARTChannelOptions *)options {
 
 - (void)setOptions_nosync:(ARTChannelOptions *)options {
     _options = options;
+    [self recreateDataEncoderWith:options.cipher];
+}
+
+- (void)recreateDataEncoderWith:(ARTCipherParams*)cipher {
+    dispatch_barrier_async(_queue, ^{
```
lawrence-forooghian, line 58: I believe that `_queue` here ultimately comes from [`-[ARTClientOptions internalDispatchQueue]`](https://github.com/ably/ably-cocoa/blob/902c112d02a34c79d48f00ebf3845ed48d6f08bd/Source/ARTClientOptions.h#L103-L109), which is always a serial queue. That means that `dispatch_barrier_async` behaves the same as `dispatch_async`. I'd suggest that we use the latter to avoid confusion, and for consistency with the rest of the codebase.

Source/ARTChannel.m
```diff
@@ -51,6 +51,19 @@ - (void)setOptions:(ARTChannelOptions *)options {
 
 - (void)setOptions_nosync:(ARTChannelOptions *)options {
     _options = options;
+    [self recreateDataEncoderWith:options.cipher];
```
lawrence-forooghian, line 54: I worry that here we're changing the behaviour of the `-setOptions:` and `-setOptions_noSync:` methods. Where before they performed their work entirely synchronously, the `dispatch_barrier_async` of `-recreateDataEncoderWith:` makes them asynchronous. This change is, however, not reflected outside of these methods' implementations – if we wanted to make these methods asynchronous, then we'd need to add callbacks to tell the caller when their requested update has completed. That might presumably ultimately have effects on the SDK's public API, so we'd want to tread carefully here.

lawrence-forooghian (commented):
I think I'd like to better understand the motivation behind the synchronisation that you've added here. Could you please explain a little what you think was the cause of the crash that @ben-xD was seeing, and how this fixes it?

Perhaps I could also check that I've understood the meaning of the `_nosync` method pattern that I've seen through the codebase. Is the idea of these methods that these should only be called on the SDK's internal queue, and hence shouldn't need any sort of synchronisation? If so, then shouldn't that mean that `-setOptions_nosync` here is already being called on `_queue` and hence doesn't need to be dispatched there?

maratal:
@ben-xD can you reproduce this crash one more time and type "po <address>", just curious what object hides under that address 0x*****8080. Triggering any synchronisation in all `*_noSync:` methods doesn't look good. Probably previous commit is better, but we need to understand the nature of this crash better.

Source/ARTChannel.m
lukasz-szyszkowski: agree
```objc
_internalDispatchQueue = dispatch_queue_create("io.ably.main", DISPATCH_QUEUE_SERIAL)
```

lukasz-szyszkowski: removed here 0b507e371ef454c8b5e6290d1e87afd3db047614

Source/ARTChannel.m
lukasz-szyszkowski: changed here 0b507e371ef454c8b5e6290d1e87afd3db047614

lawrence-forooghian (approved):

maratal (approved):

ben-xD:
@maratal, after trying to reproduce the issue, the crashes no longer happened. I also reported this in slack: https://ably-real-time.slack.com/archives/CSQEKCE81/p1636137481110900?thread_ts=1636104128.094400&cid=CSQEKCE81

Therefore, the latest state of the PR looks good to merge, without any of the thread safety additions.

QuintinWillison (approved):

lukasz-szyszkowski:
image

PR #1210, opened 2021-11-07
Added methods in ARTRestProtocol for corresponding http methods

### LINKED ISSUE #780: Ensure request method accepts UPDATE, PATCH & DELETE verbs
This will be hard to test sadly without mocking or sending requests to a non-realtime server, but we should do our best to confirm request support these verbs.

See https://github.com/ably/docs/pull/493

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-376) by [Unito](https://www.unito.io)

maratal:
Closes #780 

Since ably-cocoa supports all methods indirectly, because it accepts `NSString` as http method, I decided to add each http method shortcut as it was implemented in some other ably platforms (see https://github.com/ably/docs/pull/493)

Methods names are in uppercase, because `delete` is a keyword.

### DIFF
--- a/Source/ARTRest.h
+++ b/Source/ARTRest.h
@@ -17,6 +17,16 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@interface ARTHttpMethods : NSObject
+
+@property (class, readonly) NSString *GET;
+@property (class, readonly) NSString *PATCH;
+@property (class, readonly) NSString *PUT;
+@property (class, readonly) NSString *POST;
+@property (class, readonly) NSString *DELETE;
+
+@end
+
 @protocol ARTRestProtocol
 
 - (instancetype)init NS_UNAVAILABLE;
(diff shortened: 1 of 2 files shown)


lukasz-szyszkowski:
I wonder do we really need all this boilerplate? like:
```objc
- (BOOL)PUT:(NSString *)path
     params:(nullable NSStringDictionary *)params
       body:(nullable id)body
    headers:(nullable NSStringDictionary *)headers
   callback:(ARTHTTPPaginatedCallback)callback
      error:(NSError *_Nullable *_Nullable)errorPtr {
    return [_internal PUT:path params:params body:body headers:headers callback:callback error:errorPtr];
}

- (BOOL)POST:(NSString *)path
      params:(nullable NSStringDictionary *)params
        body:(nullable id)body
     headers:(nullable NSStringDictionary *)headers
    callback:(ARTHTTPPaginatedCallback)callback
       error:(NSError *_Nullable *_Nullable)errorPtr {
    return [_internal POST:path params:params body:body headers:headers callback:callback error:errorPtr];
}
```
it should be managed with one or two methods, one for calls with `body` and another one without.

maratal:
> it should be managed with one or two methods, one for calls with `body` and another one without.

The idea was to introduce all http methods instead of using simple `NSString`.

Source/ARTRest.h
```diff
@@ -17,6 +17,16 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@interface ARTHttpMethods : NSObject
```
lawrence-forooghian, line 20: Please could you explain your thinking behind introducing this class into the public interface?

lawrence-forooghian (commented):
I'm a bit confused about the aim behind this PR. It sounds like #780 is already met by the existing method

```objective-c
- (BOOL)request:(NSString *)method
           path:(NSString *)path
         params:(nullable NSStringDictionary *)params
           body:(nullable id)body
        headers:(nullable NSStringDictionary *)headers
        callback:(ARTHTTPPaginatedCallback)callback
           error:(NSError *_Nullable *_Nullable)errorPtr;
```

— is that right? In which case this PR seems like an enhancement to the existing API, but doesn't add any new functionality. Please let me know whether I've understood correctly.

Source/ARTRest.h
maratal: >In which case this PR seems like an enhancement to the existing API, but doesn't add any new functionality

Yes, exactly. It's just wrapper functions.

maratal: >keep the existing interface or add a new one?

Removing existing one probably will be very painful for users...

lawrence-forooghian: Yeah, we definitely can't remove the existing one.

Does that mean we can already mark #780 as closed? I think that I'd like to get @QuintinWillison's opinion on whether it's worth adding these extra convenience methods or not.

maratal:
Closing due to unnecessary implementation. "PATCH" is supported implicitly.

PR #1213, opened 2021-11-07
Turned on static analyser after build

### LINKED ISSUE #1202: Run static analysis for builds in CI
Either via Xcode build setting (`RUN_CLANG_STATIC_ANALYZER`?) or as explicit step in workflow(s) (`xcodebuild analyze`?).

See [this comment](https://github.com/ably/ably-cocoa/pull/1198#pullrequestreview-794962849) and the pull request it sits in for context.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1336) by [Unito](https://www.unito.io)

maratal:
Closes #1202


lukasz-szyszkowski (approved):

maratal:
Analyzing turned on (screenshot from "View raw logs"):

Screenshot 2021-11-11 at 16 08 32

Screenshot 2021-11-11 at 16 10 49

lukasz-szyszkowski:
> Will this fail the CI build if the static analyser spots a problem? It seems necessary to me — I don't think anybody's going to come and look at the build logs. (@QuintinWillison, what do you think? Was this implicit in #1202?)

agree, but analyze can't fail xcodebuild so it's a bit tricky, there should be some output other than console log, something "human readable" like HTML files, to achieve this you need to pass config params to `xcodebuild` command and then check the existence of files from the analyzer, you can read about this [here](https://stackoverflow.com/a/39635929)

QuintinWillison:
> analyze can't fail xcodebuild

Really? It's been quite a number of years since I set this up in a project but my memory is that this was possible. Maybe things have changed significantly, but regardless, if CI doesn't fail when analyze fails then it's impotent and meaningless in reality.

maratal:
>Will this fail the CI build if the static analyser spots a problem?

We should turn on "Treat warnings as errors" to achieve this, as was pointed out by @QuintinWillison in one of our calls.

>Additionally, you should add artifact step to upload the result of the analysis.

Isn't it in the same log file? @lukasz-szyszkowski

lawrence-forooghian:
My memory of setting this up a long time ago is that out of the box it doesn't fail a build which produces analyzer warnings, and I think that's true even with "treat warnings as errors" enabled. My rough memory is that you need to do what @lukasz-szyszkowski [said](https://github.com/ably/ably-cocoa/pull/1213#issuecomment-966477222) — set up something separate to check the output of `xcodebuild`.

lukasz-szyszkowski:
> Isn't it in the same log file?

it produce something like this
[report-2b2517.html.zip](https://github.com/ably/ably-cocoa/files/7522194/report-2b2517.html.zip)

.github/workflows/check.yml
```diff
@@ -57,6 +57,16 @@ jobs:
...
         run: bundle exec fastlane ${{ matrix.lane }}
       
+      - name: Check Static Analyzer Output
+        run:  |
+          if [[ -z `find clang -name "*.html"` ]]; then
+            echo "Static Analyzer found no issues. Fail for demo purpose."
+            exit 123
+          else
+            echo "Static Analyzer found some issues"
+            exit 1234
```
QuintinWillison, line 67: I'm not sure anything but a value of `1` adds any value here. `1234` suggests meaning that doesn't exist.

.github/workflows/check.yml
```diff
@@ -57,6 +57,16 @@ jobs:
       - name: Run Tests
         run: bundle exec fastlane ${{ matrix.lane }}
       
+      - name: Check Static Analyzer Output
+        run:  |
+          if [[ -z `find clang -name "*.html"` ]]; then
+            echo "Static Analyzer found no issues. Fail for demo purpose."
+            exit 123
```
QuintinWillison, line 64: We're going to remove this line, clearly, before this PR can land.

fastlane/Fastfile
```diff
@@ -6,7 +6,7 @@ platform :ios do
     run_tests(
       scheme: "Ably-iOS-Tests",
       devices: ["iPhone XS (12.0)"],
-      xcargs: { ABLY_ENV: ENV['ABLY_ENV'] }
+      xcargs: { ABLY_ENV: ENV['ABLY_ENV'], CLANG_ANALYZER_OUTPUT: 'plist-html', CLANG_ANALYZER_OUTPUT_DIR: '$(pwd)/clang' }
```
QuintinWillison, line 9: This pattern is quite repetitive. Could we DRY this up? I also worry that the apparent, explicit choice of `clang` as the path segment name could overload / overlap with other similar folders. Perhaps a longer name like `clang-analyzer-output` would make this clearer - both here and in the upstream (workflow-located) `find` statement?

QuintinWillison (changes requested):

.github/workflows/check.yml
maratal: Sorry, it is a draft, I've just forgotten to convert it. I've picked these values randomly (possibly avoiding overlap with something else). `exit` in both paths of `if else` was for test purpose only.

lukasz-szyszkowski (approved):

.github/workflows/check.yml
maratal: 806e5c9ad6c1138bf0e3a98422695348b6574db7

.github/workflows/check.yml
maratal: 806e5c9ad6c1138bf0e3a98422695348b6574db7

fastlane/Fastfile
maratal: d048bdfc405458524b48de4221e7cde066665947

.github/workflows/check.yml
```diff
@@ -56,7 +56,23 @@ jobs:
...
+        run: |
+          if [[ -z $(find ./derived_data -name "report-*.html") ]]; then
+            echo "Static Analyzer found no issues."
+          else
+            echo "Static Analyzer found some issues. Failing build."
+            exit 1
+          fi
       
+      - name: Static Analyzer Reports Artifact
+        if: failure()
```
QuintinWillison, line 70: This is, unfortunately, a slightly flawed approach as this step will execute no matter which of the steps above it failed. So, for example, in [this run](https://github.com/ably/ably-cocoa/runs/4275016222?check_suite_focus=true):

```
Run actions/upload-artifact@v2
  with:
    name: static-analyzer-reports-test_macOS
    path: ./derived_data/**/StaticAnalyzer/**/*.html
    if-no-files-found: warn
  env:
    LC_CTYPE: en_US.UTF-8
    LANG: en_US.UTF-8
    ABLY_ENV: sandbox
Warning: No files were found with the provided path: ./derived_data/**/StaticAnalyzer/**/*.html. No artifacts will be uploaded.
```

Where the build failed at "Run Tests", not "Check Static Analyzer Output".

QuintinWillison (changes requested):
This can still be improved, revealed by the unrelated failure on the latest run of checks.

lukasz-szyszkowski (approved):

.github/workflows/check.yml
maratal: https://github.com/ably/ably-cocoa/pull/1213/commits/16dd4de638cf99b45810d9d33d45f2d426e575ff

QuintinWillison: I'm not really sure how https://github.com/ably/ably-cocoa/commit/16dd4de638cf99b45810d9d33d45f2d426e575ff solves this, as the later step guarded by `if: always()` will still execute - which was what I was commented on. 😕 

Have I missed something?

The `exit 1` in your new step will fail that step, failing the job, but then the job was already failed.

maratal: >later step guarded by if: always() will still execute - which was what I was commented on
I believe your initial comment was on line 70, which is "Static Analyzer Reports Uploading" step now, but "later step" you're referring is "Xcodebuild Logs Artifact" which is different thing (uploads all build logs, not SA html reports).

Now how it works: I fail build if reports found in the "Check Static Analyzer Output" step. Then, in case of failure next step "Static Analyzer Reports Uploading" is activated. And then, because it's impossible to use `run` section with pre-build step `uses` section, I've added one more step only to spit the message about failure and exit with... failure. Now I see that it is a complete nonsense, and it seems I've just failed to fail build failure here 🤷‍♂️😆 The real fix was in replacing `if: failure()` with `if: steps.analyzer-output.outcome == 'failure'` in "Static Analyzer Reports Artifact" (renamed to "Static Analyzer Reports Uploading").

maratal: @QuintinWillison 122720c2ee2c1b0030cd00b6dc8fa12025f627dd

QuintinWillison (changes requested):
Sorry, @maratal - perhaps I should have spotted this on first pass but I still think this isn't doing what you expect it might.

.github/workflows/check.yml
```diff
@@ -56,6 +56,24 @@ jobs:
 
       - name: Run Tests
         run: bundle exec fastlane ${{ matrix.lane }}
+
+      - name: Check Static Analyzer Output
+        id: analyzer-output
+        continue-on-error: true
```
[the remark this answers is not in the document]
maratal: 3d3e878b

Workflow `steps` control is very misleading TBH: https://github.community/t/workflow-skips-steps-besides-if-condition/115956

.github/workflows/check.yml
```diff
@@ -69,7 +68,7 @@ jobs:
           fi
       
       - name: Static Analyzer Reports Uploading
-        if: steps.analyzer-output.outcome == 'failure'
+        if: ${{ failure() }}
```
lukasz-szyszkowski:
```suggestion
        if: failure()
```
should be enough, brackets are necessary if you, for example, want to negate it: `if: ${{ !failure() }}`

maratal: Look at 3d3e878bd080519710493ffc4751c128993c3b47

lukasz-szyszkowski: ok, now it makes sense 👍

jamienewcomb:
@maratal what is the status of this one? are you wanting on another review from Quintin. If so, please re-request

.github/workflows/check.yml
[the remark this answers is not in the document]
QuintinWillison: I see - in concert with https://github.com/ably/ably-cocoa/pull/1213/commits/fcfc0711bc7605114f48d6e48623225e358c281f 😉 

Thanks 😄

QuintinWillison:
Sorry, @jamienewcomb - my bad - I was somewhat overloaded last few weeks so didn't come back to this. Reviewing again now.

QuintinWillison (approved):
Looks sensible from a workflows perspective, though I'll admit the solution as a whole feels a little fragile (i.e. failing due to second hand information, being presence of reports output). Either way, presumably this is better than what we had before so I do think it'll move the codebase forwards (even though I predict we'll probably end up looking at this again in the future... time will tell).

.github/workflows/check.yml
```diff
@@ -56,6 +56,23 @@ jobs:
 
       - name: Run Tests
         run: bundle exec fastlane ${{ matrix.lane }}
+
+      - name: Check Static Analyzer Output
+        id: analyzer-output
+        run: |
+          if [[ -z $(find ./derived_data -name "report-*.html") ]]; then
```
lawrence-forooghian, line 63: Shouldn't the search pattern used in this `find` (which is the same as the glob `./derived_data/**/report-*.html`) be consistent with the glob used in the "Static Analyzer Reports Uploading" step (`./derived_data/**/StaticAnalyzer/**/*.html`)?

(Of the two, probably better to use the more permissive one – a false positive is probably better than a false negative here, since at least they’ll be brought to our attention by a build failure, and we can then refine the glob if we discover false positives.)

.github/workflows/check.yml
```diff
@@ -56,6 +56,23 @@ jobs:
...
+            echo "Static Analyzer found no issues."
+          else
+            echo "Static Analyzer found some issues. HTML report will be available in Artifacts section. Failing build."
+            exit 1
+          fi
+      
+      - name: Static Analyzer Reports Uploading
+        if: ${{ failure() && steps.analyzer-output.outcome == 'failure' }}
```
lawrence-forooghian, line 71: Possibly just my lack of knowledge about GitHub Actions, but is the `failure()` check not redundant here, given we're also checking for failure of a specific step?

maratal: Yes, GitHub Actions are weird, all previous commits were basically fighting against this nonsense. https://github.community/t/workflow-skips-steps-besides-if-condition/115956

.github/workflows/check.yml
lawrence-forooghian: It might also be handy to add a comment to the script with an example of a path that you're expecting this glob to match.

.github/workflows/check.yml
lawrence-forooghian: 😱 👍

.github/workflows/check.yml
lawrence-forooghian: > probably better to use the more permissive one

Also, reading this again, I don't think there is a "more permissive one" – each is permissive in different ways.

maratal: 77f834ef2961f8f956ba93f65b144699d38e2d9e
I tried with less permissive with no luck (`find` spits that path doesn't exist). But more permissive works and all tests passed. Also it's obvious (just `/derived_data`), so is example comment there really needed?

lawrence-forooghian: OK, don't worry about the comment then.

lawrence-forooghian (approved):

PR #1218, opened 2021-11-08
Example app: Use relative path of Ably Cocoa

ben-xD:
Our example app currently uses an environment variable to find the SDK (Ably Cocoa). It's simpler to use a relative path, resulting in smaller `Package.swift` and `Package.resolved` files.

You may notice that this example app uses AblyDeltaCodec v1.3.2. Even though Ably Cocoa has not been released to use AblyDeltaCodec, `1.3.2` this example app uses it because Ably Cocoa only specifies a **minimum** version of AblyDeltaCodec.

### DIFF
--- a/Examples/SPM/Package.swift
+++ b/Examples/SPM/Package.swift
@@ -4,24 +4,6 @@
 import PackageDescription
 import Darwin.C
 
-// Setup ENV variables when using with Xcode IDE
-//
-//setenv("PACKAGE_URL", "", 1)
-//setenv("PACKAGE_REVISION", "", 1)
-
-func env(_ name: String) -> String? {
-    if let envPointer = getenv(name) {
-        return String(cString: envPointer)
-    } else {
-        print("Can't find environment \(name)")
-        return nil
-    }
-}
-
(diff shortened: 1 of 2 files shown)


QuintinWillison (approved):

maratal (approved):

Examples/SPM/Package.swift
```diff
@@ -4,32 +4,14 @@
...
-
 let package = Package(
     name: "SPMIntegration",
     platforms: [
         .iOS(.v9),
         .macOS(.v10_11)
     ],
     dependencies: [
-        .package(name:"ably-cocoa", url: packageURL, .revision(revision))
+        .package(path: "../../")
```
lukasz-szyszkowski, line 14: Since when is this possible? Is it works on Xcode 12?

lukasz-szyszkowski: Ok it works since 11.4 https://developer.apple.com/documentation/swift_packages/package/dependency/3582846-package nice 🙈

lukasz-szyszkowski (approved):

PR #1225, opened 2021-11-24
[Groundwork for #1201] Unroll test-defining loops in test files into `reusableTests` function calls

lawrence-forooghian:
## What does this do?

Unrolls some test-defining loops into individual calls to a reusable function. See first commit message for more details.

No, it's just a refactor.

## Why are we doing this?

It’s groundwork for removing the Quick testing framework using an automated migrator tool (#1201). The
migrator that I’m writing won’t be able to handle arbitrary statements like `forEach` calls or loops, so we need to replace them with something it knows how to handle, namely a call to a function named with a special `reusableTests` prefix.

### DIFF
--- a/Spec/RealtimeClientPresence.swift
+++ b/Spec/RealtimeClientPresence.swift
@@ -2889,75 +2889,80 @@ class RealtimeClientPresence: QuickSpec {
             }
 
             // RTP15e
-            let cases: [String:(ARTRealtimePresence, Optional<(ARTErrorInfo?)->Void>)->()] = [
-                "enterClient": { $0.enterClient("john", data: nil, callback: $1) },
-                "updateClient": { $0.updateClient("john", data: nil, callback: $1) },
-                "leaveClient": { $0.leaveClient("john", data: nil, callback: $1) }
-            ]
-            for (testCase, performMethod) in cases {
-                context(testCase) {
-                    it("should implicitly attach the Channel") {
-                        let client = ARTRealtime(options: AblyTests.commonAppSetup())
-                        defer { client.dispose(); client.close() }
-                        let channel = client.channels.get("test")
-
-                        expect(channel.state).to(equal(ARTRealtimeChannelState.initialized))
-                        waitUntil(timeout: testTimeout) { done in
-                            //Call: enterClient, updateClient and leaveClient
-                            performMethod(channel.presence) { errorInfo in
-                                expect(errorInfo).to(beNil())
-                                done()
-                            }
-                            expect(channel.state).to(equal(ARTRealtimeChannelState.attaching))
+            func reusableTestsTestPresencePerformMethod(_ performMethod: @escaping (ARTRealtimePresence, Optional<(ARTErrorInfo?)->Void>)->()) {
+                it("should implicitly attach the Channel") {
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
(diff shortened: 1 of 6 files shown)


Spec/RealtimeClientPresence.swift
```diff
@@ -2889,76 +2889,81 @@ class RealtimeClientPresence: QuickSpec {
...
-                            //Call: enterClient, updateClient and leaveClient
-                            performMethod(channel.presence) { errorInfo in
-                                expect(errorInfo).to(beNil())
-                                done()
-                            }
-                            expect(channel.state).to(equal(ARTRealtimeChannelState.attaching))
+            func reusableTestsTestPresencePerformMethod(_ performMethod: @escaping (ARTRealtimePresence, Optional<(ARTErrorInfo?)->Void>)->()) {
+                it("should implicitly attach the Channel") {
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                    defer { client.dispose(); client.close() }
```
lukasz-szyszkowski, line 2895: I don't like these inline calls and the use of the `;`. It looks better in new lines.
```suggestion
                    defer { 
                        client.dispose()
                        client.close() 
                    }
```

Spec/RealtimeClientPresence.swift
```diff
@@ -2889,76 +2889,81 @@ class RealtimeClientPresence: QuickSpec {
...
-                                    fail("Reason is empty"); done(); return
-                                }
-                                expect(reason.message).to(equal(expectedErrorMessage))
-                                done()
+                    expect(channel.state).to(equal(ARTRealtimeChannelState.attached))
+                }
+                
+                it("should result in an error if the channel is in the FAILED state") {
+                    let client = ARTRealtime(options: AblyTests.commonAppSetup())
+                    defer { client.dispose(); client.close() }
```
lukasz-szyszkowski, line 2912:
```suggestion
                    defer { 
                        client.dispose()
                        client.close() 
                    }
```

Spec/RealtimeClientPresence.swift
```diff
@@ -2889,76 +2889,81 @@ class RealtimeClientPresence: QuickSpec {
...
-                            (client.internal.transport as! TestProxyTransport).actionsIgnored += [.attached]
-                            //Call: enterClient, updateClient and leaveClient
-                            performMethod(channel.presence) { errorInfo in
-                                expect(errorInfo).to(equal(error.error))
-                                done()
+                }
+                
+                it("should result in an error if the channel moves to the FAILED state") {
+                    let client = AblyTests.newRealtime(AblyTests.commonAppSetup())
+                    defer { client.dispose(); client.close() }
```
lukasz-szyszkowski, line 2936:
```suggestion
                    defer { 
                        client.dispose()
                        client.close() 
                    }
```

Spec/RealtimeClientPresence.swift
lawrence-forooghian: I agree with all your comments about the formatting here. However, I didn't want to touch the formatting of the existing code, to make it clear that all I've done is move the existing code into a new function.

I'm planning to run a code formatter over all of the test code once the migration is complete, anyway, so this will get fixed then. Are you happy with this?

maratal (approved):
Seems nothing except lite refactoring, so not going deeply into the details I approve.

lukasz-szyszkowski (approved):

PR #1232, opened 2021-11-25
[Groundwork for #1201] Remove floating statements at `spec` / `describe` / `context` level

lawrence-forooghian:
## What does this do?

Removes some arbitrary statements from `spec` / `describe` / `context` level, moving them into related variable declarations.

No, it's just a refactor.

## Why are we doing this?

It’s groundwork for removing the Quick testing framework using an automated migrator tool (#1201). The migrator that I’m writing won’t be able to handle these statements, so I'm moving them inside statements that it knows how to handle. See the first commit message for more details.

### DIFF
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -4114,8 +4114,11 @@ class Auth : QuickSpec {
 
             // RSA8g RSA8c
             context("when using authUrl") {
-                let options = AblyTests.clientOptions()
-                options.authUrl = URL(string: echoServerAddress)!
+                let options: ARTClientOptions = {
+                    let options = AblyTests.clientOptions()
+                    options.authUrl = URL(string: echoServerAddress)!
+                    return options
+                }()
 
                 var keys: [String: String]!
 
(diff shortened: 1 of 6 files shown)


maratal (approved):

Spec/Auth.swift
```diff
@@ -4114,8 +4114,11 @@ class Auth : QuickSpec {
 
             // RSA8g RSA8c
             context("when using authUrl") {
-                let options = AblyTests.clientOptions()
-                options.authUrl = URL(string: echoServerAddress)!
+                let options: ARTClientOptions = {
```
QuintinWillison, line 4117: Gosh, quite a lot of boilerplate. This construction pattern got me thinking about [Kotlin's `apply` function](https://kotlinlang.org/docs/scope-functions.html#apply) - for example as used in one of other codebases [here](https://github.com/ably/ably-asset-tracking-android/blob/67afa76861f43885d1a4dbe280156b8334dc3846/common/src/main/java/com/ably/tracking/common/Ably.kt#L275).

I do wonder if there's an opportunity to make this less cumbersome. On a quick Google I found [this suggestion](https://stackoverflow.com/questions/43830927/how-to-write-a-generic-apply-function-in-swift/43831365#43831365), for example. 🤔

QuintinWillison (approved):
Happy to approve as the contents is pretty straightforward, however there may be an opportunity to improve if you think it's worthwhile.

Spec/Auth.swift
lawrence-forooghian: I'm in two minds about it. It's neater, but it's also something for a person new to the codebase to need to understand (although it's relatively self-explanatory). I think in the interests of not going off on a bit of a tangent, I'll keep it as it is.

PR #1233, opened 2021-11-25
[Groundwork for #1201, but also a fix] Fix setting of fallback retry timeout in tests

lawrence-forooghian:
## What does this do?

Moves the setting of `ARTDefault.fallbackRetryTimeout` from the top level of `context` to inside `beforeEach` blocks. See commit message for explanation of why I think it was an error to put this code at `context` level.

In theory, yes, although I think in practice it doesn’t since the existing code was accidentally working (see commit message).

## Why are we doing this?

Original motivation that caused me to notice this was #1232 (removing arbitrary statements at `context` level).

### DIFF
--- a/Spec/RestClient.swift
+++ b/Spec/RestClient.swift
@@ -1456,7 +1456,9 @@ class RestClient: QuickSpec {
                     
                     context("should restore default primary host after fallbackRetryTimeout expired") {
                         
-                        ARTDefault.setFallbackRetryTimeout(1.0)
+                        beforeEach {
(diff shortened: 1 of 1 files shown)


lawrence-forooghian:
@lukasz-szyszkowski I would, in particular, appreciate your opinion on this one, since I think you wrote the original test.

maratal (approved):
Commit message on point 👌

QuintinWillison (approved):

lukasz-szyszkowski (approved):

Issue #1137, opened 2021-06-15
Remove `.xcodeproj`, drop carthage support, Use `swift build` in Github workflow


ben-xD:
# Background

As part of adding SPM support, a `Package.swift` was added to Ably cocoa, allowing Xcode projects and other swift packages to easily add Ably as a dependency. Swift packages are separate entities to Xcode projects, and do not co-exist with Xcode projects very well:
- Xcode can already open a Swift Package, but you won't use project tabs (e.g. Build phases, build settings, Summary/ build configuration, etc). This is better, because configuration does not live in XML/ Xcode files anymore.
- Changes to the Xcode project will not be reflected in the `Package.swift`, and so we have to sources of configuration which can diverge in the future. 
- Finally, the `swift package generate-xcodeproj` has been deprecated, which was previously used to generate an ephemeral Xcode project based on the `Package.swift` so that you can still open the Swift Package as a Xcode project. It no longer has the updates.

# Tasks: 

In @lukasz-szyszkowski's PR, we are keeping the Ably.xcodeproj temporarily to avoid making too many changes. But these changes should happen soon to avoid developer confusion:
- Delete `Ably.xcodeproject`, relying solely on `Package.swift` for configuration (and other automatically generated files `.swiftpm`). Xcode schemes are autogenerated based on the `Package.swift`.
  - This will actually drop support for Carthage, because Carthage uses the shared schemes within `.xcodeproj` to build the output framework. Therefore, we can delete a lot of carthage related files (Cartfile/ .resolved), lines related to updating carthage dependencies in the `.github` yaml.
  - We would need to update the CI (Github Workflow) to use `swift build` instead of `xcodebuild`, or modify existing `xcodebuild` code.

lawrence-forooghian:
Just happened to be passing by here - one thing to mention is that we should probably do this as two separate pieces of work (for ease of review): first remove Carthage support, then remove the `.xcodeproj`.

Issue #1070, opened 2020-10-01
Replace Carthage with SPM for Ably Project and test suite


ricardopereira:
### Context

Regarding https://github.com/Carthage/Carthage/issues/3019, Carthage builds will fail in Xcode 12 when using beta or latest stable, the installation fails with something like: `error: Unable to determine target or target not included`.

We are using Carthage to build our dependencies while developing and running tests. Our customers can use Carthage to add our library to their project.

### The reason

The reason of this issue is the new Xcode 12 behaviour, which Carthage doesn't handle correctly right now. See https://github.com/Carthage/Carthage/issues/3019 for details.

### Current solution

We can wait until Carthage fix the issue or use the Swift Package Manager or CocoaPods. Since we need to add support to Swift Package Manager (https://github.com/ably/ably-cocoa/issues/819), I propose to use SPM in Ably project and test suite.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-543) by [Unito](https://www.unito.io)

QuintinWillison:
Can be closed if we decide to remove carthage entirely under #1137.

PR #1235, opened 2021-11-25
Fix Carthage #3019 workaround for Xcode 13

lawrence-forooghian:
This fixes the `carthage-with-workaround-for-issue-3019.sh` script to work for Xcode 13 as well as Xcode 12. I noticed this because GitHub seem to have upgraded our macOS version, causing the default Xcode version to increase (see commit message).

Whilst I was investigating, I also split the `check-pod` workflow into multiple steps, to make it easier to see which stage of the job failed.

### DIFF
--- a/.github/workflows/check-pod.yml
+++ b/.github/workflows/check-pod.yml
@@ -12,15 +12,23 @@ jobs:
 
       # Run the steps we document in the Release Process.
       # unzip commands included as proof-of-life for the Carthage output.
-      - run: |
-          ruby --version
-          echo -n "carthage version: " && carthage version
-          echo -n "pod version: " && pod --version --verbose
-          make --version
-          make update
-          make carthage_package
+      - name: Print Ruby version
+        run: ruby --version
+      - name: Print Carthage version
+        run: 'echo -n "carthage version: " && carthage version'
+      - name: Print CocoaPods version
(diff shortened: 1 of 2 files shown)


maratal (approved):

lukasz-szyszkowski (approved):

.github/workflows/check-pod.yml
```diff
@@ -12,15 +12,23 @@ jobs:
...
       # Run the steps we document in the Release Process.
       # unzip commands included as proof-of-life for the Carthage output.
-      - run: |
-          ruby --version
-          echo -n "carthage version: " && carthage version
-          echo -n "pod version: " && pod --version --verbose
-          make --version
-          make update
-          make carthage_package
+      - name: Print Ruby version
```
QuintinWillison, line 15: This is perhaps a bit OTT to have as multiple steps. I would have perhaps preferred a single step `name`d "Print Tooling Versions" or similar. This'll look a bit busy in the run UI.

QuintinWillison (approved):
Relates to #1070 and #1137.

.github/workflows/check-pod.yml
lawrence-forooghian: As agreed on Slack, we’ll merge to get CO fixed, and I’ll address this later.

PR #1239, opened 2021-11-29
[Groundwork for #1201] Rename or remove clashing variables

lawrence-forooghian:
## What does this do?

Removes or renames some variables at the `spec` / `describe` / `context` level.

No, it's just a refactor (moving variable declarations inside other ones, or renaming variables).

## Why are we doing this?

It’s groundwork for removing the Quick testing framework using an automated migrator tool (#1201). The migrator that I’m writing will pull all `spec` / `describe` / `context` level variable and function declarations in a file into a flattened file-wide list of global declarations. This might lead to clashes if multiple contexts use the same variable name. So we rename / remove things to avoid these clashes.

## Note

I had hoped to avoid this step, by making the migrator automatically rename variables to avoid clashes. I had a solution _almost_ working with [`swift-refactor`](https://github.com/apple/swift/tree/main/tools/swift-refactor) but for some reason that tool wasn't correctly updating variable references inside helper function declarations. I decided not to investigate this further, since it would probably end up taking a lot of time for something that could be fixed relatively quickly by hand. It's a shame, though.

### DIFF
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -4073,12 +4073,12 @@ class Auth : QuickSpec {
             let messageName = "message_JWT"
             
             context("client initialized with a JWT token in ClientOptions") {
-                let options = AblyTests.clientOptions()
+                let jwtTestsOptions = AblyTests.clientOptions()
 
                 context("with valid credentials") {
                     xit("pulls stats successfully") {
-                        options.token = getJWTToken()
-                        let client = AblyTests.newRealtime(options)
+                        jwtTestsOptions.token = getJWTToken()
+                        let client = AblyTests.newRealtime(jwtTestsOptions)
                         defer { client.dispose(); client.close() }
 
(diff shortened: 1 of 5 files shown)


lukasz-szyszkowski (approved):
I know it's just 'beautification' of the code but if it's possible you should create extension for 'JSON' type that will encode it. I don't like code repetition for encoding.
If it's not possible just ignore this comment 😁

maratal (approved):

lawrence-forooghian:
> I know it's just 'beautification' of the code but if it's possible you should create extension for 'JSON' type that will encode it. I don't like code repetition for encoding.
> If it's not possible just ignore this comment 😁

@lukasz-szyszkowski I'd quite like to keep the scope of this PR limited to just the necessary refactorings, since it's already quite a large one as it is. Perhaps you'd like to create an issue if you think it's a problem worth fixing?

PR #1240, opened 2021-11-29
Remove the Quick testing framework

lawrence-forooghian:
## What does this pull request do?

Removes the [Quick testing framework](https://github.com/Quick/Quick).

## Why are we doing this?

See #1201 for details. The main two reasons are:

- to get the ability to run individual tests using Xcode’s UI;
- using the Apple native test framework instead of a third-party library that will be unfamiliar to many Swift developers.

## How do we do it?

I wrote an [automated migrator tool](https://github.com/ably/quick-to-xctest-migrator) to do the bulk of the work here. (The code for the migrator is still rather messy. We're thinking about tidying it up to make it publicly available, but it's not there yet.)

## How is this pull request structured?

Each of the steps below is a commit which I recommend reviewing individually (else you'll struggle to make _any_ sense of this PR, I think):

1. (Migrator tool) Move all of the `spec` / `describe` / `context` level local variable and function declarations into `private` variables at the top of the file.
2. (Migrator tool) Convert each `QuickSpec` subclass into a subclass of `XCTestCase`. We replace the `spec` method declaration with a set of reusable functions and `func test…` instance methods.
3. (Manual) Step 2 has left the code quite messy (see [my comment on this PR about indentation](https://github.com/ably/ably-cocoa/pull/1240#discussion_r759249368)) so we run [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) (a code formatting tool) on it to sort this out.
4. (Manual) We replace the use of Quick's `configuration.beforeSuite` hook.
5. (Manual) We remove Quick as a dependency of the project.

## Does this change the behaviour of the tests?

For the most part, no. One of my main aims was for the migrated tests to behave exactly the same as the originals running under Quick. This turned out to be important, because we have tests that are dependent upon things like order of variable initialization and order of test execution (see my diff-level comments on this PR, and https://github.com/ably/ably-cocoa/issues/1241).

The only difference is in the behaviour of local variables in the `reusableTests` functions. See [this comment](https://github.com/ably/ably-cocoa/pull/1240#discussion_r759360311).

### How did I verify that the behaviour is unchanged?

The migrator has a `--add-logging` option, which adds logging statements to all of the `before/afterEach` and `it` calls. I used the `ably-cocoa-tests-linter-log-comparison` tool in the migrator repo to compare the logs emitted by the pre and post migration tests, and confirmed that everything runs in the same order.

(The tool doesn't verify the order of variable initialization; that I did in a more ad-hoc way, adding a few logging statements selectively by hand.)

## Notes

I've left comments on the PR for things that I wanted to point out, or which I'd like opinions on. Please ignore the "Outdated" tag for now; nothing is outdated yet 🙂

### DIFF
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -4,16 +4,114 @@ import Nimble
 import Quick
 import Aspects
 
+        
+        private var testHTTPExecutor: TestProxyHTTPExecutor!
+                private func testOptionsGiveDefaultAuthMethod(_ caseSetter: (ARTAuthOptions) -> Void) {
+                    let options = ARTClientOptions()
+                    caseSetter(options)
+                    
+                    let client = ARTRest(options: options)
+                    
+                    expect(client.auth.internal.method).to(equal(ARTAuthMethod.token))
+                }
+                // Cases:
+                //  - useTokenAuth is specified and thus a key is not provided
+                //  - authCallback and authUrl are both specified
+                private func testStopsClientWithOptions(caseSetter: (ARTClientOptions) -> ()) {
+                    let options = ARTClientOptions()
+                    caseSetter(options)
+                    
+                    expect{ ARTRest(options: options) }.to(raiseException())
+                }
+
+                private let currentClientId = "client_string"
+
+                private var options: ARTClientOptions!
+                private var rest: ARTRest!
+
+                private func tokenParamsTestsSetupDependencies() {
+                    if (options == nil) {
+                        options = AblyTests.commonAppSetup()
+                        options.clientId = currentClientId
+                        rest = ARTRest(options: options)
+                    }
+                }
+                private let json = "{" +
+                "    \"token\": \"xxxxxx.yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy\"," +
+                "    \"issued\": 1479087321934," +
+                "    \"expires\": 1479087363934," +
+                "    \"capability\": \"{\\\"test\\\":[\\\"publish\\\"]}\"," +
+                "    \"clientId\": \"myClientId\"" +
+                "}"
+
+                private func check(_ details: ARTTokenDetails) {
+                    expect(details.token).to(equal("xxxxxx.yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy"))
+                    expect(details.issued).to(equal(Date(timeIntervalSince1970: 1479087321.934)))
+                    expect(details.expires).to(equal(Date(timeIntervalSince1970: 1479087363.934)))
+                    expect(details.capability).to(equal("{\"test\":[\"publish\"]}"))
+                    expect(details.clientId).to(equal("myClientId"))
+                }
+            private let channelName = "test_JWT"
+            private let messageName = "message_JWT"
+                private let jwtTestsOptions = AblyTests.clientOptions()
+                private let authUrlTestsOptions: ARTClientOptions = {
+                    let options = AblyTests.clientOptions()
+                    options.authUrl = URL(string: echoServerAddress)!
+                    return options
+                }()
+
+                private var keys: [String: String]!
+
+                private func rsa8gTestsSetupDependencies() {
+                    if (keys == nil) {
+                        keys = getKeys()
+                    }
+                }
+                private let authCallbackTestsOptions = AblyTests.clientOptions()
+            private let jwtAndRestTestsOptions = AblyTests.clientOptions()
+                private var client: ARTRest!
+
+                private func jwtContentTypeTestsSetupDependencies() {
+                    if (client == nil) {
+                        let options = AblyTests.clientOptions()
+                        let keys = getKeys()
+                        options.authUrl = URL(string: echoServerAddress)!
+                        options.authParams = [URLQueryItem]()
+                        options.authParams?.append(URLQueryItem(name: "keyName", value: keys["keyName"]))
+                        options.authParams?.append(URLQueryItem(name: "keySecret", value: keys["keySecret"]))
+                        options.authParams?.append(URLQueryItem(name: "returnType", value: "jwt"))
+                        client = ARTRest(options: options)
+                    }
+                }
+
 class Auth : QuickSpec {
-    override func spec() {
+
+// XCTest invokes this method before executing the first test in the test suite. We use it to ensure that the global variables are initialized at the same moment, and in the same order, as they would have been when we used the Quick testing framework.
+override class var defaultTestSuite : XCTestSuite {
+    let _ = testHTTPExecutor
+    let _ = currentClientId
+    let _ = options
+    let _ = rest
+    let _ = json
+    let _ = channelName
+    let _ = messageName
+    let _ = jwtTestsOptions
+    let _ = authUrlTestsOptions
+    let _ = keys
+    let _ = authCallbackTestsOptions
+    let _ = jwtAndRestTestsOptions
+    let _ = client
+
+    return super.defaultTestSuite
+}
+
         
         struct ExpectedTokenParams {
             static let clientId = "client_from_params"
             static let ttl = 1.0
             static let capability = "{\"cansubscribe:*\":[\"subscribe\"]}"
         }
-        
-        var testHTTPExecutor: TestProxyHTTPExecutor!
+    override func spec() {
 
         describe("Basic") {
 
@@ -173,14 +271,6 @@ class Auth : QuickSpec {
 
             // RSA4
             context("authentication method") {
-                func testOptionsGiveDefaultAuthMethod(_ caseSetter: (ARTAuthOptions) -> Void) {
-                    let options = ARTClientOptions()
-                    caseSetter(options)
-                    
-                    let client = ARTRest(options: options)
-                    
-                    expect(client.auth.internal.method).to(equal(ARTAuthMethod.token))
-                }
                 
                 it("should be default auth method when options’ useTokenAuth is set") {
                     testOptionsGiveDefaultAuthMethod { $0.useTokenAuth = true; $0.key = "fake:key" }
@@ -525,15 +615,6 @@ class Auth : QuickSpec {
             
             // RSA14
             context("options") {
-                // Cases:
-                //  - useTokenAuth is specified and thus a key is not provided
-                //  - authCallback and authUrl are both specified
-                func testStopsClientWithOptions(caseSetter: (ARTClientOptions) -> ()) {
-                    let options = ARTClientOptions()
-                    caseSetter(options)
-                    
-                    expect{ ARTRest(options: options) }.to(raiseException())
-                }
                 
                 it("should stop client when useTokenAuth and no key occurs") {
                     testStopsClientWithOptions { $0.useTokenAuth = true }
@@ -1740,19 +1821,6 @@ class Auth : QuickSpec {
 
             // RSA8b
             context("should support all TokenParams") {
-
-                let currentClientId = "client_string"
-
-                var options: ARTClientOptions!
-                var rest: ARTRest!
-
-                func tokenParamsTestsSetupDependencies() {
-                    if (options == nil) {
-                        options = AblyTests.commonAppSetup()
-                        options.clientId = currentClientId
-                        rest = ARTRest(options: options)
-                    }
-                }
 
                 it("using defaults") {
                     tokenParamsTestsSetupDependencies()
@@ -4032,21 +4100,6 @@ class Auth : QuickSpec {
         describe("TokenDetails") {
             // TD7
             describe("fromJson") {
-                let json = "{" +
-                "    \"token\": \"xxxxxx.yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy\"," +
-                "    \"issued\": 1479087321934," +
-                "    \"expires\": 1479087363934," +
-                "    \"capability\": \"{\\\"test\\\":[\\\"publish\\\"]}\"," +
-                "    \"clientId\": \"myClientId\"" +
-                "}"
-
-                func check(_ details: ARTTokenDetails) {
-                    expect(details.token).to(equal("xxxxxx.yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy"))
-                    expect(details.issued).to(equal(Date(timeIntervalSince1970: 1479087321.934)))
-                    expect(details.expires).to(equal(Date(timeIntervalSince1970: 1479087363.934)))
-                    expect(details.capability).to(equal("{\"test\":[\"publish\"]}"))
-                    expect(details.clientId).to(equal("myClientId"))
-                }
 
                 it("accepts a string, which should be interpreted as JSON") {
                     check(try! ARTTokenDetails.fromJson(json as ARTJsonCompatible))
@@ -4069,11 +4122,8 @@ class Auth : QuickSpec {
         }
 
         describe("JWT and realtime") {
-            let channelName = "test_JWT"
-            let messageName = "message_JWT"
             
             context("client initialized with a JWT token in ClientOptions") {
-                let jwtTestsOptions = AblyTests.clientOptions()
 
                 context("with valid credentials") {
                     xit("pulls stats successfully") {
@@ -4114,19 +4164,6 @@ class Auth : QuickSpec {
 
             // RSA8g RSA8c
             context("when using authUrl") {
-                let authUrlTestsOptions: ARTClientOptions = {
-                    let options = AblyTests.clientOptions()
-                    options.authUrl = URL(string: echoServerAddress)!
-                    return options
-                }()
-
-                var keys: [String: String]!
-
-                func rsa8gTestsSetupDependencies() {
-                    if (keys == nil) {
-                        keys = getKeys()
-                    }
-                }
 
                 context("with valid credentials") {
                     it("fetches a channels and posts a message") {
@@ -4237,7 +4274,6 @@ class Auth : QuickSpec {
 
             // RSA8g
             context("when using authCallback") {
-                let authCallbackTestsOptions = AblyTests.clientOptions()
 
                 context("with valid credentials") {
                     xit("pulls stats successfully") {
@@ -4390,7 +4426,6 @@ class Auth : QuickSpec {
         
         // RSC1 RSC1a RSC1c RSA3d
         describe("JWT and rest") {
-            let jwtAndRestTestsOptions = AblyTests.clientOptions()
             
             context("when the JWT token embeds an Ably token") {
                 it ("pulls stats successfully") {
@@ -4420,20 +4455,6 @@ class Auth : QuickSpec {
             
             // RSA4f, RSA8c
             context("when the JWT token is returned with application/jwt content type") {
-                var client: ARTRest!
-
-                func jwtContentTypeTestsSetupDependencies() {
-                    if (client == nil) {
-                        let options = AblyTests.clientOptions()
-                        let keys = getKeys()
-                        options.authUrl = URL(string: echoServerAddress)!
-                        options.authParams = [URLQueryItem]()
-                        options.authParams?.append(URLQueryItem(name: "keyName", value: keys["keyName"]))
-                        options.authParams?.append(URLQueryItem(name: "keySecret", value: keys["keySecret"]))
-                        options.authParams?.append(URLQueryItem(name: "returnType", value: "jwt"))
-                        client = ARTRest(options: options)
-                    }
-                }
 
                 beforeEach {
                     jwtContentTypeTestsSetupDependencies()
--- a/Spec/Auth.swift
+++ b/Spec/Auth.swift
@@ -1,4514 +1,4440 @@
 import Ably
 import Ably.Private
+import Aspects
 import Nimble
 import Quick
-import Aspects
 
-        
-        private var testHTTPExecutor: TestProxyHTTPExecutor!
-                private func testOptionsGiveDefaultAuthMethod(_ caseSetter: (ARTAuthOptions) -> Void) {
-                    let options = ARTClientOptions()
-                    caseSetter(options)
-                    
-                    let client = ARTRest(options: options)
-                    
-                    expect(client.auth.internal.method).to(equal(ARTAuthMethod.token))
-                }
-                // Cases:
-                //  - useTokenAuth is specified and thus a key is not provided
-                //  - authCallback and authUrl are both specified
-                private func testStopsClientWithOptions(caseSetter: (ARTClientOptions) -> ()) {
-                    let options = ARTClientOptions()
-                    caseSetter(options)
-                    
-                    expect{ ARTRest(options: options) }.to(raiseException())
-                }
+private var testHTTPExecutor: TestProxyHTTPExecutor!
+private func testOptionsGiveDefaultAuthMethod(_ caseSetter: (ARTAuthOptions) -> Void) {
+    let options = ARTClientOptions()
+    caseSetter(options)
 
-                private let currentClientId = "client_string"
+    let client = ARTRest(options: options)
 
-                private var options: ARTClientOptions!
-                private var rest: ARTRest!
-
-                private func tokenParamsTestsSetupDependencies() {
-                    if (options == nil) {
-                        options = AblyTests.commonAppSetup()
-                        options.clientId = currentClientId
-                        rest = ARTRest(options: options)
-                    }
-                }
-                private let json = "{" +
-                "    \"token\": \"xxxxxx.yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy\"," +
-                "    \"issued\": 1479087321934," +
-                "    \"expires\": 1479087363934," +
-                "    \"capability\": \"{\\\"test\\\":[\\\"publish\\\"]}\"," +
-                "    \"clientId\": \"myClientId\"" +
-                "}"
-
-                private func check(_ details: ARTTokenDetails) {
-                    expect(details.token).to(equal("xxxxxx.yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy"))
-                    expect(details.issued).to(equal(Date(timeIntervalSince1970: 1479087321.934)))
-                    expect(details.expires).to(equal(Date(timeIntervalSince1970: 1479087363.934)))
-                    expect(details.capability).to(equal("{\"test\":[\"publish\"]}"))
-                    expect(details.clientId).to(equal("myClientId"))
-                }
-            private let channelName = "test_JWT"
-            private let messageName = "message_JWT"
-                private let jwtTestsOptions = AblyTests.clientOptions()
-                private let authUrlTestsOptions: ARTClientOptions = {
-                    let options = AblyTests.clientOptions()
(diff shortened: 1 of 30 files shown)


Spec/Auth.swift
```diff
@@ -3,17 +3,113 @@ import Ably.Private
 import Nimble
 import Quick
 import Aspects
+        
+        private var testHTTPExecutor: TestProxyHTTPExecutor!
```
lawrence-forooghian: You might be wondering why we pull the variables to global level instead of making them instance variables of the `XCTestCase` subclass. It's because this way, we can access them from inside each `test_` method as if they were local variables, without having to worry about memory management. If they were instance variables, then we’d have to worry about capturing `self` inside closures that refer to these variables.

Spec/Auth.swift
```diff
@@ -3,17 +3,113 @@ import Ably.Private
...
+                private func testOptionsGiveDefaultAuthMethod(_ caseSetter: (ARTAuthOptions) -> Void) {
+                    let options = ARTClientOptions()
+                    caseSetter(options)
+                    
+                    let client = ARTRest(options: options)
+                    
+                    expect(client.auth.internal.method).to(equal(ARTAuthMethod.token))
+                }
+                // Cases:
```
lawrence-forooghian: If the spacing here bothers people, I can update the migrator to add some newlines between declarations.

Spec/Auth.swift
```diff
@@ -3,17 +3,113 @@ import Ably.Private
...
+        
+        private var testHTTPExecutor: TestProxyHTTPExecutor!
+                private func testOptionsGiveDefaultAuthMethod(_ caseSetter: (ARTAuthOptions) -> Void) {
+                    let options = ARTClientOptions()
+                    caseSetter(options)
+                    
+                    let client = ARTRest(options: options)
+                    
+                    expect(client.auth.internal.method).to(equal(ARTAuthMethod.token))
+                }
```
lawrence-forooghian, line 1190: If the spacing here bothers people, I can update the migrator to add some newlines between declarations.

Spec/Auth.swift
```diff
@@ -3,17 +3,113 @@ import Ably.Private
...
+                        options.authParams?.append(URLQueryItem(name: "keySecret", value: keys["keySecret"]))
+                        options.authParams?.append(URLQueryItem(name: "returnType", value: "jwt"))
+                        client = ARTRest(options: options)
+                    }
+                }
 
 class Auth : QuickSpec {
-    override func spec() {
+
+override class var defaultTestSuite : XCTestSuite {
```
lawrence-forooghian: This `defaultTestSuite` override serves to make sure that all of the variables that were originally local variables at `spec` / `describe` / `context` level get initialized at the start of the test suite, to be consistent with the behaviour in Quick. We have some tests which will fail if this initialization order is not respected.

Why override `defaultTestSuite`? It's a bit of a hack, really. I just needed some method that'd be called at the start of the test suite, and I couldn't find any better override. I could have instead placed something _outside_ the individual test classes and used `XCTestObservation` to call some initialization method on each of the classes, but it would have meant that the migrator didn't output self-contained test files any longer. I think we can live with it, but if you disagree or would find it useful to add a comment to the code explaining what's going on, let me know.

We probably should fix this dependency on initialization order anyway, at which point we can get rid of this code.

Spec/Auth.swift
```diff
@@ -109,20 +109,19 @@ override class var defaultTestSuite : XCTestSuite {
...
             static let capability = "{\"cansubscribe:*\":[\"subscribe\"]}"
         }
-    override func spec() {
 
-        describe("Basic") {
+        
 
             // RSA1
-            it("should work over HTTPS only") {
+            func test__003__Basic__should_work_over_HTTPS_only() {
```
lawrence-forooghian: The number in the test name (`test__003__`) ensures that XCTest will run the tests in the same order as Quick would have run them. We have some tests which will fail if they are not run in the same order as they are in our Quick test suite.

lawrence-forooghian: Also notice that I've preserved the indentation level of the original `it` calls. This is to make the diff easy to review. We fix this later in this PR by running a code formatter.

Spec/Auth.swift
```diff
@@ -3503,13 +3466,11 @@ override class var defaultTestSuite : XCTestSuite {
...
                 }
-
-            }
             
             // RSA10k
-            context("server time offset") {
+            
 
-                xit("should obtain server time once and persist the offset from the local clock") {
+                func skipped__test__115__authorize__server_time_offset__should_obtain_server_time_once_and_persist_the_offset_from_the_local_clock() {
```
lawrence-forooghian: Notice that `xit` gets turned into a `skipped__test` method. What do you think about this? I know there also exists [`XCTSkip`](https://developer.apple.com/documentation/xctest/xctskip), an error which when thrown will skip the test case and I think mark it as skipped in the Xcode test navigator. Do you think it's worth me investigating or is this fine?

Spec/Auth.swift
```diff
@@ -4030,24 +3982,45 @@ override class var defaultTestSuite : XCTestSuite {
...
+}
+
+                func reusableTestsTestTokenRequestFromJson(_ json: String, testCase: TestCase_ReusableTestsTestTokenRequestFromJson, context: (beforeEach: (() -> ())?, afterEach: (() -> ())?), check: @escaping (_ request: ARTTokenRequest) -> Void) {
+                    func test__accepts_a_string__which_should_be_interpreted_as_JSON() {
+context.beforeEach?()
 
-        describe("TokenRequest") {
-            // TE6
-            describe("fromJson") {
-                func reusableTestsTestTokenRequestFromJson(_ json: String, check: @escaping (_ request: ARTTokenRequest) -> Void) {
```
lawrence-forooghian, line 3972: I'll explain how we handle `reusableTests…` functions, which are functions that we specially marked in #1225 as functions that contain calls to `context` / `it`. The migrator recognises these functions and does the following:

### How we transform the function declaration

In this example, that'd be the `func reusableTestsTestTokenRequestFromJson` declaration.

1. It flattens all of the `it` calls contained within the function declaration into a list of individual function declarations (e.g. `test__accepts_a_NSDictionary` here).
2. It creates a `TestCase_…` enum which describes each of the test cases contained within the function declaration (see `TestCase_ReusableTestsTestTokenRequestFromJson` here).
3. It adds a `testCase` param to the function declaration, allowing the caller to specify which test case they want to execute, and it adds a `switch` statement to the body of the function declaration which invokes the specified test case

### How we translate the function call

Spec/Auth.swift
```diff
@@ -4030,24 +3982,45 @@ override class var defaultTestSuite : XCTestSuite {
...
+    test__accepts_a_string__which_should_be_interpreted_as_JSON()
+case .accepts_a_NSDictionary:
+    test__accepts_a_NSDictionary()
+}
+
                 }
 
-                context("with TTL") {
+        
+            // TE6
```
lawrence-forooghian: TODO this should be above `reusableTestsTestTokenRequestFromJson`

Spec/Auth.swift
```diff
@@ -4449,16 +4415,17 @@ override class var defaultTestSuite : XCTestSuite {
...
                     }
                 }
-            }
             
             // RSA4f, RSA8c
-            context("when the JWT token is returned with application/jwt content type") {
+            
 
-                beforeEach {
+                func beforeEach__JWT_and_rest__when_the_JWT_token_is_returned_with_application_jwt_content_type() {
```
lawrence-forooghian: Here's an example of how the migrator handles `beforeEach` hooks. It turns each call into a function, and then invokes this function at the start of all of the affected `test__…` methods.

(It does the same kind of thing for `afterEach`, too.)

TODO link to more complicated example.

Spec/PushActivationStateMachine.swift
```diff
@@ -39,27 +37,37 @@ override class var defaultTestSuite : XCTestSuite {
...
+beforeEach()
+
                 let storage = MockDeviceStorage(startWith: ARTPushActivationStateWaitingForDeviceRegistration(machine: initialStateMachine))
                 rest.internal.storage = storage
                 let stateMachine = ARTPushActivationStateMachine(rest: rest.internal, delegate: StateMachineDelegate())
                 expect(stateMachine.current).to(beAKindOf(ARTPushActivationStateWaitingForDeviceRegistration.self))
                 expect(storage.keysRead).to(haveCount(2))
                 expect(storage.keysRead.filter({ $0.hasSuffix("CurrentState") })).to(haveCount(1))
                 expect(storage.keysWritten).to(beEmpty())
+afterEach()
```
lawrence-forooghian: Here's an example of an `afterEach` hook being called.

TODO put space before this

Spec/PushActivationStateMachine.swift
```diff
@@ -68,19 +76,25 @@ override class var defaultTestSuite : XCTestSuite {
...
                 expect(stateMachine.current).to(beAKindOf(ARTPushActivationStateAfterRegistrationSyncFailed.self))
+afterEach()
+
             }
 
             // RSH3a
-            context("State NotActivated") {
+            
+
+                func beforeEach__Activation_state_machine__State_NotActivated() {
```
lawrence-forooghian: Here's an example of a `beforeEach` hook inside a context, when the scope it’s nested within _also_ has a `beforeEach` hook. You can see that we turn this hook into an appropriately-named `beforeEach__…` method (`beforeEach__Activation_state_machine__State_NotActivated` here), and at the start of _this_ method, we call the `beforeEach` of its parent scope (`beforeEach` here), if any.

And then, at the start of the test case `test__…` methods inside this context, we call the appropriate `beforeEach_…` function.

Spec/PushActivationStateMachine.swift
```diff
@@ -91,41 +105,70 @@ override class var defaultTestSuite : XCTestSuite {
...
                 }
 
                 // RSH3a2
-                context("on Event CalledActivate") {
+                
+                    // RSH3a2a
+                    func test__Activation_state_machine__State_NotActivated__on_Event_CalledActivate__reusableTestsRsh3a2a(testCase: TestCase_ReusableTestsRsh3a2a) {
                     // RSH3a2a
-                    reusableTestsRsh3a2a()
+                    reusableTestsRsh3a2a(testCase: testCase, context: (beforeEach: beforeEach__Activation_state_machine__State_NotActivated, afterEach: afterEach))}
```
lawrence-forooghian: Here's an example of non-nil `before/afterEach` hooks being passed to a `reusableTests…` function.

Spec/PushActivationStateMachine.swift
```diff
@@ -91,41 +105,70 @@ override class var defaultTestSuite : XCTestSuite {
...
                             guard let deviceSecret = rest.device.secret else {
                                 fail("Device Secret should be available because it's loaded when the getter of the property is called"); return
                             }
                             guard let data = Data(base64Encoded: deviceSecret) else {
                                 fail("Device Secret should be encoded with Base64"); return
                             }
-                            expect(data.count) == 32 //32 bytes digest
+                            expect(data.count) == 32 
+afterEach()
+//32 bytes digest
```
lawrence-forooghian: TODO this comment has gotten messed up

Spec/PushActivationStateMachine.swift
```diff
@@ -544,26 +649,56 @@ override class var defaultTestSuite : XCTestSuite {
...
+afterEach()
+
                 }
 
                 // RSH3d2
-                context("on Event CalledDeactivate") {
-                    reusableTestsRsh3d2()
-                }
+                
+                    func test__Activation_state_machine__State_WaitingForNewPushDeviceDetails__on_Event_CalledDeactivate__reusableTestsRsh3d2(testCase: TestCase_ReusableTestsRsh3d2) {
```
lawrence-forooghian: TODO this indentation a little messed up

Spec/PushActivationStateMachine.swift
```diff
@@ -544,26 +649,56 @@ override class var defaultTestSuite : XCTestSuite {
...
+                    reusableTestsRsh3d2(testCase: testCase, context: (beforeEach: beforeEach__Activation_state_machine__State_WaitingForNewPushDeviceDetails, afterEach: afterEach))}
+func test__031__Activation_state_machine__State_WaitingForNewPushDeviceDetails__on_Event_CalledDeactivate__should_use_custom_deregisterCallback_and_fire_Deregistered_event() {
```
lawrence-forooghian: TODO newline before this

Spec/RealtimeClientConnection.swift
```diff
@@ -1461,13 +1448,11 @@ override class var defaultTestSuite : XCTestSuite {
...
 
-            }
-
             // RTN10
-            context("serial") {
+            
 
                 // RTN10a
-                it("should be -1 once connected") {
+                func test__042__Connection__serial__should_be__1_once_connected() {
```
lawrence-forooghian: This is an unfortunate renaming, not sure what to do about it, maybe we just go and fix it up manually afterwards. I'll see how many other names have symbols that give meaning.

Spec/RealtimeClientConnection.swift
```diff
@@ -4019,10 +4047,10 @@ override class var defaultTestSuite : XCTestSuite {
                 expect(realtime.connection.key).to(equal(connectedProtocolMessage.connectionDetails!.connectionKey))
             }
 
-            xcontext("Transport disconnected side effects") {
```
lawrence-forooghian: Notice that an `xcontext` gets flattened into a list of skipped test cases.

Spec/Stats.swift
```diff
@@ -66,49 +66,120 @@ override class var defaultTestSuite : XCTestSuite {
...
-                reusableTestsTestAttribute("persisted")
-            }
+                func test__Stats__all__reusableTestsTestAttribute(testCase: TestCase_ReusableTestsTestAttribute) {
+                reusableTestsTestAttribute("all", testCase: testCase, context: (beforeEach: nil, afterEach: nil))}
+func test__001__Stats__all__should_return_a_MessagesTypes_object() {
+test__Stats__all__reusableTestsTestAttribute(testCase: .should_return_a_MessagesTypes_object)
+}
+
+func test__002__Stats__all__should_return_value_for_message_counts() {
+test__Stats__all__reusableTestsTestAttribute(testCase: .should_return_value_for_message_counts)
```
lawrence-forooghian: TODO intendation

Spec/Stats.swift
```diff
@@ -66,49 +66,120 @@ override class var defaultTestSuite : XCTestSuite {
...
+func test__007__Stats__persisted__should_return_value_for_all_data_transferred() {
+test__Stats__persisted__reusableTestsTestAttribute(testCase: .should_return_value_for_all_data_transferred)
+}
+
+func test__008__Stats__persisted__should_return_zero_for_empty_values() {
+test__Stats__persisted__reusableTestsTestAttribute(testCase: .should_return_zero_for_empty_values)
+}
+
+enum TestCase_ReusableTestsTestDirection {
+case should_return_a_MessageTraffic_object
```
lawrence-forooghian: TODO indentation

Spec/Stats.swift
```diff
@@ -119,103 +190,199 @@ override class var defaultTestSuite : XCTestSuite {
...
                     expect(channelsTestsSubject?.peak).to(equal(10))
                 }
 
                 // TS2
-                it("should return zero for empty values") {
+                func test__022__Stats__channels__should_return_zero_for_empty_values() {
                     expect(channelsTestsSubject?.refused).to(equal(0))
                 }
-            }
+enum TestCase_ReusableTestsTestRequestType {
```
lawrence-forooghian: TODO newline before

Spec/ObjectLifetimes.swift
```diff
@@ -1,143 +1,136 @@
 import Ably
-import Quick
 import Nimble
-            private let options: ARTClientOptions = {
-                let options = ARTClientOptions(key: "fake:key")
-                options.autoConnect = false
-                return options
-            }()
+import Quick
+private let options: ARTClientOptions = {
```
lawrence-forooghian, line 4: TODO there's no breathing space here

Spec/Stats.swift
lawrence-forooghian: (SwiftFormat will indent these cases, don't worry.)

Spec/Stats.swift
lawrence-forooghian: (SwiftFormat will indent these statements, don't worry.)

Spec/PushActivationStateMachine.swift
lawrence-forooghian: (Observation, doesn't block review)

The migrator has moved this comment to after the `afterEach` call for some reason. I'll see if I can figure out why, but if it's a huge amount of work then I'll just fix this manually (it seems to be the only such affected comment).

Spec/Auth.swift
lawrence-forooghian: (Observation, doesn't block review)

The migrator has moved put this comment in the wrong place for some reason. It should be above  the declaration of `reusableTestsTestTokenRequestFromJson`. I'll see if I can figure out why, but if it's a huge amount of work then I'll just fix this manually (it seems to be the only such affected comment).

Spec/PushActivationStateMachine.swift
lawrence-forooghian: (Observation, doesn't block review)

I'll update the migrator to put the newline _after_ `afterEach`, not before. We'd like this breathing space and SwiftFormat won't help us here.

Spec/Auth.swift
lawrence-forooghian: It's also worth pointing out that the `reusableTests…` functions are the only place where the behaviour of the migrated tests differs from that of the original Quick tests. In the original Quick specs, the local variables inside the `reusableTests…` function declaration are only initialized once. In the migrated specs, they're initialized once for _each test case within the function declaration_. If I had unlimited time, I would have fixed this, but I didn't consider it worthwhile since it doesn't seem to have introduced any issues.

Spec/Auth.swift
maratal: It's fine IMHO

Spec/PushActivationStateMachine.swift
maratal: Probably it could be just two additional parameters instead of one tuple. Just a suggestion.

lawrence-forooghian: Yeah, that's a good suggestion. I can change that.

Spec/Auth.swift
maratal: Sometimes repeating code is not a crime. Especially testing code. I would just unwrap it into individual `test_*` functions. Just a suggestion.

Spec/Auth.swift
```diff
@@ -1,4527 +1,4437 @@
...
+
+private let json = "{" +
+    "    \"token\": \"xxxxxx.yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy\"," +
+    "    \"issued\": 1479087321934," +
+    "    \"expires\": 1479087363934," +
+    "    \"capability\": \"{\\\"test\\\":[\\\"publish\\\"]}\"," +
+    "    \"clientId\": \"myClientId\"" +
+    "}"
+
+private func check(_ details: ARTTokenDetails) {
```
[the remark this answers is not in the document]
lawrence-forooghian: I think that refactoring the existing code is outside the scope of this PR, which is already big as it is…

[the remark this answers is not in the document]
lawrence-forooghian: I agree that perhaps some function names lose their meaning when pulled out of their context, but I feel trying to fix up these names could take a lot of time.

Spec/Auth.swift
maratal: > you'd expand the entire reusableTests function, removing the function declaration entirely

Yes.

> which is not bad just because it's repetition, but also because it'll make this diff a lot harder to review.

Agreed. It's already there.

Spec/Auth.swift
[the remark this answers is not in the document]
maratal: Yes, it's already a heck of a boat load of changes, such minutiae can be processed later.

maratal (approved):

Spec/RealtimeClientConnection.swift
lawrence-forooghian: I added functionality to the migrator to warn when it thinks it's encountered a test where removing a symbol might change the name. And I've opened https://github.com/ably/ably-cocoa/pull/1242 to address these warnings by changing the test descriptions.

Spec/Auth.swift
lukasz-szyszkowski: What about `keeping states` by vars? Are these global vars stateless?
We have to be 100% sure that none of the var keeps state that could disrupt any other tests.

This problem isn't probably in the scope of these changes, but we have to re-check it later.

Spec/Auth.swift
lukasz-szyszkowski: what about this https://developer.apple.com/documentation/xctest/xctestcase/1496262-setup ?

>The setUp() class method is called exactly once for a test case, before its first test method is called. Override this method to customize the initial state for all tests in the test case.

Spec/Auth.swift
lukasz-szyszkowski: That's bad, I've created Issue for this https://github.com/ably/ably-cocoa/issues/1243

Spec/Auth.swift
lukasz-szyszkowski: I'm curious why not https://developer.apple.com/documentation/xctest/xctest/1500341-setup ? 
>Before each test begins, XCTest calls setUp(completion:) for asynchronous state preparation, then setUpWithError(), followed by this method. Override this method to reset state for each test method.

Spec/PushActivationStateMachine.swift
```diff
@@ -39,27 +37,37 @@ override class var defaultTestSuite : XCTestSuite {
             initialStateMachine = ARTPushActivationStateMachine(rest: rest.internal, delegate: StateMachineDelegate())
         }
 
-        afterEach {
+        func afterEach() {
```
lukasz-szyszkowski: this one could be converted to https://developer.apple.com/documentation/xctest/xctest/1500463-teardown
>After each test completes, XCTest calls tearDown(), then tearDownWithError(), followed by tearDown(completion:). Override this method to perform any per-test cleanup.

when using it, there is no need to call this method in every test method

lukasz-szyszkowski (approved):

lukasz-szyszkowski (approved):

lukasz-szyszkowski (approved):

Spec/Auth.swift
```diff
@@ -1,4510 +1,4437 @@
...
+                    fail("tokenDetails is nil"); done(); return
+                }
+                expect(tokenDetails.clientId).to(equal(expectedTokenParams.clientId))
+                done()
+            }
+        }
+    }
+
+    // RSA8f1
+    func test__058__requestToken__ensure_the_message_published_does_not_have_a_clientId() {
```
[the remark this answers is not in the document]
lawrence-forooghian: Yep, it's a similar situation to what I mentioned in [this comment](https://github.com/ably/ably-cocoa/pull/1240#discussion_r759247799).

Quick will run all of the `it`s directly contained within a `spec` / `describe` / `context` before diving into any nested `context`s. In this particular case, `test__075…` is nested in an additional `context` inside that which contains `test__058…`, hence this ordering.

Spec/Auth.swift
lawrence-forooghian: Hey @lukasz-szyszkowski, is your question "do these global variables, which are accessed from multiple test cases, mean that we have state which is shared between test cases and which might cause one test case to be affected by another"? If so, the answer is yes. And yes, it's a problem, and it's probably a large part of the reason why the tests are so order-dependent (#1241, #1243). But this is not a new problem – it's the same behaviour as before, where multiple `it` blocks shared the same `spec` / `describe` / `context`- level local variables.

Spec/Auth.swift
lawrence-forooghian: The difference is that `setUp()` is called before the tests in _that file_ are run. To be consistent with the behaviour of Quick, I need to set up the global variables _for all test files_ before even the first test of the test suite runs. Does that make sense?

Spec/Auth.swift
lawrence-forooghian: Good question. The reason I didn't use `setUp` / `tearDown` is that, since Quick allows you to have `context`-level `before/afterEach` hooks, each test case might have its own setup and teardown steps, depending on which `context`s it's nested within. (See the link in the grandparent comment for an example of this.)

Now, that doesn't fully answer your question, I realise. We _could_ turn the _file-wide_ `before/afterEach` hooks (e.g. those which are contained within `spec`) into `setUp` / `tearDown` methods, and then in the test cases we'd only have to call any `context`-specific hooks. But I don't think we can get away from having to call _some_ hooks directly in the test cases. Would you like me to look into making this change?

Spec/PushActivationStateMachine.swift
lawrence-forooghian: Updated the migrator.

Spec/PushActivationStateMachine.swift
lawrence-forooghian: Please see [this answer](https://github.com/ably/ably-cocoa/pull/1240#discussion_r760095128).

Spec/Auth.swift
lukasz-szyszkowski: Yes, that's what I meant. Is there any Issue with this case?

Spec/Auth.swift
lukasz-szyszkowski: It makes sense, but it's called four times from four different classes with different globals. In this case, I don't understand it.

Spec/Auth.swift
lukasz-szyszkowski: This is something that could be changed later. Leave it as it is. Thanks for the explanation.

Spec/ObjectLifetimes.swift
lawrence-forooghian: I've updated the migrator to add a newline before the first global declaration that it inserts.

Spec/Auth.swift
lawrence-forooghian: Sorry, but I don't think I understand the question, do you think you could explain a bit more what you'd like to know?

Spec/PushActivationStateMachine.swift
lawrence-forooghian: I've made this change to the migrator.

lawrence-forooghian: I've also made it so that these parameters have a default value of `nil`, which tidies things up at the point where we call these functions in a lot of cases.

lawrence-forooghian:
@lukasz-szyszkowski in the end I decided to implement [your suggestion of using `setUp` / `tearDown`](https://github.com/ably/ably-cocoa/pull/1240#discussion_r760108555), in the manner that I described there (convert `spec`-level `beforeEach` to `setUp`, and then only make explicit calls to `context`-level hooks). It makes the output quite a bit more concise, and since we're going to have to live with this code for a long time, it seemed worth the time.

@maratal I made the `reusableTests` calls more concise by making the `beforeEach` and `afterEach` params have a default value of `nil`. This clears up the call sites considerably.

PR #1253, opened 2021-12-12
Reset local device

### LINKED ISSUE #1177: Push device deregistration does not clear/ reset device details
This is quite low priority, but here it is:

As per [RSH3g2a](https://docs.ably.io/client-lib-development-guide/features/#RSH3g2a):
> On On event `Deregistered`, Clears all local DeviceDetails.

Unfortunately, ably-cocoa does not clear local device details.

In Java, this can be seen by:
https://github.com/ably/ably-java/blob/2b8e3c7e65ee7109f225cad115df38d61869fc3f/android/src/main/java/io/ably/lib/push/ActivationStateMachine.java#L483

Ably-cocoa does not even have a method to reset the device details, and so is definitely not calling this. 

This can be seen functionality in the Ably flutter app, in the following recording:

https://user-images.githubusercontent.com/24711048/133272418-87678115-c3be-49b4-a6c6-55ac8d23d112.mov

* On the left is Ably-flutter running on Android. I activate and deactivate, and the device details clearly changes.
* On the right is Ably-flutter running on iOS, and activation/ deactivation doesn't change the device details.
* iOS :bug:: When deactivating the device, the LocalDevice should be **reset**. It should not keep the same local device information. This means the deviceId and ~APNs device token (registration tokens on Android)~ could be **different**. The APNs device token on iOS and FCM registration token on Android remains the same. A different APNs device token might be possible on iOS, potentially by calling [`unregisterForRemoteNotifications`](https://developer.apple.com/documentation/uikit/uiapplication/1623093-unregisterforremotenotifications). You could test this.

Note on the flutter app: The app is just calling the device method to show the device information.

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-1181) by [Unito](https://www.unito.io)

maratal:
This PR substitutes #1184 (see explanation why in [there](https://github.com/ably/ably-cocoa/pull/1184#issuecomment-991985312)).
Closes #1177

### DIFF
--- a/Source/ARTLocalDevice.m
+++ b/Source/ARTLocalDevice.m
@@ -125,4 +125,8 @@ NSString *const ARTDevicePushTransportType = @"apns";
     [self clearStorage];
 }
 
+- (BOOL)isReset {
+    return [self.id isEqualToString:@""];
+}
+
 @end
--- a/Source/ARTLocalDevice.m
+++ b/Source/ARTLocalDevice.m
@@ -110,4 +110,19 @@ NSString *const ARTDevicePushTransportType = @"apns";
     return _identityTokenDetails != nil;
 }
 
+- (void)clearStorage {
+    for (NSString *key in @[ ARTDeviceIdKey, ARTDeviceIdentityTokenKey ]) {
+        [self.storage setObject:nil forKey:key];
+    }
+}
+
+- (void)reset {
+    self.id = @"";
+    self.secret = nil;
+    self.clientId = nil;
+    _identityTokenDetails = nil;
+    [self.push.recipient removeAllObjects];
+    [self clearStorage];
+}
+
 @end
--- a/Source/ARTPush+Private.h
+++ b/Source/ARTPush+Private.h
@@ -36,4 +36,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+@interface NSData (APNSToken)
+
+- (NSString *)apnsTokenString;
+
+@end
+
 NS_ASSUME_NONNULL_END
--- a/Source/ARTRest+Private.h
+++ b/Source/ARTRest+Private.h
@@ -65,6 +65,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 #if TARGET_OS_IOS
 - (void)resetDeviceSingleton;
+- (ARTLocalDevice *)loadDevice;
 #endif
 
 @end
--- a/Source/ARTRest+Private.h
+++ b/Source/ARTRest+Private.h
@@ -85,6 +85,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (nullable NSString *)apnsDeviceToken;
 - (void)setAndPersistAPNSDeviceToken:(nullable NSString *)deviceToken;
+- (void)resetDevice;
 
 @end
 #endif
--- a/Source/ARTRest+Private.h
+++ b/Source/ARTRest+Private.h
@@ -78,6 +78,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+#if TARGET_OS_IOS
 @interface ARTRestInternal (Storage)
 
 + (dispatch_queue_t)storageQueue;
@@ -86,5 +87,6 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)setAndPersistAPNSDeviceToken:(nullable NSString *)deviceToken;
 
 @end
+#endif
 
 NS_ASSUME_NONNULL_END
--- a/Source/ARTRest+Private.h
+++ b/Source/ARTRest+Private.h
@@ -78,4 +78,13 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+@interface ARTRestInternal (Storage)
+
++ (dispatch_queue_t)storageQueue;
+
+- (nullable NSString *)apnsDeviceToken;
+- (void)setAndPersistAPNSDeviceToken:(nullable NSString *)deviceToken;
+
+@end
+
 NS_ASSUME_NONNULL_END
--- a/Source/ARTRest.m
+++ b/Source/ARTRest.m
@@ -762,7 +762,7 @@ static ARTLocalDevice *_sharedDevice;
 
 @end
 
-
+#if TARGET_OS_IOS
 @implementation ARTRestInternal (Storage)
 
 + (dispatch_queue_t)storageQueue {
@@ -789,3 +789,4 @@ static ARTLocalDevice *_sharedDevice;
 }
 
 @end
+#endif
--- a/Source/ARTRest.m
+++ b/Source/ARTRest.m
@@ -716,6 +716,9 @@ dispatch_async(_queue, ^{
 }
 
 #if TARGET_OS_IOS
+
+static ARTLocalDevice *_sharedDevice;
+
 - (ARTLocalDevice *)device {
     __block ARTLocalDevice *ret;
     dispatch_sync(_queue, ^{
@@ -723,9 +726,6 @@ dispatch_async(_queue, ^{
     });
     return ret;
 }
-
-// Store address of once_token to access it in debug function.
-static dispatch_once_t *device_once_token;
 
 - (ARTLocalDevice *)device_nosync {
     // The device is shared in a static variable because it's a reflection
@@ -736,17 +736,14 @@ static dispatch_once_t *device_once_token;
     // As a side effect, the first instance "wins" at setting the device's
     // client ID.
 
-    static dispatch_once_t once;
-    device_once_token = &once;
-    static id device;
-    dispatch_once(&once, ^{
-        device = [ARTLocalDevice load:self.auth.clientId_nosync storage:self.storage];
-    });
-    return device;
+    if (_sharedDevice == nil) {
+        _sharedDevice = [ARTLocalDevice load:self.auth.clientId_nosync storage:self.storage];
+    }
+    return _sharedDevice;
 }
 
 - (void)resetDeviceSingleton {
-    if (device_once_token) *device_once_token = 0;
+    _sharedDevice = nil;
 }
 #endif
 
--- a/Source/ARTRest.m
+++ b/Source/ARTRest.m
@@ -761,3 +761,31 @@ static ARTLocalDevice *_sharedDevice;
 #endif
 
 @end
+
+
+@implementation ARTRestInternal (Storage)
+
++ (dispatch_queue_t)storageQueue {
+    static dispatch_queue_t queue;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        queue = dispatch_queue_create("io.ably.device-storage", DISPATCH_QUEUE_SERIAL);
+    });
+    return queue;
+}
+
+- (NSString *)apnsDeviceToken {
+    __block NSString *value;
+    dispatch_sync(ARTRestInternal.storageQueue, ^{
+        value = [[self device_nosync] apnsDeviceToken];
+    });
+    return value;
+}
+
+- (void)setAndPersistAPNSDeviceToken:(NSString *)deviceToken {
+    dispatch_sync(ARTRestInternal.storageQueue, ^{
+        [[self device_nosync] setAndPersistAPNSDeviceToken:deviceToken];
+    });
+}
+
+@end
--- a/Source/ARTRest.m
+++ b/Source/ARTRest.m
@@ -742,6 +742,19 @@ static ARTLocalDevice *_sharedDevice;
     return _sharedDevice;
 }
 
+/*
+ This function is needed to return functioning local device, since after `reset` call
+ you will end up with empty string as a device's `id` on the server.
+ */
+- (ARTLocalDevice *)loadDevice {
+    ARTLocalDevice *device = [self device_nosync];
+    if (device.isReset) {
+        [self resetDeviceSingleton];
+        device = [self device_nosync];
+    }
+    return device;
+}
+
 - (void)resetDeviceSingleton {
     _sharedDevice = nil;
 }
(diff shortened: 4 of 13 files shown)


Source/ARTLocalDevice.m
```diff
@@ -110,4 +110,23 @@ - (BOOL)isRegistered {
...
+    self.id = @"";
+    self.secret = nil;
+    self.clientId = nil;
+    _identityTokenDetails = nil;
+    [self.push.recipient removeAllObjects];
+    [self clearStorage];
+}
+
+- (BOOL)isReset {
+    return [self.id isEqualToString:@""];
```
maratal, line 129: @lawrence-forooghian I know this looks weird, but is a `nullable` `id` any better? Also additional flag, as I said earlier, probably also worse than this. Until the example app be able to handle `nil` device this is the best I could come up with. WDYT?

Source/ARTLocalDevice.m
```diff
@@ -110,4 +110,23 @@ - (BOOL)isRegistered {
     return _identityTokenDetails != nil;
 }
 
+- (void)clearStorage {
```
lukasz-szyszkowski, line 113: Do we have 100% sure that `self.storage` operations will be always synchronized?

Source/ARTLocalDevice.m
lukasz-szyszkowski: Maybe using `extern NSString * const` as a `not set value` will be more elegant? You can even compare it using `==` (memory address)

maratal: or `self.secret == nil` WDYT @lukasz-szyszkowski  @lawrence-forooghian @ben-xD

Source/ARTLocalDevice.m
maratal: https://github.com/ably/ably-cocoa/pull/1253/commits/a88c282c0b88e58eb22adf7b272c49b6e649e264

Source/ARTLocalDevice.m
lukasz-szyszkowski: for me `nil` is way better than the empty string

Source/ARTRest.m
```diff
@@ -742,6 +742,19 @@ - (ARTLocalDevice *)device_nosync {
     return _sharedDevice;
 }
 
+/*
+ This function is needed to return functioning local device, since after `reset` call
+ you will end up with empty string as a device's `id` on the server.
+ */
+- (ARTLocalDevice *)loadDevice {
```
lawrence-forooghian, line 749: Since this calls `device_noSync`, should this also have `noSync` in its name to indicate that it's the caller's responsibility to ensure that this method is called on the correct queue? (And then make sure that we respect this requirement in the places where we call this method.)

Source/ARTRest.m
```diff
@@ -742,6 +742,19 @@ - (ARTLocalDevice *)device_nosync {
     return _sharedDevice;
 }
 
+/*
+ This function is needed to return functioning local device, since after `reset` call
```
lawrence-forooghian, line 746: `-[ARTRest device]` is still part of the public interface of the library, so I think that method needs to always return a "functioning"  local device without needing to first call `loadDevice` (which isn't even a part of the public interface). Could we not move the logic of "if the device has been reset, then create a new one" into `-[ARTRest device_noSync]`?

Source/ARTPush+Private.h
```diff
@@ -36,4 +36,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+@interface NSData (APNSToken)
+
+- (NSString *)apnsTokenString;
```
lawrence-forooghian, line 41: Might it be a good idea to prefix this name with `art_`?

Source/ARTRest.m
```diff
@@ -788,5 +788,12 @@ - (void)setAndPersistAPNSDeviceToken:(NSString *)deviceToken {
     });
 }
 
+- (void)resetDevice {
+    dispatch_sync(ARTRestInternal.storageQueue, ^{
+        ARTLocalDevice *device = [self device_nosync];
```
lawrence-forooghian, line 793: My understanding of the `*_noSync` methods is that they need to be manually synchronised _with the rest of the library_ – i.e. they need to be called on the `clientOptions`’ `internalDispatchQueue`. Although you're calling `device_noSync` on _a_ queue here, I don't _think_ that's enough.

Source/ARTRest+Private.h
```diff
@@ -78,4 +78,13 @@ NS_ASSUME_NONNULL_BEGIN
...
 @end
 
+@interface ARTRestInternal (Storage)
+
++ (dispatch_queue_t)storageQueue;
+
+- (nullable NSString *)apnsDeviceToken;
+- (void)setAndPersistAPNSDeviceToken:(nullable NSString *)deviceToken;
+
+@end
```
lawrence-forooghian, line 90: I'm not 100% sure what the purpose of this new, synchronised storage is. But, just going by symmetry, would it not need to also include `setAndPersistIdentityTokenDetails:`?

Source/ARTLocalDevice.m
lawrence-forooghian: I think my answer to this question probably depends on @maratal's answer to [this question](https://github.com/ably/ably-cocoa/pull/1253#discussion_r772363251).

Source/ARTRest.m
maratal: I tried that, but it immediately re-creates the device and thus it's never can be in a `isReset` state which was the whole point of these changes and underlying issue.

Source/ARTRest+Private.h
maratal: > what the purpose of this new, synchronised storage is
It was discussed in slack here https://ably-real-time.slack.com/archives/CSQEKCE81/p1638465303293500?thread_ts=1638354865.280100&cid=CSQEKCE81

> would it not need to also include setAndPersistIdentityTokenDetails:
Agreed, but it wasn't used anywhere, so I decided to implement only the bare minimum.

Source/ARTRest.m
maratal: Originally the caller in `ARTPushActivationStateWaitingForDeregistration` did this:

```
ARTLocalDevice *local = self.machine.rest.device_nosync;
[local setAndPersistIdentityTokenDetails:nil];
```

So I thought that 

```
- (void)resetDevice {
    dispatch_sync(ARTRestInternal.storageQueue, ^{
        ARTLocalDevice *device = [self device]; // <-- ARTRest queue sync call
```

Would be too much. I'm still not sure about this though.

Source/ARTRest+Private.h
lawrence-forooghian: Sorry, perhaps I'm not understanding - what do you mean by "it wasn't used anywhere"? `setAndPersistIdentityTokenDetails` is still called in `ARTPushActivationState.m`.

lawrence-forooghian: > It was discussed in slack here

OK, thanks for the link. I think I understand the motivation for the new queue a bit better now, but is it directly relevant to this PR, which is about resetting device details? It seems to me like it might be a separate change

Source/ARTRest.m
lawrence-forooghian: The difference is that the original `ARTPushActivationStateWaitingForDeregistration` code 
[was always called on the `ARTRest` internal queue](https://github.com/ably/ably-cocoa/blob/3b6e33508450395c0b64357632a2897a4c585d06/Source/ARTPushActivationStateMachine.m#L88-L92) and hence could use the `*_noSync` methods. So I think you need to use the `device` method here — why do you think it might be "too much"?

Source/ARTRest.m
lawrence-forooghian: I think that the issue being described by #1177 actually consists of two separate things:

1. When the device deregisters itself for push notifications, it keeps its existing device details (`id`, `clientId`, etc), hence violating [RSH3g2a](https://docs.ably.io/client-lib-development-guide/features/#RSH3g2a).
2. Once we've fixed point 1, then we should make sure we respect [RSH8b](https://ably.com/documentation/client-lib-development-guide/features#RSH8b):

   > The `LocalDevice` `id` and `deviceSecret` attributes are generated, and persisted as part of the LocalDevice state, when required by step (RSH3a2b) in the Activation State Machine.

   That is, after the device has deregistered itself for push notifications, it should have "empty" (whatever that actually means) `id` and `deviceSecret` until the user tries to register the device for push notifications again. This is what @ben-xD showed happening on Android in his video.

So, I think that point 2 above is what you're referring to by "thus it's never can be in a `isReset` state", but I disagree that it "was the whole point of these changes" – we're also addressing point 1 above.

Furthermore, I think that trying to address point 2 as part of this PR brings additional complication, because we have _never_ respected RSH8b in our codebase. That is, at the moment, we generate the device `id` and `deviceSecret` the first time that `-[ARTRest device]` is called, instead of waiting until the user tries to register the device for push notifications. That's why we're now finding ourselves trying to answer questions like "should a reset device have an `id` of `@""`?" If we were already respecting RSH8b, then we'd already have answers to these questions.

I think that perhaps respecting RSH8b should be a separate task. And for me (possibly from lack of knowledge of the spec) it raises questions like those you've already identified – if there are API calls that require the device’s ID, which can be called before the push activation process is started, then where are we meant to get the ID from?

Source/ARTRest+Private.h
maratal: > It seems to me like it might be a separate change.

Please see this comment https://github.com/ably/ably-cocoa/pull/1253#discussion_r773931486

maratal: > `setAndPersistIdentityTokenDetails` is still called in `ARTPushActivationState.m`

Yes, I should inspect more places where storage queue access should be implemented, thanks.

Source/ARTRest.m
maratal: Just to clarify to whoever will read this:

[RSH3g2a](https://docs.ably.io/client-lib-development-guide/features/#RSH3g2a) - reset device after deactivation (nullify all fields)
[RSH8b](https://docs.ably.io/client-lib-development-guide/features/#RSH8b) - re-initializing device's fields with the new `id` and `deviceSecret` on event `CalledActivate`.

maratal: > I think that perhaps respecting RSH8b should be a separate task

Agreed

> if there are API calls that require the device’s ID, which can be called before the push activation process is started

I don't think so, it wouldn't make any sense. Or maybe I misunderstood what you've meant here. All functions that compose requests with the device `id` doesn't do any checks on it though:

`NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[[NSURL URLWithString:@"/push/deviceRegistrations"] URLByAppendingPathComponent:local.id]];`

maratal: @lawrence-forooghian https://github.com/ably/ably-cocoa/issues/1256

lawrence-forooghian: > Or maybe I misunderstood what you've meant here.

So, my comment was in response to [your code comment on `-[ARTRest localDevice]`](https://github.com/ably/ably-cocoa/blob/a88c282c0b88e58eb22adf7b272c49b6e649e264/Source/ARTRest.m#L746-L747):

>  This function is needed to return functioning local device, since after `reset` call
 you will end up with empty string as a device's `id` on the server.

I interpreted that comment as meaning the following: 

After calling `reset`, the `localDevice`'s `id` becomes an empty string. If we make an API request that uses the device's `id` whilst in this empty string state, then we'll submit an empty `id` to the server, hence we "will end up with empty string as a device's `id` on the server".

Did I misunderstand?

Source/ARTRest+Private.h
lawrence-forooghian: I've spent a while trying to understand this storage queue better, but it might be best if we have a call to talk about it, to prevent this becoming a long back-and-forth. What do you think?

But, here's where my current thinking is at. I still don't understand why it's necessary to introduce a new queue in order to achieve the overall objective of this pull request.

Looking at what's contained in the `ARTRestInternal (Storage)` category:

- `apnsDeviceToken` / `setAndPersistAPNSDeviceToken` – access to the APNS device token was not synchronised in the past, why does the fact that we might now reset the device introduce a need to synchronise it?
- `resetDevice` – this method is already being called on a serial queue because it's only called inside a push activation state machine transition method, so why do we need an extra serial queue?

It might be that, if you agree with my [analysis](https://github.com/ably/ably-cocoa/pull/1253#discussion_r773872600) that we don't need to put the device into an intermediate "reset" state, then some of this might simplify when you make that change, too.

maratal: > then some of this might simplify

Sure, upon splitting this PR into two, this one is going to contain only the very first [commit](https://github.com/ably/ably-cocoa/pull/1253/commits/a240961239e9bab77eddf31cb8cd6402fe8378fe) or so.

> we don't need to put the device into an intermediate "reset" state

I agreed with the splitting this PR, we still need something that marks the device as "reset", no? Lets discuss it in a call.

Source/ARTRest.m
maratal: > Did I misunderstand?

No, all is correct, but:

> so I think that method (`device`) needs to always return a "functioning" local device without needing to first call loadDevice

How is it possible to have an empty `id` then? It's either this or that ("reset" state or always "functioning"). That's why `loadDevice` exist - to load it on demand (when the function _needs_ it). Anyways lets further discuss it in a call.

maratal: > if there are API calls that require the device’s ID, which can be called before the push activation process is started

Or maybe you've meant wether `id` might be required somewhere else outside of the scope of push notifications? Then no, AFAIK.

maratal:
Closed in favor of https://github.com/ably/ably-cocoa/pull/1259

PR #1258, opened 2021-12-24
XCTUnwrap instead of guard

maratal:
Since the purpose of this PR is to improve readability of the tests, I've only replaced places that were overwhelmed with `guard` statements followed one after another. Also standalone cases where clause were one of a bunch mentioned in previous sentence (in most cases).

### DIFF
--- a/Spec/Tests/AuthTests.swift
+++ b/Spec/Tests/AuthTests.swift
@@ -73,6 +73,11 @@ private func rsa8gTestsSetupDependencies() {
 private let authCallbackTestsOptions = AblyTests.clientOptions()
 private let jwtAndRestTestsOptions = AblyTests.clientOptions()
 private var client: ARTRest!
+private var jsonEncoder: ARTJsonLikeEncoder = {
+    let encoder = ARTJsonLikeEncoder()
+    encoder.delegate = ARTJsonEncoder()
+    return encoder
+}()
 
 private func jwtContentTypeTestsSetupDependencies() {
     if client == nil {
@@ -209,10 +214,7 @@ class AuthTests: XCTestCase {
             }
         }
 
-        guard let currentToken = client.internal.options.token else {
-            fail("No access token")
-            return
-        }
+        let currentToken = try XCTUnwrap(client.internal.options.token, "No access token")
 
         let expectedAuthorization = "Bearer \(currentToken)"
 
@@ -596,7 +598,7 @@ class AuthTests: XCTestCase {
     // RSA4c
 
     // RSA4c1 & RSA4c2
-    func test__032__Token__options__if_an_attempt_by_the_realtime_client_library_to_authenticate_is_made_using_the_authUrl_or_authCallback__the_request_to_authUrl_fails__if_the_connection_is_CONNECTING__then_the_connection_attempt_should_be_treated_as_unsuccessful() {
+    func test__032__Token__options__if_an_attempt_by_the_realtime_client_library_to_authenticate_is_made_using_the_authUrl_or_authCallback__the_request_to_authUrl_fails__if_the_connection_is_CONNECTING__then_the_connection_attempt_should_be_treated_as_unsuccessful() throws {
         let options = AblyTests.clientOptions()
         options.autoConnect = false
         options.authUrl = URL(string: "http://echo.ably.io")!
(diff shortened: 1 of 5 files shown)


Spec/Tests/AuthTests.swift
```diff
@@ -73,6 +73,11 @@ private func rsa8gTestsSetupDependencies() {
 private let authCallbackTestsOptions = AblyTests.clientOptions()
 private let jwtAndRestTestsOptions = AblyTests.clientOptions()
 private var client: ARTRest!
+private var jsonEncoder: ARTJsonLikeEncoder = {
+    let encoder = ARTJsonLikeEncoder()
+    encoder.delegate = ARTJsonEncoder()
+    return encoder
+}()
```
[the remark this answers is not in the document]
maratal: Since it was used to improve readability, I think yes

lukasz-szyszkowski (approved):

Spec/Tests/AuthTests.swift
```diff
@@ -2970,18 +2936,11 @@ class AuthTests: XCTestCase {
         }
     }
 
-    func test__107__authorize__should_adhere_to_all_requirements_relating_to__authUrl_with_json() {
-        guard let tokenDetails = getTestTokenDetails() else {
-            XCTFail("TokenDetails is empty")
-            return
-        }
-
-        let encoder = ARTJsonLikeEncoder()
```
lawrence-forooghian, line 2991: I was wondering how this change (extracting the `encoder` variable) is related to the rest of the changes in this commit (using `XCTUnwrap`)? Also, could you explain the motivation for the change? It introduces shared state between test cases, which is something we should be trying to _reduce_, not add, I think.

lawrence-forooghian (commented):
Nice changes 👍 – just one question.

Spec/Tests/AuthTests.swift
maratal: About motivation behind this change - https://github.com/ably/ably-cocoa/pull/1258#discussion_r775505729
Since `ARTJsonLikeEncoder`/`ARTJsonEncoder` doesn't keep any state by themselves and are merely just a wrappers around `NSJSONSerialization` and `NSDictionary` I think it's fine to hide them behind this global var. WDYT @lukasz-szyszkowski  @QuintinWillison

Spec/Tests/AuthTests.swift
```diff
@@ -3045,19 +2992,8 @@ class AuthTests: XCTestCase {
...
-        do {
-            encodedTokenRequest = try encoder.encode(tokenRequest)
-        } catch {
-            fail("Encode failure: \(error)")
-            return
-        }
-        guard let tokenRequestJSON = String(data: encodedTokenRequest, encoding: .utf8) else {
-            XCTFail("JSON Token Request is empty")
-            return
-        }
```
lukasz-szyszkowski, line 3060: I approved this PR before but I have doubts looking at this?
1. it should be changed in a separate PR
2. don't we need to check if we can encode `tokenRequest` anymore?

Spec/Tests/AuthTests.swift
lukasz-szyszkowski: [Here](https://github.com/ably/ably-cocoa/pull/1258#discussion_r780922925) is my comment

lawrence-forooghian: I think it still adds mental overhead, somebody has to know that those classes don't have any internal state, and it's an undocumented assumption which we have to make sure we don't violate by accident. I don't think it costs much to just create a new one each time:

```swift
private func createJsonEncoder() -> ARTJsonLikeEncoder {
     let encoder = ARTJsonLikeEncoder()
     encoder.delegate = ARTJsonEncoder()
     return encoder
 }
```

and in tests

```swift
let jsonTokenDetails = try XCTUnwrap(createJsonEncoder().encode(testTokenDetails), "Invalid TokenDetails")
```

I think that if we didn't already have loads of shared state in the tests, it would bother me less, but since doing #1240 we have a lot already.

Spec/Tests/AuthTests.swift
maratal: > don't we need to check if we can encode tokenRequest anymore?
Could you elaborate on this? I mean maybe you've seen changes in the docs related to this or some other hints.

maratal: With this [comment](https://github.com/ably/ably-cocoa/pull/1258#discussion_r784825822) in mind can we agree that collapsing these lines into `XCTUnwrap` here can go into this PR as well?

lukasz-szyszkowski: My bad 🙈 it's correct, I found it here @maratal

lukasz-szyszkowski (approved):

lawrence-forooghian (approved):

Issue #1207, opened 2021-11-04
Encrypt messages if a new ARTRealtimeChannelOptions or ARTChannelOptions containing cipher is set


ben-xD:
When a `ARTChannel` is instantiated, the `_dataEncoder` is initialized using the cipher passed in the options argument **correctly**.

https://github.com/ably/ably-cocoa/blob/ad22530e65a1dc6874b5c5e8c4983eeb9366315f/Source/ARTChannel.m#L18-L32

Unfortunately, if a user sets the options afterwards, this cipher argument is completely ignored. Ably cocoa only updates the field. Unfortunately, this field is not used. Ably cocoa should update the encoder to encrypt data if the channel options contains a cipher.

https://github.com/ably/ably-cocoa/blob/ad22530e65a1dc6874b5c5e8c4983eeb9366315f/Source/ARTChannel.m#L46-L54

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1348) by [Unito](https://www.unito.io)

ben-xD:
`ARTRealtimeChannelOptions` might have the same issue

PR #1266, opened 2022-01-19
When setting realtime channel options, pass through to REST channel

### LINKED ISSUE #1265: Updating `ARTRealtimeChannel`’s options doesn’t update the options of its `ARTRestChannelInternal`
When calling `-[ARTRealtimeChannel setOptions:callback:]`, the passed options are not used to update the options of this instance’s REST channel. This means that, for example, if we update the realtime channel’s cipher options, these options will not be used to decrypt messages subsequently fetches using the `-history:*` methods.

I think this is something we missed when doing #1207.

This is causing https://github.com/ably/ably-flutter/issues/296.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1496) by [Unito](https://www.unito.io)

lawrence-forooghian:
When calling `-[ARTRealtimeChannel setOptions:callback:]`, the passed options are not used to update the options of this instance’s REST channel. This means that, for example, if we update the realtime channel’s cipher options, these options will not be used to decrypt messages subsequently fetched using the `-history:*` methods.

I think this is something we missed when fixing #1207.

This issue is causing ably/ably-flutter#296.

Closes #1265.

### DIFF
--- a/Spec/Test Utilities/NSObject+TestSuite.m
+++ b/Spec/Test Utilities/NSObject+TestSuite.m
@@ -17,7 +17,7 @@
     return [self aspect_hookSelector:selector withOptions:AspectPositionAfter usingBlock:^(id<AspectInfo> info) {
         __autoreleasing id arg;
         [[info originalInvocation] getArgument:&arg atIndex:2+index];
-        callback([arg copy]);
+        callback(arg);
     } error:nil];
 }
 
--- a/Source/ARTRealtimeChannel.m
+++ b/Source/ARTRealtimeChannel.m
@@ -1235,6 +1235,7 @@ dispatch_sync(_queue, ^{
 
 - (void)setOptions_nosync:(ARTRealtimeChannelOptions *_Nullable)options callback:(nullable ARTCallback)callback {
     [self setOptions_nosync:options];
+    [self.restChannel setOptions_nosync:options];
 
     if (!options.modes && !options.params) {
         if (callback)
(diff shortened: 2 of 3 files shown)


Spec/Test Utilities/NSObject+TestSuite.m
```diff
@@ -17,7 +17,7 @@ @implementation NSObject (TestSuite)
     return [self aspect_hookSelector:selector withOptions:AspectPositionAfter usingBlock:^(id<AspectInfo> info) {
         __autoreleasing id arg;
         [[info originalInvocation] getArgument:&arg atIndex:2+index];
-        callback([arg copy]);
+        callback(arg);
```
maratal, line 20: Why would they use copy here? 🤔 Just curious

maratal (approved):

lukasz-szyszkowski (approved):

ikbalkaya (approved):

Spec/Test Utilities/NSObject+TestSuite.m
QuintinWillison: I'm guessing that @lawrence-forooghian has removed this because `copy` is generally shallow in nature. 🤔 ... Regardless, this is test suite code so not a product implementation change. 😄

QuintinWillison (approved):

Spec/Test Utilities/NSObject+TestSuite.m
QuintinWillison: I'm going to resolve this conversation on @maratal's behalf, so that I can land this PR. Please feel free to continue the conversation or create an issue if you feel there is more work to be done on the test suite later. Thanks.

lawrence-forooghian: I'd actually addressed this in my [commit message](https://github.com/ably/ably-cocoa/pull/1266/commits/dd272cfb76d5337c178a6e4a15250603db4819bb).

PR #1269, opened 2022-01-20
Fix/1268 push error field

maratal:


lawrence-forooghian (commented):
Perhaps before we start trying to parse this field, we should get it added to the [documentation for the "get details from a registered device" etc REST API endpoints](https://docs.ably.io/rest-api/#get-device-registration)?

lawrence-forooghian:
No, I think that’s describing what the response looks like if there’s an error in responding to the HTTP request. The error we’re talking about here is nested inside an object that represents a device, and I believe may be present even if the response has a 200 status.

lukasz-szyszkowski:
> No, I think that’s describing what the response looks like if there’s an error in responding to the HTTP request. The error we’re talking about here is nested inside an object that represents a device, and I believe may be present even if the response has a 200 status.

agree 👍

lawrence-forooghian:
@maratal looks like this isn't ready for review and there's been no activity on it for a while, so I've converted it to a draft.

maratal:
Blocked by docs https://github.com/ably/specification/issues/23#issuecomment-1295033935

Issue #1273, opened 2022-01-21
Put all check jobs for each platform in a different GH workflow


maratal:
Now in case of one workflow job failure it requires to re-run all jobs in this workflow. It takes a few hours before needed "combination" of all of them being green will pop up. We should de-couple it to have an ability to re-run only the ones that failed.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1507) by [Unito](https://www.unito.io)

QuintinWillison:
@maratal and I have discussed this via Slack. I've also been thinking the same for a while, across a few of our SDKs.

- needs to align with our new(er) guidance for workflows: https://github.com/ably/engineering/blob/main/sdk/github.md#workflows
- perhaps helped by this feature, but this theory is not validated yet so needs research: https://docs.github.com/en/actions/using-workflows/reusing-workflows

PR #1274, opened 2022-01-21
Split multiple jobs in one workflow to multiple workflows

maratal:
based on a single workflow

Closes https://github.com/ably/ably-cocoa/issues/1273


maratal:
Executing in parallel:
Screenshot 2022-01-22 at 01 37 10

One failed:
Screenshot 2022-01-22 at 02 08 23

Re-run the one that failed:
Screenshot 2022-01-22 at 02 09 38

.github/workflows/check-iOS11.yml
```diff
@@ -0,0 +1,12 @@
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  check:
+    uses: ably/ably-cocoa/.github/workflows/check-base.yml@feature/1273-one-job-per-workflow
```
lukasz-szyszkowski, line 9: I'm not feeling good about this? What will happen when we merge this PR and remove `feature/1273-one-job-per-workflow` branch?
It would be better if we use `composite action` for this type of workflow ([more here](https://damienaicheh.github.io/github/actions/2021/10/07/reuse-workflows-composite-action-en.html)) I know it's a bit more of work to do but IMO this is a more elegant solution.

QuintinWillison (changes requested):
This is definitely going in the right direction, also acknowledging @lukasz-szyszkowski's comment. Thanks, @maratal. 😁 I'll probably pick this PR up myself later and make some more tweaks as might just be quicker for me to make changes rather than try to instruct you as to what I would like to see.

.github/workflows/check-iOS11.yml
maratal: Sorry, I forgot to mention that after we agree upon this PR I'll change the branch into `main` there, but PR will fail and we must be landing it "as-is", because once it will be in the `main` itself it will work perfectly. WDYT @QuintinWillison

lukasz-szyszkowski: In that case LGTM.

lukasz-szyszkowski (approved):

QuintinWillison:
ok, I get you, @maratal. 😁 I'm going to push a couple of commits to the branch underlying this PR that will:

- Remove the branch reference
- Add `name` keys to root of workflows

And perhaps other tweaks.

I am planning that the checks will then fail when run in this PR context, but that it will then pass once landed to `main` branch. As administrator I can override the checks requirement when I land the PR for this special case.

QuintinWillison:
Still following [the plan I outlined earlier](https://github.com/ably/ably-cocoa/pull/1274#issuecomment-1020368145), except now waiting for checks to pass after my most recent set of commits as they were quite extensive in the changes they made. Once I'm happy with them then I shall push another commit to make the references relative (implies I _have_ to remove those version suffixes) and then land this PR to `main` branch.

QuintinWillison:
I've just re-run the tvOS and macOS jobs, both of which experienced 1 or 2 test failures... but just the fact that I was able to re-run _just_ those test runtime environments was a breath of fresh air! 😁

QuintinWillison:
tvOS succeeded this time but macOS failed again, so I've re-run it.

maratal:
Looks like macOS decided to play little bit 🤪

QuintinWillison (approved):

maratal:
According to this commit message b436640 which I've copied from CI failure it required to include branch or tag name, also according to this commit message 4413eeb they require a full repo name as well. Hence I'm not sure if `main` works now as expected.

PR #1282, opened 2022-02-07
Fix/1280 unique test channel names

### LINKED ISSUE #1280: Unique test channel names
Now all channels used in tests have a chance to be overlapped by different executing at the same time tests with unpredictable results (failure most of the time or success by accident) because channel names are not unique enough.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1548) by [Unito](https://www.unito.io)

maratal:
Closes #1280

### DIFF
--- a/Spec/Test Utilities/TestUtilities.swift
+++ b/Spec/Test Utilities/TestUtilities.swift
@@ -46,6 +46,26 @@ let testTimeout = DispatchTimeInterval.seconds(20)
 let testResourcesPath = "ably-common/test-resources/"
 let echoServerAddress = "https://echo.ably.io/createJWT"
 
+func uniqueChannelName(prefix: String = "",
+                       testIdentifier: String = #function,
+                       timestamp: TimeInterval = Date.timeIntervalSinceReferenceDate) -> String {
+    let platform: String
+#if targetEnvironment(macCatalyst)
+    platform = "macCatalyst"
+#elseif os(OSX)
+    platform = "OSX"
+#elseif os(iOS)
+    platform = "iOS"
+#elseif os(tvOS)
+    platform = "tvOS"
+#elseif os(watchOS)
+    platform = "watchOS"
+#else
+    platform = "Unknown"
+#endif
+    return "\(prefix)-\(platform)-\(testIdentifier.replacingOccurrences(of: "()", with: ""))-\(timestamp)"
+}
+
 /// Common test utilities.
 class AblyTests {
 
--- a/Spec/Test Utilities/TestUtilities.swift
+++ b/Spec/Test Utilities/TestUtilities.swift
@@ -63,7 +63,7 @@ func uniqueChannelName(prefix: String = "",
 #else
     platform = "Unknown"
 #endif
-    return "\(prefix)-\(platform)-\(testIdentifier.replacingOccurrences(of: "()", with: ""))-\(timestamp)"
+    return "\(prefix)-\(platform)-\(testIdentifier.replacingOccurrences(of: "()", with: ""))-\(timestamp)-\(NSUUID().uuidString)"
 }
 
 /// Common test utilities.
--- a/Spec/Test Utilities/TestUtilities.swift
+++ b/Spec/Test Utilities/TestUtilities.swift
@@ -441,8 +441,8 @@ class PublishTestMessage {
     var completion: ((ARTErrorInfo?) -> Void)? = nil
     var error: ARTErrorInfo? = nil
 
-    init(client: ARTRest, failOnError: Bool = true, completion: ((ARTErrorInfo?) -> Void)? = nil) {
-        client.channels.get("test").publish(nil, data: "message") { error in
+    init(client: ARTRest, channelName: String, failOnError: Bool = true, completion: ((ARTErrorInfo?) -> Void)? = nil) {
+        client.channels.get(channelName).publish(nil, data: "message") { error in
             self.error = error
             if let callback = completion {
                 callback(error)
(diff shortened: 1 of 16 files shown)


Spec/Tests/AuthTests.swift
```diff
@@ -4320,11 +4320,13 @@ class AuthTests: XCTestCase {
     func test__002__should_accept_authURL_response_with_timestamp_argument_as_string() throws {
         var originalTokenRequest: ARTTokenRequest!
         let tmpRest = ARTRest(options: AblyTests.commonAppSetup())
+        
+        let channelName = uniqueChannelName()
         waitUntil(timeout: testTimeout) { done in
             let tokenParams = ARTTokenParams()
             tokenParams.clientId = "john"
             tokenParams.capability = """
-            {"chat:*":["publish","subscribe","presence","history"]}
```
lukasz-szyszkowski, line 4327: what does this do `*`? Is it just naming or some special character/option?

maratal: It works _only_ with an exact channel name, otherwise it gives `Error 40160 - action not permitted`. So I assume that's how original author named this channel.

Spec/Test Utilities/TestUtilities.swift
```diff
@@ -46,6 +46,26 @@ let testTimeout = DispatchTimeInterval.seconds(20)
...
+#elseif os(iOS)
+    platform = "iOS"
+#elseif os(tvOS)
+    platform = "tvOS"
+#elseif os(watchOS)
+    platform = "watchOS"
+#else
+    platform = "Unknown"
+#endif
+    return "\(prefix)-\(platform)-\(testIdentifier.replacingOccurrences(of: "()", with: ""))-\(timestamp)"
```
lawrence-forooghian, line 66: Do you have an idea of the resolution of this timestamp? We have some tests that generate two channel names in a row. In that case, I think that even millisecond resolution might not guarantee uniqueness. Could we put a UUID in here somewhere?

lawrence-forooghian (commented):
A heroic job, well done 💪 – a few comments

Spec/Tests/RestClientChannelTests.swift
```diff
@@ -300,7 +300,7 @@ class RestClientChannelTests: XCTestCase {
         let client = AblyTests.newRealtime(options)
         defer { client.dispose(); client.close() }
         waitUntil(timeout: testTimeout) { done in
-            client.channels.get("RSA7e1")
+            client.channels.get(uniqueChannelName(prefix: "RSA7e1"))
```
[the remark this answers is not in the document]
maratal: I just decided to leave it as "tag", since the original test didn't use standard name "test" probably for a reason. So I've decided to add an additional recognition to where author of the original test did.

Spec/Tests/RealtimeClientTests.swift
```diff
@@ -1382,7 +1382,7 @@ class RealtimeClientTests: XCTestCase {
     // Issue https://github.com/ably/ably-cocoa/issues/640
     func test__009__RealtimeClient__should_dispatch_in_user_queue_when_removing_an_observer() {
         class Foo {
-            static let channelName = uniqueChannelName()
+            static let channelName = uniqueChannelName(testIdentifier: "test__009__RealtimeClient__should_dispatch_in_user_queue_when_removing_an_observer")
```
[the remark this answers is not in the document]
maratal: Doesn't work that way, swift compiler can't figure this out - `#function` is something else in that case, and any attempt to get it being inside nested class is a compiler error `Class declaration cannot close over value 'funcName' defined in outer scope`

Spec/Test Utilities/TestUtilities.swift
maratal: >two channel names in a row
I tried to avoid situations like that, couldn't you point them out, please..

lukasz-szyszkowski: > couldn't you point them out, please...

I'm also curious :) because this shouldn't happen anywhere

lawrence-forooghian: I think I was wrong – I had a look through the diff again and I'm struggling to find any!

lawrence-forooghian: What we _do_ have, however, are a few places where we generate multiple channel names in a row, and where we rely on the `prefix` parameter to ensure that these names are unique, e.g. [here](https://github.com/ably/ably-cocoa/blob/7ee8a3dbcfc984a10788f1ace810c66b330af690/Spec/Tests/RealtimeClientChannelTests.swift#L643-L649) and [here](https://github.com/ably/ably-cocoa/blob/7ee8a3dbcfc984a10788f1ace810c66b330af690/Spec/Tests/RealtimeClientChannelTests.swift#L832-L838) and [here](https://github.com/ably/ably-cocoa/blob/7ee8a3dbcfc984a10788f1ace810c66b330af690/Spec/Tests/RealtimeClientChannelTests.swift#L2120-L2121).

I think that since the method is called `uniqueChannelName`, it really should always return a unique value, not "a unique value as long as you either call the method in different milliseconds or pass a different prefix", else it's going to be confusing for people reading the tests, and using this method in the future. So, yeah, I'd say to put a UUID in there.

maratal: 6bfb4bab84ff44ffac3d2a36eb67c3e5cf877eef

QuintinWillison: I'll let @lawrence-forooghian close this conversation, [as etiquette states](https://github.com/ably/engineering/blob/main/sdk/pull-requests.md#conversations), but I just wanted to add that I fully approve of adding UUID.

QuintinWillison (approved):
I've only really looked closely at the addition to test utilities, and then skim read the rest, but looks good to me.

lawrence-forooghian (approved):

lawrence-forooghian:
I'm going to merge this, since I'm keen to run the tests in a loop and see if this has made any difference to our test success rate.

PR #1288, opened 2022-02-17
Don’t build Carthage libraries for unnecessary platforms

lawrence-forooghian:
Hopefully will speed CI jobs up a bit. We should probably have done this as part of #1274.

### DIFF
--- a/.github/workflows/integration-test-iOS11.yaml
+++ b/.github/workflows/integration-test-iOS11.yaml
@@ -29,7 +29,7 @@ jobs:
         run: |
           make submodules
           bundle install
-          make update_carthage_dependencies
+          make update_carthage_dependencies_ios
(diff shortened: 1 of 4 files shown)


lukasz-szyszkowski (approved):

.github/workflows/integration-test-iOS11.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
         run: |
           make submodules
           bundle install
-          make update_carthage_dependencies
+          make update_carthage_dependencies_tvos
```
maratal, line 32: integration-test-iOS11.yaml vs update_carthage_dependencies_tvos

lawrence-forooghian: D'oh, thanks! Updated.

PR #1293, opened 2022-02-28
Call to the `registerForRemoteNotifications`

### LINKED ISSUE #1271: According to spec, the SDK should be responsible for calling `-[UIApplication registerForRemoteNotifications]`, not the user
## Description

[`RSH3a2d`](https://ably.com/documentation/client-lib-development-guide/features#RSH3a2d) says:

> On event `CalledActivate` … If the local device does not have the necessary push details, it initiates a request to the underlying platform (or otherwise generates them).

I believe that on iOS, to "initiate a request to the underlying platform" is to call `-[UIApplication registerForRemoteNotifications]`. However:

- This SDK never calls that method.
- We have various pieces of documentation and example code that suggest that a user of the library needs to call this method on app startup:
  - [Documentation: "Push Notifications - Device activation and subscription"](https://ably.com/documentation/general/push/activate-subscribe):
     > In `UIApplicationDelegate#didFinishLaunchingWithOptions`, call `UIApplication.shared.registerForRemoteNotifications`.

    But then later on in the same article (section 2.1), it says that calling `activate` will:

    > Activate the device for push notifications with the underlying OS or platform and obtain a unique identifier for the device as a push recipient
  
  - [iOS push notifications tutorial, step 6](https://ably.com/tutorials/ios-push-notifications#step6-register-device-for-push) demonstrates calling this method on app startup.
  - Our [push-example-ios example app](https://github.com/ably/push-example-ios) copies this pattern, [calling the method on app startup](https://github.com/ably/push-example-ios/blob/49e4f6532927979af34507f16138a6a278330dba/push-interactive-app/AppDelegate.swift#L43). 

## Next steps

- Confirm whether we believe the behaviour described by the spec to be the right one
- If so, implement it in the library and remove the call to this method from our documentation / example code (confirm that existing code that does call this method will still work fine).

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1503) by [Unito](https://www.unito.io)

maratal:
According to the issue, a call to the `registerForRemoteNotifications:` should be included somewhere to simplify the initial setup process for the cocoa framework. But the main "headache" and a startup "noise" is this:

```
[[UNUserNotificationCenter currentNotificationCenter] requestAuthorizationWithOptions:options completionHandler:^(BOOL granted, NSError * _Nullable error) {
    if (granted) {
        dispatch_async(dispatch_get_main_queue(), ^{
            [[UIApplication sharedApplication] registerForRemoteNotifications];
        });
    } else if (error) {
        [rest.logger error:@"ARTPush: request authorization failed (%@)", [error localizedDescription]];
    }
}];
```
So, I've introduced a new `activate:` method with the `UNAuthorizationOptions` parameter to justify that, whereas its old counterpart optionless `activate` method calls the new one with the default set of options.

This has raised a concern of a possible interfere of the new `activate` method behavior with existing apps startup logic, which is true and further discussion about it is [here](https://github.com/ably/ably-cocoa/pull/1293#discussion_r815973799).

In addition to that, I've created a new handler for `application:didRegisterForRemoteNotificationsWithDeviceToken` which is rather than being "handler" is now a method for saving on a device and registering the new token within Ably infrastructure: `registerAPNSDeviceToken:`, signature of which much easier to comprehend in oppose to `didFailToRegisterForRemoteNotificationsWithError:rest:`.

The new one just calls the old one, because it already implements what is needed. In the following commits I'm going to flip their bodies, because the old handlers should eventually be removed.

These two changes allow to setup ably-cocoa and APNs in just two lines of code:

```
self.realtime.push.activate([.badges, .alert])

func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
    self.realtime.push.registerAPNSDeviceToken(deviceToken)
}
```

Which is way more welcoming in comparing to Step 6 ad Step 7 of the tutorial https://ably.com/tutorials/ios-push-notifications#step6-register-device-for-push


Source/ARTPush.m
```diff
@@ -223,13 +231,44 @@ + (void)didFailToRegisterForRemoteNotificationsWithError:(NSError *)error rest:(
...
 }
 
+- (void)activate {
+    UNAuthorizationOptions options = UNAuthorizationOptionAlert | UNAuthorizationOptionSound | UNAuthorizationOptionBadge;
+    [self activate:options];
+}
+
 - (void)deactivate {
+#if !TARGET_OS_SIMULATOR
+    [[UIApplication sharedApplication] unregisterForRemoteNotifications];
```
lukasz-szyszkowski, line 270: Shouldn't we call something like:
```objc
[ARTPushInternal didUnegisterForRemoteNotificationsWithRestInternal:rest];
```
?

Source/ARTPush.h
```diff
@@ -66,10 +67,22 @@ NS_ASSUME_NONNULL_BEGIN
 + (void)didFailToRegisterForRemoteNotificationsWithError:(NSError *)error realtime:(ARTRealtime *)realtime;
 
 /**
- Activating a device for push notifications and registering it with Ably. The registration process can be performed entirely from the device or from your own server using the optional `ablyPushCustomRegister:deviceDetails:callback` method.
+ Activating a device for push notifications (with requesting user permissions and registering with the APN service).
+ The registration process can be performed on the device or from your own server using the optional ``ablyPushCustomRegister:deviceDetails:callback`` method.
+ */
+- (void)activate:(UNAuthorizationOptions)options;
```
lawrence-forooghian, line 73: I know that this PR isn't ready for review yet, but I happened to see this and wanted to say that I don't think that requesting permissions to display notifications to the user should be part of what the Ably SDK does - that seems like a concern for the application which integrates the SDK, since it might need to be integrated into some flow within the app. The Ably SDK doesn't need these permissions to be able to function - it just needs an APNS device token, which it doesn't need any user permission to get.

maratal: > since it might need to be integrated into some flow within the app

From docs:
> If you do not request and receive authorization for your app's interactions, the system delivers all remote notifications to your app silently.

I agree too, this creates unwanted scenario of unexpected user interaction. Then I can leave optionless `activate` method just to call `registerForRemoteNotifications:` and the one with the options requesting permissions (as it does now). Or to rename `activate:` to something like `requestAuthorizationAndActivate:` to avoid confusion. I still insist on introducing method for requesting permissions as it greatly reduces initial setup code and thus increases satisfaction of trying the new framework. WDYT @lawrence-forooghian @lukasz-szyszkowski @QuintinWillison

lawrence-forooghian: I think that we should first fix https://github.com/ably/ably-cocoa/issues/1271, which only requires us to call `-registerForRemoteNotifications`, as @lukasz-szyszkowski said. The user permissions stuff is a separate thing that I think is outside the scope of that issue. I'm personally not in favour of adding further methods to handle the user permissions stuff, but it's something we could discuss as a team separately, I think.

maratal: Ok, I've created a separate issue for this https://github.com/ably/ably-cocoa/issues/1303

QuintinWillison: I think the concern with this change for me is that it looks like it involves adding a new framework dependency to this SDK, that is [User Notifications](https://developer.apple.com/documentation/usernotifications?language=objc). It begs a few questions for me:

- Do we already require any other App UI oriented frameworks for this SDK?
- Have we ever defined scope/boundaries for this SDK, in respect of baseline/foundational requirements to build and run an app developed against it?

We should be clear as to what the SDK does and doesn't do. It's not a problem to have to spend time communicating to app developers how to integrate their app with the SDK - for example, see [our Flutter PN Documentation](https://github.com/ably/ably-flutter/blob/main/PushNotifications.md) and [our .NET PN Documentation](https://github.com/ably/ably-dotnet/blob/main/PushNotifications.md). 🤯 

Sometimes providing careful guidance that leaves app developers with more flexibility is better, rather than forcing architectural requirements on them. Plus, of course, we don't know how Apple might evolve the `UN` framework going forwards, so we could just be baking in maintenance burden for ourselves going forwards.

Source/ARTPush.m
maratal: Actually no, there are no callbacks for `UIApplication.unregisterForRemoteNotifications`, as well as anything related to it in ably cocoa, it's all about `UIApplication.registerForRemoteNotifications`.

Source/ARTPush.h
maratal: > Do we already require any other App UI oriented frameworks for this SDK?

Yes, `UIKit` was added by Ricardo back in 2019 to "ARTPush.m":

```
#if TARGET_OS_IOS
#import <UIKit/UIKit.h>
#import "ARTPushActivationStateMachine+Private.h"
#endif
```

> Have we ever defined scope/boundaries for this SDK, in respect of baseline/foundational requirements to build and run an app developed against it? ... Sometimes providing careful guidance that leaves app developers with more flexibility is better, rather than forcing architectural requirements on them.

> Plus, of course, we don't know how Apple might evolve the UN framework going forwards, so we could just be baking in maintenance burden for ourselves going forwards.

I agree, that incorporating ever changing APNs API into ably cocoa is asking for a trouble. On the other hand, many frameworks for providing push notifications do this to make their users life easier. But ably cocoa is more than just push notifications service/framework, so I'm not sure whether we should go into that direction or not. I've created an issue for further discussion of this problem - #1303.

Meanwhile for this PR I'll just make minimum changes required regarding calling `registerForRemoteNotifications`.

maratal: https://github.com/ably/ably-cocoa/pull/1293/commits/a47ef44df3e93fa2d3208d4a1e342f9f8f834274

lukasz-szyszkowski (approved):

Source/ARTPush.h
QuintinWillison: I'm going to resolve this particular conversation, on behalf of @lawrence-forooghian, as the question it raised has been resolved by @maratal ... however, I have some other questions which I'll raise in a new comment shortly.

QuintinWillison (commented):
I have a couple of questions around the changes being made to the `activate` and `deactivate` methods:

1. Is there a risk that calls to these methods, or other methods altering state, could overlap one another? I'm not familiar with this code, so there may be external factors guaranteeing this, however I can't documentation indicating that (for example: these methods aren't safe to be called from any thread, or can only be called in very specific circumstances).
2. Perhaps related, do methods on the shared application instance have to be called from the main thread? Looking at [this code comment](https://github.com/ably/ably-cocoa/blob/91e55626f2ca2388583519e84003a39663ff89c5/Source/ARTPush.m#L121) it appears that calls to the delegate might be.

maratal:
Thanks @QuintinWillison, that's a good spot. Regarding overlapping, I've created this issue - https://github.com/ably/ably-flutter/issues/345, so `registerForRemoteNotifications` wouldn't be called twice. And regarding thread safety, according to this SO post - https://stackoverflow.com/questions/44391367/uiapplication-registerforremotenotifications-must-be-called-from-main-thread-o one should re-assure this call is made from the main thread.

Source/ARTPush.m
```diff
@@ -224,12 +224,18 @@ + (void)didFailToRegisterForRemoteNotificationsWithError:(NSError *)error rest:(
 }
 
 - (void)activate {
+#if TARGET_OS_IOS && !TARGET_OS_SIMULATOR
+    [[UIApplication sharedApplication] registerForRemoteNotifications];
```
QuintinWillison, line 228: Consider enforcing that this is only called from main thread - see https://github.com/ably/ably-cocoa/pull/1293#issuecomment-1061804417.

Source/ARTPush.m
```diff
@@ -224,12 +224,18 @@ + (void)didFailToRegisterForRemoteNotificationsWithError:(NSError *)error rest:(
...
+#endif
     [self getActivationMachine:^(ARTPushActivationStateMachine *stateMachine) {
         [stateMachine sendEvent:[ARTPushActivationEventCalledActivate new]];
     }];
 }
 
 - (void)deactivate {
+#if TARGET_OS_IOS && !TARGET_OS_SIMULATOR
+    [[UIApplication sharedApplication] unregisterForRemoteNotifications];
```
QuintinWillison, line 237: Consider enforcing that this is only called from main thread - see https://github.com/ably/ably-cocoa/pull/1293#issuecomment-1061804417.

QuintinWillison (changes requested):

Source/ARTPush.m
maratal: https://github.com/ably/ably-cocoa/pull/1293/commits/b8093726b8075edb39f83341b18745cc0ed29d43

Source/ARTPush.m
maratal: https://github.com/ably/ably-cocoa/pull/1293/commits/b8093726b8075edb39f83341b18745cc0ed29d43

Source/ARTPush.m
```diff
@@ -225,7 +225,9 @@ + (void)didFailToRegisterForRemoteNotificationsWithError:(NSError *)error rest:(
 
 - (void)activate {
 #if TARGET_OS_IOS && !TARGET_OS_SIMULATOR
-    [[UIApplication sharedApplication] registerForRemoteNotifications];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
lukasz-szyszkowski, line 228: Consider check if the `currentThread` isn't the `mainThread` already.
Your code will work, but won't be called immediately.

maratal: ok, will do

maratal: https://github.com/ably/ably-cocoa/pull/1293/commits/a9e92dec5885775bb96509c82e2c65f8c8303a9e

lawrence-forooghian (changes requested):

lawrence-forooghian:
@maratal in terms of testing:

- have you tested this end to end? i.e. tested in a sample app that push notifications work even without explicitly calling `registerForRemoteNotifications`
- can we also test that `registerForRemoteNotifications` is called?

maratal:
>     * have you tested this end to end? i.e. tested in a sample app that push notifications work even without explicitly calling `registerForRemoteNotifications`

Yes, I have simple demo app that does just that. I plan to convert it to SwiftUI according to https://ably.atlassian.net/browse/SDK-1655

>     * can we also test (i.e. in our test suite) that `registerForRemoteNotifications` is called?

I think yes, I need to investigate this.

Source/ARTPush.m
```diff
@@ -224,12 +224,32 @@ + (void)didFailToRegisterForRemoteNotificationsWithError:(NSError *)error rest:(
...
+#endif
     [self getActivationMachine:^(ARTPushActivationStateMachine *stateMachine) {
         [stateMachine sendEvent:[ARTPushActivationEventCalledActivate new]];
     }];
 }
 
 - (void)deactivate {
+#if TARGET_OS_IOS && !TARGET_OS_SIMULATOR
+    if ([NSThread isMainThread]) {
+        [[UIApplication sharedApplication] unregisterForRemoteNotifications];
```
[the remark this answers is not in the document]
maratal: You are right, specs tell nothing about it, will fix.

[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1293/commits/d959b7b53b7b62602a25bf075b9d60773b67259e

Source/ARTPush.m
```diff
@@ -224,12 +224,32 @@ + (void)didFailToRegisterForRemoteNotificationsWithError:(NSError *)error rest:(
 }
 
 - (void)activate {
+#if TARGET_OS_IOS && !TARGET_OS_SIMULATOR
+    if ([NSThread isMainThread]) {
```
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1293/commits/01b448f3aa568447326a2cbb35993a3e327b22a7

lawrence-forooghian:
Thanks for the updates @maratal, I've resolved my code-level comments. Did you or @QuintinWillison have thoughts on [this question](https://github.com/ably/ably-cocoa/pull/1293#issuecomment-1074163869) and [this one](https://github.com/ably/ably-cocoa/pull/1293#issuecomment-1074164953)?

QuintinWillison:
@lawrence-forooghian Regarding [the REST API PATCH endpoint](https://ably.com/documentation/rest-api#patch-device-registration), I'm going to find it very difficult to gain enough context to form an opinion without spending more time than I could possibly find right now to dig around. I'm not familiar with this code. The documented constraints on what fields can be updated via that endpoint suggest that calling it multiple times needs to be considered very carefully. 🤷 

In respect of your question around updating examples and documentation, I don't think that needs to block this pull request from landing, especially if it _does_ end up being decided that multiple calls to `registerForRemoteNotifications` are harmless. But I might be missing something.

QuintinWillison:
@lawrence-forooghian In order not to forget, we would usually create an issue in `ably/docs`.

maratal:
> Thanks for the updates @maratal, I've resolved my code-level comments. Did you or @QuintinWillison have thoughts on [this question](https://github.com/ably/ably-cocoa/pull/1293#issuecomment-1074163869) and [this one](https://github.com/ably/ably-cocoa/pull/1293#issuecomment-1074164953)?

@lawrence-forooghian @QuintinWillison Turned out it doesn't matter how many times in a raw you call `registerForRemoteNotifications`, `UIKit` gives only one callback to `UIApplicationDelegate/didRegisterForRemoteNotificationsWithDeviceToken`. Also, if you call `registerForRemoteNotifications` on already activated device, apple gives you the same `deviceToken`, so `ARTPush/didRegisterForRemoteNotificationsWithDeviceToken` exits early without sending `ARTPushActivationEventGotPushDeviceDetails` to the state machine.

maratal:
> In order not to forget, we would usually create an issue in `ably/docs`.

@lawrence-forooghian https://github.com/ably/docs/issues/1332

lawrence-forooghian (approved):

Source/ARTPushActivationStateMachine.m
```diff
@@ -379,6 +381,19 @@ - (void)callUpdateFailedCallback:(nullable ARTErrorInfo *)error {
...
 }
 
+- (void)registerForAPNS {
+#if !TARGET_OS_SIMULATOR
+    if ([NSThread isMainThread]) {
+        [[UIApplication sharedApplication] registerForRemoteNotifications];
+    }
+    else {
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [[UIApplication sharedApplication] registerForRemoteNotifications];
```
QuintinWillison, line 391: If I was being really picky then I could point out that you could declare this block once and invoke it from both call sites to reduce repetition, but it's not like this is an overly complex method implementation so that's perhaps more of a tasting note. 😁

QuintinWillison (approved):

Source/ARTPushActivationStateMachine.m
maratal: Hmm, I'm not sure I've understood, `registerForRemoteNotifications` called twice - one time directly, and the other in the `dispatch_async` block. Or you mean to transform `dispatch_async(dispatch_get_main_queue(), ^{ }` into `dispatch_async_main(^{ }` ?

QuintinWillison: The `[[UIApplication sharedApplication] registerForRemoteNotifications];` call in in two places, both of which can be encapsulated as a `dispatch_block_t` earlier on. Then you can either call it directly or pass it to `dispatch_async`.

maratal: I got your point - calling `registerForRemoteNotifications` twice near to each other looks a bit weird. But wrapping this one-liner into another one (two liner actually) would be unnecessary overhead I think. What I would do, it's to remove `isMainThread` check entirely and call it always on `dispatch_get_main_queue` because `didRegisterForRemoteNotificationsWithDeviceToken` callback is asynchronous anyways. WDYT @QuintinWillison @lukasz-szyszkowski  @lawrence-forooghian

QuintinWillison: Sounds good to me. It was just the DRY element I was commenting on - so never a problem to replace one line with a few lines, if the goal is reduction of repetition. 🤷

maratal: https://github.com/ably/ably-cocoa/pull/1293/commits/baa71f75040b97f69dccbd244dfa69428c6c11b0

lukasz-szyszkowski (approved):

PR #1294, opened 2022-02-28
Upload test results to observability server

lawrence-forooghian:
This PR gets our CI jobs to generate JUnit XML test result reports, and adds a step to upload these reports to our [test observability server](https://test-observability.herokuapp.com) after every test run.


.github/workflows/integration-test-iOS14_4.yaml
```diff
@@ -18,6 +18,7 @@ jobs:
       LC_CTYPE: en_US.UTF-8
       LANG: en_US.UTF-8
       ABLY_ENV: sandbox
+      TEST_OBSERVABILITY_SERVER_AUTH_KEY: ${{ secrets.TEST_OBSERVABILITY_SERVER_AUTH_KEY }}
```
QuintinWillison, line 21: Injecting the secret into the environment [at `job` level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idenv) means that it's available to all steps under that job, which is a leak risk - especially when those steps use third party actions.

Please move this to the specific step that needs it - that is, [at `step` level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsenv).

(same comment applies to other workflows modified in this way in this PR)

QuintinWillison (changes requested):
Looks good. Just one comment around encapsulation level of secrets...

.github/workflows/integration-test-iOS14_4.yaml
lawrence-forooghian: Ah, thanks for the info. Changed it.

maratal:
Nothing caught my attention except `local_dev_upload_test_results.sh` wasn't called from anywhere. But it's I guess me being out of context.

lawrence-forooghian:
> Nothing caught my attention except `local_dev_upload_test_results.sh` wasn't called from anywhere. But it's I guess me being out of context.

Yep, it’s just used for testing the server locally.

maratal (approved):

Scripts/upload_test_results.sh
```diff
@@ -0,0 +1,165 @@
...
+done
+
+if [[ -z $iteration ]]
+then
+  iteration=1
+fi
+
+# 4. Find the JUnit test report.
+
+test_reports=$(find fastlane/test_output/sdk -name '*.junit')
```
lukasz-szyszkowski, line 113: Is this path correct?
The path used before (for the Fastlane output) was `fastlane/test_output/examples/`
Moreover, maybe it's a better idea to pass this path as an argument to `upload_test_results.sh`?

lawrence-forooghian: There are two paths that are used for Fastlane output:

- `fastlane/test_output/sdk/` is used for the test output from the SDK’s tests
- `fastlane/test_output/examples/` is used for the test output from the example app’s tests

In this script, we only care about the SDK’s test results.

Why do you think it might be better to pass the path as an argument?

QuintinWillison (approved):

Scripts/upload_test_results.sh
lukasz-szyszkowski: > Why do you think it might be better to pass the path as an argument?

I'm guessing we can configure it from within the Fastfile, hence if someone will change it there, they have to change it here too. But I just realized this doesn't solve it, maybe better idea is to set these paths as ENV var ... I just wonder ;)

lukasz-szyszkowski: anyway I approved this pull request

lukasz-szyszkowski (approved):

PR #1307, opened 2022-03-14
Feature/1305 symmetric decryption

### LINKED ISSUE #1305: Spec alignment for symmetric decryption of messages
`ably-cocoa` is not aligned with the specs regarding manual decryption of messages according to [TP4](https://docs.ably.com/client-lib-development-guide/features/#TP4) and [TM3](https://docs.ably.com/client-lib-development-guide/features/#TM3). More context [here #207](https://github.com/ably/ably-flutter/issues/207)

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1665) by [Unito](https://www.unito.io)

maratal:
Partially closes #1305

1. I've only added decryption for `ARTMessage`, because of absence of encrypted test data for `ARTPresenceMessage`. I could use the same file but I'm not sure, lets discuss.
2. `fromEncodedJsonArray` omits error while iterating and calling `decodeWithEncoder` for each item, because in the original code (from `ARTRestChannel.history`) this error goes nowhere except logger. And in this static function I have no logger, except one I can directly pass into `fromEncodedJsonArray`, but that would be in contradiction with spec requirement for this function signature ([TM3](https://docs.ably.com/client-lib-development-guide/features/#TM3)).


Source/ARTMessage.h
```diff
@@ -16,4 +17,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+@interface ARTMessage (Decryption)
+
++ (instancetype)fromEncodedJsonObject:(NSDictionary *)json channelOptions:(ARTChannelOptions *)options error:(NSError **)error;
```
ikurek, line 22: Not sure if dictionary is a right format here, `ably-java` allows the JSON to be passed as `String` and `JsonObject` types. I think if they're saved or retrieved outside of Ably SDK, they're probably stored as Strings, so it may be a good idea to think about implementing an alternative definition of this method?

lukasz-szyszkowski: > it may be a good idea to think about implementing an alternative definition of this method?

or, change the `NSDictionary` to universal type `id` and check type inside the function body.

Source/ARTMessage.m
```diff
@@ -45,3 +49,28 @@ - (NSInteger)messageSize {
 }
 
 @end
+
+@implementation ARTMessage (Decryption)
+
++ (instancetype)fromEncodedJsonObject:(NSDictionary *)input channelOptions:(ARTChannelOptions *)options error:(NSError **)error {
```
lukasz-szyszkowski, line 55: it's just cosmetics, but the first parameter name in header declaration is `json` not `input`

Source/ARTMessage.h
lukasz-szyszkowski, line 22: I wonder if return value shouldn't be `nullable`?

Source/ARTMessage.m
```diff
@@ -45,3 +49,28 @@ - (NSInteger)messageSize {
...
+    return message;
+}
+
++ (NSArray<ARTMessage *> *)fromEncodedJsonArray:(NSArray<NSDictionary *> *)jsonArray channelOptions:(ARTChannelOptions *)options error:(NSError **)error {
+    ARTJsonLikeEncoder *jsonEncoder = [[ARTJsonLikeEncoder alloc] initWithDelegate:[[ARTJsonEncoder alloc] init]];
+    ARTDataEncoder *decoder = [[ARTDataEncoder alloc] initWithCipherParams:options.cipher error:error];
+    
+    NSArray <ARTMessage *> *messagesArray = [jsonEncoder messagesFromArray:jsonArray];
+    messagesArray = [messagesArray artMap:^(ARTMessage *message) {
+        return [message decodeWithEncoder:decoder error:nil];
```
lukasz-szyszkowski, line 71: Are you sure we can't manage this error? It's kind of risky to leave it as nil especially when there is a requirement where we should pass a pointer to it.

Source/ARTMessage.h
maratal: What about its `NSArray` variant? Should it be an array of strings or dictionaries or else? Or to leave only one function that figures out whether input is an array or not and returns an array or single object. That would also be confusing. I've just used the same types that an actual parser already uses - `ARTJsonLikeEncoder`. `NSDictionary` translates into `[String: Any]` in Swift and it's pretty conventional type for a JSON. WDYT @QuintinWillison @ikurek @lukasz-szyszkowski @lawrence-forooghian

maratal: Ok, ably-java just checks if an element of an array is a JSON object - https://github.com/ably/ably-java/blob/382829ee6df70875d372162bb84186344e2f7dfb/lib/src/main/java/io/ably/lib/types/Message.java#L243

maratal: It's because `ARTBaseMessage/decodeWithEncoder` doesn't do. I think this change should be in another PR.

Source/ARTMessage.m
maratal: You're right, I should do something about it) `artMap:` closure didn't allowed me to refer `error`, but it's a no excuse.

maratal: https://github.com/ably/ably-cocoa/pull/1307/commits/caf46b333fd9e559641e19b1a0e15ddf79248696

Source/ARTMessage.m
maratal: https://github.com/ably/ably-cocoa/pull/1307/commits/dd9a0ad288f1e64869e4977a735b69f6be95665e

Source/ARTMessage.h
maratal: Made it here anyway: https://github.com/ably/ably-cocoa/pull/1307/commits/3b58d439a6a692b4874e4b12f22c6ea3c8778b74 https://github.com/ably/ably-cocoa/pull/1307/commits/6fb0756453292dc06ff292a2e413eac9929f69b8

lukasz-szyszkowski: It depends how universal this method should be. Do we want to have one method for both `String` and `Dictionary`? Or maybe it's a better idea to have two separate methods for these types. IMO universal type aka `id` is a bad idea because there is no simple information what we should pass as an input.

lawrence-forooghian: From TP4 (emphasis mine):

> `fromEncoded` and `fromEncodedArray` are alternative constructors that take an _(already deserialized)_ `Message`-like object (or array of such objects)

and from the [IDL](https://docs.ably.com/client-lib-development-guide/features/#idl):

```
+fromEncoded(JsonObject, ChannelOptions?) -> Message
```

> `JSONObject` and `JSONArray` denote any type or interface in the target language that represents an [RFC4627](https://www.ietf.org/rfc/rfc4627.txt) object or array value respectively.

The spec doesn't explicitly define the meaning of "deserialized", but from what I can tell, it means "already decoded by a JSON or MessagePack parser".

So, I think we don't need to worry about accepting a string here, seems like we only need to accept `NSDictionary` and `NSArray`.

lawrence-forooghian:
@maratal, I don't think we can say "Closes https://github.com/ably/ably-cocoa/issues/1305" if we're not handling `PresenceMessage` here. We should either handle it here or split that issue in two.

Source/ARTMessage.m
```diff
@@ -45,3 +49,73 @@ - (NSInteger)messageSize {
 }
 
 @end
+
+@implementation ARTMessage (Decoding)
+
++ (instancetype)fromEncodedJsonObject:(NSDictionary *)json channelOptions:(ARTChannelOptions *)options error:(NSError **)error {
```
lawrence-forooghian, line 55: How does this behave with VCDIFF-encoded messages? Given that the decoding of such a message relies on having a "previous" message available, I don't know how it's meant to work. The spec provides a recovery procedure in [RTL18](https://docs.ably.com/client-lib-development-guide/features/#RTL18), but it seems designed for the case where you have a realtime connection available to request the previous message. Do we need to address this in the spec?

Source/ARTMessage.h
```diff
@@ -16,4 +17,16 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+@interface ARTMessage (Decoding)
+
++ (nullable instancetype)fromEncodedJsonObject:(NSDictionary *)json
```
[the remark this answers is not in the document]
maratal: I'm not sure what you mean, array method name is `fromEncodedJsonArray:`

Source/ARTMessage.m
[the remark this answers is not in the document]
maratal: I guess it should be `nullable` then.

maratal: Probably externally saved messages shouldn't be VCDIFF-encoded?

Source/ARTMessage.h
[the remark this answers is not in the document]
lawrence-forooghian: In the features spec it's just called `fromEncodedArray`, not `fromEncodedJsonArray`.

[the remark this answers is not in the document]
maratal: Good catch, but tbh I think `fromEncodedJsonArray` makes more sense bc it what it does. You can't pass any other array there.

Source/ARTMessage.m
QuintinWillison: As far as I know, the Ably service will only ever send VCDiff-encoded messages on a Realtime connection. It is then the responsibility of the SDK to transparently apply that delta to the previous message and then deliver the result to the application. Therefore, I'm not sure that @lawrence-forooghian is describing a use-case that is possible, but I may be wrong.

Source/ARTMessage.h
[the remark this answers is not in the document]
lawrence-forooghian: @QuintinWillison what do you think about this - should we be using a name different to the features spec? Should we be changing the name in the features spec?

Source/ARTMessage.m
lawrence-forooghian: Does that mean we're happy to leave the behaviour undefined in this case?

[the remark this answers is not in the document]
lawrence-forooghian: Maybe. I think we should add something to the spec.

Source/ARTMessage.h
ikurek: That's going to create some complexity for Dart SDK. Right now I simply pass the String to `ably-java` and leave the decoding to native implementation. If we decide to use `NSDictionary` for `ably-cocoa` I'll need to deserialize the object on Dart side and pass either `NSDictionary` or `JSONObject` to platform implementations. It's possible, but it will be some additional work on Dart side, and can be a little bit tricky because `JSONObject` comes from GSON lib and I'm not sure if I'll be Able to use it in Dart bridge 🤔

Source/ARTMessage.h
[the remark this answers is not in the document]
QuintinWillison: I would stick to alignment with the feature spec, unless there's a strong idiomatic reason why it _had_ to be different (e.g. to make an 'overload' method or something like that). But, as far as I can see, that's not the case here.

Source/ARTMessage.m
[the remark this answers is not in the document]
maratal: Searching for `.connectionId =` string across the Xcode project I didn't find any traces where described behavior could be implemented. `ARTJsonLikeEncoder/messageFromDictionary:` populates it:

`message.connectionId = [input artString:@"connectionId"]; `

Handling `@"connectionId"` being an empty string should be in another PR IMHO.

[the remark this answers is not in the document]
maratal: Definitely, but that would be changes in `ARTDataEncoder/decode:identifier:encoding:` method tweaking which wasn't the purpose of this PR. I've created a separate issue for that - https://github.com/ably/ably-cocoa/issues/1317

Source/ARTMessage.h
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1307/commits/cfb21148d3e024a7b0a54bdc09de610d086b1739

Source/ARTMessage.h
maratal: As of these methods already exist in ably-java I've created them here as well for consistency. Probably we need to add them to specs too. WDYT @QuintinWillison  @lawrence-forooghian @lukasz-szyszkowski 
https://github.com/ably/ably-cocoa/pull/1307/commits/ee20ff4781733fe25882e74c854a6fa477c801b8

maratal:
> I don't think we can say "Closes #1305" if we're not handling `PresenceMessage` here. We should either handle it here or split that issue in two.

Since there is no reliable way to detect which type of message in the provided dictionary, I have nothing to do except add similar methods to `ARTPresenceMessage` class. I'll do it once we agree upon all changes for `ARTMessage`. @lawrence-forooghian

Source/ARTMessage.m
maratal: @QuintinWillison

Source/ARTMessage.h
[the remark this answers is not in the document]
lawrence-forooghian: This method is still called `fromEncodedJsonObject`, it seems?

Source/ARTMessage.m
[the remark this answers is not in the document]
lawrence-forooghian: But the idea of this PR was to add the `fromEncoded(…)` methods, so if there are things that the spec doesn't address about how these methods should be implemented, would now not be a good time to get those things added to the spec and then incorporate them into these methods we add?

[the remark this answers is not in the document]
lawrence-forooghian: (@QuintinWillison thoughts?)

Source/ARTMessage.h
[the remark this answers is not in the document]
maratal: I thought we talked about `fromEncodedJsonArray` only, sorry - eb11c89403c961f3deb35e933668ffd74cf60c66

Source/ARTMessage.h
ikurek: In Java the `fromEncoded` method takes `String messageJson` as parameter, for MessagePack there's an alternative method `fromMsgpack`. There's also a `fromEncoded` variant that takes `JsonObject` but I don't think it's a good idea, because `JsonObject` is provided by `GSON` library, so this requires users to rely on external sources for serialization

Source/ARTMessage.m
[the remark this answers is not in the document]
maratal: Can't remember the exact thoughts at that moment, probably handling incoming JSON with absent fields is a more broad topic, but you are right, why not to take care about this particular field now. So, there are two options - make this field `nullable`, as the other fields in the `ARTBaseMessage` or return a decoding error. Since messages that are stored externally I guess have no any live connection, then the first option is I think the right one.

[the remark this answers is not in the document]
lawrence-forooghian: I would agree, the former makes sense. Could we open a spec issue to get a discussion going and get something added to the spec?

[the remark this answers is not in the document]
QuintinWillison: I agree with @lawrence-forooghian that if we're introducing a new API then we should try to ensure we get it right from the outset. However - perhaps more in line with @maratal's thinking - as far as I can tell, the method signature is not going to need to change - regardless of how it's implemented (i.e. there's already an `NSError *` return path). So, perhaps, we could defer enhancing the implementation to match up to a spec addition to a later unit of work (i.e. another pull request)? This PR is getting pretty old and it would be nice to stay agile - I believe it's taking the code from one working state to another, right?

Source/ARTMessage.h
maratal: @ikurek So what was the progress on that? Can I now remove methods accepting string as an argument?

Source/ARTMessage.m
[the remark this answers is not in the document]
maratal: @QuintinWillison @lawrence-forooghian the implementation of these new methods probably will not change either. I's `ARTDataEncoder/decode:identifier:encoding:` which is going to be changed.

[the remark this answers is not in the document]
maratal: @lawrence-forooghian https://github.com/ably/docs/issues/1329

maratal:
> Since there is no reliable way to detect which type of message in the provided dictionary, I have nothing to do except add similar methods to `ARTPresenceMessage` class. I'll do it once we agree upon all changes for `ARTMessage`.

I agree to create one more PR to actually fully address corresponding issue (I'll reopen it after this PR merged).

lawrence-forooghian (approved):

lukasz-szyszkowski (approved):

Issue #1279, opened 2022-02-04
Understand the scope of work required to stabilise ably cocoa tests


jamienewcomb:
We need to understand what needs to be done so the tests in ably-cocoa are stable

┆Issue is synchronized with this [Jira Research](https://ably.atlassian.net/browse/SDK-1545) by [Unito](https://www.unito.io)

lawrence-forooghian:
The [internal discussion](https://ably-real-time.slack.com/archives/CSQEKCE81/p1643733012856989) about #1278 might be worth looking at here, too.

maratal:
Example of unstable test - https://github.com/ably/ably-cocoa/issues/1281

QuintinWillison:
I see that there has been a lot of activity linked to this issue, however there is no clear scope or acceptance criteria that could lead to completion. I believe that since this issue was created we have a much better view of our tests, via @lawrence-forooghian's efforts around observability, as well as plenty of well-defined issues under [the `failing-test` label](https://github.com/ably/ably-cocoa/labels/failing-test). Therefore, at this point, I don't think this issue adds any value to our backlog.

PR #1327, opened 2022-04-05
Add documentation about tools for investigating test failures

lawrence-forooghian:
This will help us investigate failures as we work on ably/ably-cocoa#1279.


maratal (approved):
LGTM 👍

Issue #1331, opened 2022-04-07
Investigate failures in test case RealtimeClientPresenceTests test__078__Presence__enter__optional_data_can_be_included_when_entering_a_channel (1e5beffd-b21f-4281-85c9-b1b6ab02471d)


lawrence-forooghian:
https://test-observability.herokuapp.com/repos/ably/ably-cocoa/test_cases/1e5beffd-b21f-4281-85c9-b1b6ab02471d

`RealtimeClientPresenceTests.test__078__Presence__enter__optional_data_can_be_included_when_entering_a_channel`

Split from #1330.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1733) by [Unito](https://www.unito.io)

Issue #1309, opened 2022-03-14
Investigate failures in test case RealtimeClientPresenceTests test__037__Presence__update__should_update_the_data_for_the_present_member_with_a_value (114e01a8-8971-4c3a-9cb7-0bb128e2749e)


lawrence-forooghian:
https://test-observability.herokuapp.com/repos/ably/ably-cocoa/test_cases/114e01a8-8971-4c3a-9cb7-0bb128e2749e

`RealtimeClientPresenceTests.test__037__Presence__update__should_update_the_data_for_the_present_member_with_a_value()`

Split from #1279. Our first investigation of a failure.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1676) by [Unito](https://www.unito.io)

PR #1371, opened 2022-04-20
(#1309, #1331) Speculative presence fixes with extra logging in test case `1e5beffd-b21f-4281-85c9-b1b6ab02471d` (run for longer)

lawrence-forooghian:
A longer-running version of #1353.


lawrence-forooghian:
Replaced by #1374.

PR #1377, opened 2022-04-27
(#1279) Upload crash reports to test observability server

lawrence-forooghian:
## What does this do?

After the tests have run in CI, it uses (our fork of) the `xcparse` tool to extract any crash reports from the `.xcresult` bundle generated by Xcode, and it uploads these to the test observability server.

## Why are we doing this?

So that we can find out the reason why a test has crashed (see commit messages for more details).

## Example

Here's an example of a failure with a crash report attached: https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads/c31a9238-5b46-4799-ae1a-1d4c50a3cb53

### DIFF
--- a/.github/workflows/integration-test-iOS14_4.yaml
+++ b/.github/workflows/integration-test-iOS14_4.yaml
@@ -20,7 +20,27 @@ jobs:
       ABLY_ENV: sandbox
 
     steps:
-      - uses: actions/checkout@v2
+      - name: Check out SDK repo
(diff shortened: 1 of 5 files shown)


QuintinWillison (approved):
Interesting technique and very bespoke! Clearly this adds a lot of value and provides a great foundation for us to evolve our test analysis techniques as needs evolve. Nice work. 🎉

PR #1381, opened 2022-05-01
Premature disconnect to test reconnect within 10 seconds

maratal:
Most failures for this test are related to the test timeout. One of the causes is disconnect message failed to be received in time. To eliminate this and simulate sudden `.disconnected` event I've added manual disconnection. This change will not help in case of timeout for `connected` event.

### DIFF
--- a/Spec/Tests/RealtimeClientConnectionTests.swift
+++ b/Spec/Tests/RealtimeClientConnectionTests.swift
@@ -2483,15 +2483,13 @@ class RealtimeClientConnectionTests: XCTestCase {
     func test__064__Connection__connection_failures_once_CONNECTED__if_a_Connection_transport_is_disconnected_unexpectedly_or_if_a_token_expires__then_the_Connection_manager_will_immediately_attempt_to_reconnect() {
         let options = AblyTests.commonAppSetup()
         options.autoConnect = false
-        let ttl = 3.0
-        options.tokenDetails = getTestTokenDetails(ttl: ttl)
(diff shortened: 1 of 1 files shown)


Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2483,11 +2483,18 @@ class RealtimeClientConnectionTests: XCTestCase {
...
         options.autoConnect = false
-        options.tokenDetails = getTestTokenDetails(ttl: 3.0)
+        let ttl = 3.0
+        options.tokenDetails = getTestTokenDetails(ttl: ttl)
         let client = ARTRealtime(options: options)
         defer { client.dispose(); client.close() }
 
         waitUntil(timeout: testTimeout) { done in
+            client.connection.once(.connected) { _ in
+                // if realtime didn't disconnect after ttl, fire onDisconnected manually a second later
```
lawrence-forooghian, line 2493: It would be easier to understand this test if we could make it behave more predictably. So instead of waiting for either a real timeout (from token expiry) or inducing one ourselves, how about we:

- exclude the possibility of token expiry by not requesting a short TTL
- _always_ force a disconnection after receiving a `.connected` event?

maratal: https://github.com/ably/ably-cocoa/pull/1381/commits/846b9047a0171ab3f45004279afbaafe0d3a2e7a

lawrence-forooghian (approved):

Issue #1332, opened 2022-04-07
Investigate failures in test case PushTests test__008__LocalDevice__has_a_device_method_that_returns_a_LocalDevice (4f837671-6233-41f1-94e8-01174d1da7b8)


lawrence-forooghian:
https://test-observability.herokuapp.com/repos/ably/ably-cocoa/test_cases/4f837671-6233-41f1-94e8-01174d1da7b8

`PushTests.test__008__LocalDevice__has_a_device_method_that_returns_a_LocalDevice`

Split from #1330.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1731) by [Unito](https://www.unito.io)

lawrence-forooghian:
This one is a crash; let's wait for some `.xcresult` artifacts to become available with the crash logs.

lawrence-forooghian:
There's one example here on the GitHub job for this upload: https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads/d003f499-c414-4131-8d6f-f56e24cd2d67

PR #1392, opened 2022-05-05
(#1332) Fix crashes in test case `PushTests.test__008__LocalDevice__has_a_device_method_that_returns_a_LocalDevice` (`4f837671-6233-41f1-94e8-01174d1da7b8`)

lawrence-forooghian:
## What does this do?

Introduces various concurrency-related fixes related to the local device. Together, I believe these fix various crashes that we've observed in the test mentioned in the title (#1332). See commit messages for more details.

## How do I know it's fixed the test?

[These](https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads/compare?alternative-branches%5B%5D=1322-fix-test-case-4f837671-6233-41f1-94e8-01174d1da7b8-take-3-run-longer&alternative-createdBefore=&alternative-createdAfter=&alternative-failureMessage=) are the results of the tests running continuously with these fixes applied. At time of writing, there are 97 test runs with no failures in this test, but I'll keep it running for a while more before merging.

### DIFF
--- a/Source/ARTRest.m
+++ b/Source/ARTRest.m
@@ -724,10 +724,28 @@ dispatch_async(_queue, ^{
     return ret;
 }
 
-// Store address of once_token to access it in debug function.
-static dispatch_once_t *device_once_token;
-
 - (ARTLocalDevice *)device_nosync {
+    __block ARTLocalDevice *ret;
+    dispatch_sync(ARTRestInternal.deviceAccessQueue, ^{
+        ret = [self deviceWithClientId_onlyCallOnDeviceAccessQueue:self.auth.clientId_nosync];
+    });
+    return ret;
+}
+
++ (dispatch_queue_t)deviceAccessQueue {
+    static dispatch_queue_t queue;
+    static dispatch_once_t onceToken;
+    
+    dispatch_once(&onceToken, ^{
+        queue = dispatch_queue_create("io.ably.deviceAccess", DISPATCH_QUEUE_SERIAL);
+    });
+    
+    return queue;
+}
+
+static BOOL sharedDeviceNeedsLoading_onlyAccessOnDeviceAccessQueue = YES;
+
+- (ARTLocalDevice *)deviceWithClientId_onlyCallOnDeviceAccessQueue:(NSString *)clientId {
     // The device is shared in a static variable because it's a reflection
     // of what's persisted. Having a device instance per ARTRest instance
     // could leave some instances in a stale state, if, through another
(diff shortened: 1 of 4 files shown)


Source/ARTRest.m
```diff
@@ -724,10 +724,28 @@ - (ARTLocalDevice *)device {
     return ret;
 }
 
-// Store address of once_token to access it in debug function.
-static dispatch_once_t *device_once_token;
-
 - (ARTLocalDevice *)device_nosync {
```
maratal, line 727: So, it's no more "nosync", I think at this point we should just get rid of this method and use `device` which is scheduled via `deviceAccessQueue`. Because immidiatly re-shedule it in `device` on another queue has no sense. Maybe leave it for compatibility here (which would be a call to `device`), so not to remove call to it across codebase in this PR.

Source/ARTRest.m
```diff
@@ -724,10 +724,28 @@ - (ARTLocalDevice *)device {
...
+    dispatch_once(&onceToken, ^{
+        queue = dispatch_queue_create("io.ably.deviceAccess", DISPATCH_QUEUE_SERIAL);
+    });
+    
+    return queue;
+}
+
+static BOOL sharedDeviceNeedsLoading_onlyAccessOnDeviceAccessQueue = YES;
+
+- (ARTLocalDevice *)deviceWithClientId_onlyCallOnDeviceAccessQueue:(NSString *)clientId {
```
maratal, line 748: deviceWithClientId_nosync maybe?

lawrence-forooghian: I didn't want to use `nosync` because in the rest of the codebase it seems to have a specific meaning, which is that the method must be called from the internal queue of the `ARTRealtime/Rest/…` instance. This is a different queue so I didn't want to use the same naming.

Source/ARTRest.m
lawrence-forooghian: It needs to be `nosync` (meaning, as I mentioned below, that it needs to be called on the instance's internal queue), because it calls a `nosync` method (`clientid_nosync`).

maratal: Are you sure it cannot be just like this:

```
- (ARTLocalDevice *)device {
    __block ARTLocalDevice *ret;
    dispatch_sync(ARTRestInternal.deviceAccessQueue, ^{
        ret = [self deviceWithClientId_onlyCallOnDeviceAccessQueue:self.auth.clientId_nosync];
    });
    return ret;
}

// this is temporary for compatibility
- (ARTLocalDevice *)device_nosync {
    return [self device];
}
```

lawrence-forooghian: I'm pretty sure. The `_nosync` naming of `clientId_nosync` means (as I understand it) that it needs to be called from _queue, not from `ARTRestInternal.deviceAccessQueue`.

maratal (approved):

PR #1404, opened 2022-05-09
Fix mistake in my recent PR #1332

lawrence-forooghian:
This slipped past me and code review. I knew that `clientId_nosync` needed to be called on the same thread as its containing method (I even mentioned it in code review) and yet I didn’t actually do it 🤦.


maratal (approved):
oops)

PR #1428, opened 2022-06-05
Jazzy docs generation workflow

maratal:


Scripts/jazzy.sh
```diff
@@ -0,0 +1,13 @@
...
+# Generates Jazzy documentation: https://github.com/realm/jazzy
+
+jazzy \
+  --objc \
+  --author Ably \
+  --module-version 1.2.10 \
+  --umbrella-header Source/Ably.h \
+  --framework-root . \
+  --module Ably \
+  --exclude="*/ARTPushActivationStateMachine.h","*/ARTPushActivationStateMachine+Private.h" \
```
lawrence-forooghian, line 12: I must admit that I am rather confused about how header files work in our project. There seem to be so many different lists of header files. To name a few:

- those specified in `Ably.podspec`
- those specified in `Package.swift`
- the directory hierarchy and header file symlinks under the `Source` directory
- `Source/Ably.modulemap` contains a list of header files
- `Source/include/Ably.modulemap` contains a list of header files
- `Source/Ably.h` contains a list of header files

And I don't fully understand what each one is used for. But I imagine that, somewhere contained amongst all those, already lies a description of all of the public header files of our SDK. It would be great if we could make use of that information in this script instead of generating _another_ list of header files.

maratal: Jazzy docs say "In Objective-C mode, Jazzy documents all declarations found in the --umbrella-header header file and any other header files included by it". So even if you provide a "clean" list of public headers, if these headers include some private ones, they will be included in the output. Thus providing a list of exclusions is necessary. But it can be just one folder path if we tweak a project structure a bit. For example, `ARTPushActivationStateMachine` isn't considered private interface, but it shouldn't be used by users either, so it's rather `internal`. If we put these kind of classes into `internal folder`, its path, along with private headers path, can be fed to `jazzy` command.

maratal: Well, I've made further research and now I think there is no need to split files between internal and private. Since the state machine is not used by users directly, it makes sense to put it in the private lists (and remove from `Ably.h`). See last 5 commits for details. @lawrence-forooghian
https://sdk.ably.com/builds/ably/ably-cocoa/pull/1428/jazzydoc/

Gemfile
```diff
@@ -4,6 +4,8 @@
 
 source "https://rubygems.org"
 
+gem 'jazzy'
```
lawrence-forooghian, line 7: There should be a corresponding change to `Gemfile.lock` here which specifies the exact version of `jazzy` to use. You just need to run `bundle install` and it will generate this change.

Scripts/jazzy.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Generates Jazzy documentation: https://github.com/realm/jazzy
+
+jazzy \
+  --objc \
+  --author Ably \
+  --module-version 1.2.10 \
```
lawrence-forooghian, line 8: We need to keep the `--module-version` argument in sync with the current library version. We can do this either by:

- adding this file to `Scripts/set-version.sh`, or
- using `Scripts/get-version.sh` inside this script to populate the `--module-version` argument.

lawrence-forooghian (commented):
I'm getting an error when I try to run `Scripts/jazzy.sh` locally - does it work for you?

```
$ ./Scripts/jazzy.sh

/Users/lawrence/code/ably/ably-cocoa/Source/include/Ably/Ably.h:1:9: fatal error: could not build module 'Foundation'
/Users/lawrence/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/jazzy-0.14.2/lib/jazzy/sourcekitten.rb:657:in `block in expand_extensions': undefined method `include?' for nil:NilClass (NoMethodError)
        from /Users/lawrence/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/jazzy-0.14.2/lib/jazzy/sourcekitten.rb:656:in `map'
        from /Users/lawrence/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/jazzy-0.14.2/lib/jazzy/sourcekitten.rb:656:in `expand_extensions'
        from /Users/lawrence/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/jazzy-0.14.2/lib/jazzy/sourcekitten.rb:1080:in `parse'
        from /Users/lawrence/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/jazzy-0.14.2/lib/jazzy/doc_builder.rb:155:in `build_docs_for_sourcekitten_output'
        from /Users/lawrence/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/jazzy-0.14.2/lib/jazzy/doc_builder.rb:86:in `build'
        from /Users/lawrence/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/jazzy-0.14.2/bin/jazzy:16:in `<top (required)>'
        from /Users/lawrence/.rbenv/versions/3.0.0/bin/jazzy:23:in `load'
        from /Users/lawrence/.rbenv/versions/3.0.0/bin/jazzy:23:in `<main>'
```

Gemfile
maratal: https://github.com/ably/ably-cocoa/pull/1428/commits/ae671e231bb31486cf88756b6baf070c6d1efb7f

lawrence-forooghian:
@maratal did you see my question above about it not working locally?

maratal:
> @maratal did you see my question above about it not working locally?

I thought the issue with Gemfile.lock will do the trick. I guess you need to run `bundle install` first?

lawrence-forooghian:
Nope, I'm still getting the same issue. Is it happening for you locally? If not, I can investigate a bit on my machine.

lawrence-forooghian:
I've also noticed that in your [CI output](https://github.com/ably/ably-cocoa/runs/6863571146?check_suite_focus=true) there's a `/Users/runner/work/ably-cocoa/ably-cocoa/Source/include/Ably/ARTDeltaCodec.h:2:9: fatal error: 'AblyDeltaCodec/AblyDeltaCodec.h' file not found`, did you see that?

lawrence-forooghian:
I've spent a bit of time looking into the error I've been getting, but I think that to investigate it in depth would take more time than I can afford to put into it right now. Just to double check - you're not seeing it locally? Might be worth asking a few other people in the team to try it out too and see if it works for them.

maratal:
> I've spent a bit of time looking into the error I've been getting, but I think that to investigate it in depth would take more time than I can afford to put into it right now. Just to double check - you're not seeing it locally? Might be worth asking a few other people in the team to try it out too and see if it works for them.

Works locally, but it gives me some `Carthage` build artifacts:

Screenshot 2022-06-14 at 16 17 11

But since the docs were generated regardless this output, and `Carthage` is not in favor now I decided not to dig into that. I'll investigate `/Users/runner/work/ably-cocoa/ably-cocoa/Source/include/Ably/ARTDeltaCodec.h:2:9: fatal error: 'AblyDeltaCodec/AblyDeltaCodec.h' file not found` issue, looks like was fixed already, but here it is again!

maratal:
`/Users/runner/work/ably-cocoa/ably-cocoa/Source/include/Ably/ARTDeltaCodec.h:2:9: fatal error: 'AblyDeltaCodec/AblyDeltaCodec.h' file not found` error was due to absence of dependencies. I've added this step here https://github.com/ably/ably-cocoa/pull/1428/commits/94bfab9cbb6961ccfc79b2901d220bff8f4380b2

Also there were multiple warnings like `warning: no 'assign', 'retain', or 'copy' attribute is specified - 'assign' is assumed [-Wobjc-property-no-attribute]`. I've created a separate issue for that: https://github.com/ably/ably-cocoa/issues/1433

Regarding local `jazzy` command. Strangely enough I began experiencing the same error `/Users/lawrence/code/ably/ably-cocoa/Source/include/Ably/Ably.h:1:9: fatal error: could not build module 'Foundation'`, but only after I've installed `sourcekitten` and used it directly like that:

```
sourcekitten doc --objc $(pwd)/Source/Ably.h \
      -- -x objective-c  -isysroot $(xcrun --show-sdk-path --sdk iphoneos) \
      -I $(pwd) -fmodules > objcDoc.json
```

`jazzy` command works without issues. I still have `Big Sur` and `ruby 2.6.3` though, will update to `Monterey` and check it again. But CI build is now clean. @lawrence-forooghian

lawrence-forooghian:
@maratal great news about the CI build, thanks!

As for my local issues, I just tried with Ruby 2.6.3, didn't make a difference, still get the error. As for the rest of my environment, I'm on macOS 12.4, Xcode 13.4.1.

maratal:
> @maratal great news about the CI build, thanks!
> 
> As for my local issues, I just tried with Ruby 2.6.3, didn't make a difference, still get the error. As for the rest of my environment, I'm on macOS 12.4, Xcode 13.4.1.

Yes, after fresh install of macOS 12.4 and Xcode 13.4.1. it's now reproducible. @lawrence-forooghian

maratal:
Jazzy stopped crashing after I've replaced `--framework-root .` with  `--framework-root Source`. But it still struggled with error `'AblyDeltaCodec/AblyDeltaCodec.h' file not found`. Then there were some manipulations with the make command (`make update_carthage_dependencies_<platform>`). After that it never appeared again no matter what to cleanup and what `--framework-root` value was. So I can't say for sure what "fixed" this crash.

But like on previous macOS installation, when I installed `sourcekitten` separately, it was there again:


```
sourcekitten doc --objc $(pwd)/Source/Ably.h
  -- -x objective-c
  -isysroot $(xcrun --show-sdk-path --sdk iphoneos)
  -I $(pwd) -fmodules > objcDoc.json
```

`/Users/admin/Work/ably-cocoa/Source/Ably.h:1:9: fatal error: could not build module 'Foundation'`


After `iphoneos` was replaced to `iphonesimulator` (as suggested in this thread https://github.com/realm/jazzy/issues/840):

```
sourcekitten doc --objc $(pwd)/Source/Ably.h
  -- -x objective-c
  -isysroot $(xcrun --show-sdk-path --sdk iphonesimulator)
  -I $(pwd) -fmodules > objcDoc.json

```

`/Users/admin/Work/ably-cocoa/Source/Ably.h:9:9: fatal error: 'Ably/ARTLog.h' file not found`


And after adding symlink to `Source` directory with the module name (as suggested in this article https://nicedoc.io/realm/jazzy/ObjectiveC.md):

`/Users/admin/Work/ably-cocoa/Ably/ARTDeltaCodec.h:2:9: fatal error: 'AblyDeltaCodec/AblyDeltaCodec.h' file not found.`


Also, as that article suggests, playing with clang command might be helpful, as jazzy relies on that, but the thing is that it doesn't work even in our `main` branch:

`clang -c -x objective-c -isysroot $(xcrun --show-sdk-path) -I $(pwd) Source/Ably.h -fmodules`

```
In file included from Source/Ably.h:60:
/Users/admin/Work/ably-cocoa/Ably/ARTDeltaCodec.h:2:9: fatal error: 'AblyDeltaCodec/AblyDeltaCodec.h' file not found
#import <AblyDeltaCodec/AblyDeltaCodec.h>
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```


Adding -F to pass Carthage frameworks doesn't help (from that article: "_If you are happy to build the framework project before generating docs and all the problematic paths have the form ModuleName/PublicHeader.h then have clang resolve those includes to the built framework by passing -F <path of directory containing ModuleName.framework>_"):


```
clang -c -x objective-c -isysroot $(xcrun --show-sdk-path) \
  -I $(pwd) Source/Ably.h \
  -F Carthage/Build \
  -fmodules
```

So, at this point I kinda ran out of ideas. Let me know if the last version of jazzy.sh works for you @lawrence-forooghian (it works for me locally and on CI without errors).

lawrence-forooghian:
Hey @maratal - I haven't read your above comment in detail, but unfortunately on commit 3f7a94dde182cfaf775b5eb9eedf4a9bcde4f288 I'm still getting the exact same error as before on my machine.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,50 @@
+name: Docs Generation
+
+on:
```
lawrence-forooghian, line 3: After https://github.com/ably/sdk-upload-action/pull/40 is merged and released, we'll want to make this workflow run for all tags, too, so that we're permanently saving documentation for each release.

```yaml
tags:
    "*"
```

But you can do it as a separate PR later.

lawrence-forooghian (approved):

.github/workflows/docs.yml
maratal: I've made an issue for that https://github.com/ably/ably-cocoa/issues/1440

PR #1432, opened 2022-06-13
Fix `Scripts/set-version.sh`

lawrence-forooghian:
I missed this when I renamed the test files in 240f51f.


maratal (approved):
Looks like it should work now =)

PR #1434, opened 2022-06-15
Added missing `strong` property attributes

### LINKED ISSUE #1433: Multiple warnings generated by `jazzy` command:
Like for example in this run: https://github.com/ably/ably-cocoa/runs/6863571146

`warning: no 'assign', 'retain', or 'copy' attribute is specified - 'assign' is assumed [-Wobjc-property-no-attribute]`

Not sure why these things only visible when using `jazzy` command, but lets make these properties explicit `strong`.
ARC turned on for all targets for sure. More info: https://stackoverflow.com/questions/22613154/warning-when-omitting-removing-strong-attribute-from-property

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2052) by [Unito](https://www.unito.io)

maratal:
Closes #1433, #1435

### DIFF
--- a/Source/ARTDeviceDetails.h
+++ b/Source/ARTDeviceDetails.h
@@ -7,12 +7,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface ARTDeviceDetails : NSObject
 
-@property (nonatomic) ARTDeviceId *id;
-@property (nullable, nonatomic) NSString *clientId;
-@property (nonatomic) NSString *platform;
-@property (nonatomic) NSString *formFactor;
-@property (nonatomic) NSMutableDictionary<NSString *, NSString *> *metadata;
-@property (nonatomic) ARTDevicePushDetails *push;
+@property (strong, nonatomic) ARTDeviceId *id;
+@property (strong, nullable, nonatomic) NSString *clientId;
+@property (strong, nonatomic) NSString *platform;
+@property (strong, nonatomic) NSString *formFactor;
+@property (strong, nonatomic) NSMutableDictionary<NSString *, NSString *> *metadata;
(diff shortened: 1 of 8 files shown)


Source/ARTDeviceDetails.h
```diff
@@ -7,12 +7,12 @@ NS_ASSUME_NONNULL_BEGIN
...
-@property (nullable, nonatomic) NSString *clientId;
-@property (nonatomic) NSString *platform;
-@property (nonatomic) NSString *formFactor;
-@property (nonatomic) NSMutableDictionary<NSString *, NSString *> *metadata;
-@property (nonatomic) ARTDevicePushDetails *push;
+@property (strong, nonatomic) ARTDeviceId *id;
+@property (strong, nullable, nonatomic) NSString *clientId;
+@property (strong, nonatomic) NSString *platform;
+@property (strong, nonatomic) NSString *formFactor;
+@property (strong, nonatomic) NSMutableDictionary<NSString *, NSString *> *metadata;
```
QuintinWillison, line 14: I wonder why this was ever mutable. 🙄 ... I would have assumed this class should be immutable in aggregate.

QuintinWillison (changes requested):
In particular for synthesised properties, I think we should prefer `copy` over `strong` where the type has a mutable counterpart.

Source/ARTDeviceDetails.h
maratal: If set copy  for mutable object static analyzer shows a complain on it:
```
1. Property of mutable type 'NSMutableDictionary<NSString *,NSString *>' has 'copy' attribute; an immutable object will be stored instead
```

QuintinWillison (approved):

PR #1439, opened 2022-06-30
`ARTCipherParams` and `ARTCrypto` doc comments

maratal:
Added docstrings according to https://github.com/ably/canonical-api-reference-prototyping/pull/8

Was checked for compatibility with `jazzy`.

### DIFF
--- a/Source/ARTCrypto.h
+++ b/Source/ARTCrypto.h
@@ -41,7 +41,7 @@ NS_ASSUME_NONNULL_BEGIN
 /// The length of the key in bits; for example 128 or 256.
 @property (readonly, nonatomic) NSUInteger keyLength;
 
-/// The cipher mode. Only "CBC" is supported.
+/// The cipher mode. Only CBC is supported.
 @property (readonly, getter=getMode) NSString *mode;
 
 - (instancetype)init UNAVAILABLE_ATTRIBUTE;
--- a/Source/ARTCrypto.h
+++ b/Source/ARTCrypto.h
@@ -66,13 +66,13 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  Generates a random key to be used in the encryption of the channel.
  @param keyLength The length of the key, in bits, to be generated.
- @return `NSData` The key as a binary data.
+ @return The key as a binary data.
  */
 + (NSData *)generateRandomKey:(NSUInteger)keyLength;
 
 /**
  Same as `+generateRandomKey:`, but with the default key length of the default algorithm: for AES this is 256 bits.
- @return `NSData` The key as a binary data.
+ @return The key as a binary data.
  */
 + (NSData *)generateRandomKey;
 
--- a/Source/ARTCrypto.h
+++ b/Source/ARTCrypto.h
@@ -61,14 +61,14 @@ NS_ASSUME_NONNULL_BEGIN
  @param params Overrides the default parameters. A suitable key must be provided as a minimum.
  @return `ARTCipherParams` A ARTCipherParams object, using the default values for any field not supplied.
  */
-+ (ARTCipherParams *)getDefaultParams:(NSDictionary *)values;
++ (ARTCipherParams *)getDefaultParams:(NSDictionary *)params;
 
 /**
  Generates a random key to be used in the encryption of the channel.
  @param keyLength The length of the key, in bits, to be generated. If not specified, this is equal to the default keyLength of the default algorithm: for AES this is 256 bits.
  @return `NSData` The key as a binary data.
  */
-+ (NSData *)generateRandomKey:(NSUInteger)length;
++ (NSData *)generateRandomKey:(NSUInteger)keyLength;
 
 /**
  Same as `+getDefaultParams:`, but with the default algorithm and key length: AES, 256 bits.
--- a/Source/ARTCrypto.h
+++ b/Source/ARTCrypto.h
@@ -28,13 +28,20 @@ NS_ASSUME_NONNULL_BEGIN
 @end
 
 /**
- ARTCipherParams contains configuration options for a channel cipher, including algorithm, mode, key length and key.
- Ably client libraries currently support AES with CBC, PKCS#7 with a default key length of 256 bits. All implementations also support AES128.
+ The ARTCipherParams object contains properties to configure encryption for a channel.
  */
 @interface ARTCipherParams : NSObject <ARTCipherParamsCompatible>
+
+/// The algorithm to use for encryption. Only AES is supported.
 @property (readonly, strong, nonatomic) NSString *algorithm;
+
+/// The private key used to encrypt and decrypt payloads.
 @property (readonly, strong, nonatomic) NSData *key;
+
+/// The length of the key in bits; for example 128 or 256.
 @property (readonly, nonatomic) NSUInteger keyLength;
+
+/// The cipher mode. Only "CBC" is supported.
 @property (readonly, getter=getMode) NSString *mode;
 
 - (instancetype)init UNAVAILABLE_ATTRIBUTE;
--- a/Source/ARTCrypto.h
+++ b/Source/ARTCrypto.h
@@ -59,7 +59,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  Retrieves, or optionally sets, the `ARTCipherParams` for the channel.
  @param params Overrides the default parameters. A suitable key must be provided as a minimum.
- @return `ARTCipherParams` A ARTCipherParams object, using the default values for any field not supplied.
+ @return An `ARTCipherParams` object, using the default values for any field not supplied.
  */
 + (ARTCipherParams *)getDefaultParams:(NSDictionary *)params;
 
--- a/Source/ARTCrypto.h
+++ b/Source/ARTCrypto.h
@@ -51,11 +51,30 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+/**
+ The ARTCrypto object ensures that message payloads are encrypted, can never be decrypted by Ably, and can only be decrypted by other clients that share the same secret symmetric key.
+ */
 @interface ARTCrypto : NSObject
 
+/**
+ Retrieves, or optionally sets, the `ARTCipherParams` for the channel.
+ @param params Overrides the default parameters. A suitable key must be provided as a minimum.
+ @return `ARTCipherParams` A ARTCipherParams object, using the default values for any field not supplied.
+ */
 + (ARTCipherParams *)getDefaultParams:(NSDictionary *)values;
-+ (NSData *)generateRandomKey;
+
+/**
+ Generates a random key to be used in the encryption of the channel.
+ @param keyLength The length of the key, in bits, to be generated. If not specified, this is equal to the default keyLength of the default algorithm: for AES this is 256 bits.
+ @return `NSData` The key as a binary data.
+ */
 + (NSData *)generateRandomKey:(NSUInteger)length;
+
+/**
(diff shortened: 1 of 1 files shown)


maratal:
https://sdk.ably.com/builds/ably/ably-cocoa/pull/1439/jazzydoc/Classes/ARTCrypto.html

lawrence-forooghian:
I've added @m-hulbert as a reviewer here, too, since he might be picking up any necessary changes once @tbedford rolls off this project.

Source/ARTCrypto.h
```diff
@@ -28,13 +28,20 @@ NS_ASSUME_NONNULL_BEGIN
 @end
 
 /**
- ARTCipherParams contains configuration options for a channel cipher, including algorithm, mode, key length and key.
- Ably client libraries currently support AES with CBC, PKCS#7 with a default key length of 256 bits. All implementations also support AES128.
+ The ARTCipherParams object contains properties to configure encryption for a channel.
```
lawrence-forooghian, line 31:
```suggestion
 The `ARTCipherParams` object contains properties to configure encryption for a channel.
```

@tbedford @m-hulbert I’d suggest that this sort of formatting should be in the canonical documentation, what do you think?

Source/ARTCrypto.h
```diff
@@ -28,13 +28,20 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
 @interface ARTCipherParams : NSObject <ARTCipherParamsCompatible>
+
+/// The algorithm to use for encryption. Only AES is supported.
```
lawrence-forooghian, line 35: @tbedford @m-hulbert For consistency with the documentation for `mode`, should this not be the following?

```suggestion
/// The algorithm to use for encryption. Only `"AES"` is supported.
```

Source/ARTCrypto.h
```diff
@@ -28,13 +28,20 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property (readonly, strong, nonatomic) NSString *algorithm;
+
+/// The private key used to encrypt and decrypt payloads.
 @property (readonly, strong, nonatomic) NSData *key;
+
+/// The length of the key in bits; for example 128 or 256.
 @property (readonly, nonatomic) NSUInteger keyLength;
+
+/// The cipher mode. Only "CBC" is supported.
```
lawrence-forooghian, line 44:
```suggestion
/// The cipher mode. Only `"CBC"` is supported.
```

@tbedford @m-hulbert as I mentioned elsewhere in this PR, I think this sort of formatting should be part of the canonical docs?

Source/ARTCrypto.h
```diff
@@ -44,11 +51,30 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+/**
+ The ARTCrypto object ensures that message payloads are encrypted, can never be decrypted by Ably, and can only be decrypted by other clients that share the same secret symmetric key.
```
lawrence-forooghian, line 55:
```suggestion
 The `ARTCrypto` object ensures that message payloads are encrypted, can never be decrypted by Ably, and can only be decrypted by other clients that share the same secret symmetric key.
```

@tbedford @m-hulbert as I mentioned elsewhere in this PR, I think this sort of formatting should be part of the canonical docs?

lawrence-forooghian, line 55:
```suggestion
 The ARTCrypto class ensures that message payloads are encrypted, can never be decrypted by Ably, and can only be decrypted by other clients that share the same secret symmetric key.
```

Source/ARTCrypto.h
lawrence-forooghian, line 31:
```suggestion
 An ARTCipherParams instance contains properties to configure encryption for a channel.
```

Source/ARTCrypto.h
lawrence-forooghian: I'm not sure whether I agree with this description entirely. I think that the description of how encryption works in Ably is a good one, but it's not the `ARTCrypto` class itself that's responsible for performing the encryption. It’s more of a utility class, which helps you to create the parameters used to configure encryption elsewhere in the SDK.

Source/ARTCrypto.h
```diff
@@ -44,11 +51,30 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
 @interface ARTCrypto : NSObject
 
-+ (ARTCipherParams *)getDefaultParams:(NSDictionary *)values;
+/**
+ Retrieves, or optionally sets, the `ARTCipherParams` for the channel.
```
lawrence-forooghian, line 60: I don't think I agree with this description. My understanding is that `getDefaultParams` is a factory method for creating an `ARTCipherParams` instance. I don't think it queries or manipulates a channel in any way (bear in mind that this is a class method, so there isn't even a "the channel" to speak of here).

Source/ARTCrypto.h
```diff
@@ -44,11 +51,30 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param params Overrides the default parameters. A suitable key must be provided as a minimum.
+ @return `ARTCipherParams` A ARTCipherParams object, using the default values for any field not supplied.
```
lawrence-forooghian, line 62:
```suggestion
 @return `ARTCipherParams` A `ARTCipherParams` object, using the default values for any field not supplied.
```

lawrence-forooghian, line 62: I think it should be "**An** ARTCipherParams object", not "A".

Source/ARTCrypto.h
```diff
@@ -44,11 +51,30 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
++ (ARTCipherParams *)getDefaultParams:(NSDictionary *)params;
+
+/**
+ Generates a random key to be used in the encryption of the channel.
+ @param keyLength The length of the key, in bits, to be generated. If not specified, this is equal to the default keyLength of the default algorithm: for AES this is 256 bits.
```
lawrence-forooghian, line 68:
```suggestion
 @param keyLength The length of the key, in bits, to be generated. If not specified, this is equal to the default key length of the default algorithm: for AES this is 256 bits.
```

@tbedford @m-hulbert I think an encryption algorithm is an abstract concept and not a class in our SDK, so I'm not sure that using `keyLength`, which suggests a property name, is appropriate here. Not sure though.

lawrence-forooghian, line 68: @maratal what does it mean to "not specify" a `keyLength` here?

Source/ARTCrypto.h
lawrence-forooghian, line 62: @maratal this question applies to all of the `@return` tags where you've written the return type as part of the tag – do you think this is necessary? The return type is already displayed in the Jazzy documentation as part of the method signature:

> Screenshot 2022-07-05 at 17 42 22

If you look at Apple's documentation ([example](https://developer.apple.com/documentation/foundation/nsstring/1410710-lengthofbytes)), the "Return Value" section doesn't repeat the return type; rather, it's used to provide semantic information about the return value:

> Screenshot 2022-07-05 at 17 44 50

Source/ARTCrypto.h
m-hulbert: Agree with the formatting. It is already styled as code in the canonical docs though.

Source/ARTCrypto.h
m-hulbert: This should just be `AES` (no " ") as per the canonical table.

Source/ARTCrypto.h
m-hulbert: This should just be `CBC` (no " ") as per the canonical table.

Source/ARTCrypto.h
m-hulbert: Agree with the change and this is also part of the canonical table already.

m-hulbert: Is something like the following more accurate?

```
The `ARTCrypto` class enables you to set the parameters required for configuring the encryption of message payloads.
```

Source/ARTCrypto.h
m-hulbert: Is this more accurate?

```
Creates an `ARTCipherParams` object to use for encryption. 
```

Source/ARTCrypto.h
m-hulbert: @QuintinWillison are you able to confirm this if possible?

Source/ARTCrypto.h
```diff
@@ -44,11 +51,30 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @return `NSData` The key as a binary data.
+ */
++ (NSData *)generateRandomKey:(NSUInteger)keyLength;
+
+/**
+ Same as `+getDefaultParams:`, but with the default algorithm and key length: AES, 256 bits.
```
m-hulbert, line 74: The description for `generateRandomKey` is the following in the canonical table:

```
Generates a random key to be used in the encryption of the channel.
```

I'm assuming this should be something like "Generates a random key use for encrypting message payloads." instead given @lawrence-forooghian 's comment on `getDefaultParams`.

Source/ARTCrypto.h
maratal: This will link to the page that user is already on. Is this what was expected? @lawrence-forooghian

Source/ARTCrypto.h
maratal: How about

> The `ARTCrypto` class enables you to generate the parameters required for configuring the encryption of message payloads.

@lawrence-forooghian @m-hulbert

Source/ARTCrypto.h
maratal: Maybe 
>The `ARTCipherParams` properties allows you to configure encryption for a channel.
@lawrence-forooghian @m-hulbert

Source/ARTCrypto.h
maratal: But we have two methods with this name: one with parameter `keyLength`, and one without parameters. Canonical table contains only one description and notes that "_If not specified, this is equal to the default keyLength of the default algorithm: for AES this is 256 bits._". So I decided to produce different description for method without parameter. Also I've messed up with the referred method: `+getDefaultParams:` instead of `generateRandomKey:`. Duh! @lawrence-forooghian @m-hulbert

Source/ARTCrypto.h
maratal: https://github.com/ably/ably-cocoa/pull/1439/commits/0deb0a0ef9bd63bd007f7c36ffcac863575920a7

Source/ARTCrypto.h
maratal: https://github.com/ably/ably-cocoa/pull/1439/commits/3e0875b98ece697f3b01cc346f3b803c886fa423

Source/ARTCrypto.h
maratal: https://github.com/ably/ably-cocoa/pull/1439/commits/3cce6135c1776c6802ab5a640dd0331fc46b0894

Source/ARTCrypto.h
maratal: https://github.com/ably/ably-cocoa/pull/1439/commits/3cce6135c1776c6802ab5a640dd0331fc46b0894

Source/ARTCrypto.h
maratal: https://github.com/ably/ably-cocoa/pull/1439/commits/451185143a27f803a68b015fdd798df5e5e369f4

maratal:
Closed in favor of https://github.com/ably/ably-cocoa/pull/1491

Issue #1380, opened 2022-04-29
SCNetworkReachabilitySetCallback (Crashed: io.ably.main)


MrNiceMrGood:
### Which version of the Ably SDK are you using? 
1.2.10

### On which platform does the issue happen? 
Multiple

### Are you using Cocoapods? 
1.11.2

### Which version of Xcode are you using? 
Xcode 13.3.1
Build version 13E500a

### What did you do?
I don't have too much info. I have found this to happen to some users in production. Never managed to reproduce directly.
You can check the firebase stack trace.

This looks like a similar issue as https://github.com/ably/ably-cocoa/issues/593. 

```
Crashed: io.ably.main
0  libsystem_pthread.dylib        0x4924 <redacted> + 126
1  SystemConfiguration            0x28938 SCNetworkReachabilitySetCallback + 64
2  Ably                           0x27570 -[ARTOSReachability listenForHost:callback:] + 348
3  Ably                           0x3df10 -[ARTRealtimeInternal transitionSideEffects:] + 4516
4  Ably                           0x3ca90 -[ARTRealtimeInternal transition:withErrorInfo:] + 388
5  Ably                           0x17210 -[ARTEventListener timeout] + 84
6  Ably                           0x1738c __30-[ARTEventListener startTimer]_block_invoke + 36
7  Ably                           0x1ae88 __53-[ARTScheduledBlockHandle initWithDelay:queue:block:]_block_invoke + 100
8  libdispatch.dylib              0x103e4 _dispatch_block_async_invoke2 + 104
9  libdispatch.dylib              0x61298 _dispatch_client_callout + 16
10 libdispatch.dylib              0x6028 _dispatch_continuation_pop$VARIANT$mp + 412
11 libdispatch.dylib              0x16d64 _dispatch_source_invoke$VARIANT$mp + 1304
12 libdispatch.dylib              0x9908 _dispatch_lane_serial_drain$VARIANT$mp + 300
13 libdispatch.dylib              0xa518 _dispatch_lane_invoke$VARIANT$mp + 420
14 libdispatch.dylib              0x13fac _dispatch_workloop_worker_thread + 712
15 libsystem_pthread.dylib        0xb5bc _pthread_wqthread + 272
16 libsystem_pthread.dylib        0xe86c start_wqthread + 8
```

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-1816) by [Unito](https://www.unito.io)

maratal:
[...]

Issue #593, opened 2017-04-17
Crash in ARTOSReachability


escfrya:
Seems like ARTOSReachability_Callback has referance to deallocated object.
ably version: 1.0.1

{code}
Crashed: com.apple.main-thread
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000010

0	Ably					-[ARTOSReachability internalCallback:] + 80
1 	Ably					ARTOSReachability_Callback + 124
2	SystemConfiguration		reachPerformAndUnlock + 580
3	CoreFoundation			__CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 24
4	CoreFoundation			__CFRunLoopDoSources0 + 540
5	CoreFoundation			__CFRunLoopRun + 744
6	CoreFoundation			CFRunLoopRunSpecific + 424
7	GraphicsServices		GSEventRunModal + 100
8 	UIKit					UIApplicationMain + 208
9	test					AppDelegate.swift line 15
10	libdyld.dylib	 		start + 4

{code}

tcard:
@escfrya Thanks for the report! Is this happening always/often? If so, could you tell us how? Logs with ARTClientOptions.logLevel set to .Debug would be ideal.

escfrya:
@tcard I can't reproduce this bug on my device.
Stacktrace from Crashlytics, and i see 18 crashes in dashboard (not so often).

tcard:
I've added 8297ea4 (1.0) and bada06b (0.8). That should prevent a crash, but the underlying condition causing the crash is probably not fixed. We also can't reproduce this issue, so this requires further looking into.

@escfrya Could you tell us what you have at AppDelegate.swift, line 15?

escfrya:
@tcard line 15 - class AppDelegate: UIResponder, UIApplicationDelegate {

escfrya:
@tcard Unfortunately, crash reproduced in new app version(ably version 1.0.3)

```
Crashed: com.apple.main-thread
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000010

0	Ably	-[ARTOSReachability internalCallback:] + 84
1       Ably	ARTOSReachability_Callback + 144
2	SystemConfiguration	 reachPerformAndUnlock + 640
[...]

PR #1452, opened 2022-07-28
Fix dereference issue on ARTOSReachability

ikbalkaya:
This PR follows [this blog post](https://msolarana.netlify.app/2018/08/04/monitoring-network-changes/) to potentially fix  for https://github.com/ably/ably-cocoa/issues/593 and https://github.com/ably/ably-cocoa/issues/1380

### DIFF
--- a/Source/Private/ARTOSReachability.m
+++ b/Source/Private/ARTOSReachability.m
@@ -78,6 +78,10 @@ static void ARTOSReachability_Callback(SCNetworkReachabilityRef target, SCNetwor
     if (_reachabilityRef != NULL) {
         [NSNotificationCenter.defaultCenter removeObserver:self name:kARTOSReachabilityNetworkIsReachableNotification object:nil];
         [NSNotificationCenter.defaultCenter removeObserver:self name:kARTOSReachabilityNetworkIsDownNotification object:nil];
+
+        [NSNotificationCenter.defaultCenter removeObserver:self name:kARTOSReachabilityNetworkIsReachableNotification object:nil];
+        [NSNotificationCenter.defaultCenter removeObserver:self name:kARTOSReachabilityNetworkIsDownNotification object:nil];
+
         SCNetworkReachabilityUnscheduleFromRunLoop(_reachabilityRef, CFRunLoopGetCurrent(), kCFRunLoopDefaultMode);
         SCNetworkReachabilitySetDispatchQueue(_reachabilityRef, NULL);
         SCNetworkReachabilitySetCallback(_reachabilityRef, NULL, NULL);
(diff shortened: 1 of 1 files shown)


Source/Private/ARTOSReachability.m
```diff
@@ -44,11 +48,18 @@ - (void)listenForHost:(NSString *)host callback:(void (^)(BOOL))callback {
     _host = host;
     _callback = callback;
 
-    _reachabilityRef = SCNetworkReachabilityCreateWithName(NULL, [host UTF8String]);
+    dispatch_queue_attr_t attrs = dispatch_queue_attr_make_with_qos_class(DISPATCH_QUEUE_SERIAL, QOS_CLASS_UTILITY, DISPATCH_QUEUE_PRIORITY_DEFAULT);
+    _queue = dispatch_queue_create("com.ably.reachability-monitor", attrs);
+
+    _reachabilityRef = CFAutorelease(SCNetworkReachabilityCreateWithName(kCFAllocatorDefault, [host UTF8String]));
```
JakubJankowski, line 54: To be honest - It's been a while since I last worked with the manual objC memory management, so apologies if I'm completely wrong here. But - doesn't `CFAutorelease` mean that the object is added to the autoreleasepool, and will be cleared up sometime after the end of the current scope?

Soo... won't the `_reachabilityRef` be nullified at some point, even though it's not our intention?

We're already firing `CFRelease(_reachabilityRef);` in `dealloc`, which makes sense to me.

ikbalkaya: That was my thinking as well and probably the thinking of original author 😄 But apparently core foundation objects needs to be manually released. I have read it in a couple of places, one example is t[his SO answer](https://stackoverflow.com/a/31435846). So I suspect that the problem might have been that this object might have remained in memory pointing to objects that has already been deallocated. But mine is also just a guess. I just wanted to check whether this would make sense to you too.

JakubJankowski: Yeah, I also recall that Core Foundation objects have to be manually released. In this case I think that the `CFRelease(_reachabilityRef)` in the dealloc should work for that purpose, and using `CFAutorelease` here as well might introduce some other issues that we're not expecting. 

Other than that point everything else makes sense to me, so I'll throw in my approval for this PR.

JakubJankowski (approved):

Source/Private/ARTOSReachability.m
ikbalkaya: Resolving this on behalf of @JakubJankowski as I removed `CFAutoRelease` https://github.com/ably/ably-cocoa/pull/1452/commits/742ffbf4f309850100bc7799ca7ae156c28d8edf

maratal:
I'm closing this PR, changes in `ARTOSReachability` requires more deep investigation @ikbalkaya

maratal:
New PR https://github.com/ably/ably-cocoa/pull/1453

PR #1517, opened 2022-11-03
Enforce RTP8j spec point

lawrence-forooghian:
I've split this from #1515, to extract the part that's needed for fixing https://github.com/ably/ably-asset-tracking-swift/issues/401.


maratal (approved):
LGTM

PR #1519, opened 2022-11-03
Use CI to build the `Ably.framework.zip` that we attach to releases

lawrence-forooghian:
I was unable to build this file on my machine since I (foolishly) upgraded to macOS 13 and hence can no longer use Xcode 13.4.1. I think that building this file on CI instead of the developers’ machines is kinda better anyway since it gives us a more consistent environment for the build.

I’ve kept the `pod lib lint` local part though, since at the moment we’ll still be running `pod trunk push` from developers’ machines, and `pod lib lint` gives us a way of checking whether that command is likely to succeed.


.github/workflows/check-pod.yaml
```diff
@@ -40,3 +40,8 @@ jobs:
           unzip -l Ably.framework.zip | grep 'iOS/Ably.framework'
       - name: Validate pod
         run: pod lib lint
+      - name: Archive built framework
+        uses: actions/upload-artifact@v3
+        with:
+          name: carthage-built-framework
+          path: Ably.framework.zip
```
maratal, line 47: I downloaded it, and during unzip this happens - it produces `Ably.framework.zip` and immediately unzip it further into Carthage folder:

https://user-images.githubusercontent.com/45491869/199824768-4d8abd49-aabb-4718-9b6d-450645742eb3.mov

lawrence-forooghian: I noticed that, and you're right, it's not good. I should have done something about it in the first place. Will see what I can do.

maratal: I think it's macos sees that it's an embed archive and continues to unzip it. Probably you should include a command that unzips it and zips again with a proper name.

lawrence-forooghian: I've introduced a directory at the top level of the artifact zip file; this means that Archive Utility won't try to continue unzipping. (Explained it in the action YAML file too.)

lawrence-forooghian: Hopefully this is how it'll behave now:

https://user-images.githubusercontent.com/53756884/199828198-63874010-2465-413b-8b09-ee05fb674180.mov

maratal (approved):

PR #1520, opened 2022-11-06
Incremental backoff and jitter

### LINKED ISSUE #1431: Implement incremental backoff and jitter
Please see the example implementation in js here https://github.com/ably/ably-js/issues/997

Incremental backoff and jitter was added to the features spec in https://github.com/ably/docs/pull/1445

To implement this we need to change the retry behaviour for:

* When the client connection is in the `DISCONNECTED` state
* When a `RealtimeChannel` is in the `SUSPENDED` state

The new retry behaviour is specified in RTB1 of the features spec.

### RTB1a
The backoff coefficient described in `RTB1a` can be calculated in javascript like so:

```ts
function getBackoffCoefficient(n) {
  return Math.min((n + 2) / 3, 2);
}
```

### RTB1b
The jitter coefficient described in `RTB1b` can be calculated in javascript like so:

```ts
function getJitterCoefficient() {
  return 1 - Math.random() * 0.2;
}
```

### RTB1
The overall retry time calculation should look like:

```ts
function getRetryTime(initValue, n) {
  return initValue * getBackoffCoefficient(n) * getJitterCoefficient();
}
```

The following code can be used to test it:

```ts
[1, 2, 3, 4, 5].forEach((n) => {
  console.log(getRetryTime(initValue, n));
});
```

Which, with an `initValue` of 15 (seconds) should print something like:

```
13.917470451245713
18.415226855678757
20.444851032898747
26.650729887092275
27.803382948778786
```

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-1993) by [Unito](https://www.unito.io)

maratal:
Closes #1431 
There is an unresolved conversation in the reference implementation for channel - https://github.com/ably/ably-js/pull/1008#discussion_r925971338 I've decided to take `state` check from Owen's implementation.


owenpearson (changes requested):
The implementation LGTM, although I don't know much objective-c so it's hard to say for sure.

A couple of things to mention:
- There are some newly skipped tests here which shouldn't be the case - is there a particular reason why these tests can't be fixed?
- It might be worth considering using a seed for random number generation in the tests here to make retry intervals deterministic (this wasn't possible in ably-js because `Math.random()` doesn't take a seed argument).

lawrence-forooghian (commented):
Will review the content of the PR next, but re the skipped tests, I think ideally we should investigate and fix this here, and not leave it for #1526 and #1527. The newly-skipped tests are tests that were passing before, and have only started to fail after we added these new tests. Is there something that makes you think that the problem lies with the existing tests, and not the new ones that we've added?

maratal:
I guess most of all other failing/flaky tests were reported like that, but
I agree we should stop this practice and try to fix it in place.

чт, 17 нояб. 2022 г. в 14:50, Lawrence Forooghian ***@***.***

lawrence-forooghian:
Are the tests that you've marked as skipped failing consistently after adding the new tests?

Source/ARTRealtime.m
```diff
@@ -1581,4 +1583,25 @@ - (ARTLocalDevice *)device {
...
++ (double)jitterDelta {
+    return 0.2;
+}
+
++ (double)jitterCoefficient {
+    double coeff = 1.0 - 0.1 * arc4random_uniform(10) * self.jitterDelta;
+    return coeff;
+}
+
++ (NSTimeInterval)retryDelayFromTimeout:(NSTimeInterval)timeout retryCount:(NSInteger)retryCount {
```
lawrence-forooghian, line 1602: I find `retryCount` to be a bit unclear – does it mean the number of retries that have already occurred (in which case we are now on the `retryCount + 1`<sup>th</sup> retry) or the number of the current retry?

Also, instead of just "timeout", how about using the language "initial retry timeout" from RTB1?

```suggestion
+ (NSTimeInterval)retryDelayFromInitialRetryTimeout:(NSTimeInterval)initialRetryTimeout forRetryNumber:(NSInteger)retryNumber {
```

Source/ARTRealtime+Private.h
```diff
@@ -113,6 +116,10 @@ NS_ASSUME_NONNULL_BEGIN
 // Message sending
 - (void)send:(ARTProtocolMessage *)msg sentCallback:(nullable ARTCallback)sentCallback ackCallback:(nullable ARTStatusCallback)ackCallback;
 
++ (double)backoffCoefficientWithNumber:(NSInteger)number;
```
lawrence-forooghian, line 119: I'm not sure whether `ARTRealtimeInternal` is the right place to put this. Given that the functionality is also used for deciding channel retry logic, the choice feels a bit arbitrary. I wonder whether it might be better to introduce a new utility class to contain this timeout calculation logic.

lawrence-forooghian: Also, I think the naming could be clearer:

```suggestion
+ (double)backoffCoefficientForRetryNumber:(NSInteger)retryNumber;
```

Source/ARTRealtime+Private.h
```diff
@@ -113,6 +116,10 @@ NS_ASSUME_NONNULL_BEGIN
...
++ (double)jitterDelta;
++ (double)jitterCoefficient;
```
lawrence-forooghian, line 121: Perhaps the naming could make it clear that this is doing something and isn't just returning a static value – maybe `generateRandomJitterCoefficient` or at least `generateJitterCoefficient`?

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2234,41 +2234,40 @@ class RealtimeClientConnectionTests: XCTestCase {
...
+            var disconnectedAt: Date!
+            
+            client.connection.on(.disconnected) { stateChange in
                 expect(stateChange.reason!.message).to(contain("timed out"))
                 expect(stateChange.previous).to(equal(ARTRealtimeConnectionState.connecting))
-                expect(stateChange.retryIn).to(beCloseTo(options.disconnectedRetryTimeout))
-                partialDone()
-                start = NSDate()
+                
+                let expectedDelay = ARTRealtimeInternal.retryDelay(fromTimeout: options.disconnectedRetryTimeout, retryCount: retryCount + 1)
```
lawrence-forooghian, line 2245: I think that if we're going to use our own code inside the tests, then we probably need to have some unit tests for that code separately too.

lawrence-forooghian (commented):
I need to dive into this properly and understand the tests, but here are some initial thoughts from looking over the code.

lawrence-forooghian:
I just tried (locally) un-skipping the tests mentioned in #1526 and #1527, and I didn't see them failing. But what I did see was some other failures:

- RealtimeClientConnectionTests.swift:2193 test__058__Connection__connection_request_fails__connection_attempt_should_fail_if_not_connected_within_the_default_realtime_request_timeout(): Waited more than 20.0 seconds
- RealtimeClientConnectionTests.swift:2206 test__058__Connection__connection_request_fails__connection_attempt_should_fail_if_not_connected_within_the_default_realtime_request_timeout(): Start date or end date are empty
- RealtimeClientConnectionTests.swift:2246 test__059__Connection__connection_request_fails__if_connection_attempt_fails_connection_state_will_transition_to_DISCONNECTED_with_periodic_connection_attempts_with_incremental_backoff_and_jitter_until_SUSPENDED(): expected to be close to <1.0933> (within 0.2187), got <1.3333>

maratal:
> Are the tests that you've marked as skipped failing consistently after adding the new tests?

Failures are gone now, fix [here](https://github.com/ably/ably-cocoa/pull/1520/commits/d958772d841a093f2c82635737bcbdc612140776)

Source/ARTRealtime.m
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/171efb801fb805bb3d47064c713221affaf6405a

Source/ARTRealtime+Private.h
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/171efb801fb805bb3d47064c713221affaf6405a

Source/ARTRealtime+Private.h
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/171efb801fb805bb3d47064c713221affaf6405a

Spec/Tests/RealtimeClientConnectionTests.swift
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/5ce76ddd9b961c657e411f91c4beb26f5b10e386

Source/ARTRealtime+Private.h
maratal: >a new utility class to contain this timeout calculation logic
https://github.com/ably/ably-cocoa/pull/1520/commits/c11f2b2e6890338d8f233d27c000241ce47f0812
https://github.com/ably/ably-cocoa/pull/1520/commits/fe3139983f2d0d2bb0502c118deb37889e5e268a

Source/ARTRealtime.m
maratal: >retryCount to be a bit unclear

https://github.com/ably/ably-cocoa/pull/1520/commits/aed866f7ecfb7e1ed9fa4841dea5845337158d89

maratal:
> * There are some newly skipped tests here which shouldn't be the case - is there a particular reason why these tests can't be fixed?
> 
> * It might be worth considering using a seed for random number generation in the tests here to make retry intervals deterministic (this wasn't possible in ably-js because `Math.random()` doesn't take a seed argument).

Both fixed now @owenpearson

Spec/Tests/UtilsTests.swift
```diff
@@ -0,0 +1,42 @@
+import Ably
+import Nimble
+import XCTest
+
+class UtilsTests: XCTestCase {
+    
+    struct Constants {
+        static let backoffValues = [ 1.0, 1.3333333333333333, 1.6666666666666667, 2.0, 2.0 ]
```
owenpearson, line 8: Feel free to ignore because I don't know if this is a standard thing in Objective-C, but if it's possible to use computed values here (ie `4/3` rather than `1.3333333333333333`) that would look more readable IMO.

owenpearson (dismissed):
LGTM (with one minor suggestion)

Spec/Tests/UtilsTests.swift
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/12b1cdc9c48990cc0d35cb01947480a4b212d8c7

Spec/Tests/UtilsTests.swift
```diff
@@ -0,0 +1,42 @@
...
+        let delays = (1...5).map { n in ARTUtils.retryDelay(fromInitialRetryTimeout: initialRetryTimeout, forRetryNumber: n) }
+        
+        for i in 0..<delays.count {
+            let delay = delays[i]
+            let targetValue = Constants.delayValues[i]
+            let backoff = Constants.backoffValues[i]
+            let maxValue = initialRetryTimeout * backoff
+            let minValue = initialRetryTimeout * backoff * (1.0 - ARTUtils.jitterDelta())
+            let tolerance = (maxValue - minValue) / 2.0
+            expect(delay).to(beCloseTo(targetValue, within: tolerance))
```
lawrence-forooghian, line 39: I'm a bit confused about the purpose of the `Constants.delayValues` array here – instead of this assertion could we not assert that `delay` is between `minValue` and `maxValue`?

Spec/Tests/UtilsTests.swift
```diff
@@ -0,0 +1,42 @@
...
+        static let delayValues = [ 13.5, 18.0, 22.5, 27.0, 27.0 ] // calculated for randomFraction = 0.5 (jitterCoeff = 0.9)
+    }
+    
+    // MARK: Backoff and jitter
+    
+    func testRetryDelayCalculationForRetryNumberWithConstantRandomSeeding() {
+        let initialRetryTimeout = 15.0
+        
+        ARTUtils.seedRandom(withNumber: 5)
+        let delays1 = (1...5).map { n in initialRetryTimeout * ARTUtils.backoffCoefficient(forRetryNumber: n) * ARTUtils.generateJitterCoefficient() }
```
lawrence-forooghian, line 18: Should this and the test below not be consistent in how they access the n<sup>th</sup> backoff coefficient? Either using `Constants.backoffValues` or using `ARTUtils.backoffCoefficient(forRetryNumber:)`.

Source/Private/ARTUtils.h
```diff
@@ -0,0 +1,24 @@
...
+
++ (void)seedRandomWithNumber:(unsigned int)number;
++ (void)seedRandomWithSystemTime;
++ (double)randomFraction;
++ (long)randomFromZeroToNumber:(long)number;
+
+#pragma mark - Incremental backoff and jitter // RTB1
+
++ (double)jitterDelta;
++ (double)generateJitterCoefficient;
```
lawrence-forooghian, line 17: Does this need to be part of the public interface of this class?

Source/Private/ARTUtils.h
```diff
@@ -0,0 +1,24 @@
...
++ (double)backoffCoefficientForRetryNumber:(NSInteger)retryNumber;
```
lawrence-forooghian, line 18: Does this need to be part of the public interface of this class?

Source/Private/ARTUtils.h
```diff
@@ -0,0 +1,24 @@
...
+
++ (NSTimeInterval)retryDelayFromInitialRetryTimeout:(NSTimeInterval)timeout forRetryNumber:(NSInteger)retryNumber;
```
lawrence-forooghian, line 20:
```suggestion
+ (NSTimeInterval)generateRetryDelayFromInitialRetryTimeout:(NSTimeInterval)timeout forRetryNumber:(NSInteger)retryNumber;
```

Source/ARTRealtimeChannel.m
```diff
@@ -603,14 +610,11 @@ - (void)transition:(ARTRealtimeChannelState)state status:(ARTStatus *)status {
...
             [_attachedEventEmitter emit:nil with:status.errorInfo];
             if (self.realtime.shouldSendEvents) {
-                channelRetryListener = [self unlessStateChangesBefore:self.realtime.options.channelRetryTimeout do:^{
+                NSTimeInterval retryDelay = [ARTUtils retryDelayFromInitialRetryTimeout:self.realtime.options.channelRetryTimeout
+                                                                         forRetryNumber:++_retryCount];
+                channelRetryListener = [self unlessStateChangesBefore:retryDelay do:^{
                     [self.realtime.logger debug:__FILE__ line:__LINE__ message:@"RT:%p C:%p (%@) reattach initiated by retry timeout", self->_realtime, self, self.name];
-                    [self reattachWithReason:nil callback:^(ARTErrorInfo *errorInfo) {
-                        if (errorInfo) {
-                            ARTStatus *status = [ARTStatus state:ARTStateError info:errorInfo];
```
lawrence-forooghian, line 610: Why have we removed this code that sets the channel to suspended? I haven't looked into its purpose but at a glance it doesn't seem related to the intention of this pull request.

Spec/Tests/RealtimeClientChannelTests.swift
```diff
@@ -3889,7 +3889,67 @@ class RealtimeClientChannelTests: XCTestCase {
...
+        let detachedMessageWithError = AblyTests.newErrorProtocolMessage()
+        detachedMessageWithError.action = .detached
+        detachedMessageWithError.channel = channel.name
+        
+        ARTUtils.seedRandom(withNumber: 5)
+        
+        var retryNumber = 1
+        let maxRetryCount = 5
+        
+        let predefinedDelays = [ 0.85, 1.16, 1.633333333333333, 1.712, 1.696 ]
```
lawrence-forooghian, line 3927: The meaning of these numbers isn’t clear to me – is there a way to make this clearer, by naming or commenting or both?

This applies to the similar code in `Spec/Tests/RealtimeClientConnectionTests.swift` too.

Spec/Tests/RealtimeClientChannelTests.swift
```diff
@@ -3889,7 +3889,67 @@ class RealtimeClientChannelTests: XCTestCase {
...
+                if retryNumber > maxRetryCount {
+                    done()
+                }
+            }
+            channel.on(.attached) { _ in
+                fail("Should not receive attached message")
+            }
+            client.internal.transport?.receive(detachedMessageWithError) // force to .suspended
+        }
+        expect(actualDelays).to(beCloseTo(predefinedDelays, within: 0.15))
```
lawrence-forooghian, line 3950: The meaning of 0.15 isn’t clear to me – is there a way to make this clearer?

This applies to the similar code in `Spec/Tests/RealtimeClientConnectionTests.swift` too.

lawrence-forooghian (commented):
Thanks for the changes 🙂 Another round of feedback for you. I also noticed that one of the `ARTUtils` tests [seems to be failing in CI](https://github.com/ably/ably-cocoa/actions/runs/3568141358/jobs/5996671659).

maratal:
> one of the `ARTUtils` tests [seems to be failing in CI](https://github.com/ably/ably-cocoa/actions/runs/3568141358/jobs/5996671659).

_expected to be close to <22.5> (within 2.5), got <25>_

I've just used calc to check whether it should work, but it doesn't 😅

Ably.xcodeproj/project.pbxproj
```diff
@@ -2517,7 +2535,7 @@
 					};
 					96BF61301A35B2AB004CF2B3 = {
 						CreatedOnToolsVersion = 6.1.1;
-						LastSwiftMigration = 1250;
+						LastSwiftMigration = 1410;
```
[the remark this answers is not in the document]
maratal: I think that's just a noise committed by accident (and similar below)...

Source/Private/ARTUtils.h
```diff
@@ -0,0 +1,24 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ARTUtils : NSObject
+
+#pragma mark - Random generator
+
++ (void)seedRandomWithNumber:(unsigned int)number;
++ (void)seedRandomWithSystemTime;
++ (double)randomFraction;
```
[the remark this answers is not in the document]
maratal: What if someone would want to use random 0...1.0 out of the box in Objective-C?

Spec/Tests/UtilsTests.swift
maratal: Good catch, thanks. What would you use in both places, constants or method?

Spec/Tests/UtilsTests.swift
maratal: Thanks, is there a one-liner for `expect` for this, just curious.

Source/ARTRealtimeChannel.m
maratal: Here is an explanation https://github.com/ably/ably-cocoa/pull/1520/commits/21b207f5f2926bf72ef7c961e8a0ea27feb46047

Spec/Tests/RealtimeClientChannelTests.swift
maratal: That was the best accuracy I could get (never hits exact values). With `0.1` fails often.

Source/ARTRealtimeChannel.m
lawrence-forooghian: Could you open this one as a separate PR? I think that if it's a bug fix then we should give it a proper review separately.

Source/Private/ARTUtils.h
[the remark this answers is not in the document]
lawrence-forooghian: Do you mean "what if a developer of ably-cocoa wanted to have access to a random number in 0...1.0"? If that's what you mean, then I think that we can expose this functionality if it's ever needed.

Ably.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
lawrence-forooghian: Can we revert these changes please?

Spec/Tests/UtilsTests.swift
lawrence-forooghian: According to [the docs](https://quick.github.io/Nimble/#collection-membership) you can do `expect(5).to(beWithin(1...10))`, but that method doesn't seem to exist for me (we're on quite an old version of Nimble), so I guess just `expect((minValue...maxValue).contains(delay)).to(beTrue())`?

Spec/Tests/UtilsTests.swift
lawrence-forooghian: It's the sort of thing that I'm never sure about when testing, to be honest. Do you have any thoughts on it?

Spec/Tests/RealtimeClientChannelTests.swift
lawrence-forooghian: OK, I'll look at the test again once I've got a better idea of [what `predefinedDelays` is](https://github.com/ably/ably-cocoa/pull/1520#discussion_r1036421743).

Spec/Tests/UtilsTests.swift
lawrence-forooghian: Maybe, since we're not testing the calculation of the backoff coefficient elsewhere, we should use the `Constants.backoffValues`. But as I say, keen to hear your thoughts.

Spec/Tests/RealtimeClientChannelTests.swift
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/a929718d7bb75780a5375ba56dcf8aef8573b93a

Source/ARTRealtimeChannel.m
maratal: https://github.com/ably/ably-cocoa/pull/1551

Spec/Tests/UtilsTests.swift
maratal: While cleaning all this up, I've decided to put those values in the comment and go with an actual function https://github.com/ably/ably-cocoa/pull/1520/commits/7b35df981bdb2e1cd61012a161f1072aa48d3ae3

Spec/Tests/UtilsTests.swift
maratal: Neat solution
https://github.com/ably/ably-cocoa/commit/7b35df981bdb2e1cd61012a161f1072aa48d3ae3

Ably.xcodeproj/project.pbxproj
```diff
@@ -3414,13 +3441,15 @@
 				SDKROOT = macosx;
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
 				SWIFT_OBJC_BRIDGING_HEADER = "Spec/AblySpec-Bridging-Header.h";
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
```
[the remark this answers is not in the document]
maratal: 5446de1aeb2f945e642897bd377deaa9c9edebb9

Ably.xcodeproj/project.pbxproj
```diff
@@ -3155,6 +3179,7 @@
 		856AAC931B6E304B00B07119 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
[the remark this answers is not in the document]
maratal: 5446de1aeb2f945e642897bd377deaa9c9edebb9

Ably.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
maratal: 5446de1aeb2f945e642897bd377deaa9c9edebb9

Source/Private/ARTUtils.h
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/b17d6b572f236a2f2147f83324ad0fb336f32ec1

Source/Private/ARTUtils.h
```diff
@@ -0,0 +1,24 @@
...
++ (long)randomFromZeroToNumber:(long)number;
```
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/b17d6b572f236a2f2147f83324ad0fb336f32ec1

Source/Private/ARTUtils.h
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/b17d6b572f236a2f2147f83324ad0fb336f32ec1

Source/Private/ARTUtils.h
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/b17d6b572f236a2f2147f83324ad0fb336f32ec1

Ably.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
lawrence-forooghian: Hey Marat, I'm not sure what the commit you've linked to has to do with this comment? These `LastSwiftMigration` changes still seem to be in the PR.

Source/Private/ARTUtils.h
lawrence-forooghian: I see your point. I guess I don't have very strong opinions on it. Will resolve this comment.

Source/ARTRealtimeChannel.m
lawrence-forooghian: 👍 thanks!

Spec/Test Utilities/TestUtilities.swift
```diff
@@ -360,6 +360,17 @@ class AblyTests {
         let items = jsonItems.map{ $0 }.map(CryptoTestItem.init)
         return (keyData, ivData, items)
     }
+    
+    class func backoffWithJitterDelaysForTimeout(_ timeout: TimeInterval) -> [TimeInterval] {
+        // This is calculated for seeding number = 5 (randomFraction gives [ 0.75, 0.65, 0.1, 0.72, 0.76 ])
```
lawrence-forooghian, line 365: From [reading the documentation for `random`](https://pubs.opengroup.org/onlinepubs/007908799/xsh/initstate.html) it's not very clear to me whether `random` is guaranteed to always return the same numbers for all implementations (i.e. whether the algorithm could change in an OS update). But I don't think it's a huge problem, we can cross that bridge if we need to later.

Ably.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
lawrence-forooghian: > those are added by Xcode every time files added to the project

Oh really? I thought normally Xcode does this when you click the "Update to recommended settings" button (I forgot the naming exactly). I'm sure I've added a file with Xcode 14.something recently and it didn't do that…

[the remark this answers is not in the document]
maratal: 🤷‍♂️

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2283,95 +2288,52 @@ class RealtimeClientConnectionTests: XCTestCase {
...
+                    expect(connectingAt).to(beNil())
+                }
+                else if stateChange.previous == .suspended {
+                    connectingAt = Date()
+                    expect(connectingAt.timeIntervalSince(suspendedAt)).to(beCloseTo(options.suspendedRetryTimeout, within: 0.5))
+                }
+                else {
+                    expect(stateChange.previous).to(equal(ARTRealtimeConnectionState.disconnected))
+                }
+                retryCount += 1
```
lawrence-forooghian, line 2331: In this test, do we care about the overall number of retries, or do we just care about the number of retries that happened after the connection became `SUSPENDED`? The name of the test suggests it's the latter.

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2283,95 +2288,52 @@ class RealtimeClientConnectionTests: XCTestCase {
         client.internal.shouldImmediatelyReconnect = false
         defer { client.dispose(); client.close() }
 
-        let ttlHookToken = client.overrideConnectionStateTTL(0.3)
+        let ttlHookToken = client.overrideConnectionStateTTL(4)
         defer { ttlHookToken.remove() }
 
+        var retryCount = 0
+        let maxRetryCount = 8 // RTN14f: reasonable amount within testTimeout
```
lawrence-forooghian, line 2295: Perhaps a comment explaining how you calculated this would make this number a bit less magical and give people more confidence to change the test in the future?

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2206,22 +2206,22 @@ class RealtimeClientConnectionTests: XCTestCase {
...
-    func test__059__Connection__connection_request_fails__connection_attempt_fails_for_any_recoverable_reason() {
+    
+    // RTN14d, RTB1
+    func test__059__Connection__connection_request_fails__if_connection_attempt_fails_connection_state_will_transition_to_DISCONNECTED_with_periodic_connection_attempts_with_incremental_backoff_and_jitter_until_SUSPENDED() {
         let options = AblyTests.commonAppSetup()
         options.realtimeHost = "10.255.255.1" // non-routable IP address
         options.disconnectedRetryTimeout = 1.0
         options.autoConnect = false
-        let expectedTime = 3.0
+        let connectionStateTtl = 7.0
```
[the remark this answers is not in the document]
lawrence-forooghian: Also, perhaps a comment explaining how you calculated this (in relation to the contents of `AblyTests.backoffWithJitterDelaysForTimeout(options.disconnectedRetryTimeout)`) would make this number a bit less magical and give people more confidence to change the test in the future?

Source/ARTRealtime.m
```diff
@@ -643,6 +644,7 @@ - (ARTEventListener *)transitionSideEffects:(ARTConnectionStateChange *)stateCha
             break;
         }
         case ARTRealtimeSuspended: {
+            _retryCount = 0;
```
lawrence-forooghian, line 647: When I was looking at the feature spec, it wasn't clear to me where it says that the retry counter should be reset in certain circumstances. ([RTB1a](https://sdk.ably.com/builds/ably/specification/main/features/#RTB1a) talks about the "nth retry" but the nth retry since what?) 

I was wondering what made you choose to reset it in the places that you have (both in the case of `ARTRealtime` and `ARTRealtimeChannel`? I think that the choice that you've made (reset it once a `CONNECTING` / `ATTACHING` <-> `SUSPENDED` cycle ends) makes sense intuitively, but was there something in the spec that led you to it?

Spec/Tests/RealtimeClientConnectionTests.swift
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/ce7c86ad0063dc4b8d5aeda3bc927abef58df11a

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2233,42 +2233,47 @@ class RealtimeClientConnectionTests: XCTestCase {
...
             client.close()
         }
-
-        var totalRetry = 0
+        
+        ARTUtils.seedRandom(withNumber: 5)
+        
+        var actualDelays = [TimeInterval]()
+        
+        var retryNumber = 1
```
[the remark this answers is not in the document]
maratal: Forgotten from previous iteration... https://github.com/ably/ably-cocoa/pull/1520/commits/ce7c86ad0063dc4b8d5aeda3bc927abef58df11a

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2233,42 +2233,47 @@ class RealtimeClientConnectionTests: XCTestCase {
             client.connection.off()
             client.close()
         }
-
-        var totalRetry = 0
+        
+        ARTUtils.seedRandom(withNumber: 5)
+        
+        var actualDelays = [TimeInterval]()
```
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/ce7c86ad0063dc4b8d5aeda3bc927abef58df11a

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2233,42 +2233,47 @@ class RealtimeClientConnectionTests: XCTestCase {
...
             client.connect()
 
             client.connection.on(.connecting) { stateChange in
                 expect(stateChange.previous).to(equal(ARTRealtimeConnectionState.disconnected))
-                totalRetry += 1
             }
         }
-
-        expect(totalRetry).to(equal(Int(expectedTime / options.disconnectedRetryTimeout)))
+        expect(actualDelays).to(beCloseTo(AblyTests.backoffWithJitterDelaysForTimeout(options.disconnectedRetryTimeout), within: 0.15))
```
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/ce7c86ad0063dc4b8d5aeda3bc927abef58df11a

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2233,42 +2233,47 @@ class RealtimeClientConnectionTests: XCTestCase {
...
+        var actualDelays = [TimeInterval]()
+        
+        var retryNumber = 1
+        
         waitUntil(timeout: testTimeout) { done in
-            let partialDone = AblyTests.splitDone(2, done: done)
-            var start: NSDate?
-
-            client.connection.once(.disconnected) { stateChange in
+            var disconnectedAt: Date!
```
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/ce7c86ad0063dc4b8d5aeda3bc927abef58df11a

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2283,95 +2288,52 @@ class RealtimeClientConnectionTests: XCTestCase {
...
         defer { ttlHookToken.remove() }
 
+        var retryCount = 0
+        let maxRetryCount = 8 // RTN14f: reasonable amount within testTimeout
+        
         waitUntil(timeout: testTimeout) { done in
-            client.connection.on(.suspended) { _ in
-                expect(client.connection.errorReason!.message).to(contain("timed out"))
+            var firstDisconnectedAt: Date!
+            var suspendedAt: Date!
```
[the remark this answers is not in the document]
maratal: "Last" sounds like there will be no more .suspended. More like "currentSuspendedAt", but I'm not sure, maybe leave it as-is? https://github.com/ably/ably-cocoa/pull/1520/commits/dc4a95579e0fe35e60591bc34fdb98c0c88c88fa

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2283,95 +2288,52 @@ class RealtimeClientConnectionTests: XCTestCase {
...
+            var connectingAt: Date!
```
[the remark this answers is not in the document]
maratal: Removed in https://github.com/ably/ably-cocoa/pull/1520/commits/dc4a95579e0fe35e60591bc34fdb98c0c88c88fa

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2283,95 +2288,52 @@ class RealtimeClientConnectionTests: XCTestCase {
...
 
-                let start = NSDate()
-                client.connection.once(.connecting) { _ in
-                    let end = NSDate()
-                    expect(end.timeIntervalSince(start as Date)).to(beCloseTo(options.suspendedRetryTimeout, within: 0.5))
+            client.connection.on(.disconnected) { stateChange in
+                expect(suspendedAt).to(beNil()) // https://github.com/ably/ably-cocoa/issues/913
```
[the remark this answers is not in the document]
maratal: Asserting it in .suspended https://github.com/ably/ably-cocoa/pull/1520/commits/dc4a95579e0fe35e60591bc34fdb98c0c88c88fa

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2283,95 +2288,52 @@ class RealtimeClientConnectionTests: XCTestCase {
...
-        let ttlHookToken = client.overrideConnectionStateTTL(3.0)
-        defer { ttlHookToken.remove() }
-
-        waitUntil(timeout: testTimeout) { done in
-            client.connection.once(.connected) { stateChange in
-                expect(stateChange.reason).to(beNil())
-                done()
+            client.connection.on(.connecting) { stateChange in
+                if stateChange.previous == .initialized {
+                    expect(firstDisconnectedAt).to(beNil())
```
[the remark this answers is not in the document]
maratal: Removed in https://github.com/ably/ably-cocoa/commit/dc4a95579e0fe35e60591bc34fdb98c0c88c88fa

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2283,95 +2288,52 @@ class RealtimeClientConnectionTests: XCTestCase {
...
+                    expect(suspendedAt).to(beNil())
+                    expect(connectingAt).to(beNil())
+                }
+                else if stateChange.previous == .suspended {
+                    connectingAt = Date()
+                    expect(connectingAt.timeIntervalSince(suspendedAt)).to(beCloseTo(options.suspendedRetryTimeout, within: 0.5))
```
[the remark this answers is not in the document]
maratal: Removed in https://github.com/ably/ably-cocoa/commit/dc4a95579e0fe35e60591bc34fdb98c0c88c88fa

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2283,95 +2288,52 @@ class RealtimeClientConnectionTests: XCTestCase {
...
+                }
+                else {
+                    expect(stateChange.previous).to(equal(ARTRealtimeConnectionState.disconnected))
```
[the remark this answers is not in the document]
maratal: Removed in https://github.com/ably/ably-cocoa/commit/dc4a95579e0fe35e60591bc34fdb98c0c88c88fa

Spec/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -2283,95 +2288,52 @@ class RealtimeClientConnectionTests: XCTestCase {
...
-                expect(stateChange.reason).to(beNil())
-                done()
+            client.connection.on(.connecting) { stateChange in
+                if stateChange.previous == .initialized {
+                    expect(firstDisconnectedAt).to(beNil())
+                    expect(suspendedAt).to(beNil())
+                    expect(connectingAt).to(beNil())
+                }
+                else if stateChange.previous == .suspended {
+                    connectingAt = Date()
```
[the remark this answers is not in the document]
maratal: Removed in https://github.com/ably/ably-cocoa/commit/dc4a95579e0fe35e60591bc34fdb98c0c88c88fa

Source/ARTRealtime.m
maratal: I took it from [this discussion](https://github.com/ably/ably-js/pull/1008#discussion_r925947944)

lawrence-forooghian: I think it would be good for this to be more properly specified instead of us relying on the implementation of another SDK.

Would you be able to open an issue in the specification repo for this, please? I don't think it needs to block us merging this PR, though.

Spec/Tests/RealtimeClientConnectionTests.swift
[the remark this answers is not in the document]
lawrence-forooghian: Do we not want to check that the client enters the `.connecting` state, though? That's the sign that it is indeed retrying the connection, which is the point of the test, I thought. I'm not sure we can just remove this whole `.on(.connecting)`.

Spec/Tests/RealtimeClientConnectionTests.swift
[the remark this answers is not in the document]
lawrence-forooghian: I don't see where that's happening. From what I'm seeing, the expectations are placed like this:

```swift
waitUntil(timeout: testTimeout) {
    client.connection.once(.disconnected) { stateChange in
        // some expectations
    }

    client.connection.on(.disconnected) { stateChange in
        // some expectations
    }

    client.connection.once(.suspended) { stateChange in
        // some expectations
    }

    client.connection.on(.suspended) { stateChange in
        // some expectations
    }

    client.connect()
}
```

What happens if the SDK doesn't call any of the `once` / `on` callbacks? Then the test will pass and none of the expectations will have been verified.

Spec/Tests/RealtimeClientConnectionTests.swift
lawrence-forooghian: If we're going to do it this way, can we have an assertion or something that checks that `maxSuspendedRetryCount` is  greater than, say, 2? We definitely want to be sure it's non-zero.

The alternative would be, instead of calculating `maxSuspendedRetryCount` based on { `testTimeout`, `connectionStateTtl`, `options.suspendedRetryTimeout` }, to choose the following:

- how long do we want the test to run for? (which then means that the test doesn't need to run for `testTimeout` (20 seconds) which is a long time)
- how many retries do we want to observe in order to get confidence that everything is working properly?

and then to calculate { `connectionStateTtl`, `options.suspendedRetryTimeout` } based on those. I think that this would make the test more understandable.

ikbalkaya (commented):
Some minor comments from me

Source/ARTRealtime.m
```diff
@@ -629,12 +632,10 @@ - (ARTEventListener *)transitionSideEffects:(ARTConnectionStateChange *)stateCha
                 _connectionLostAt = [NSDate date];
                 [self.logger verbose:@"RT:%p set connection lost time; expected suspension at %@ (ttl=%f)", self, [self suspensionTime], self.connectionStateTtl];
             }
-            NSTimeInterval retryInterval = self.options.disconnectedRetryTimeout;
-            // RTN15a - retry immediately if client was connected
-            if (stateChange.previous == ARTRealtimeConnected && _shouldImmediatelyReconnect) {
-                retryInterval = 0.1;
-            }
-            [stateChange setRetryIn:retryInterval];
+            BOOL immediately = stateChange.previous == ARTRealtimeConnected && _shouldImmediatelyReconnect; // RTN15a - retry immediately if client was connected
```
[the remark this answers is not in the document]
maratal: Yes, longer name would be better, but the next line will become so long that it's hard to format. Silly reason I know, but at least @lawrence-forooghian was convinced! 😝
```
NSTimeInterval retryDelay = immediately ? 0.1 : [ARTUtils retryDelayFromInitialRetryTimeout:self.options.disconnectedRetryTimeout
                                                                                         forRetryNumber:++_retryCount];
```
vs
```
NSTimeInterval retryDelay = shouldImmediatelyRetry ? 0.1 : [ARTUtils 
retryDelayFromInitialRetryTimeout:self.options.disconnectedRetryTimeout
                                                                                         forRetryNumber:++_retryCount];
```
on a 15' macbook.

Source/Private/ARTUtils.m
```diff
@@ -0,0 +1,54 @@
...
+    long r = [self randomFromZeroToNumber:100];
+    return r / 100.0;
+}
+
++ (long)randomFromZeroToNumber:(long)number {
+    if (_needsSeeding) {
+        [self seedRandomWithSystemTime];
+    }
+    return random() % number;
+}
```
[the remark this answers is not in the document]
maratal: `seedRandomWithNumber:` is needed for tests to force them to give a predictable (the same for each pass) value. Whereas `seedRandomWithSystemTime` is for the real scenario. `randomFraction` just utilizes functionality of the main method - `randomFromZeroToNumber:`. Does it make sense? @ikbalkaya

Source/ARTRealtime.m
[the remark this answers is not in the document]
QuintinWillison: > on a 15' macbook

I'm somewhat jealous that @maratal was able to obtain a laptop with a 15 foot screen... you must have to stand well back from that thing! 😆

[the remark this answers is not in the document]
maratal: Well... what options do I have? It's either 16' monster that will not fit into my backpack or 14' WITH A NOTCH! 😭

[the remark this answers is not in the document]
maratal: Well... what options do I have? It's either 16' monster that will not fit into my backpack or 14' WITH A NOTCH! 😭 @QuintinWillison

Spec/Tests/RealtimeClientConnectionTests.swift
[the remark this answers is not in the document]
maratal: I've made a combined test for both .disconnected and .suspended states https://github.com/ably/ably-cocoa/pull/1520/commits/564498baacfcf91dbc6ac805ad645ee788f81118

Spec/Tests/RealtimeClientConnectionTests.swift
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1520/commits/564498baacfcf91dbc6ac805ad645ee788f81118

Spec/Tests/RealtimeClientConnectionTests.swift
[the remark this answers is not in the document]
maratal: It checks if previous stated was `.connected` - https://github.com/ably/ably-cocoa/pull/1520/commits/564498baacfcf91dbc6ac805ad645ee788f81118

Spec/Tests/RealtimeClientConnectionTests.swift
maratal: I've chosen 3 times, I think that's enough 🧐 https://github.com/ably/ably-cocoa/pull/1520/commits/564498baacfcf91dbc6ac805ad645ee788f81118

Source/ARTRealtime.m
maratal: I've created an issue https://github.com/ably/specification/issues/127

Source/ARTRealtime.m
QuintinWillison, line 635:
```suggestion
            const BOOL immediately = stateChange.previous == ARTRealtimeConnected && _shouldImmediatelyReconnect; // RTN15a - retry immediately if client was connected
```

Source/ARTRealtime.m
```diff
@@ -629,12 +632,10 @@ - (ARTEventListener *)transitionSideEffects:(ARTConnectionStateChange *)stateCha
...
+            NSTimeInterval retryDelay = immediately ? 0.1 : [ARTUtils retryDelayFromInitialRetryTimeout:self.options.disconnectedRetryTimeout
```
QuintinWillison, line 636:
```suggestion
            const NSTimeInterval retryDelay = immediately ? 0.1 : [ARTUtils retryDelayFromInitialRetryTimeout:self.options.disconnectedRetryTimeout
```

Source/ARTRealtimeChannel.m
```diff
@@ -603,7 +610,9 @@ - (void)transition:(ARTRealtimeChannelState)state status:(ARTStatus *)status {
         case ARTRealtimeChannelSuspended:
             [_attachedEventEmitter emit:nil with:status.errorInfo];
             if (self.realtime.shouldSendEvents) {
-                channelRetryListener = [self unlessStateChangesBefore:self.realtime.options.channelRetryTimeout do:^{
+                NSTimeInterval retryDelay = [ARTUtils retryDelayFromInitialRetryTimeout:self.realtime.options.channelRetryTimeout
```
QuintinWillison, line 613:
```suggestion
                const NSTimeInterval retryDelay = [ARTUtils retryDelayFromInitialRetryTimeout:self.realtime.options.channelRetryTimeout
```

Source/Private/ARTUtils.m
```diff
@@ -0,0 +1,54 @@
...
+    srandom(number);
+    _needsSeeding = NO;
+}
+
++ (void)seedRandomWithSystemTime {
+    [self seedRandomWithNumber:(unsigned)time(NULL)];
+}
+
++ (double)randomFraction {
+    long r = [self randomFromZeroToNumber:100];
```
QuintinWillison, line 22: I'm not sure what value the intermediary local variable adds here.

```suggestion
    return [self randomFromZeroToNumber:100] / 100.0;
```

Source/Private/ARTUtils.m
```diff
@@ -0,0 +1,54 @@
...
+}
+
+#pragma mark - Incremental backoff and jitter (RTB1)
+
++ (double)jitterDelta {
+    return 0.2;
+}
+
++ (double)generateJitterCoefficient {
+    double coeff = 1.0 - self.randomFraction * self.jitterDelta;
```
QuintinWillison, line 40: No need for intermediary local variable (also, we know it's a coefficient from the name of the method we're implementing 🤷).

```suggestion
    return 1.0 - self.randomFraction * self.jitterDelta;
```

Source/Private/ARTUtils.m
```diff
@@ -0,0 +1,54 @@
...
+    return coeff;
+}
+
++ (double)backoffCoefficientForRetryNumber:(NSInteger)retryNumber {
+    double coeff = MIN((NSTimeInterval)(retryNumber + 2.0) / 3.0, 2.0);
```
QuintinWillison, line 45: No need for intermediary local variable (also, we know it's a coefficient from the name of the method we're implementing 🤷).

```suggestion
    return MIN((NSTimeInterval)(retryNumber + 2.0) / 3.0, 2.0);
```

QuintinWillison (approved):
Just some minor tweaks, so feel free to ignore them. Some of my suggested changes will require further changes (removal of subsequent `return` statements).

Source/ARTRealtime.m
lawrence-forooghian: I'm going to resolve Quintin’s `const` comments here — we aren't doing it in general in the rest of the codebase and haven't been vigilant about it in PRs in general.

Source/Private/ARTUtils.m
lawrence-forooghian: Done.

Source/Private/ARTUtils.m
lawrence-forooghian: Done.

Source/Private/ARTUtils.m
lawrence-forooghian: Done.

Source/ARTRealtime.m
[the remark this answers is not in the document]
lawrence-forooghian: I've changed it back to `shouldImmediatelyRetry` and removed the conditional operator to reduce the line length issue.

lawrence-forooghian:
@maratal I've been reworking this PR to make it more testable. I’m going split it into a sequence of smaller PRs. First one is #1589.

lawrence-forooghian:
I'm going to close this PR — I intend to land the remainder of this work in #1611, #1612, #1613.

PR #1524, opened 2022-11-09
Fix CI step that prepares built framework artifact

lawrence-forooghian:
Commit 938fe91 contained two different issues:

1. There was a typo `carthage-build-framework` instead of `carthage-built-framework`;
2. We actually required _two_ levels of directories to provide the desired effect of the GitHub artifact not containing a zip file at the top level.

This fixes those issues.

### DIFF
--- a/.github/workflows/check-pod.yaml
+++ b/.github/workflows/check-pod.yaml
@@ -49,10 +49,10 @@ jobs:
         # described in CONTRIBUTING.md.
       - name: Prepare built framework for archiving
         run: |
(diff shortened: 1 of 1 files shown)


maratal (approved):
Adding a hint where this file can be found ("on the Summary page of the workflow action run") could potentially save a minute or two for someone who is looking for it =)

lawrence-forooghian:
Thanks for the feedback @maratal, I'll be sure to do that next time 🙂

PR #1534, opened 2022-11-15
Removed deprecated `SecKeychainItemDelete`

### LINKED ISSUE #1521: Replace usage of `SecKeychainItemDelete` to fix deprecation error in Xcode 14
[Our usage of `SecKeychainItemDelete` ](https://github.com/ably/ably-cocoa/blob/d0cd1603716e02d50c534400fded2120da31387f/Source/ARTLocalDeviceStorage.m#L163) is causing the following deprecation error when building Ably-macOS using Xcode 14:

> 'SecKeychainItemDelete' is deprecated: first deprecated in macOS 10.10 - SecKeychain is deprecated

We need to find out what we should be using instead and start using it.

maratal:
Closes #1521 

We don't use `ARTLocalDeviceStorage` on macos, hence `SecKeychainItemDelete` can be safely removed.

The code itself was copied from `SAMKeychain` library. More info: https://github.com/soffes/SAMKeychain/blob/d3d64f8c5fa14bf2adb2d51a18ee8b6639232ca5/Sources/SAMKeychainQuery.m#L83

### DIFF
--- a/Source/ARTLocalDeviceStorage.m
+++ b/Source/ARTLocalDeviceStorage.m
@@ -152,23 +152,10 @@
 
 - (BOOL)keychainDeletePasswordForService:(NSString *)serviceName account:(NSString *)account error:(NSError *__autoreleasing *)error {
     NSMutableDictionary *query = [self newKeychainQueryForService:serviceName account:account];
-    OSStatus status;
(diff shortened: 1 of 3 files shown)


lawrence-forooghian (approved):
Looks like a good solution 👍

PR #1543, opened 2022-11-29
Run all actions on macos latest image.

### LINKED ISSUE #1542: macos-10.15 image for github action is deprecated
https://github.com/actions/runner-images/issues/5583

maratal:
Closes #1542

### DIFF
--- a/.github/workflows/integration-test-iOS16.yaml
+++ b/.github/workflows/integration-test-iOS16.yaml
@@ -1,4 +1,4 @@
-name: "Integration Test: iOS 16.1"
+name: "Integration Test: iOS 16"
 
 on:
   pull_request:
@@ -51,7 +51,7 @@ jobs:
           make submodules
           bundle install
           make update_carthage_dependencies_ios
-          bundle exec fastlane test_iOS16_1
+          bundle exec fastlane test_iOS16
 
       - name: Check Static Analyzer Output
         id: analyzer-output
@@ -67,7 +67,7 @@ jobs:
         if: ${{ failure() && steps.analyzer-output.outcome == 'failure' }}
         uses: actions/upload-artifact@v2
         with:
-          name: static-analyzer-reports-test_iOS16_1
+          name: static-analyzer-reports-test_iOS16
           path: ./derived_data/**/report-*.html
 
       # This is the script specified as the pod’s prepare_command in its Podspec.
@@ -82,7 +82,7 @@ jobs:
         run: |
           pod repo update
           pod install
-          bundle exec fastlane scan -s Tests --output-directory "fastlane/test_output/examples/test_iOS16_1"
+          bundle exec fastlane scan -s Tests --output-directory "fastlane/test_output/examples/test_iOS16"
 
       - name: Build APNS Example Project
         working-directory: ./Examples/AblyPush
--- a/.github/workflows/integration-test-iOS16.yaml
+++ b/.github/workflows/integration-test-iOS16.yaml
@@ -1,4 +1,4 @@
-name: "Integration Test: iOS 14.4"
+name: "Integration Test: iOS 16"
 
 on:
   pull_request:
@@ -12,7 +12,7 @@ on:
 
 jobs:
   check:
-    runs-on: macos-10.15
+    runs-on: macos-latest
 
     env:
       LC_CTYPE: en_US.UTF-8
@@ -51,7 +51,7 @@ jobs:
           make submodules
           bundle install
           make update_carthage_dependencies_ios
-          bundle exec fastlane test_iOS14_4
+          bundle exec fastlane test_iOS16
 
       - name: Check Static Analyzer Output
         id: analyzer-output
@@ -67,7 +67,7 @@ jobs:
         if: ${{ failure() && steps.analyzer-output.outcome == 'failure' }}
         uses: actions/upload-artifact@v2
         with:
-          name: static-analyzer-reports-test_iOS14_4
+          name: static-analyzer-reports-test_iOS16
           path: ./derived_data/**/report-*.html
 
       # This is the script specified as the pod’s prepare_command in its Podspec.
@@ -82,12 +82,12 @@ jobs:
         run: |
           pod repo update
           pod install
-          bundle exec fastlane scan -s Tests --output-directory "fastlane/test_output/examples/test_iOS14_4"
+          bundle exec fastlane scan -s Tests --output-directory "fastlane/test_output/examples/test_iOS16"
 
       - name: Build APNS Example Project
         working-directory: ./Examples/AblyPush
         run: |
-          xcodebuild build -scheme "AblyPushExample" -destination "platform=iOS Simulator,name=iPhone 12" -configuration "Debug"
+          xcodebuild build -scheme "AblyPushExample" -destination "platform=iOS Simulator,name=iPhone 13" -configuration "Debug"
       
       - name: Xcodebuild Logs Artifact
         if: always()
--- a/.github/workflows/integration-test-macOS.yaml
+++ b/.github/workflows/integration-test-macOS.yaml
@@ -1,4 +1,4 @@
-name: "Integration Test: macOS 10.15"
+name: "Integration Test: macOS Latest"
 
 on:
   pull_request:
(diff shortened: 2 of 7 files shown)


fastlane/Fastfile
```diff
@@ -2,24 +2,24 @@ default_platform(:ios)
 
 platform :ios do
 
-  lane :test_iOS14_4 do
+  lane :test_iOS16 do
     run_tests(
       scheme: "Ably-iOS-Tests",
       derived_data_path: "derived_data",
-      devices: ["iPhone 12 (14.4)"],
+      devices: ["iPhone 14 (16.0)"],
```
lawrence-forooghian, line 9: How come we're using iOS 16.0 and tvOS 16.1? Can't we use 16.1 for both?

.github/workflows/integration-test-iOS16.yaml
```diff
@@ -12,7 +12,7 @@ on:
 
 jobs:
   check:
-    runs-on: macos-10.15
+    runs-on: macos-latest
```
lawrence-forooghian, line 15: I know that @QuintinWillison [mentioned on Slack](https://ably-real-time.slack.com/archives/CSQEKCE81/p1669707060644419?thread_ts=1669666028.231469&cid=CSQEKCE81) that he'd prefer to use `latest` instead of locking a version. I'd be curious to understand the reason. Personally, I'd prefer that we remain in control of when our CI environment changes (but we _should_ be making these changes more frequently than we currently do, to keep up as new runner versions are released).

Also, as mentioned in another comment here, the macOS version chosen for the runners also affects the macOS version which we use for testing the macOS version of the SDK, which I think should be locked for consistency with how we choose iOS and tvOS versions.

.github/workflows/integration-test-tvOS16_1.yaml
```diff
@@ -1,4 +1,4 @@
-name: "Integration Test: tvOS 14.3"
+name: "Integration Test: tvOS 16.1"
```
lawrence-forooghian, line 1: We need to update the Continuous Integration Testing of the README with this new version.

fastlane/Fastfile
```diff
@@ -2,24 +2,24 @@ default_platform(:ios)
 
 platform :ios do
 
-  lane :test_iOS14_4 do
+  lane :test_iOS16 do
```
lawrence-forooghian, line 5: We need to update the Continuous Integration Testing of the README with this new version.

.github/workflows/integration-test-iOS16.yaml
QuintinWillison: My preference is just a personal opinion, in the absence of a better plan. In my experience we've generally seen less ongoing maintenance overhead and friction by sticking with _convention over configuration_, thus not explicitly overriding defaults unless we have a strong reason to. In theory, our builds should always be able to execute and succeed on the latest tooling, as well as more recent versions of the tooling older than that (macOS test target is a separate concern, albeit related due to the way Apple's tooling works)

If we are going to proceed with a policy of pinning down the version of macOS being used by our CI jobs then I would like to see documentation as to why we're selecting a specific version now, plus a strategy for ensuring we proactively upgrade it from time to time on an appropriate cadence presumably aligned with Apple (macOS) or GitHub (`macos-` runners) releases.

By opting for `macos-latest`, sticking with convention, the counter argument is that we're automatically gaining a "fail early" behaviour in CI when GitHub update the runners to newer versions of the OS. 🤔 

I think there's pros and cons of both approaches.

lawrence-forooghian: >  In theory, our builds should always be able to execute and succeed on the latest tooling

Definitely. I think my concern is that choosing `macos-latest` doesn't allow us to _choose how we prioritise_ the work needed to make sure we satisfy this requirement. If CI starts failing because of a runner version change, we have no choice but to immediately address that failure, even if there are more urgent things we would prefer to be working on. This is because CI failure blocks us from doing any other work.

> plus a strategy for ensuring we proactively upgrade it from time to time

That's very reasonable. I don't have a good idea for exactly what that would look like, other than documenting a process that expects developers to raise an issue when a new version of Xcode is released or something along those lines. (Unless we wanted to be fancier and have some sort of automation that checks available runner versions.)

> macOS test target is a separate concern

Do you have any thoughts on how we should address that one? Would you prefer that we move to running macOS tests on the latest version available?

fastlane/Fastfile
maratal: I just took value script was looking for. Will fix.

maratal: I just took value from "Using the first of multiple matching destinations:" warning. Will fix.

.github/workflows/integration-test-iOS16.yaml
QuintinWillison: Thanks, @lawrence-forooghian. I can't disagree with anything you've stated above, however right now I feel that we should leave this at `macos-latest` which will solve the issue being addressed by @maratal and allow this pull request to merge and mitigate the immediate term risk (that being the `macos-10.15` image disappearing from under us!). Longer term I've created https://github.com/ably/ably-cocoa/issues/1544 which is probably the most appropriate venue to continue this conversation on, plus of course the issue under which we can work to remove `macos-latest` from the mix.

lawrence-forooghian: Cool, thanks for creating that issue @QuintinWillison. In that case, @maratal, I think it's just a case of addressing [this comment](https://github.com/ably/ably-cocoa/pull/1543#discussion_r1034979446).

.github/workflows/integration-test-macOS.yaml
```diff
@@ -12,7 +12,7 @@ on:
 
 jobs:
   check:
-    runs-on: macos-10.15
```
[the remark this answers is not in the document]
lawrence-forooghian: [Sounds like](https://github.com/ably/ably-cocoa/pull/1543#discussion_r1035924973) we're going to use `macos-latest`, so I think we need to update it to say, instead of "macOS 10.15", "the version of macOS specified by [GitHub Actions’ `macos-latest` runner image](https://github.com/actions/runner-images#available-images)".

fastlane/Fastfile
maratal: I've tried - didn't work https://github.com/ably/ably-cocoa/pull/1543/commits/83d2f98565a6caf17e9f23000ae77702715cc5ed

fastlane/Fastfile
maratal: https://github.com/ably/ably-cocoa/pull/1543/commits/2812a5e55478bb324d280c9b9db49dfd09f6ba59

.github/workflows/integration-test-tvOS16_1.yaml
maratal: https://github.com/ably/ably-cocoa/pull/1543/commits/2812a5e55478bb324d280c9b9db49dfd09f6ba59

fastlane/Fastfile
lawrence-forooghian: What do you mean that it didn't work? I'm not sure how to use the linked commit to help me understand.

lawrence-forooghian: Oh, that's weird. [Looks like](https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md) Xcode 14.1 has iOS 16.1 simulators, but the default is Xcode 14.0.1. I guess since we're going with the "don't configure anything" approach, then we should leave it using Xcode 14.0.1 / iOS 16.0.

.github/workflows/integration-test-iOS16.yaml
```diff
@@ -1,4 +1,4 @@
-name: "Integration Test: iOS 14.4"
+name: "Integration Test: iOS 16"
```
lawrence-forooghian, line 1: Could we keep this named as iOS 16.0 for consistency with the other files (and the README)? And change the file name to `…iOS16_0.yaml`. It'll make it easier to bump the version number to 16.something in the future.

maratal: Good point. Here goes another round of cancelling workflows game 😆

maratal: https://github.com/ably/ably-cocoa/pull/1543/commits/3e62e0d3ef7a9ec95a479059a0a491f95fdbaf40

lawrence-forooghian (approved):

QuintinWillison (approved):

PR #1551, opened 2022-12-04
Fix/1550 extra callback with set suspended

### LINKED ISSUE #1550: Remove extra callback with `setSuspended`
>Could you open this one as a separate PR? I think that if it's a bug fix then we should give it a proper review separately.

_Originally posted by @lawrence-forooghian in https://github.com/ably/ably-cocoa/pull/1520#discussion_r1036570736_
      
>I think it's a bug calling `setSuspended` here, bc it's being called from `attachAfterChecks`. Also it leads to the channel emit `.suspended` twice in a raw, which I guess is not intended.

https://github.com/ably/ably-cocoa/commit/21b207f5f2926bf72ef7c961e8a0ea27feb46047

maratal:
Closes #1550 

If you revert [this commit](f58f731a8d545d4f7f55d6025554f07fad4a99bb), you'll get newly added test ([131b](https://github.com/ably/ably-cocoa/pull/1551/commits/8b017d691de96e71d072ec63304cffe68028d989)) failed every time.


Source/ARTRealtimeChannel.m
```diff
@@ -605,12 +605,7 @@ - (void)transition:(ARTRealtimeChannelState)state status:(ARTStatus *)status {
             if (self.realtime.shouldSendEvents) {
                 channelRetryListener = [self unlessStateChangesBefore:self.realtime.options.channelRetryTimeout do:^{
                     [self.realtime.logger debug:__FILE__ line:__LINE__ message:@"RT:%p C:%p (%@) reattach initiated by retry timeout", self->_realtime, self, self.name];
-                    [self reattachWithReason:nil callback:^(ARTErrorInfo *errorInfo) {
-                        if (errorInfo) {
-                            ARTStatus *status = [ARTStatus state:ARTStateError info:errorInfo];
-                            [self setSuspended:status];
-                        }
-                    }];
+                    [self reattachWithReason:nil callback:nil];
```
lawrence-forooghian, line 608: I think we can now remove the `callback` parameter from this method; it seems like nobody is using it any more.

Spec/Tests/RealtimeClientChannelTests.swift
```diff
@@ -3889,6 +3889,57 @@ class RealtimeClientChannelTests: XCTestCase {
...
+        channel.attach()
+        
+        expect(channel.state).to(equal(ARTRealtimeChannelState.attaching))
+
+        let detachedMessageWithError = AblyTests.newErrorProtocolMessage()
+        detachedMessageWithError.action = .detached
+        detachedMessageWithError.channel = channel.name
+        
+        var retryNumber = 1
+        let maxRetryCount = 5
```
lawrence-forooghian, line 3923: Could we call this `numberOfRetriesToWaitFor` and put a comment saying that it's an arbitrary number chosen to give us confidence that the retries are continuing indefinitely? When I first read it I thought that 5 had some special meaning.

Spec/Tests/RealtimeClientChannelTests.swift
```diff
@@ -3889,6 +3889,57 @@ class RealtimeClientChannelTests: XCTestCase {
...
+    func test__131b__Channel__if_the_channel_receives_a_server_initiated_DETACHED_message_and_if_the_attempt_to_reattach_fails_then_the_channel_will_transition_to_SUSPENDED_state_with_periodic_reattach_with_channelRetryTimeout() {
+        
+        let options = AblyTests.commonAppSetup()
+        options.channelRetryTimeout = 1.0
+        options.autoConnect = false
+        let client = ARTRealtime(options: options)
+        client.internal.setTransport(TestProxyTransport.self)
+        
+        let previousRealtimeRequestTimeout = ARTDefault.realtimeRequestTimeout()
+        ARTDefault.setRealtimeRequestTimeout(1.0)
```
lawrence-forooghian, line 3903: I _think_ that this plus the `transport.actionsIgnored += [.attached]` lines are intended to make sure that each attach attempt times out after one second, is that right? Perhaps a comment or two would make this clearer.

lawrence-forooghian (commented):
The fix looks good, thanks for spotting that! Just a few small comments.

Source/ARTRealtimeChannel.m
maratal: https://github.com/ably/ably-cocoa/pull/1551/commits/5ea06e1d219adae068b01fc2d467ffab4c6d983b

Spec/Tests/RealtimeClientChannelTests.swift
maratal: https://github.com/ably/ably-cocoa/pull/1551/commits/4fa70650b6f05265b682e767a027646daa48fdb3

lawrence-forooghian (approved):

PR #1552, opened 2022-12-05
Remove `pod lib lint` from release instructions

lawrence-forooghian:
It’s already run as part of the check-pod workflow.

I think I also added these Xcode version instructions because they were needed in order for the later `pod trunk push` step (which internally runs `pod lib lint`) to succeed. But now that step is working with Xcode 14 so we don’t need it.


CONTRIBUTING.md
```diff
@@ -63,8 +63,6 @@ For each release, the following needs to be done:
 * Steps to perform *before* pushing a release tag up:
```
maratal, line 63: I would unwind this to just two parental bullet points tbh. It made sense when it had four sub-items, but now it looks unnecessary imho.

maratal (approved):

CONTRIBUTING.md
lawrence-forooghian: Resolving this one on your behalf @maratal.

PR #1562, opened 2022-12-11
Updated example project with basic messaging functionality

maratal:
This change adds basic subscribe/publish buttons to the only UI example provided with the framework. It's better to have it there since it allows manual testing quickly.


maratal:
Closing since the example changed a lot.

PR #1565, opened 2022-12-18
Schedule `ARTOSReachability_Callback` on `_queue` instead of a run loop.

### LINKED ISSUE #1380: SCNetworkReachabilitySetCallback (Crashed: io.ably.main)

maratal:
Closes #1380 

Creation of the `ARTRealtime` object on the main thread were scheduling `ARTOSReachability_Callback` on the main run loop. While releasing was scheduled on the `_queue` (via `ARTQueuedDealloc`) which lead to a thread race and crash.

This bug was "unlocked" by this [PR](https://github.com/ably/ably-cocoa/pull/1453), because before it that crash was occurring earlier for another reason.


maratal:
This bug was "unlocked" by this [PR](https://github.com/ably/ably-cocoa/pull/1453), because before that the crash occurred earlier for another reason.

lawrence-forooghian:
Were you able to reproduce the crash @maratal? Do you have any steps I could follow to try reproducing?

maratal:
> Were you able to reproduce the crash @maratal? Do you have any steps I could follow to try reproducing?

From the crash report provided ([internal discussion](https://ably-real-time.slack.com/archives/CSQEKCE81/p1671033299051899?thread_ts=1671011739.412719&cid=CSQEKCE81)) crashed line is:

`ARTOSReachability *reachability = (__bridge ARTOSReachability *)info;`

Which is an attempt to access removed `ARTOSReachability` object on the "com.apple.main-thread" queue, which was deallocated on another queue ("io.ably.main"). I'm pretty confident that scheduling `ARTOSReachability_Callback` on the "io.ably.main" queue will solve this problem. @lawrence-forooghian

lawrence-forooghian:
Hey @maratal, I've taken a look at this but I don't really understand the problem that it's fixing. You've said that the `SCNetworkReachability` callback was scheduled for the main thread and that the `ARTOSReachability` instance was released on the main thread, but I don't think it's enough information for me to understand how that lead to a "thread race and crash". Would you be able to elaborate to explain the sequence of events that led to the crash and why it caused a crash?

maratal:
> Hey @maratal, I've taken a look at this but I don't really understand the problem that it's fixing. You've said that the `SCNetworkReachability` callback was scheduled for the main thread and that the `ARTOSReachability` instance was released on the main thread, but I don't think it's enough information for me to understand how that lead to a "thread race and crash". Would you be able to elaborate to explain the sequence of events that led to the crash and why it caused a crash?

To put less guesses into my reply [I've asked](https://github.com/ably/ably-cocoa/issues/593#issuecomment-1370972675) client directly about their setup. @lawrence-forooghian

lawrence-forooghian:
OK, but was there not at least one concrete scenario that you imagined when you wrote the fix?

maratal:
@lawrence-forooghian I did - if you call `close`, and then re-create `ARTRealtime` object than exactly this might happen - the object will be released on the `_queue`, but `SCNetworkReachability` callback will still be scheduled on `main` and have a good chance being called there after the object doesn't exist anymore.

maratal:
> OK, but was there not at least one concrete scenario that you imagined when you wrote the fix?

He replied with [this](https://github.com/ably/ably-cocoa/issues/593#issuecomment-1370989673). @lawrence-forooghian

lawrence-forooghian:
It seems to me like the underlying issue you describe is not one of threading but of memory management. If we're managing memory correctly, we should never be in a position where we try and access an object that has already been deallocated.

In general, [ARC](https://clang.llvm.org/docs/AutomaticReferenceCounting.html) takes care of memory management for us. But there are circumstances in which ARC needs hints from us about how to manage memory. One of these situations is when we store an Objective-C object in a C struct, which is what we're doing when we [store the `ARTOSReachability` instance in the `info` field of a `SCNetworkReachabilityContext` struct](https://github.com/ably/ably-cocoa/blob/eeebc41931b480972593b92626a813ae47bc0c5c/Source/Private/ARTOSReachability.m#LL44). We're using the `(__bridge T)` cast ([docs](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#bridged-casts)), which tells ARC not to perform any additional memory management on the Objective-C object. The net result of this is that the `ARTOSReachability` can end up being deallocated even whilst the `SCNetworkReachability` struct is still being used.

I think that to solve this potential crash, it suffices to ensure one of the following:

- that the `ARTOSReachability` instance stored as the `info` field in the `SCNetworkReachability` struct will be kept alive for as long as it's possible that the `SCNetworkReachability` callback might occur
- or, that if the `SCNetworkReachability` callback occurs after the `ARTOSReachability` instance has been deallocated, it will not cause a crash

Without looking at it in more detail I'm not sure exactly what the best solution would be or how to implement it, but the other `__bridge` casts mentioned on the other page might be an avenue, or perhaps storing a weak reference to the `ARTOSReachability` object (possibly via some other strongly-captured proxy object or Objective-C block).

maratal:
I think that using `__bridge_retained`/`__bridge_transfer` is not going to solve the issue, because there still will be unsynchronized access to this variable, as the docs [state](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#optimization):

>There is very little point in trying to guarantee correctness in the presence of race conditions. ARC does not have a stack-scanning garbage collector, and guaranteeing the atomicity of every load and store operation would be prohibitive and preclude a vast amount of optimization.

And [using `weak`](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#semantics) I believe is synchronized similar to `atomic` properties via thread locks:
>For __weak objects, the current pointee is retained and then released at the end of the current full-expression. This must execute atomically with respect to assignments and to the final release of the pointee.

So we are ending up with thread synchronization anyways, so why not to use dispatch then? WDYT @QuintinWillison  @lawrence-forooghian

lawrence-forooghian:
I unfortunately don't have the time to read that document in enough detail to really understand what those statements mean. However, I have always been under the impression that memory management in Objective-C is in general thread-safe. That is, the calls to `release` and `retain` (which, ultimately, is what ARC is generating) are thread-safe.

I think that if that weren't the case, all multi-threaded Objective-C programs would be littered with _lots_ of synchronisation code (it would massively limit the way you could pass objects between threads), in a way that I haven't seen them to be.

I have a feeling that the bit about race conditions that you've mentioned is probably talking about race conditions involving the variable that stores the pointer to an object, as opposed to the object itself.

I'm still of the opinion (even if one that I can't currently justify with reference to that document) that if we use the memory management tools provided to us by the language, we will not need to introduce any further synchronisation code. Do you have any thoughts on this, @QuintinWillison?

maratal:
> I unfortunately don't have the time to read that document in enough detail to really understand what those statements mean. However, I have always been under the impression that memory management in Objective-C is in general thread-safe. That is, the calls to `release` and `retain` (which, ultimately, is what ARC is generating) are thread-safe.
> 
> I think that if that weren't the case, all multi-threaded Objective-C programs would be littered with _lots_ of synchronisation code (it would massively limit the way you could pass objects between threads), in a way that I haven't seen them to be.
> 
> I have a feeling that the bit about race conditions that you've mentioned is probably talking about race conditions involving the variable that stores the pointer to an object, as opposed to the object itself.
> 
> I'm still of the opinion (even if one that I can't currently justify with reference to that document) that if we use the memory management tools provided to us by the language, we will not need to introduce any further synchronisation code. Do you have any thoughts on this, @QuintinWillison?

There is a great summary on thread safety in Apple's framework infrastructure - https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html
@lawrence-forooghian you are right re retain/release, I'll change this PR, but still I feel like doing these things on different threads is something that might be worth eliminating for the sake of clarity at least, especially if the tool for this already there (`SCNetworkReachabilitySetDispatchQueue`). If you'll find this reasonable let me know whether an issue for this should be backlogged.

Source/Private/ARTOSReachability.m
```diff
@@ -38,10 +32,24 @@ - (instancetype)initWithLogger:(ARTLog *)logger queue:(dispatch_queue_t)queue {
...
+                    callback(reachable);
+                }
+            });
+        }
+    };
+    
     _reachabilityRef = SCNetworkReachabilityCreateWithName(NULL, [host UTF8String]);
-    SCNetworkReachabilityContext context = {0, (__bridge void *)(self), NULL, NULL, NULL};
+    
+    SCNetworkReachabilityContext context = { .version = 0, .info = (void *)CFBridgingRetain(callbackBlock), .release = CFRelease };
```
lawrence-forooghian, line 52: I don't really understand how this memory management works. Do you know whether `SCNetworkReachabilityContext` expects to be passed an `info` that's already been retained? On a related note, under what circumstances will it call its `.release` if you don't also pass a `.retain` callback?

Here are the retains and releases performed during a run of `RealtimeClientTests` as this stands (I added logging [here](https://github.com/ably/ably-cocoa/commit/0cee3217279332e18f7f6d27fcf313aac359957b) and then piped the result through `egrep 'Retain|Release' | cut -f 4,5 -d ' '`):

```
0x600003a6d650 CFBridgingRetain
0x600003a74b40 CFBridgingRetain
0x600003a6e190 CFBridgingRetain
0x600003a6e190 myRelease
0x600003a6ae50 CFBridgingRetain
0x600003a6ed60 CFBridgingRetain
0x600003a6ae50 myRelease
0x600003a6ed60 myRelease
0x600003a7c5d0 CFBridgingRetain
0x600003a7c5d0 myRelease
0x600003a86280 CFBridgingRetain
0x600003a7cab0 CFBridgingRetain
0x600003a86280 myRelease
0x600003a7cab0 myRelease
0x600003a79a10 CFBridgingRetain
0x600003a79a10 myRelease
0x600003a6e610 CFBridgingRetain
0x600003a8d080 CFBridgingRetain
0x600003a8d080 myRelease
0x600003a6e610 myRelease
0x600003a799b0 CFBridgingRetain
0x600003a799b0 myRelease
0x600003a79a40 CFBridgingRetain
0x600003a6dbf0 CFBridgingRetain
0x600003a79a40 myRelease
0x600003a6dbf0 myRelease
0x600003a87390 CFBridgingRetain
0x600003a87390 myRelease
0x600003a79e30 CFBridgingRetain
0x600003a79e30 myRelease
0x600003a7b2d0 CFBridgingRetain
0x600003a7b2d0 myRelease
0x600003a7edc0 CFBridgingRetain
0x600003a7edc0 myRelease
0x600003a71c50 CFBridgingRetain
0x600003a71c50 myRelease
0x600003a9d200 CFBridgingRetain
0x600003a9d200 myRelease
0x600003a913b0 CFBridgingRetain
0x600003a913b0 myRelease
0x600003a7f690 CFBridgingRetain
0x600003a7f690 myRelease
0x600003a8a280 CFBridgingRetain
0x600003a8a280 myRelease
0x600003a639f0 CFBridgingRetain
0x600003a639f0 myRelease
0x600003aac540 CFBridgingRetain
0x600003aac540 myRelease
0x600003aaee80 CFBridgingRetain
0x600003aaee80 myRelease
0x600003aac990 CFBridgingRetain
0x600003a8f3f0 CFBridgingRetain
0x600003a8f3f0 myRelease
0x600003ab4900 CFBridgingRetain
0x600003ab4900 myRelease
0x600003ab0ba0 CFBridgingRetain
0x600003a5f960 CFBridgingRetain
0x600003a9dc80 CFBridgingRetain
0x600003a9dc80 myRelease
0x600003a7e7f0 CFBridgingRetain
0x600003ab2880 CFBridgingRetain
0x600003a833f0 CFBridgingRetain
0x600003a9d0e0 CFBridgingRetain
0x600003ab3e40 CFBridgingRetain
0x600003a9c7b0 CFBridgingRetain
0x600003ac40c0 CFBridgingRetain
0x600003a74270 CFBridgingRetain
0x600003ac03f0 CFBridgingRetain
0x600003ac0600 CFBridgingRetain
0x600003ac1950 CFBridgingRetain
```

You can see that whilst some of the calls are balanced, some of them are not. I wonder what's going on here? I also tried not retaining the `callbackBlock` up front and instead passing a `.retain` callback ([here](https://github.com/ably/ably-cocoa/commit/1e13c1d16131dcb0ba23933f5a51350b79abb45f)) and that gave this:

```
0x60000184e100 myRetain
0x60000184e100 myRetain
0x60000184e100 myRelease
0x600001831c80 myRetain
0x600001864f00 myRetain
0x600001864f00 myRetain
0x600001864f00 myRelease
0x600001864f00 myRelease
0x600001843090 myRetain
0x600001849710 myRetain
0x600001843090 myRetain
0x600001843090 myRelease
0x600001849710 myRetain
0x600001849710 myRelease
0x600001843090 myRelease
0x600001849710 myRelease
0x600001855890 myRetain
0x600001855890 myRelease
0x6000018bc1b0 myRetain
0x6000018bc1b0 myRetain
0x6000018bc1b0 myRelease
0x6000018b9950 myRetain
0x6000018bc1b0 myRelease
0x6000018b9950 myRetain
0x6000018b9950 myRelease
0x6000018b9950 myRelease
0x60000184f150 myRetain
0x60000184f150 myRetain
0x60000184f150 myRelease
0x60000184f150 myRelease
0x600001852f40 myRetain
0x600001852f40 myRetain
0x600001852f40 myRelease
0x6000018ba610 myRetain
0x6000018ba610 myRetain
0x6000018ba610 myRelease
0x6000018ba610 myRelease
0x600001852f40 myRelease
0x6000018bd200 myRetain
0x6000018bd200 myRetain
0x6000018bd200 myRelease
0x6000018bd200 myRetain
0x6000018bd200 myRelease
0x6000018bd200 myRelease
0x600001849710 myRetain
0x600001854f90 myRetain
0x600001849710 myRelease
0x600001854f90 myRelease
0x600001864cc0 myRetain
0x600001864cc0 myRelease
0x600001853630 myRetain
0x600001853630 myRetain
0x600001853630 myRelease
0x600001853630 myRelease
0x60000184fae0 myRetain
0x60000184fae0 myRetain
0x60000184fae0 myRelease
0x60000184fae0 myRelease
0x6000018bf330 myRetain
0x6000018bf330 myRetain
0x6000018bf330 myRelease
0x6000018bf330 myRelease
0x6000018b5a70 myRetain
0x6000018b5a70 myRetain
0x6000018b5a70 myRelease
0x6000018b5a70 myRelease
0x600001867b10 myRetain
0x600001867b10 myRetain
0x600001867b10 myRelease
0x600001867b10 myRelease
0x6000018a0240 myRetain
0x6000018a0240 myRetain
0x6000018a0240 myRelease
0x6000018a0240 myRelease
0x6000018a48d0 myRetain
0x6000018a48d0 myRetain
0x6000018a48d0 myRelease
0x6000018a48d0 myRelease
0x6000018ada40 myRetain
0x6000018ada40 myRetain
0x6000018ada40 myRelease
0x6000018ada40 myRelease
0x6000018aa550 myRetain
0x6000018aa550 myRetain
0x6000018aa550 myRelease
0x6000018aa550 myRelease
0x6000018a26d0 myRetain
0x6000018a26d0 myRetain
0x6000018a26d0 myRelease
0x6000018a26d0 myRelease
0x60000189e9a0 myRetain
0x60000189e9a0 myRetain
0x60000189e9a0 myRelease
0x60000189e9a0 myRelease
0x600001895080 myRetain
0x60000188a2b0 myRetain
0x60000188a2b0 myRetain
0x60000188a2b0 myRelease
0x60000188a2b0 myRelease
0x60000188de60 myRetain
0x60000188de60 myRetain
0x60000188de60 myRelease
0x60000188de60 myRelease
0x60000188e0a0 myRetain
0x60000188e0a0 myRetain
0x60000188e0a0 myRelease
0x600001857a50 myRetain
0x60000188a700 myRetain
0x60000188a700 myRetain
0x60000188a700 myRelease
0x60000188a700 myRelease
0x6000018ac6c0 myRetain
0x6000018b39f0 myRetain
0x6000018b39f0 myRetain
0x6000018b39f0 myRelease
0x6000018af870 myRetain
0x6000018a46c0 myRetain
0x6000018a46c0 myRetain
0x6000018a46c0 myRelease
0x6000018afc60 myRetain
0x60000188bc90 myRetain
0x60000188bc90 myRetain
0x60000188bc90 myRelease
0x6000018a42d0 myRetain
0x6000018a42d0 myRetain
0x6000018a42d0 myRelease
0x600001885110 myRetain
0x600001885110 myRetain
0x600001885110 myRelease
0x60000188dec0 myRetain
0x60000188dec0 myRetain
0x60000188dec0 myRelease
0x6000018f00f0 myRetain
0x6000018f00f0 myRetain
0x6000018f00f0 myRelease
0x6000018f4cc0 myRetain
0x6000018f4cc0 myRetain
0x6000018f4cc0 myRelease
```

This is just as confusing — some calls are balanced, some are not.

Do you have any thoughts on what's going on here?

lawrence-forooghian: Mike Ash normally knows what he's talking about, so let's go with it 😀 Can you add a code comment explaining where you got this strategy from? https://www.mikeash.com/pyblog/friday-qa-2013-06-14-reachability.html

maratal: https://github.com/ably/ably-cocoa/pull/1565/commits/036be28852b3ef1b3a112aea7052c463c5b8792a Some interesting comments about the behavior of Reachability there 🧐

lawrence-forooghian: I'm still uneasy about this one, actually. Given the potential consequences of getting memory management wrong (e.g. over-release leading to app crash) I want to be sure we're careful about this and don't just take somebody else's solution, even if we believe them trustworthy. Just by symmetry, if nothing else, it seems really weird that you'd tell `SCNetworkReachability` how to release an object but now how to retain it. It shouldn't be releasing anything unless one of the following is true:

- the release balances some corresponding action that it has performed itself (i.e. a retain);
- the release balances some corresponding action that was performed externally, about which a guarantee is made (e.g. that ownership has been in some way transferred)

Take a look at these examples of third-party libraries that wrap `SCNetworkReachability` (the first couple to come up when I Googled "swift reachability"):

- https://github.com/ashleymills/Reachability.swift/blob/a81b7367f2c46875f29577e03a60c39cdfad0c8d/Sources/Reachability.swift#L190-L208
- https://github.com/vsanthanam/NetworkReachability/blob/918943ba196327e1256a8f2b7e8b922651488607/Sources/NetworkReachability/API/ReachabilityMonitor/ReachabilityMonitor.swift#L529-L547

You can see they both pass `retain` and `release` callbacks.

Please could you do some more investigation to find out about how memory management is meant to work with `SCNetworkReachability`? I think Apple made a well-known example project once, and there might also be some documentation.

maratal: https://github.com/ably/ably-cocoa/pull/1565/commits/b72f778c5934b5a24243b52dd5fed409ce8f34a7

Arwin-17media:
Hi @lawrence-forooghian
We have the same problem cause by this issue, does this fix including in the v1.2.20?
Since I don't find #1565 in the bugs fix list in the v1.2.20, but it seems there are some issues fixed relate to #1565

maratal:
@Arwin-17media it's "SCNetworkReachabilitySetCallback (Crashed: io.ably.main) https://github.com/ably/ably-cocoa/issues/1380" line in changes. So, yes, this fix was included in v1.2.20.

PR #1567, opened 2022-12-18
Call to `connect` instead of `_connect` to avoid potential race conditions

### LINKED ISSUE #1566: Call to `_connect` in `ARTRealtimeInternal` constructor
I think that `_connect` shouldn't be called directly. It violates threading contract described [here](https://github.com/ably/ably-cocoa/commit/57f0e77f0388b8650b2ab2de2b6a3a0bda1d8591).

Also this call should be identical to the following code:

```
let options = AblyTests.clientOptions()
options.autoConnect = false
client.connect()
```

But calling `_connect` makes it not.
WDYT @lawrence-forooghian @QuintinWillison

maratal:
Closes #1566

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -210,7 +210,7 @@
         self.rest.prioritizedHost = nil;
         
         if (options.autoConnect) {
-            [self _connect];
+            [self connect];
         }
     }
     return self;


lawrence-forooghian (approved):
Good catch!

PR #1571, opened 2023-02-26
Fixed CI error due to changed iOS in the environment

### LINKED ISSUE #1570: CI for iOS 16.0 fails
with error "No device found with name 'iPhone 14 (16.0)'"

maratal:
Closes #1570

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -2,14 +2,14 @@ default_platform(:ios)
 
 platform :ios do
 
-  lane :test_iOS16_0 do
+  lane :test_iOS16_2 do
     run_tests(
       scheme: "Ably-iOS-Tests",
       derived_data_path: "derived_data",
-      devices: ["iPhone 14 (16.0)"],
+      devices: ["iPhone 14"],
       test_without_building: false,
       xcargs: { ABLY_ENV: ENV['ABLY_ENV'], CLANG_ANALYZER_OUTPUT: 'plist-html' },
-      output_directory: "fastlane/test_output/sdk/test_iOS16_0"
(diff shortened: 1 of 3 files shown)


fastlane/Fastfile
```diff
@@ -2,14 +2,14 @@ default_platform(:ios)
 
 platform :ios do
 
-  lane :test_iOS16_0 do
+  lane :test_iOS16_2 do
     run_tests(
       scheme: "Ably-iOS-Tests",
       derived_data_path: "derived_data",
-      devices: ["iPhone 14 (16.0)"],
+      devices: ["iPhone 14"],
```
lawrence-forooghian, line 9: Why haven't we replaced 16.0 by 16.2?

maratal: It gave the list what it has in the logs, now they named it just "iPhone 14"

lawrence-forooghian: But then we can't make any claims about what version of iOS we're running the tests on, making https://github.com/ably/ably-cocoa#continuous-integration-testing inaccurate.

We need to be sure about which version of the OS we're running on. If the name of the simulator doesn't tell us this any more, then we need to do something else, like creating the simulator ourselves or querying the existing simulators to find one with the right OS version.

maratal: Turned out it works with iPhone (16.2) https://github.com/ably/ably-cocoa/pull/1571/commits/5726d4a0ae2423f2d669d005cf7a1dca2441ad46

lawrence-forooghian: We need to also update the linked section in the README.

lawrence-forooghian (approved):

PR #1582, opened 2023-03-27
Replace header symlinks and further organise headers

lawrence-forooghian:
The aim of this PR is to reduce the number of steps needed in order to add a new Objective-C header file to the SDK, and to document the new process. See commit messages for more details.


maratal:
I wonder how it will affect installation using Cocoapods, Carthage and SPM, because I remember that there were some issues with all these private headers and symlinks when Lukasz were making those changes...

maratal (commented):
I wonder how it will affect installation using Cocoapods, Carthage and SPM, because I remember that there were some issues with all these private headers and symlinks when Lukasz were making those changes...

lawrence-forooghian:
@maratal I've tested it out using CocoaPods, SPM, and Carthage in standalone test projects, and I’m able to use Ably from Objective-C and Swift with no problems. I’m not sure why we ended up with all of these symlinks in the first place and unfortunately the commit history doesn't really explain it.

maratal (approved):

PR #1589, opened 2023-03-30
Add protocol for generating `RTB1b`’s jitter coefficient

lawrence-forooghian:
We’ll use this in the implementation of #1431.


maratal (approved):
Personally I would still prefer `random()` with predictable result by providing seeding number, but since this implementation is already here and works just fine I'm approving.

lawrence-forooghian:
@maratal that relies on global state, though — what if somebody else wants to call `random` as well? Then your sequence stops being predictable.

PR #1592, opened 2023-03-31
Add retry delay calculator

lawrence-forooghian:
This provides a calculator which determines the amount of time to wait before retrying an operation. We’ll use it in the implementation of #1431.


Source/ARTIncrementalBackoffAndJitterRetryDelayCalculator.m
```diff
@@ -0,0 +1,38 @@
+#import "ARTIncrementalBackoffAndJitterRetryDelayCalculator.h"
+#import "ARTJitterCoefficientGenerator.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ARTIncrementalBackoffAndJitterRetryDelayCalculator ()
```
maratal, line 6: For the sake of balance between "to be obvious" and the width of the screen to accommodate all these lengthy words I would just drop it to "ARTIncrementalRetryDelayCalculator"

lawrence-forooghian: Makes sense, will change it.

lawrence-forooghian: Changed it to `ARTBackoffRetryDelayCalculator` which I think best captures the essence of what this class does.

maratal (approved):

PR #1596, opened 2023-04-02
Refactor transitions

lawrence-forooghian:
This creates objects that encapsulate all of the metadata related to a request to perform an operation that may cause a connection or channel to emit a state change. See commit messages for more details.

This is a refactor, which will form part of our implementation of #1431.

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -39,6 +39,7 @@
 #import "ARTQueuedDealloc.h"
 #import "ARTErrorChecker.h"
 #import "ARTConnectionStateChangeMetadata.h"
+#import "ARTChannelStateChangeMetadata.h"
 
 @interface ARTConnectionStateChange ()
 
@@ -519,7 +520,7 @@
 }
 
 - (ARTEventListener *)transitionSideEffects:(ARTConnectionStateChange *)stateChange {
-    ARTStatus *status = nil;
+    ARTChannelStateChangeMetadata *channelStateChangeMetadata = nil;
     ARTEventListener *stateChangeEventListener = nil;
     
     [self.logger debug:@"RT:%p realtime is transitioning from %tu - %@ to %tu - %@", self, stateChange.previous, ARTRealtimeConnectionStateToStr(stateChange.previous), stateChange.current, ARTRealtimeConnectionStateToStr(stateChange.current)];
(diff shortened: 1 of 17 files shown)


Source/ARTRealtime.m
```diff
@@ -283,7 +286,7 @@ - (void)auth:(ARTAuthInternal *)auth didAuthorize:(ARTTokenDetails *)tokenDetail
         default: {
             // Client state is NOT Connecting or Connected, so it should start a new connection
             [self.logger debug:__FILE__ line:__LINE__ message:@"RS:%p new connection from successfull authorize %@", self.rest, tokenDetails];
-            [self transition:ARTRealtimeConnecting];
+            [self transition:ARTRealtimeConnecting withMetadata:[[ARTConnectionStateChangeMetadata alloc] init]];
```
maratal, line 289: Why not to call `transition:` overload here and further in case of an empty `ARTConnectionStateChangeMetadata`?

lawrence-forooghian: As I think I mentioned in one of the commit messages, I wanted to get rid of the overloads and have a single method.

I did consider allowing `nil` metadata, but decided it simpler in the end to just always require a metadata (especially since in the future it might be useful to give the metadata an ID that can be used for logging).

Source/PrivateHeaders/Ably/ARTAttachRequest.h
```diff
@@ -0,0 +1,34 @@
...
+
+@class ARTErrorInfo;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ A request to perform an operation that may ultimately call `ARTChannelRealtimeInternal`’s `internalAttach:callback:` method.
+ */
+NS_SWIFT_NAME(AttachRequest)
+@interface ARTAttachRequest: NSObject
```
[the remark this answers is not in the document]
lawrence-forooghian: I get the confusion but I _think_ it's OK — there is no URL request related to attaching a channel (you attach a channel by sending an `ATTACH` `ProtocolMessage` on the WebSocket transport). But if you feel strongly about it then I’m happy to rethink the name, what do you think?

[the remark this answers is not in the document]
maratal: I think that `AttachRequestMetadata` would do the job since we have its counterpart for channel state.

Source/ARTRealtime.m
maratal: I would consider to introduce a new static `empty` method then.

lawrence-forooghian: I'm curious why? Given it's a type all of whose properties are nullable, it seems pretty intuitive to me how a bare `init` method would behave.

maratal: It's more about style, I don't insist on that of course, `[ARTConnectionStateChangeMetadata empty]` looks more prettier then `[[ARTConnectionStateChangeMetadata alloc] init]]` imho.

Source/PrivateHeaders/Ably/ARTAttachRequest.h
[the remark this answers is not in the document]
lawrence-forooghian: 👍 done

Source/ARTRealtime.m
lawrence-forooghian: I think I'm going to keep it the way it is, if you don't mind?

maratal: Sure, it's cosmetics.

maratal (approved):

PR #1599, opened 2023-04-04
Emit retry attempt in state changes

lawrence-forooghian:
We’ll use this for testing the upcoming implementation of #1431.


maratal (approved):
LGTM

PR #1606, opened 2023-04-05
Remove compatibility macros

lawrence-forooghian:
Removes some macros that aren't serving any useful purpose. See commit messages for more details.


maratal (approved):
👍

PR #1611, opened 2023-04-05
Improve channel retry tests

lawrence-forooghian:
This is preparation for modifying `test__131b__Channel__if_the_channel_receives_a_server_initiated_DETACHED_message_and_if_the_attempt_to_reattach_fails_then_the_channel_will_transition_to_SUSPENDED_state_with_periodic_reattach_with_channelRetryTimeout` in #1431.

I didn't find the test very readable in its current state, and this is an attempt at improving that. I've also added assertions about the retry delays, which will become more important when we introduce the backoff and jitter logic.


maratal (approved):
LGTM

I like the idea of hiding particular test SDK under utilities classes. It would be very handy to have a standardized testing tool set and a requirement for specs team to emit a test(s) in pseudo code using this tool set accomodating spec changes. Then implementers would just take that pseudo code, tweak it for their particular SDK and viola - you have tests for a new spec.

lawrence-forooghian:
I know that @jaley was thinking about how we could implement a single test suite to be used by all of the SDKs — I believe he started writing a test suite in Lua which would then interface with each SDK. I don't know what the plan is for that work, though. But yes, I agree that having a single list of suggested test implementations for each spec point would be great.

PR #1615, opened 2023-04-05
Introduce new logging types

### LINKED ISSUE #1619: Create an `InternalLog` type that will evolve to implement the required internal logging API
#1617 will eventually define the API that the SDK will internally use for logging.

Let’s create an `InternalLog` type which, as the API becomes defined, will evolve to implement it. Initially its interface will match that of `ARTLog`, for compatibility with existing internal SDK code.

In #1622 we’ll update the internal SDK code to use this type.

### LINKED ISSUE #1620: Create a `Version2Log` protocol that will evolve to describe the SDK’s public logging API
#1628 will eventually define the API that the SDK will publicly use for logging.

Let’s create a `LogVersion2` protocol which, as the API becomes defined, will evolve to implement it. Initially its interface will be similar to that of `ARTLog`.

### LINKED ISSUE #1621: Create an adapter class to bridge between `ARTLog` and `ARTVersion2Log`
#1620 introduces the `Version2Log` protocol, which will become the public logging API in the next major version of the SDK. Until then, though, the SDK still needs to be able to accept a user-supplied `ARTLog` instance.

To allow `InternalLog` (added in #1619) to target `Version2Log`, let’s write an adapter class that conforms to the `Version2Log` protocol and which writes all of its output to a provided `ARTLog` instance.

lawrence-forooghian:
This introduces the types that will form the basis of our work for the upcoming [Logging Milestone](https://github.com/ably/ably-cocoa/milestone/6). See commit messages for more details.

Resolves #1619, resolves #1620, resolves #1621.

### DIFF
--- a/Test/Test Utilities/MockVersion2Log.swift
+++ b/Test/Test Utilities/MockVersion2Log.swift
@@ -1,13 +1,19 @@
 import Ably.Private
 
-class MockVersion2Log: Version2Log {
+@objc(ARTMockVersion2Log)
+class MockVersion2Log: NSObject, Version2Log {
     var logLevel: ARTLogLevel = .none
 
     var lastReceivedLogMessageArguments: (message: String, level: ARTLogLevel)?
-    var lastReceivedLogErrorArgument: ARTErrorInfo?
+    @objc var lastReceivedLogMessageArgumentMessage: String?
+    @objc var lastReceivedLogMessageArgumentLevel: ARTLogLevel = .none
+
+    @objc var lastReceivedLogErrorArgument: ARTErrorInfo?
 
     func log(_ message: String, with level: ARTLogLevel) {
         lastReceivedLogMessageArguments = (message: message, level: level)
+        lastReceivedLogMessageArgumentMessage = message
+        lastReceivedLogMessageArgumentLevel = level
     }
 
     func logWithError(_ error: ARTErrorInfo) {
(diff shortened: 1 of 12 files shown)


Test/Test Utilities/MockVersion2Log.swift
```diff
@@ -0,0 +1,21 @@
+import Ably.Private
+
+@objc(ARTMockVersion2Log) class MockVersion2Log: NSObject, Version2Log {
```
maratal, line 3: Not sure whether this was intentional, but usually `@objc` has a dedicated line before class declaration.

Test/Test Utilities/MockVersion2Log.swift
```diff
@@ -0,0 +1,21 @@
...
+
+    @objc var lastReceivedLogErrorArgument: ARTErrorInfo?
+
+    func log(_ message: String, with level: ARTLogLevel) {
+        lastReceivedLogMessageArguments = (message: message, level: level)
+        lastReceivedLogMessageArgumentMessage = message
+        lastReceivedLogMessageArgumentLevel = level
+    }
+
+     func logWithError(_ error: ARTErrorInfo) {
```
maratal, line 18: extra indentation space

maratal (approved):
Some minor comments

Test/Test Utilities/MockVersion2Log.swift
lawrence-forooghian: Didn't know that! Fixed it, thanks :)

Test/Test Utilities/MockVersion2Log.swift
lawrence-forooghian: Thanks, fixed it 👍

PR #1616, opened 2023-04-05
Use `ARTInternalLogHandler` internally for logging

lawrence-forooghian:


Source/ARTCrypto.m
```diff
@@ -10,6 +11,12 @@ - (BOOL)ccAlgorithm:(CCAlgorithm *)algorithm error:(NSError **)error;
 
 @end
 
+@interface ARTCrypto ()
+
+@property (nonatomic, strong) ARTInternalLog *logger;
```
maratal, line 16: I wonder why the class name ends with "Log" and var is always "logger". Maybe it's a perfect moment to settle this down 🙃

maratal (approved):
Minor suggestion

Source/ARTCrypto.m
lawrence-forooghian: Good point — I'm not going to change it here since we're changing from one class ending in `Log` to another ending in `Log`. But I’ve created https://github.com/ably/ably-cocoa/issues/1640 to make sure we address it. Thanks!

lawrence-forooghian: (We're also constrained by the fact that the property is called `logHandler` in the feature spec!)

PR #1638, opened 2023-04-10
Log line number and file

### LINKED ISSUE #1643: Emit file name and line number with every logged message
Split from #1643. Every log message should contain information about the code location from which it was emitted.

lawrence-forooghian:
This makes every emitted log message include the code location (i.e. file name and line number) of the corresponding log statement. See commit messages for more details.

Resolves #1643.

### DIFF
--- a/Source/ARTInternalLog.m
+++ b/Source/ARTInternalLog.m
@@ -29,10 +29,6 @@ NS_ASSUME_NONNULL_END
 
 // MARK: Logging
 
-- (void)log:(NSString *)message withLevel:(ARTLogLevel)level {
-    [self.core log:message withLevel:level];
-}
-
 - (void)log:(NSString *)message withLevel:(ARTLogLevel)level file:(const char *)fileName line:(NSInteger)line {
     [self.core log:message withLevel:level file:fileName line:line];
 }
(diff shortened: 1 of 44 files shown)


Source/ARTInternalLog.m
```diff
@@ -1,36 +1,46 @@
 #import "ARTInternalLog.h"
+#import "ARTInternalLogBackend.h"
 #import "ARTVersion2Log.h"
 
+NS_ASSUME_NONNULL_BEGIN
+
 @interface ARTInternalLog ()
 
-@property (nonatomic, readonly) id<ARTVersion2Log> logger;
+@property (nonatomic, readonly) id<ARTInternalLogBackend> backend;
```
maratal, line 9: The word "Backend" is mostly associated with a server side code, I wonder why it was chosen?

lawrence-forooghian: Fair point — I didn't have a good word and had seen "backend" being used a lot in https://github.com/apple/swift-log, so thought I’d steal that. But I’ve changed it to `ARTInternalLogCore`, what do you think?

maratal: `ARTInternalLogCore` is much better, thanks 👍

maratal (approved):

PR #1645, opened 2023-04-12
Remove unnecessary property attributes

lawrence-forooghian:
This removes unnecessary `strong` and `assign` property attributes. See commit messages for more details.


maratal (approved):
It's strange why they've decided to make `atomic` the default one when it's clearly should be explicitly asked.

lawrence-forooghian:
> It's strange why they've decided to make `atomic` the default one when it's clearly should be explicitly asked.

Yeah! That's why I decided not to touch those ones.

PR #1651, opened 2023-04-12
Pass logger into places where used

### LINKED ISSUE #1644: Pass a logger instance into all places from which we emit log messages
The aim of this issue is for every object to log to its own logger, instead of grabbing a logger from some other object it has access to (e.g. a channel grabbing its `ARTRealtime` instance’s logger). This will allow us to subsequently embed contextual information in the logger such as the identity of the object it belongs to.

lawrence-forooghian:
Resolves #1644. See commit messages for more details.


maratal (approved):
LGTM

PR #1654, opened 2023-04-13
 Populate `ARTCbcCipher.logger` property

lawrence-forooghian:
This property was added (with the class itself) in c377cae, but seems like we never actually populated it.


Source/ARTInternalLog.m
```diff
@@ -1,19 +1,25 @@
...
-@property (nonatomic, readonly) id<ARTInternalLogCore> core;
-
-@end
-
-NS_ASSUME_NONNULL_END
+#import "ARTLogAdapter.h"
 
 @implementation ARTInternalLog
 
++ (ARTInternalLog *)classMethodLogger {
```
maratal, line 9: Usually it's "shared" or "default", hence `sharedLogger` or `defaultLogger`.

maratal (commented):
Name changing request

Source/ARTInternalLog.m
lawrence-forooghian: Could it be `sharedLogger_onlyForUseInClassMethods`? I want it to be very obvious when we're misusing it (since the number of situations in which we're meant to use it is tiny).

maratal: Maybe to add a comment that states that it should be used only in class methods?

lawrence-forooghian: It is already documented, explaining the limited situations in which it's meant to be used.

I've renamed it to `sharedClassMethodLogger_readDocumentationBeforeUsing`. Given we use it in only two places, I think the long name is fine.

maratal (approved):

Issue #1604, opened 2023-04-04
Don't have global state in test suite


lawrence-forooghian:
TODO fill this one out

PR #1661, opened 2023-04-13
Remove shared `TestUtilities.userQueue`

lawrence-forooghian:
Part of #1604 (removing global mutable state in test suite).

The makes it less likely that work enqueued by one test will impact work enqueued by other tests. (I say "less likely" and not "impossible" because separate dispatch queues will impact each other if you have too many of them performing work at the same time.)

### DIFF
--- a/Test/Tests/PushChannelTests.swift
+++ b/Test/Tests/PushChannelTests.swift
@@ -14,12 +14,15 @@ class PushChannelTests: XCTestCase {
         return super.defaultTestSuite
     }
 
+    private var userQueue: DispatchQueue!
+
     override func setUp() {
         super.setUp()
 
         mockHttpExecutor = MockHTTPExecutor()
         let options = ARTClientOptions(key: "xxxx:xxxx")
-        options.dispatchQueue = AblyTests.userQueue
+        userQueue = AblyTests.createUserQueue()
+        options.dispatchQueue = userQueue
         options.internalDispatchQueue = AblyTests.queue
         rest = ARTRest(options: options)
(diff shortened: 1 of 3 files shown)


Test/Tests/PushChannelTests.swift
```diff
@@ -33,13 +36,13 @@ class PushChannelTests: XCTestCase {
...
     func test__001__Push_Channel__subscribeDevice__should_fail_if_the_LocalDevice_doesn_t_have_an_deviceIdentityToken() {
-        waitUntil(timeout: testTimeout) { done in
+        waitUntil(timeout: testTimeout) { [userQueue] done in
             rest.channels.get(uniqueChannelName()).push.subscribeDevice { error in
                 guard let error = error else {
                     fail("Error is nil"); done(); return
                 }
                 expect(error.message).to(contain("cannot use device before device activation has finished"))
-                XCTAssertTrue(AblyTests.currentQueueLabel() == AblyTests.userQueue.label)
+                XCTAssertTrue(AblyTests.currentQueueLabel() == userQueue!.label)
```
maratal, line 45: I guess it should work without "!" as well (here and below).

lawrence-forooghian: It doesn’t — for some reason, when you capture an implicitly unwrapped optional in a capture list (i.e. the `[userQueue]` at the start of the closure) the compiler then starts to see it as an optional instead (try removing the `!` and you'll get a compiler error). Weird, right?

maratal (approved):

PR #1682, opened 2023-04-18
Fix wait logic in `PushAdminTests.setUp`

lawrence-forooghian:
The current logic assumes that the `deviceRegistrations.save` calls will complete in the order that they are started (and hence, that when the final device in the list has been registered, all of the devices have been registered). This is not correct, and means that sometimes we end up trying to create a channel subscription for a device that has not been registered yet, causing the `channelSubscriptions.save` call to fail.

### DIFF
--- a/Test/Tests/PushAdminTests.swift
+++ b/Test/Tests/PushAdminTests.swift
@@ -110,24 +110,20 @@ class PushAdminTests: XCTestCase {
         rest.internal.storage = MockDeviceStorage()
         let group = DispatchGroup()
 
-        group.enter()
         for device in allDeviceDetails {
+            group.enter()
             rest.push.admin.deviceRegistrations.save(device) { error in
                 assert(error == nil, error?.message ?? "no message")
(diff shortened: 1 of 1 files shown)


Test/Tests/PushAdminTests.swift
```diff
@@ -110,24 +110,20 @@ class PushAdminTests: XCTestCase {
         rest.internal.storage = MockDeviceStorage()
         let group = DispatchGroup()
 
-        group.enter()
         for device in allDeviceDetails {
+            group.enter()
             rest.push.admin.deviceRegistrations.save(device) { error in
                 assert(error == nil, error?.message ?? "no message")
-                if allDeviceDetails.last == device {
-                    group.leave()
```
maratal, line 118: How it even worked if for `allDeviceDetails.count` enters you have only one leave?

maratal: Oh I see, enter was out of the loop

maratal (approved):
👍

PR #1684, opened 2023-04-19
Fix premature discarding of fallback host

### LINKED ISSUE #1683: Fallback host is being discarded before `fallbackRetryTimeout` elapses
Running `RestClientTests` on my development machine reveals multiple test failures which are caused by the fact that the SDK incorrectly believes that the timeout described by [`RSC15f`](https://sdk.ably.com/builds/ably/specification/main/features/#RSC15f) has elapsed.

The underlying cause is that [we're not initializing the value (`size`) pointed to by the `oldlenp` pointer passed to `sysctl`](https://github.com/ably/ably-cocoa/blob/b1a62413e08ab3702c246f69f2f9af3c519e72ca/Source/ARTTime.m#L29).

These tests are not failing in CI, which leads me to wonder if the symptoms only manifest on ARM devices (e.g. my development machine). We don't have an quick way of running our unit tests on actual iOS devices so I haven't been able to confirm if they're affected.

lawrence-forooghian:
Resolves #1683. See commit messages for more details.


l-szyszkowski (approved):
lgtm

PR #1690, opened 2023-04-20
Log observability server URL of created upload

lawrence-forooghian:
This is useful when you start on a GitHub action run page and want to find the corresponding upload in the observability server UI.


l-szyszkowski (approved):
LGTM

PR #1701, opened 2023-04-25
Error handling in test setup

lawrence-forooghian:
A few improvements to error handling in test setup. See commit messages for more details.

### DIFF
--- a/Test/Test Utilities/TestUtilities.swift
+++ b/Test/Test Utilities/TestUtilities.swift
@@ -72,6 +72,9 @@ func uniqueChannelName(prefix: String = "",
 
 /// Common test utilities.
 class AblyTests {
+    enum Error: Swift.Error {
+        case timedOut
+    }
 
     class func base64ToData(_ base64: String) -> Data {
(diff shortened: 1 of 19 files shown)


Test/Test Utilities/TestUtilities.swift
```diff
@@ -279,14 +279,21 @@ class AblyTests {
         }
     }
 
-    class func waitFor<T>(timeout: DispatchTimeInterval, file: FileString = #file, line: UInt = #line, f: @escaping (@escaping (T?) -> Void) -> Void) -> T? {
+    enum WaitError: Error {
```
maratal, line 278: I don't think this is a good name for enum, because there are no wait errors other than a timeout. Maybe to introduce a common for the whole test suit enum and collect all error cases there?

maratal (commented):
Minor request

Test/Test Utilities/TestUtilities.swift
lawrence-forooghian: Thanks @maratal — I've followed your suggestion and changed the type to `AblyTests.Error`.

maratal (approved):

PR #1703, opened 2023-04-30
Fix/1626 handle reachability events

### LINKED ISSUE #1626: Implement RTN20C: Handle os connectivity event while `CONNECTING`
See https://github.com/ably/specification/pull/135 and [RTN20c](https://sdk.ably.com/builds/ably/specification/main/features/#RTN20c)

> When `CONNECTING`, if the operating system indicates that the underlying internet connection is now available, the client should restart the pending connection attempt

As part of this work, we should also ensure that RTN20a and RTN20c are implemented correctly. Also, "Connection: Handle OS connectivity events" should be added to the `features.yml` file.

maratal:
This is based on https://github.com/ably/ably-cocoa/pull/1691
Closes #1626

See commit comments for more details.

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -180,12 +180,14 @@ NS_ASSUME_NONNULL_BEGIN
 NS_ASSUME_NONNULL_END
 
 const NSTimeInterval _immediateReconnectionDelay = 0.1;
+const NSTimeInterval _reachabilityReconnectionAttemptThreshold = 0.1;
 
 @implementation ARTRealtimeInternal {
     BOOL _resuming;
     BOOL _renewingToken;
     BOOL _shouldImmediatelyReconnect;
     ARTEventEmitter<ARTEvent *, ARTErrorInfo *> *_pingEventEmitter;
+    NSDate *_reachabilityActivatedAt;
     NSDate *_connectionLostAt;
     NSDate *_lastActivity;
     Class _reachabilityClass;
@@ -540,6 +542,16 @@ const NSTimeInterval _immediateReconnectionDelay = 0.1;
 - (void)handleNetworkIsReachable:(BOOL)reachable {
     if (reachable) {
         switch (_connection.state_nosync) {
+            case ARTRealtimeConnecting: {
+                // There is no certain way to detect whether reachability callback was called because you've just set it up,
+                // or because it really has detected reachability of the host. So we rely on a short timeout here - if the callback was called
+                // shortly after the setting up reachability, then reconnection shouldn't be done. Whereas if it was called after the certain delay (a threshold) - than it means you are in the middle of the dead-end connection attemt and it should be restarted.
+                // See https://github.com/ably/ably-cocoa/issues/1713 for more details.
+                if ([self shouldRestartPendingConnectionAttempt]) {
+                    [self transportReconnectWithExistingParameters];
+                }
+                break;
+            }
             case ARTRealtimeDisconnected:
             case ARTRealtimeSuspended:
                 [self transition:ARTRealtimeConnecting withMetadata:[[ARTConnectionStateChangeMetadata alloc] init]];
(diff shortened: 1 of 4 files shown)


Source/PrivateHeaders/Ably/ARTTestClientOptions.h
```diff
@@ -31,6 +31,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic) id<ARTRealtimeTransportFactory> transportFactory;
 
+/**
+ Initial value is `nil`.
```
lawrence-forooghian, line 35: Can you please document what this property does and what it's for? It sounds like it's used to provide a way for test code to be able to simulate the case where a reconnection attempt results in a different outcome to the original connection attempt, is that right?

Source/ARTRealtime.m
```diff
@@ -540,6 +542,12 @@ - (void)updateWithErrorInfo:(nullable ARTErrorInfo *)errorInfo {
 - (void)handleNetworkIsReachable:(BOOL)reachable {
     if (reachable) {
         switch (_connection.state_nosync) {
+            case ARTRealtimeConnecting: {
+                if ([self shouldRestartPendingConnectionAttempt]) {
```
lawrence-forooghian, line 546: The behaviour that you describe in the commit message should be described here too, I think — as you mentioned in Slack, it's a workaround and it should be brought to people's attention since it's not a behaviour described by the spec. Also I'd suggest creating an issue that describes the underlying reason for this workaround and what we need to do to remove it, and link to it here.

Source/ARTRealtime.m
```diff
@@ -180,12 +180,14 @@ @interface ARTRealtimeInternal ()
...
 
 const NSTimeInterval _immediateReconnectionDelay = 0.1;
+const NSTimeInterval _reachabilityReconnectionAttemptThreshold = 0.1;
 
 @implementation ARTRealtimeInternal {
     BOOL _resuming;
     BOOL _renewingToken;
     BOOL _shouldImmediatelyReconnect;
     ARTEventEmitter<ARTEvent *, ARTErrorInfo *> *_pingEventEmitter;
+    NSDate *_connectStartedAt;
```
lawrence-forooghian, line 190: From what you're saying, we need a piece of state that we can use upon receiving a notification that the host is reachable, to determine how long it's been since we told the reachability class to start emitting reachability notifications. This concept doesn't seem to be directly related to connection state or the state transition, so I'm wondering if this name is the right one and whether you're setting up and tearing it down in the best place?

lawrence-forooghian (commented):
I've put a few comments on the code. Also, I found the commits and their messages a bit confusing — 40303c0c30bfd82ab84281db07b969de685b30de’s message doesn't make it clear that it's implementing a new feature, it sounds like a bug fix or a refactor or something. And wouldn't it then make sense to put the test in the same commit that implements the feature?

Source/ARTRealtime.m
maratal: https://github.com/ably/ably-cocoa/pull/1703/commits/08ba4c8fded2d94ec7efcecba95ccf979b9c80ca

Source/ARTRealtime.m
maratal: https://github.com/ably/ably-cocoa/pull/1703/commits/08ba4c8fded2d94ec7efcecba95ccf979b9c80ca

Issue https://github.com/ably/ably-cocoa/issues/1713

Source/PrivateHeaders/Ably/ARTTestClientOptions.h
maratal: https://github.com/ably/ably-cocoa/pull/1703/commits/08ba4c8fded2d94ec7efcecba95ccf979b9c80ca

lawrence-forooghian (approved):

PR #1706, opened 2023-05-02
Create a unique identifier for each test method execution

lawrence-forooghian:


Test/Test Utilities/TestUtilities.swift
```diff
@@ -114,8 +114,8 @@ class AblyTests {
         return queue
     }()
 
-    static func createUserQueue() -> DispatchQueue {
-        let queue = DispatchQueue(label: "io.ably.tests.callbacks.\(UUID().uuidString)", qos: .userInitiated)
+    static func createUserQueue(for test: Test) -> DispatchQueue {
+        let queue = DispatchQueue(label: "io.ably.tests.callbacks.\(test.id).\(UUID().uuidString)", qos: .userInitiated)
```
maratal, line 118: Wouldn't additional `uuidString` here be a bit excessive?

maratal (commented):
Not a blocker, but a couple of suggestions:
- Would be nice to have a single `TestEnvironment` structure for the whole test suit, currently one has a `rest` property, and the other one a `client` property, both of which are of type `ARTRest` which has little sense. 
- Move `commonAppSetup` method into this new  `TestEnvironment` structure.
- Move `uniqueChannelName` into the `Test` structure.

lawrence-forooghian:
Thanks for the suggestions @maratal.

> Would be nice to have a single `TestEnvironment` structure for the whole test suit, currently one has a `rest` property, and the other one a `client` property, both of which are of type `ARTRest` which has little sense. 

The `TestEnvironment` structures are unique to each test case since they have different setup logic.

> Move `uniqueChannelName` into the `Test` structure.

Done 👍

PR #1747, opened 2023-05-12
Fix tests broken by `cb7af9d`

### LINKED ISSUE #1746: Commit `cb7af9d` broke two tests locally for me
These two tests are failing locally for me:

- `test__039__Token__options__if_an_attempt_by_the_realtime_client_library_to_authenticate_is_made_using_the_authUrl_or_authCallback__the_attempt_times_out_after_realtimeRequestTimeout__if_the_connection_is_CONNECTED__then_the_connection_should_remain_CONNECTED`
- `test__011__Connection__should_disconnect_the_transport_when_no_activity_exist`

`git bisect` shows that cb7af9db2c74f05020232c173e14b4624b12fed5 broke them.

Another one we should fix before doing a release.

lawrence-forooghian:
Resolves #1746. See commit messages for more details.

### DIFF
--- a/Test/Tests/AuthTests.swift
+++ b/Test/Tests/AuthTests.swift
@@ -832,7 +832,9 @@ class AuthTests: XCTestCase {
         options.authCallback = { _, completion in
             getTestTokenDetails(for: test, completion: completion)
         }
-        options.testOptions.realtimeRequestTimeout = 0.5
+
+        // This needs to be sufficiently long such that we can expect to receive a CONNECTED ProtocolMessage within this duration after starting a connection attempt (there's no "correct" value since it depends on network conditions, but 1.5s seemed to work locally and in CI at time of writing). But we also don't want it to be longer than necessary since that would impact test execution time.
+        options.testOptions.realtimeRequestTimeout = 1.5
 
         let realtime = ARTRealtime(options: options)
         defer { realtime.dispose(); realtime.close() }
--- a/Test/Tests/RealtimeClientConnectionTests.swift
+++ b/Test/Tests/RealtimeClientConnectionTests.swift
@@ -4407,7 +4407,8 @@ class RealtimeClientConnectionTests: XCTestCase {
     func test__011__Connection__should_disconnect_the_transport_when_no_activity_exist() throws {
         let test = Test()
         let options = try AblyTests.commonAppSetup(for: test)
-        let realtimeRequestTimeout = 0.5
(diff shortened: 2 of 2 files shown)


maratal:
>we were only changing the global realtimeRequestTimeout value to 0.5 _after_ instantiating the ARTRealtime object, which meant that the client would use the default timeout of 10s

So it means, that before that commit, setting realtimeRequestTimeout to 0.5 was actually doing nothing, right?

lawrence-forooghian:
> > we were only changing the global realtimeRequestTimeout value to 0.5 _after_ instantiating the ARTRealtime object, which meant that the client would use the default timeout of 10s
> 
> So it means, that before that commit, setting realtimeRequestTimeout to 0.5 was actually doing nothing, right?

You might be right there, actually — will check and if so I'll update the commit message, thanks!

lawrence-forooghian:
> > > we were only changing the global realtimeRequestTimeout value to 0.5 _after_ instantiating the ARTRealtime object, which meant that the client would use the default timeout of 10s
> > 
> > 
> > So it means, that before that commit, setting realtimeRequestTimeout to 0.5 was actually doing nothing, right?

OK, I looked into it and the value of 0.5 _was_ being used after the `ARTRealtime`’s instantiation:

- in `test__011__Connection__should_disconnect_the_transport_when_no_activity_exist` to set the connection idle timer after receiving the `CONNECTED` `ProtocolMessage`
- in `test__039__Token__options__if_an_attempt_by_the_realtime_client_library_to_authenticate_is_made_using_the_authUrl_or_authCallback__the_attempt_times_out_after_realtimeRequestTimeout__if_the_connection_is_CONNECTED__then_the_connection_should_remain_CONNECTED` to set the connection idle timer after seeing activity on the connection, and also something to do with setting a timeout when reconnecting after an auth failure

— so, long story short, the 0.5s timeout was being used before. Thanks for asking, though, I hadn't considered it 👍

maratal (approved):
> the 0.5s timeout was being used before

I guess changing that to 1.5 in those places doesn't affect tests much, does it? Discard this approval if I'm wrong.

lawrence-forooghian:
> > the 0.5s timeout was being used before
> 
> I guess changing that to 1.5 in those places doesn't affect tests much, does it? Discard this approval if I'm wrong.

I believe it does not affect them — I think that we just want a small number so the tests don't last long. The tests where the exact timeout value matters refer to it using a variable, so if we adjust it then they still work, with the exception of the one assertion that I had to change here.

Issue #1559, opened 2022-12-08
App Store publication issues when using Carthage


bhargavbajani:
There're a few issues with publishing to App Store if you add Ably through.

The app builds and runs but when you're making an archive and uploading it to App Store it won't be validated. Errors while validating: `CFBundleIdentifier Collision. There is more than one bundle with the CFBundleIdentifier value 'com.remyvirin.msgpack-iOS' under the iOS application 'XXXXXX.app'. With error code STATE_ERROR.VALIDATION_ERROR.90685 for id 042b0740-7b7f-4012-a121-ee888e479aa9`

this is a similar issue https://github.com/ably/ably-cocoa/issues/698

The framework contains a duplicate framework 

![](https://user-images.githubusercontent.com/13344552/205957347-96a15ff9-9654-43fc-bc99-7dfd4cff9d84.png)

### Which version of the Ably SDK are you using? 
>= 1.2.26 (Not sure about previous versions)

### On which platform does the issue happen? 
iOS

### Are you using Carthage? 
0.38.0

### Which version of Xcode are you using? 
XCode 13.4.1

iOSDev-Auction:
Any update?

maratal:
Thanks @bhargavbajani, we are looking into it.

PR #1754, opened 2023-06-05
Fix linker error upon following `README`’s Carthage instructions

### LINKED ISSUE #1183: App crash and dyld error message (AblyDeltaCodec cannot be loaded, image not found) when using XCFrameworks on macOS app.
A customer has reported that they’re getting a dyld error (AblyDeltaCodec cannot be loaded, image not found.) when attempting to use XCFrameworks instead of Carthage, whilst attempting to follow the instructions in

https://github.com/ably/ably-cocoa to build their macOS app.

The application runs fine from Xcode but upon installing the app on the test machine, the app crashes with dyld error message because AblyDeltaCodec cannot be loaded, image not found.

Steps followed prior to the error & crash were:

1-Changed Carthage update to include --use-xcframeworks
2-Removed the old framework references from the project
3-Dragged the three XCFrameworks from the Carthage build folder into the project
4-Confirmed that all three are in the Frameworks, Libraries and Embedded Content list on the General tab and that they have Do Not Embed selected
5-Confirmed all three are in the Link Binary with Libraries list in Build Phases

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-1256) by [Unito](https://www.unito.io)

lawrence-forooghian:
Resolves #1183. See commit messages for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,12 +95,9 @@ to build the framework and drag the built (in `[PROJECT_ROOT]/Carthage/Build`)
 - `AblyDeltaCodec.xcframework`
 - `msgpack.xcframework`
 
-into the "Frameworks, Libraries, and Embedded Content" section of the General tab of your Xcode target's settings.
+into the "Frameworks, Libraries, and Embedded Content" section of the General tab of your Xcode target's settings. If your target is an application, select "Embed & Sign", otherwise "Do Not Embed".
 
 If you see, for example, a `dyld: Library not loaded: @rpath/AblyDeltaCodec.framework/AblyDeltaCodec` error, then most likely you forgot to add all the dependencies to your project. You have more detailed information [here](https://github.com/Carthage/Carthage#adding-frameworks-to-an-application).
-
-*NOTE:*
-*For **macOS** target you have to select `Do Not Embed` for `Ably.xcframework` in `General` tab for your target, and make sure the `Ably.xcframework` is on the `Link Binary With Libraries` list in `Build Phases` tab.*
 
 ### Manual installation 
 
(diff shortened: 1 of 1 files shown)


maratal:
>These instructions were added in https://github.com/ably/ably-cocoa/commit/9a1e84cb10aace173bd43348b23ce75fdecf2f98, as part of the PR for adding
Swift Package Manager support

I bet it should be another commit hash here.

>This is the issue described by https://github.com/ably/ably-cocoa/pull/1599, which was resolved in https://github.com/ably/ably-cocoa/commit/9a1e84cb10aace173bd43348b23ce75fdecf2f98
by no longer embedding the dependency frameworks inside Ably.framework.

Not sure if "Emit retry attempt in state changes" has anything related to embedding frameworks.

lawrence-forooghian:
Sorry @maratal — the commit hash is right but the linked issue is wrong, should be #1594. I've updated the commit message.

maratal (approved):
>These instructions were added in https://github.com/ably/ably-cocoa/commit/9a1e84cb10aace173bd43348b23ce75fdecf2f98, as part of the PR for adding
Swift Package Manager support

I don't agree (if I understood right), because https://github.com/ably/ably-cocoa/pull/1594 was only addressing https://github.com/ably/ably-cocoa/issues/1559. Which in its turn was an outcome of the PR for adding Swift Package Manager support. Since it's a minor commentary issue I'm approving nevertheless.

lawrence-forooghian:
Ah yes, sorry! The first commit was meant to be 9691ebc. I've updated. Does that look right now?

lawrence-forooghian:
@maratal pointed out that the commit message is still wrong — the mention of 9a1e84c should actually be 9691ebc. Unfortunately this is now merged so there's not loads that I can do, but leaving a message here in case somebody ends up on this PR in the future.

PR #1786, opened 2023-07-26
Release/1.2.21

maratal:
This should go as a `alpha`, `beta` or `rc` since location pushes feature is implemented as a preview. I would choose a `rc`, but I'd rather discuss this @lawrence-forooghian @paddybyers 

According to [this](https://github.com/ably/engineering/blob/main/best-practices/product-lifecycle.md) the full version tag should be `1.2.21-rc.1`


lawrence-forooghian (commented):
Your linked documentation mentions that a Release Candidate "will not be changed in terms of API or its behavior". Given that we're doing a pre-release to verify that the location push API we've built is adequate for the customer's needs, there is a chance that we will need to change the API and hence it seems a Release Candidate is not appropriate. I think a Beta sounds right.

Also, your PR as implemented will create a normal release e.g. on CocoaPods. I think that all of the references to `1.2.21` in this PR should probably in fact be `1.2.21-beta`. I don't think that our current release process accounts for this so you'll probably need to do it manually. Same for the name of this PR and its branch, too.

maratal:
> Also, your PR as implemented will create a normal release e.g. on CocoaPods. I think that all of the references to `1.2.21` in this PR should probably in fact be `1.2.21-beta`. I don't think that our current release process accounts for this so you'll probably need to do it manually. Same for the name of this PR and its branch, too.

So we are not going to use `<build>` after `beta`?

lawrence-forooghian:
> > Also, your PR as implemented will create a normal release e.g. on CocoaPods. I think that all of the references to `1.2.21` in this PR should probably in fact be `1.2.21-beta`. I don't think that our current release process accounts for this so you'll probably need to do it manually. Same for the name of this PR and its branch, too.
> 
> So we are not going to use `<build>` after `beta`?

Ah, yes, sorry — we should, so should be `1.2.21-beta.1`, is that right?

maratal:
Closing this in favor of https://github.com/ably/ably-cocoa/pull/1787 (branch name has changed)

PR #1791, opened 2023-08-02
[SDK-3769] Respect rtl12

### LINKED ISSUE #1790: Library doesn't respect RTL12
[(RTL12)](https://sdk.ably.com/builds/ably/specification/main/features/#RTL12) An attached channel may receive an additional ATTACHED ProtocolMessage from Ably at any point. (This is typically triggered following a transport being resumed to indicate a partial loss of message continuity on that channel, in which case the ProtocolMessage will have a resumed flag set to false). **If and only if the resumed flag is false, this should result in the channel emitting an UPDATE event** with a ChannelStateChange object. The ChannelStateChange object should have both previous and current attributes set to attached, the reason attribute set to to the error member of the ATTACHED ProtocolMessage (if any), and the resumed attribute set per the RESUMED bitflag of the ATTACHED ProtocolMessage. (Note that UPDATE should be the only event emitted: in particular, the library must not emit an ATTACHED event if the channel was already attached, see RTL2g).

But library [emits](https://github.com/ably/ably-cocoa/blob/273277e566fc1641cc8fc5b6473999a3e9ab6c4e/Source/ARTRealtimeChannel.m#L748) an UPDATE event regardless said condition.

maratal:
Closes #1790

### DIFF
--- a/Source/ARTRealtimeChannel.m
+++ b/Source/ARTRealtimeChannel.m
@@ -741,11 +741,13 @@ dispatch_sync(_queue, ^{
     }
 
     if (self.state_nosync == ARTRealtimeChannelAttached) {
-        if (message.error != nil) {
-            _errorReason = message.error;
+        if (!message.resumed) { // RTL12
+            if (message.error != nil) {
+                _errorReason = message.error;
+            }
+            ARTChannelStateChange *stateChange = [[ARTChannelStateChange alloc] initWithCurrent:self.state_nosync previous:self.state_nosync event:ARTChannelEventUpdate reason:message.error resumed:message.resumed];
+            [self emit:stateChange.event with:stateChange];
         }
-        ARTChannelStateChange *stateChange = [[ARTChannelStateChange alloc] initWithCurrent:self.state_nosync previous:self.state_nosync event:ARTChannelEventUpdate reason:message.error resumed:message.resumed];
-        [self emit:stateChange.event with:stateChange];
         return;
     }
 
--- a/Test/Tests/RealtimeClientChannelTests.swift
+++ b/Test/Tests/RealtimeClientChannelTests.swift
@@ -501,6 +501,7 @@ class RealtimeClientChannelTests: XCTestCase {
     func test__011__Channel__EventEmitter__channel_states_and_events__ChannelStateChange_will_contain_a_resumed_boolean_attribute_with_value__true__if_the_bit_flag_RESUMED_was_included() throws {
         let test = Test()
         let options = try AblyTests.commonAppSetup(for: test)
+        options.testOptions.transportFactory = TestProxyTransportFactory()
         options.tokenDetails = try getTestTokenDetails(for: test, ttl: 5.0)
         let client = ARTRealtime(options: options)
         defer { client.dispose(); client.close() }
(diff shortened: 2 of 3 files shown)


Test/Tests/RealtimeClientChannelTests.swift
```diff
@@ -521,15 +522,12 @@ class RealtimeClientChannelTests: XCTestCase {
...
-
-                channel.on { stateChange in
-                    if stateChange.current == .attached {
-                        XCTAssertTrue(stateChange.resumed)
-                        XCTAssertNil(stateChange.reason)
-                        XCTAssertEqual(stateChange.current, ARTRealtimeChannelState.attached)
-                        XCTAssertEqual(stateChange.previous, ARTRealtimeChannelState.attached)
-                        done()
-                    }
+                XCTAssertEqual(channel.state, ARTRealtimeChannelState.attached)
```
lawrence-forooghian, line 525: I don't understand the idea behind these test changes — why is it no longer making assertions about the `ATTACHED` `ChannelStateChange`, which is what RTL2f talks about? And how is it related to the above change, which relates to when an `UPDATE` `ChannelStateChange` will be emitted?

maratal: Looks like RTL2f talks about `ChannelStateChange` missing details of `RTL12` which states that only in case of `resumed == false` `channel.on` will be triggered. This test checks the case when `resumed == true`, that's why I've removed `channel.on` (bc it will not be triggered). So the new assertions claim that channel states remains the same before and after ATTACHED protocol message and that it contains `resumed == true`.

lawrence-forooghian: `RTL2f` is about what value to use for the `resumed` value of a `ChannelStateChange`. That means that this test is no longer testing `RTL2f` as it advertises, since as you've said there is no channel state change any more.

maratal: I've [removed](https://github.com/ably/ably-cocoa/pull/1791/commits/1a5f933f4068398009cf0641bfdfb76844fa5baa) `RTL2f` test and created [issue](https://github.com/ably/ably-cocoa/issues/1812) to create a proper one.

maratal: Also updated `RTL12` test - https://github.com/ably/ably-cocoa/pull/1791/commits/9184efa3eece85f60db12d0cf69cada8a6ba5319

Test/Tests/RealtimeClientChannelTests.swift
```diff
@@ -3791,10 +3721,13 @@ class RealtimeClientChannelTests: XCTestCase {
             attachedMessageWithError.action = .attached
             attachedMessageWithError.channel = channel.name
 
-            channel.once(.update) { stateChange in
+            channel.on { stateChange in
```
lawrence-forooghian, line 3724: Why this change from `once` to `on`?

lawrence-forooghian (commented):
Which test changes reflect the changed behaviour of `ARTRealtimeChannel`? It's not clear to me.

Test/Tests/RealtimeClientChannelTests.swift
lawrence-forooghian, line 3724: Why this change from `once` to `on`?

Source/ARTRealtimeChannel.m
```diff
@@ -741,11 +741,13 @@ - (void)setAttached:(ARTProtocolMessage *)message {
     }
 
     if (self.state_nosync == ARTRealtimeChannelAttached) {
-        if (message.error != nil) {
-            _errorReason = message.error;
+        if (!message.resumed) { // RTL12
```
lawrence-forooghian, line 744: Where is this new behaviour tested? It's not clear to me.

Test/Tests/RealtimeClientChannelTests.swift
maratal: I think it should be `on`, bc in case of invalid event the test will fail (on event assertion), which will signal about improper library behavior.

lawrence-forooghian: OK, makes sense.

Source/ARTRealtimeChannel.m
maratal: https://github.com/ably/ably-cocoa/pull/1791/commits/18f5f6ca27f47eaccb234d57a3a6350e03e504f6

lawrence-forooghian (approved):

PR #1793, opened 2023-08-20
[SDK-3814] Refactoring for "test__023__Presence__Channel_state_change_side_effects…"

### LINKED ISSUE #1792: Refactoring for test "test__023__Presence__Channel_state_change_side_effects__channel_enters_the_SUSPENDED..."
Test for [`RTP5f`](https://sdk.ably.com/builds/ably/specification/main/features/#RTP5f) should ensure, that for every member whose state changed during SUSPENDED an appropriate event should be generated after client reconnects:

>This ensures that if the channel later becomes @ATTACHED@, it will only emit presence events for the changes in the @PresenceMap@ that have occurred whilst the client was disconnected.

How [leaving](https://github.com/ably/ably-cocoa/blob/273277e566fc1641cc8fc5b6473999a3e9ab6c4e/Test/Tests/RealtimeClientPresenceTests.swift#L812C37-L812C37) after channel was attached demonstrates that is unclear for me.

maratal:
Closes #1792

Now a separate client is created for the member that will leave during SUSPENDED and ensure that a leave event for this member will be generated upon new attach.

### DIFF
--- a/Test/Tests/RealtimeClientPresenceTests.swift
+++ b/Test/Tests/RealtimeClientPresenceTests.swift
@@ -813,7 +813,10 @@ class RealtimeClientPresenceTests: XCTestCase {
         }
 
         var presenceEvents = [ARTPresenceMessage]()
-        
+        var mainSuspendedAt = Date()
+        var mainAttachedAt = Date()
+        var clientLeftAt = Date()
+
         waitUntil(timeout: testTimeout) { done in
             let partialDone = AblyTests.splitDone(4, done: done)
             mainChannel.presence.subscribe { presence in
@@ -823,19 +826,21 @@ class RealtimeClientPresenceTests: XCTestCase {
                 }
             }
             mainChannel.once(.suspended) { _ in
+                mainSuspendedAt = Date()
                 mainChannel.internalSync { _internal in
                     XCTAssertEqual(_internal.presenceMap.members.count, 4) // "main", "user1", "user2", "leaves"
                     XCTAssertEqual(_internal.presenceMap.localMembers.count, 1) // "main"
                 }
                 leavesChannel.presence.leave(nil) { error in
                     XCTAssertNil(error)
-                    XCTAssertEqual(mainChannel.state, .suspended)
+                    clientLeftAt = Date()
                     partialDone()
                 }
                 partialDone()
             }
             mainChannel.once(.attached) { stateChange in
                 XCTAssertNil(stateChange.reason)
+                mainAttachedAt = Date()
                 partialDone()
             }
             mainChannel.internalAsync { _internal in
@@ -845,6 +850,9 @@ class RealtimeClientPresenceTests: XCTestCase {
         XCTAssertEqual(presenceEvents.count, 1)
         XCTAssertEqual(presenceEvents[0].action, ARTPresenceAction.leave)
         XCTAssertEqual(presenceEvents[0].clientId, "leaves")
+        
+        XCTAssertTrue(mainSuspendedAt < clientLeftAt)
+        XCTAssertTrue(clientLeftAt < mainAttachedAt)
 
         mainChannel.presence.unsubscribe()
         
--- a/Test/Tests/RealtimeClientPresenceTests.swift
+++ b/Test/Tests/RealtimeClientPresenceTests.swift
@@ -829,6 +829,7 @@ class RealtimeClientPresenceTests: XCTestCase {
                 }
                 leavesChannel.presence.leave(nil) { error in
                     XCTAssertNil(error)
+                    XCTAssertEqual(mainChannel.state, .suspended)
                     partialDone()
                 }
                 partialDone()
--- a/Test/Tests/RealtimeClientPresenceTests.swift
+++ b/Test/Tests/RealtimeClientPresenceTests.swift
@@ -780,11 +780,11 @@ class RealtimeClientPresenceTests: XCTestCase {
             let partialDone = AblyTests.splitDone(4, done: done)
             mainChannel.presence.subscribe { message in
                 if message.clientId == "main" {
-                    XCTAssertEqual(message.action, ARTPresenceAction.enter)
+                    XCTAssertTrue(message.action == ARTPresenceAction.enter || message.action == ARTPresenceAction.present)
                     partialDone()
                 }
                 else if message.clientId == "leaves" {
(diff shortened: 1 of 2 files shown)


Test/Tests/RealtimeClientPresenceTests.swift
```diff
@@ -756,86 +756,106 @@ class RealtimeClientPresenceTests: XCTestCase {
...
-            channel.presence.get { members, error in
+            leavesChannel.presence.enter(nil) { error in
+                XCTAssertNil(error)
+                partialDone()
+            }
+            mainChannel.presence.get { members, error in
                 XCTAssertNil(error)
                 if let members {
-                    XCTAssertEqual(members.count, 3)
+                    XCTAssertEqual(members.count, 3) // "user1", "user2", "leaves"
```
lawrence-forooghian, line 792: Is this guaranteed to contain `"leaves"`, given that line 781's `leavesChannel.presence.enter` may not yet even have completed?

Test/Tests/RealtimeClientPresenceTests.swift
```diff
@@ -756,86 +756,106 @@ class RealtimeClientPresenceTests: XCTestCase {
...
-                if message.clientId == "tester" {
-                    channel.presence.unsubscribe()
+            mainChannel.presence.subscribe { message in
+                if message.clientId == "main" {
+                    XCTAssertEqual(message.action, ARTPresenceAction.enter)
+                    mainChannel.presence.unsubscribe()
+                    partialDone()
+                }
+                else if message.clientId == "leaves" {
+                    XCTAssertEqual(message.action, ARTPresenceAction.present) // .enter was replaced by .present (RTP2d)
```
lawrence-forooghian, line 805: What do you mean ".enter was replaced by .present (RTP2d)"? RTP2d says "with an event name set to its original action" — there's nothing there to suggest that anything should be "replaced".

Test/Tests/RealtimeClientPresenceTests.swift
```diff
@@ -756,86 +756,106 @@ class RealtimeClientPresenceTests: XCTestCase {
...
             }
-            channel.once(.attached) { stateChange in
-                XCTAssertNil(stateChange.reason)
-                channel.presence.leave(nil) { error in
+            mainChannel.once(.suspended) { _ in
+                mainChannel.internalSync { _internal in
+                    XCTAssertEqual(_internal.presenceMap.members.count, 4) // "main", "user1", "user2", "leaves"
+                    XCTAssertEqual(_internal.presenceMap.localMembers.count, 1) // "main"
+                }
+                leavesChannel.presence.leave(nil) { error in
```
lawrence-forooghian, line 825: What makes you sure that this presence leave occurs whilst `mainChannel` is `SUSPENDED`? What if `mainChannel` reattaches before this line’s presence leave completes?

lawrence-forooghian (commented):
I agree that the motivation for the behaviour in the existing test is not very clear, and that it makes sense to replace it with something that makes more sense. However, I still think that the test, as it stands in this PR, is very difficult to understand (at least for me). What do you think?

lawrence-forooghian:
Note: this needs to be merged before #1714, so that we can remove commit fdbd4ac from that PR.

Test/Tests/RealtimeClientPresenceTests.swift
maratal: https://github.com/ably/ably-cocoa/pull/1793/commits/5220e18b399095293657b1e269566226d4c9ce5f

Test/Tests/RealtimeClientPresenceTests.swift
maratal: https://github.com/ably/ably-cocoa/pull/1793/commits/9a11e4383d16fd72c65554a7a65123283bde0788

maratal (commented):
I think it's better now @lawrence-forooghian

Test/Tests/RealtimeClientPresenceTests.swift
lawrence-forooghian: Sorry, yes, I meant RTP2g.

What I'm saying is that as far as I can see there is nothing in the spec which says that `RealtimePresence` should "replace" the action before emitting it. RTP2d talks about what should be stored in the `PresenceMap`, not what should be emitted. And then RTP2g explicitly says that the received presence message should be emitted.

I think that what's probably happening here is timing-dependent, and that in the situations where you ran it, the `mainChannel.presence.enter()`’s side effects completed after the initial channel `SYNC`, and those of `leavesChannel.presence.enter()` after the initial `SYNC`, but I don't think this is guaranteed. I imagine that if you run this test multiple times, or if you put a delay between adding the presence subscriber and entering the channels, then you'll see that these events could be `PRESENT` or `ENTER`.

Whilst there are probably things you can do to get consistent behaviour here, I'd argue that it doesn't matter; you really just care that both members are present, so I'd suggest just checking (for both client IDs) that the action is `PRESENT` or `ENTER`.

Test/Tests/RealtimeClientPresenceTests.swift
lawrence-forooghian: I don't think that the situation is different in your new version of the code.

You're triggering the `leavesChannel` presence leave when `mainChannel` enters the `SUSPENDED` state. But, upon entering the `SUSPENDED` state, `mainChannel` will trigger a channel attach operation. How do you know that the presence leave will complete before the attach?

I'm not sure of a good way of being sure of this. The ideal would be to make sure that the presence leave operation completes before `mainChannel` tries to reattach. Is there a way, for example, that we could delay the sending of the `ATTACH` `ProtocolMessage` until the presence enter has completed? Perhaps using `TestProxyTransport`?

Test/Tests/RealtimeClientPresenceTests.swift
maratal: https://github.com/ably/ably-cocoa/pull/1793/commits/e20310d2c94bd8f870460e05973ac06dc6a44e15

Test/Tests/RealtimeClientPresenceTests.swift
maratal: https://github.com/ably/ably-cocoa/pull/1793/commits/e04e5865f0bfc2091aaaa8417e60634baa1aa248

lawrence-forooghian (approved):

PR #1783, opened 2023-07-08
[Experimental] Transition to Xcode 14.3

maratal:
See #1801

PR #1797, opened 2023-09-10
[Experimental] Transition to macos 13 and Xcode 14.3

### LINKED ISSUE #1764: Transition to Xcode 14.3 (umbrella issue)
Turned out there are number of problems popped out in Xcode 14.3:

Issue #1765
Build fails on the `Build Carthage dependencies` step during building Nimble for watchOS.

Issue #1766
After watchOS removal build fails on the `Build Ably framework` step.

<strike>Issue #1767</strike>
<strike>A lot of new test failures. Probably the result of changes in the Nimble framework (how it does comparisons).</strike>

Issue #1796:
(Replaces #1767 for CI)

[Draft PR](https://github.com/ably/ably-cocoa/pull/1760) with experiments on macOS 13.4 and Xcode 14.3

<strike>Turned out Xcode 14.2 also can't compile Nimble - https://github.com/ably/ably-cocoa/issues/1768</strike>

maratal:
See #1801

Contains work from https://github.com/ably/ably-cocoa/pull/1783

The reason why tests stop failing after `ttl` time has been increased in #1783 remains unknown, but I think it's due to different hardware Github uses for its beta servers where low performance can affect how test executes with smaller timeout values. So I would run those workflows again when macOS 13 (by reverting explicit switch to macos 13 3bb354db67ef476603ff0a75b03c3f856a1f3b93) will become the latest available os and see how tests will behave there.

The test "test__81..." was modified to reflect what it says in the title and to be similar to the "test__80..." which has similar title and conditions. But I left the previous version of the "test__80..." with the "skip" prefix and created an [issue](https://github.com/ably/ably-cocoa/issues/1798) for dealing with it later, since I think that behavior of this test is rather confusing and is in contradiction to its own title.

### DIFF
--- a/.github/workflows/check-pod.yaml
+++ b/.github/workflows/check-pod.yaml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   check:
-    runs-on: macos-latest
+    runs-on: macos-13
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/check-pod.yaml
+++ b/.github/workflows/check-pod.yaml
@@ -12,9 +12,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Select Specific Xcode Version (13.4.1)
+      - name: Select Specific Xcode Version (14.3)
         run: |
-          sudo xcode-select -s /Applications/Xcode_13.4.1.app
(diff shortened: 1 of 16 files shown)


maratal:
This work is done in https://github.com/ably/ably-cocoa/pull/1801

PR #1799, opened 2023-09-10
[Experimental] Transition to Xcode 15

maratal:
See #1802

PR #1760, opened 2023-06-12
Try CI on macOS 13 explicitly.

maratal:
Not for merging

PR #1802, opened 2023-09-15
[SDK-3831] CI job for iOS 17

### LINKED ISSUE #1796: Add CI job for latest Xcode 15 / iOS 17 and fix issues
Add a new CI job that runs in the latest version of Xcode and iOS. Fix any issues discovered.

maratal:
Closes #1796

On top of #1801.

What's interesting, even while just co-existing (in this PR) with a workflow, containing macOS 13 (added iOS 17), some of [those](https://github.com/ably/ably-cocoa/pull/1760#issuecomment-1606290881) tests might [fail](https://github.com/ably/ably-cocoa/actions/runs/6197805473/job/16827067736?pr=1802).

So my suggestions to handle this would be one of the following:

1) Skip those tests (and create an issue) until macOS 13 will be the default version and see what will happen.
2) Increase ttl (bad idea, because it will just hide the problem).
3) Rewrite tests not to use token authentication when possible (bad the same way as 2).

WDYT @lawrence-forooghian 

Previous [attemt](https://github.com/ably/ably-cocoa/pull/1799) with [increased](https://github.com/ably/ably-cocoa/pull/1799/commits/a1e270a235abd6171dc5804ddd0a0197870874b7) `ttl`.


lawrence-forooghian:
> On top of https://github.com/ably/ably-cocoa/pull/1801.

When you have a PR that's based on top of an another one, the "from" branch of the second PR should be the branch of the first PR. I've changed it in this PR.

lawrence-forooghian:
@maratal I think now that iOS / tvOS 17 are now released, we can just change the existing CI jobs to use them (and update [the documentation in the README](https://github.com/ably/ably-cocoa/blob/273277e566fc1641cc8fc5b6473999a3e9ab6c4e/README.md?plain=1#L35-L36)), instead of adding new ones. What do you think?

Test/Tests/RestClientTests.swift
```diff
@@ -1822,7 +1822,7 @@ class RestClientTests: XCTestCase {
         let url = try XCTUnwrap(request.url, "No request url found")
         let acceptHeaderValue = try XCTUnwrap(request.allHTTPHeaderFields?["Accept"], "Accept HTTP Header is missing")
         
-        XCTAssertEqual(request.httpMethod, "PATCH")
+        XCTAssertEqual(request.httpMethod!.uppercased(), "PATCH")
```
lawrence-forooghian, line 1825: I don't understand this second change. What's it trying to achieve? I also can't replicate the failure locally — it passes using `"PATCH"` as the second argument for me.

maratal: In Xcode 15 `XCTAssertEqual(request.httpMethod, "patch")` fails with `Optional("PATCH")") is not equal to ("Optional("patch")`. So I've changed it in 028cde0c0f1518fada4a0c81e5bb1b867ec04407 to uppercased "PATCH".

But on CI in Xcode 14.2 it began to fail with opposite message `Optional("patch")") is not equal to ("Optional("PATCH")`.

Thus to satisfy both Xcodes I've forced to use uppercase in this assertion.

maratal:
> @maratal I think now that iOS / tvOS 17 are now released, we can just change the existing CI jobs to use them (and update [the documentation in the README](https://github.com/ably/ably-cocoa/blob/273277e566fc1641cc8fc5b6473999a3e9ab6c4e/README.md?plain=1#L35-L36)), instead of adding new ones. What do you think?

But for that macos-13 should be selected as runner image OS. Which leads to all the failures described in https://github.com/ably/ably-cocoa/pull/1801

lawrence-forooghian:
> But for that macos-13 should be selected as runner image OS (default is still macos-12). Which leads to all the failures described in https://github.com/ably/ably-cocoa/pull/1801

I don't understand what you mean. What I'm saying is that instead of having a CI job for iOS 17 and also a CI job for iOS 16, we only need one for iOS 17 (since in general we only perform CI testing on the latest version of iOS). What is the connection between this and macOS version or test failures?

Test/Tests/RestClientTests.swift
maratal: This goes into a separate [PR](https://github.com/ably/ably-cocoa/pull/1808)

lawrence-forooghian:
@maratal are you sure that these test failures are dependent on iOS version? I've noticed that we seem to have some tests that fail intermittently. From what I've seen, it seems to depend on the CI runner — i.e. if the tests fail once on that runner, then they'll fail if you run them again (I've noticed this times when I've been running tests in a loop; some runners continuously fail and some don't fail at all). See [here](https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads?branches%5B%5D=2023-11-07-run-tests-in-loop-before-merging-1823&createdBefore=&createdAfter=&failureMessage=) for examples, for example [this one](https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads/554d09f1-71e2-4e79-b641-9f33a8382220) seems to match the list of tests you mentioned above.

Is it possibly something to do with the clock on the runner?

maratal:
> @maratal are you sure that these test failures are dependent on iOS version? I've noticed that we seem to have some tests that fail intermittently. From what I've seen, it seems to depend on the CI runner — i.e. if the tests fail once on that runner, then they'll fail if you run them again (I've noticed this times when I've been running tests in a loop; some runners continuously fail and some don't fail at all). See [here](https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads?branches%5B%5D=2023-11-07-run-tests-in-loop-before-merging-1823&createdBefore=&createdAfter=&failureMessage=) for examples, for example [this one](https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads/554d09f1-71e2-4e79-b641-9f33a8382220) seems to match the list of tests you mentioned above.

Not iOS, macOS 13 back when I've created this PR. Recently I began to see those failures on macOS 12 as well from time to time. Now it happens a few days in a raw very often both on macOS 12 and 13.

> Is it possibly something to do with the clock on the runner?

I don't know yet, but increasing token ttl helped to eliminate this. Do you see any improvement by adding tolerance to ttl?

maratal:
Solved https://github.com/ably/ably-cocoa/pull/1851

PR #1806, opened 2023-09-24
[SDK-3868] Introducing `nil` of type `String` for Xcode 15

### LINKED ISSUE #1805: Xcode 15 requires contextual type for nils in swift
Some optional args with passed `nil` show a compilation error `'nil' requires a contextual type` (but still allow the project to be built). This appears in the currently opened file (among other errors). I think it's related to Xcode indexing process.

Screenshot 2023-09-25 at 03 20 11

maratal:
Closes #1805

### DIFF
--- a/Test/Test Utilities/TestUtilities.swift
+++ b/Test/Test Utilities/TestUtilities.swift
@@ -1957,3 +1957,8 @@ extension ARTErrorCode {
         return NSInteger(rawValue)
     }
 }
+
+// Xcode 15 requires contextual type for nils in swift
+extension String {
+    static let `nil`: Self? = nil
+}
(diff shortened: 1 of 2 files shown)


lawrence-forooghian (commented):
It's a surprising error; shouldn't the compiler be able to infer the type of the `nil` literal? I was also unable to reproduce it (just tried opening `RealtimeClientChannelTests.swift` in Xcode and then ran the tests); anything specific I should try? I'm hesitant about merging this "fix" until we understand why the issue is happening, because it makes the code less readable.

I'm also not a fan of extending `String` with a method named `nil`; my preference would be to write `nil as String?` if needed (or maybe to keep the method but make its name more descriptive with a good comment explaining why it exists).

maratal:
It's now [closed](https://github.com/ably/ably-cocoa/issues/1805#issuecomment-1742112447) @lawrence-forooghian

maratal:
I've just realized I hadn't an email notification on this, thus late response, sorry @lawrence-forooghian

PR #1822, opened 2023-11-05
[SDK-3930] Removed unnecessary re-attach attempt.

### LINKED ISSUE #1821: Extra re-attach forgotten to be removed.
My initial comment `// Attach should be 2 here, as one above and one reattach` was misleading (I've made erroneous conclusion about what gives me the second `ATTACH` in sent messages). I've understood it when was trying to address you comment about other tests with "as one above and one reattach" pattern. The count wasn't always 2 for sent messages in other tests, and after some debugging I've located this place where the second `ATTACH` message was generated. I believe Ricardo forgot to remove this in 784da01, because earlier re-attach would not happen,  because `RTL4i` block was in the `else if (![self shouldQueueEvents])` part of `performTransitionWithStateChange:` and re-attaching was made upon emitting `[_connectedEventEmitter emit:nil with:nil]` in `case ARTRealtimeConnected:`. Once Recardo moved it to `if ([self shouldSendEvents])` re-attaching from emitter should have been be deleted.

_Originally posted by @maratal in https://github.com/ably/ably-cocoa/pull/1714#discussion_r1359903665_

maratal:
Closes #1821

### DIFF
--- a/Source/ARTRealtimeChannel.m
+++ b/Source/ARTRealtimeChannel.m
@@ -1050,16 +1050,6 @@ dispatch_sync(_queue, ^{
             [self setSuspended:metadata];
         }] startTimer];
     } ackCallback:nil];
-
-    if (![self.realtime shouldQueueEvents]) {
-        ARTEventListener *reconnectedListener = [self.realtime.connectedEventEmitter once:^(NSNull *n) {
-            // Disconnected and connected while attaching, re-attach.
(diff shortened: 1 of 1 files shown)


lawrence-forooghian (approved):
I've got no idea what the original code was trying to do. I'm approving based on the fact that it hasn't caused any test failures.

Issue #1824, opened 2023-11-06
App build fails due to missing dependencies (xcode 15.0.1)


bhargavbajani-simformsolutions:
### Which version of the Ably SDK are you using? 
Version `[1.2.21](https://github.com/ably/ably-cocoa/releases/tag/1.2.21)` 

### On which platform does the issue happen? 
`iOS17`

### Are you using Carthage? 
Yes, github "ably/ably-cocoa" "1.2.21"

### Which version of Xcode are you using? 
yes, Xcode 15.0.1

### What did you do?
The build fails when including the framework - tried with the attached framework via Github releases, Carthage.
![image](https://github.com/ably/ably-cocoa/assets/63727314/b5d8691b-835c-458d-bae4-8a31b8cf6a6a)

'ARTTypes.h' file not found

### Command used to install a framework
carthage update ably-cocoa --use-xcframeworks --no-use-binaries --platform iOS

### What did you expect to happen?
The build should succeed.

### What happened instead?
The build should fail with the above error message.

bhargavbajani-simformsolutions:
@maratal  It's not possible for now due to project limitations.

maratal:
@lawrence-forooghian shouldn't we give this another shot? https://github.com/ably/ably-cocoa/pull/1756

maratal:
It is now fixed in https://github.com/ably/ably-cocoa/releases/tag/1.2.23 @bhargavbajani-simformsolutions

PR #1833, opened 2023-11-13
Carthage install fix

### LINKED ISSUE #1755: App build fails due to missing dependencies
1.2.20

iOS 16.4

### Are you using Carthage? 
Yes, 0.38.0

### Are you using Cocoapods? 
Yes, 1.11.3

14.3

### What did you do?
The build fails when including the framework - tried with the attached framework via Github releases, Carthage and Cocoapods.
Screenshot 2023-06-08 at 10 39 57

> 'ARTTypes.h' file not found

The build should succeed.

### What happened instead?
The build should fails with the above error message.

maratal:
Closes #1755, #1824

This PR doesn't change runner image to macos 13 to run tests on Xcode 14.3 or 15, although it will fix Carthage installation problem under those versions. Experiments under these versions see [here](https://github.com/ably/ably-cocoa/pull/1799).

Issue was reproduced [here](https://github.com/ably/ably-cocoa/actions/runs/6801574304/job/18494243611).

### DIFF
--- a/.github/workflows/integration-test-iOS16_2.yaml
+++ b/.github/workflows/integration-test-iOS16_2.yaml
@@ -121,3 +121,11 @@ jobs:
           export PACKAGE_BRANCH_NAME=$GITHUB_HEAD_REF
           export PACKAGE_REVISION=$GITHUB_SHA
           swift test --package-path Examples/SPM -v
+      
+      - name: Carthage - Installation Test
+        working-directory: ./Examples/AblyCarthage
+        run:  |
+          echo 'Installing cathage dependencies...'
+          carthage update --use-xcframeworks --platform iOS --no-use-binaries
+          echo 'Building AblyCarthage example...'
+          xcodebuild build -scheme "AblyCarthage" -destination "platform=iOS Simulator,name=iPhone 14" -configuration "Debug"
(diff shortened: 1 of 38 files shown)


.github/workflows/integration-test-iOS16_2.yaml
```diff
@@ -121,3 +121,11 @@ jobs:
           export PACKAGE_BRANCH_NAME=$GITHUB_HEAD_REF
           export PACKAGE_REVISION=$GITHUB_SHA
           swift test --package-path Examples/SPM -v
+      
+      - name: Carthage - Installation Test
```
lawrence-forooghian, line 125: I don't think we need to run this three times (especially as it doesn't seem to vary by platform) — could we have a new workflow for it? It will also mean that it can run in parallel to the other build steps.

.github/workflows/integration-test-iOS16_2.yaml
```diff
@@ -121,3 +121,11 @@ jobs:
...
+        working-directory: ./Examples/AblyCarthage
+        run:  |
+          echo 'Installing cathage dependencies...'
```
lawrence-forooghian, line 128:
```suggestion
          echo 'Installing Carthage dependencies...'
```

maratal: https://github.com/ably/ably-cocoa/pull/1833/commits/9dba2a9770064d4aa5eba36d9f6ccd94e704bc1c

.github/workflows/integration-test-iOS16_2.yaml
maratal: https://github.com/ably/ably-cocoa/pull/1833/commits/9dba2a9770064d4aa5eba36d9f6ccd94e704bc1c

lawrence-forooghian (approved):

PR #1836, opened 2023-11-14
Try to reduce failures in some tests that exercise token expiry

lawrence-forooghian:
I’ve noticed these tests intermittently failing on CI. They all wait for the Ably service to consider a token as having expired by waiting for its `ttl` to elapse. However, clock discrepancies between Ably servers might mean that this doesn’t always work. So let’s try adding a bit of a tolerance to the wait and see if it helps.

There are probably other tests that are affected similarly; these are just the ones that caught my eye.


maratal (approved):
LGTM

PR #1840, opened 2023-11-19
[SDK-1683] Single workflow with matrix strategy

### LINKED ISSUE #1311: Revert the work done to shard workflows in order to work around GitHub's lack of capability to re-run specific jobs, rather than entire workflows
owenpearson spotted a new blog post from GitHub: [Save time with partial re-runs in GitHub Actions](https://github.blog/2022-03-16-save-time-partial-re-runs-github-actions/).

If that :point_up: solves the issue for us in terms of re-runs, then we should undo the work done under https://github.com/ably/ably-cocoa/pull/1274 in order to get rid of the current repetition we have between workflow files.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1683) by [Unito](https://www.unito.io)

maratal:
Closes #1311


lawrence-forooghian (commented):
Something that I noticed is that there is now no way to tell which OS a corresponding observability server upload corresponds to. Compare the following:

- [an upload from before these changes](https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads/b12c25dc-c3ac-4912-a6cc-6106ed026267)
- [an upload from after these changes](https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads/358cbc52-d669-4e6a-827d-d3ac7f46e685)

Notice that in the first one, clicking the "GitHub run ID" or "GitHub run attempt" link takes you to the iOS job. In the second, it takes you to an overview page that shows all 3 OSs.

There is probably some other piece of data that we need to get from the [environment](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) and include in the upload to the observability server to allow it to display a link to the correct job.

maratal:
> Something that I noticed is that there is now no way to tell which OS an observability server upload corresponds to. Compare the following:
> 
>     * [an upload from before these changes](https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads/b12c25dc-c3ac-4912-a6cc-6106ed026267)
> 
>     * [an upload from after these changes](https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads/358cbc52-d669-4e6a-827d-d3ac7f46e685)
> 
> 
> Notice that in the first one, clicking the "GitHub run ID" or "GitHub run attempt" link takes you to the iOS job. In the second, it takes you to an overview page that shows all 3 OSs.
> 
> There is probably some other piece of data that we need to get from the [environment](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) and include in the upload to the observability server to allow it to display a link to the correct job.

Turned out it's not trivial. `GITHUB_JOB` is actually an alphabetical id ("check" in our case), and to get numerical id one need to use a custom parser action - [this](https://github.com/marketplace/actions/github-actions-job_id-parser) for example. Source https://stackoverflow.com/questions/71240338/obtain-job-id-from-a-workflow-run-using-contexts

lawrence-forooghian:
I think we need to do that then. Sounds like the steps are the following:

1. before uploading to the observability server, [fetch the jobs for the current workflow attempt](https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28#list-jobs-for-a-workflow-run-attempt)
2. find the currently-running job from the response's `jobs` array (I believe, according to what [this](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy) says about the order of job creation, that the $n^{th}$ item there corresponds to the $n^{th}$ item in the matrix, so we can fetch it by using the index of the current job in the matrix, which we might need to pass as an additional variable from the matrix to the job)
3. extract the `html_url` from that job and upload it as a new parameter (e.g. `GITHUB_JOB_URL`, I can add support to the server for storing and displaying this value) to the observability server

lawrence-forooghian:
I've actually just needed to do the same thing for ably-js. Here's the PR for adding this functionality to the upload; it should help you here. https://github.com/ably/test-observability-action/pull/20

lawrence-forooghian:
I've implemented it for ably-cocoa in https://github.com/ably/ably-cocoa/pull/1845. You'll just need to pass the `--job-index` argument in this PR (in your case it'll be `${{ strategy.job-index }}`).

maratal:
> I've implemented it for ably-cocoa in #1845. You'll just need to pass the `--job-index` argument in this PR (in your case it'll be `${{ strategy.job-index }}`).

This is wonderful, thanks Lawrence!

.github/workflows/integration-test.yaml
```diff
@@ -14,25 +14,39 @@ jobs:
...
           sudo xcode-select -s /Applications/Xcode_14.2.app
           echo "Selected Xcode version:"
           xcodebuild -version
 
-      - name: Log environment information
+      - name: Environment Info
         run: ./Scripts/log-environment-information.sh
 
-      - name: Check out xcparse repo
+      - name: Checkout xcparse repo
```
lawrence-forooghian, line 49: "Check out" is the correct wording here (see e.g. the wording in `man git-checkout`). Ditto above.

.github/workflows/integration-test.yaml
```diff
@@ -72,48 +88,35 @@ jobs:
...
-        working-directory: ./
-        run:  |
-          echo 'Current Branch: ' $GITHUB_HEAD_REF
-          echo 'Current Revision (SHA):' $GITHUB_SHA
-          echo Current Path: $(pwd)
-          export PACKAGE_URL=file://$(pwd)
-          export PACKAGE_BRANCH_NAME=$GITHUB_HEAD_REF
-          export PACKAGE_REVISION=$GITHUB_SHA
-          swift test --package-path Examples/SPM -v
+          TEST_RESULTS_WEB_UI_URL: undefined
```
lawrence-forooghian, line 117: Is this dance with `TEST_RESULTS_WEB_UI_URL` (it's not a common pattern for a script to write its output to an environment variable) and the extra job necessary? Can't we just wrap `upload_test_results.sh`’s `printf "Created request body..."` inside a `::group::Body of request`?

.github/workflows/integration-test.yaml
maratal: https://github.com/ably/ably-cocoa/pull/1840/commits/b9e17f004d54bd851c950e953d278127744e4728

.github/workflows/integration-test.yaml
maratal: https://github.com/ably/ably-cocoa/pull/1840/commits/6bd3ea8a8adf35b3290e4362ad9a4c9663cead07

lawrence-forooghian: Nice, looks good

lawrence-forooghian (commented):
Looks good — let's not merge this until https://github.com/ably/test-observability/pull/76 is merged and deployed though.

.github/workflows/integration-test.yaml
```diff
@@ -14,22 +14,36 @@ jobs:
   check:
     runs-on: macos-latest
 
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+
+          - platform: iOS
```
lawrence-forooghian, line 22: Can we keep the version numbers in the `platform` property so that they appear alongside the CI job, like they did before?

maratal: I guess this value is used as a var later - ${{ matrix.platform }}

maratal: maybe rename a lane to just "iOS_XXX"? "test_" prefix occupies too much space in GH UI.

lawrence-forooghian:
> Looks good — let's not merge this until [ably/test-observability#76](https://github.com/ably/test-observability/pull/76) is merged and deployed though.

This is now deployed.

.github/workflows/integration-test.yaml
lawrence-forooghian: I think it's best to keep the `test_` prefix. Perhaps we call the param `lane_suffix` and then do `lane: test_${{ matrix.lane_suffix }}`?

lawrence-forooghian: Oh, sorry, I was confused, I didn't notice the `lane` param. Ignore this thread.

lawrence-forooghian:
Re-running CI to check that the URL gets correctly populated in observability server. If it does, I'll approve.

lawrence-forooghian (approved):

PR #1847, opened 2023-12-21
Proper RSH8b and RSH3g2a behavior.

maratal:
See https://github.com/ably/ably-cocoa/pull/1842, https://github.com/ably/ably-cocoa/pull/1253 and commit messages for context.


Source/ARTLocalDevice.m
```diff
@@ -102,6 +95,48 @@ + (ARTLocalDevice *)load:(NSString *)clientId storage:(id<ARTDeviceStorage>)stor
     return device;
 }
 
+- (BOOL)setupDetailsWithClientId:(NSString *)clientId {
+    NSString *deviceId = self.id;
+    NSString *deviceSecret = self.secret;
+    
+    if ([deviceId isEqualToString:@""] || deviceSecret == nil) { // generate both at the same time
```
ttypic, line 102: why we check for empty string now?

Source/ARTLocalDevice.m
```diff
@@ -102,6 +95,48 @@ + (ARTLocalDevice *)load:(NSString *)clientId storage:(id<ARTDeviceStorage>)stor
...
+        return NO;
+    }
+    if (clientId) {
+        self.clientId = clientId;
+    }
+    return YES;
+}
+
+- (void)resetDetails {
+    self.id = @"";
```
ttypic, line 126: Why it's empty string and not nil? Is empty string shows that id hasn't been reseted?

Source/ARTLocalDevice.m
```diff
@@ -72,24 +71,18 @@ + (ARTLocalDevice *)load:(NSString *)clientId storage:(id<ARTDeviceStorage>)stor
     #endif
     device.push.recipient[@"transportType"] = ARTDevicePushTransportType;
 
-    NSString *deviceId = [storage objectForKey:ARTDeviceIdKey];
+    NSString *deviceId = [storage objectForKey:ARTDeviceIdKey] ?: @""; // PCD2, not nullable
```
ttypic, line 74: I guess statement below that checks for `nil` will be affected. I think we shouldn't replace `nil` with empty string just because in the spec id is not nullable.

maratal: Does it mean it should become nullable in the spec? Or should we ignore the spec because an empty string check in objective-c looks ugly?

ttypic: I think it should be nullable in spec. It's not because it's ugly, it's because it's inconsistent. If id is not set we use `nil` everywhere, what is the reason to do differently here?

maratal: >I think it should be nullable in spec.
Agree. Will change it back then. Nice catch!

ttypic: Thank you!

maratal: It will be a breaking change tho... I guess we should discuss it with the rest of the team and this shouldn't be done in this PR.

maratal: Lets rely on what already was approved by Lawrence https://github.com/ably/ably-cocoa/pull/1259

maratal: https://github.com/ably/ably-cocoa/pull/1847/commits/8381e102e261e302cb3e986c53f3971ecd9b2e5b

ttypic (approved):
LGTM

Source/ARTLocalDevice.m
```diff
@@ -102,6 +95,33 @@ + (ARTLocalDevice *)load:(NSString *)clientId storage:(id<ARTDeviceStorage>)stor
...
+        [_storage setSecret:deviceSecret forDevice:deviceId];
+    }
+    
+    self.id = deviceId;
+    self.secret = deviceSecret;
+    
+    NSString *localClientId = self.clientId;
+    
+    if (localClientId && clientId && ![localClientId isEqualToString:clientId]) {
+        ARTLogError(self.logger, @"ARTLocalDevice: `clientId`s don't match: %@ vs %@.", clientId, localClientId);
```
lawrence-forooghian, line 116: What spec point does this correspond to?

Source/ARTLocalDevice.m
lawrence-forooghian: How does Android handle this — what value does `LocalDevice#id` return before RSH3a2b has happened?

Source/ARTLocalDevice.m
```diff
@@ -72,24 +71,18 @@ + (ARTLocalDevice *)load:(NSString *)clientId storage:(id<ARTDeviceStorage>)stor
...
-    }
-    
     device.id = deviceId;
     device.secret = deviceSecret;
 
     id identityTokenDetailsInfo = [storage objectForKey:ARTDeviceIdentityTokenKey];
     ARTDeviceIdentityTokenDetails *identityTokenDetails = [ARTDeviceIdentityTokenDetails unarchive:identityTokenDetailsInfo withLogger:logger];
     device->_identityTokenDetails = identityTokenDetails;
 
+    device.clientId = identityTokenDetails.clientId;
```
lawrence-forooghian, line 84: I don't think this matches the behaviour of the spec — my understanding of [RSH8a](https://sdk.ably.com/builds/ably/specification/main/features/#RSH8a), [RSH8b](https://sdk.ably.com/builds/ably/specification/main/features/#RSH8b), and [RSH8d](https://sdk.ably.com/builds/ably/specification/main/features/#RSH8d) is that `clientId` is meant to be persisted as a standalone attribute. (Again, I'd suggest checking how Android does it, too.) We also need to update the tests for RSH8d to make sure that the updated value gets persisted.

(For existing installations of the SDK we will however need to also populate this persisted value from the existing data, and I guess that we can use `identityTokenDetails.clientId` for that.)

Source/ARTPushActivationState.m
```diff
@@ -117,6 +117,9 @@ - (ARTPushActivationState *)transition:(ARTPushActivationEvent *)event {
     }
     else if ([event isKindOfClass:[ARTPushActivationEventCalledActivate class]]) {
         [self.machine registerForAPNS];
+#if TARGET_OS_IOS
+        [self.machine.rest setupLocalDevice];
```
lawrence-forooghian, line 121: This order doesn't match that of RSH3a2b and RSH3a2d.

Source/ARTLocalDevice.m
```diff
@@ -102,6 +95,33 @@ + (ARTLocalDevice *)load:(NSString *)clientId storage:(id<ARTDeviceStorage>)stor
     return device;
 }
 
+- (BOOL)setupDetailsWithClientId:(NSString *)clientId {
```
lawrence-forooghian, line 98: What's the return value of this method for? Nobody uses it.

Source/ARTRest.m
```diff
@@ -776,12 +775,20 @@ - (ARTLocalDevice *)deviceWithClientId_onlyCallOnDeviceAccessQueue:(NSString *)c
...
     static id device;
     if (sharedDeviceNeedsLoading_onlyAccessOnDeviceAccessQueue) {
-        device = [ARTLocalDevice load:clientId storage:self.storage logger:self.logger];
+        device = [ARTLocalDevice deviceWithStorage:self.storage logger:self.logger];
         sharedDeviceNeedsLoading_onlyAccessOnDeviceAccessQueue = NO;
     }
     return device;
 }
 
+- (void)setupLocalDevice {
```
lawrence-forooghian, line 784:
```suggestion
- (void)setupLocalDevice_nosync {
```

Source/ARTRest.m
```diff
@@ -789,6 +789,13 @@ - (void)setupLocalDevice {
     });
 }
 
+- (void)resetLocalDevice {
```
lawrence-forooghian, line 792:
```suggestion
- (void)resetLocalDevice_nosync {
```

lawrence-forooghian (commented):
Do the tests not need updating (or new ones adding) for the new behaviours that you've implemented? There are existing tests that claim to test RSH8b and RSH3g2a but presumably they're inadequate given that those spec points weren't implemented.

Source/ARTLocalDevice.m
maratal: Android [uses](https://github.com/ably/ably-java/blob/72992c943498ee2ebefb8acec303718ed7f35ae4/android/src/main/java/io/ably/lib/push/LocalDevice.java#L64) a separate attribute for `clientId`, but I would vote against it, how does it make any sense?

Source/ARTLocalDevice.m
maratal: This is for [RSH3a2a1](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3a2a1). I've avoided to make further changes in this PR due to some urgency for the customer, but in case of mismatched clientId I would generate error [earlier](https://github.com/ably/ably-cocoa/blob/6c6f0fbd12da95c1b47bc474f791fbd854ec3ff0/Source/ARTPushActivationState.m#L121) then it's implemented now in `validateAndSync` function. That's why the return value is needed.

Source/ARTLocalDevice.m
maratal: See response [here](https://github.com/ably/ably-cocoa/pull/1847/files#r1442138958)

Source/ARTRest.m
maratal: https://github.com/ably/ably-cocoa/pull/1847/commits/b7eda52452d386aa979e089d73eed2860a0784c3

Source/ARTRest.m
maratal: https://github.com/ably/ably-cocoa/pull/1847/commits/b7eda52452d386aa979e089d73eed2860a0784c3

Source/ARTPushActivationState.m
maratal: https://github.com/ably/ably-cocoa/pull/1847/commits/9277ab083ae1fc5aec060ba98632a4083986ce93

Source/ARTLocalDevice.m
maratal: Android uses null(s).

maratal:
> Do the tests not need updating (or new ones adding) for the new behaviours that you've implemented? There are existing tests that claim to test RSH8b and RSH3g2a but presumably they're inadequate given that those spec points weren't implemented.

https://github.com/ably/ably-cocoa/pull/1847/commits/d5a9d823799941a7a2bbd72df02def17683251e1

Source/ARTLocalDevice.m
lawrence-forooghian: What do you mean by "it doesn't make any sense"? Do you not agree that it's the specified behaviour? Bear in mind that the `LocalDevice`’s `clientId` doesn't necessarily match that of the currently saved `identityTokenDetails` (see e.g. [RSH8e](https://sdk.ably.com/builds/ably/specification/main/features/#RSH8e), where `LocalDevice.clientId` has been updated but the registration has not).

Source/ARTLocalDevice.m
lawrence-forooghian: But I don't understand what this new `clientId` check that you've introduced in the current PR is achieving?

Source/ARTLocalDevice.m
lawrence-forooghian: Yeah, I mean, there isn't a good answer to this situation; the spec is wrong. After this current PR is done, it might be worth you spending time making a PR for the spec changes necessary to fix it.

I think that the best option for now might be to ignore [RSH8b](https://sdk.ably.com/builds/ably/specification/main/features/#RSH8b) and to generate the new `LocalDevice.id` as soon as we clear the old one, so that the `LocalDevice` always has an `id`. I can't see any harm in doing this; what do you think?

Source/ARTLocalDevice.m
lawrence-forooghian: > We also need to update the tests for RSH8d to make sure that the updated value gets persisted.

(Same for [RSH8f](https://sdk.ably.com/builds/ably/specification/main/features/#RSH8f)).

Source/ARTLocalDevice.m
maratal: I think that temporarily not resetting `deviceId` in `resetDetails` and generate the new one for a new installation without persisting it might "do the trick". - https://github.com/ably/ably-cocoa/pull/1847/commits/094c9659b6c2318d132b62709a7e30083b78eadc and https://github.com/ably/ably-cocoa/pull/1847/commits/91183b9e4a1829e3d087cd23fdb9afca4fdc1410

Source/ARTLocalDevice.m
maratal: https://github.com/ably/ably-cocoa/pull/1847/commits/3098e3ae010a86aa37cf525af6b92f4a5b0ae079

Source/ARTLocalDevice.m
maratal: https://github.com/ably/ably-cocoa/pull/1847/commits/c64c56192644dfc2a865c2b41a411be72d2bfb71

Source/ARTLocalDevice.m
maratal: agree, better to add all of this when needed - https://github.com/ably/ably-cocoa/pull/1847/commits/c64c56192644dfc2a865c2b41a411be72d2bfb71

Source/ARTPushActivationState.m
maratal: Unresolved due to code changed - https://github.com/ably/ably-cocoa/pull/1847/commits/7159d42f255b9c593bc0f94f0b32eb2c57afee17

Source/ARTLocalDevice.m
lawrence-forooghian: The aim of my suggestion was to make sure that the `LocalDevice` instance always has a non-nil `id`. In your implementation, it's possible for it to have a nil `id` after the device details have been cleared per RSH3g2a, no?

lawrence-forooghian: But I don't think we want the `LocalDevice` to continue to have the same `id` after the device details have been cleared post-RSH3g2a, do we? Also, I think that someone looking at the code might not understand that we're relying on the `secret` having been cleared in order to generate a new device `id`.

I think it would better to, inside `resetDetails`, generate and persist a new `id`.

Source/ARTLocalDevice.m
```diff
@@ -75,21 +75,20 @@ + (ARTLocalDevice *)load:(NSString *)clientId storage:(id<ARTDeviceStorage>)stor
...
-    if (deviceId == nil || deviceSecret == nil) { // generate both at the same time
-        deviceId = [self generateId];
-        deviceSecret = [self generateSecret];
-        
-        [storage setObject:deviceId forKey:ARTDeviceIdKey];
-        [storage setSecret:deviceSecret forDevice:deviceId];
-    }
-    
-    device.id = deviceId;
+    device.id = deviceId ?: [self.class generateId]; // temporarily ignore RSH8a/b, see https://github.com/ably/ably-cocoa/pull/1847#discussion_r1441954284 thread
```
lawrence-forooghian, line 78: I think that it would be better to create an issue for this in https://github.com/ably/specification (with a proper description) and link to that.

Source/ARTLocalDevice.m
```diff
@@ -75,21 +75,20 @@ + (ARTLocalDevice *)load:(NSString *)clientId storage:(id<ARTDeviceStorage>)stor
...
     device.secret = deviceSecret;
 
     id identityTokenDetailsInfo = [storage objectForKey:ARTDeviceIdentityTokenKey];
     ARTDeviceIdentityTokenDetails *identityTokenDetails = [ARTDeviceIdentityTokenDetails unarchive:identityTokenDetailsInfo withLogger:logger];
     device->_identityTokenDetails = identityTokenDetails;
 
+    NSString *clientId = [storage objectForKey:ARTClientIdKey];
+    if (clientId == nil && identityTokenDetails.clientId != nil) {
```
lawrence-forooghian, line 86: This is not a specified behaviour — please could you add a comment explaining why we do it?

Source/ARTLocalDevice.m
lawrence-forooghian: Thanks — leaving this comment unresolved so I remember to look at the test changes (will look at the tests once implementation comments addressed).

Source/ARTPushActivationState.m
lawrence-forooghian: Would you mind explaining why this code moved into `validateAndSync`, please?

Source/ARTLocalDevice.m
lawrence-forooghian: Let's continue to generate `secret` at the point where the spec says that we should, yeah.

Source/ARTLocalDevice.m
maratal: https://github.com/ably/specification/issues/180

Source/ARTLocalDevice.m
maratal: What we have in this setup is `id` = `nil` until device activation which is inconsistent with what current property is (non nullable).

lawrence-forooghian: That's a good point. So in that case we need to restore some of the previous behaviour - populate the `id` when the `LocalDevice` instance is first accessed, right?

Source/ARTLocalDevice.m
maratal: https://github.com/ably/ably-cocoa/pull/1847/commits/e9cab44214014d599353bb8b740fd1c1e5ba77c6

Source/ARTLocalDevice.m
maratal: https://github.com/ably/ably-cocoa/pull/1847/commits/79e741b269e7ab7486292fc0acc09b5146c2badf

Still generating id/secret pair together, because once issue with nullable `id` resolved we can just delete extra `generateAndPersistPairOfDeviceIdAndSecret` calls, wdyt? Also it allows not to violate [RSH3a2b](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3a2b) which states that those values should be generated as a pair.

Source/ARTPushActivationState.m
maratal: This is to keep a proper call order `setupLocalDevice_nosync` - `registerForAPNS` and also call `setupLocalDevice_nosync` after `61002` error check. That's why I wanted that return value, turned out it can be solved by moving everything inside `validateAndSync`. Simple and doesn't break anything, why not then?

Source/ARTLocalDevice.m
maratal: https://github.com/ably/ably-cocoa/pull/1847/commits/79e741b269e7ab7486292fc0acc09b5146c2badf

Source/ARTLocalDevice.m
```diff
@@ -75,21 +83,25 @@ + (ARTLocalDevice *)load:(NSString *)clientId storage:(id<ARTDeviceStorage>)stor
...
-    device.id = deviceId;
-    device.secret = deviceSecret;
 
     id identityTokenDetailsInfo = [storage objectForKey:ARTDeviceIdentityTokenKey];
     ARTDeviceIdentityTokenDetails *identityTokenDetails = [ARTDeviceIdentityTokenDetails unarchive:identityTokenDetailsInfo withLogger:logger];
     device->_identityTokenDetails = identityTokenDetails;
 
+    NSString *clientId = [storage objectForKey:ARTClientIdKey];
+    if (clientId == nil && identityTokenDetails.clientId != nil) {
+        clientId = identityTokenDetails.clientId; // pickup some value that exists for the updated installations with already registered device
```
lawrence-forooghian, line 100: This is not very easy to understand.

```suggestion
        clientId = identityTokenDetails.clientId; // Older versions of the SDK did not persist clientId, so as a fallback when loading data persisted by these versions we use the clientId of the stored identity token
```

Source/ARTLocalDevice.m
lawrence-forooghian: Yep, sounds good.

lawrence-forooghian (approved):

PR #1851, opened 2024-01-07
Transition to Xcode 15.1

maratal:
On top of https://github.com/ably/ably-cocoa/pull/1840

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -22,8 +22,8 @@ end
 platform :ios do
 
   LANE_CONFIGS = [
-    LaneConfig.new(:test_iOS16_2, "Ably-iOS", ["iPhone 14 (16.2)"]),
-    LaneConfig.new(:test_tvOS16_1, "Ably-tvOS", ["Apple TV 4K (2nd generation) (16.1)"]),
+    LaneConfig.new(:test_iOS17_2, "Ably-iOS", ["iPhone 15"]),
+    LaneConfig.new(:test_tvOS17_2, "Ably-tvOS", ["Apple TV"]),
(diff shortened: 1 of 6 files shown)


fastlane/Fastfile
```diff
@@ -22,8 +22,8 @@ end
 platform :ios do
 
   LANE_CONFIGS = [
-    LaneConfig.new(:test_iOS16_2, "Ably-iOS", ["iPhone 14 (16.2)"]),
-    LaneConfig.new(:test_tvOS16_1, "Ably-tvOS", ["Apple TV 4K (2nd generation) (16.1)"]),
+    LaneConfig.new(:test_iOS17_2, "Ably-iOS", ["iPhone 15"]),
```
lawrence-forooghian, line 25: Why have you dropped the version numbers?

lawrence-forooghian (commented):
This isn't just "transition to Xcode 15.1", right? It's also changing the platform versions on which we test in CI. If so, we need to update the corresponding information in the `README`.

fastlane/Fastfile
maratal: I wanted to check what version it will select by default but somehow left it there. Sorry, fixed here - https://github.com/ably/ably-cocoa/pull/1851/commits/f443b48ae46d6dc0c2537d5ff6a87de7419a7d3b

maratal:
> This isn't just "transition to Xcode 15.1", right? It's also changing the platform versions on which we test in CI. If so, we need to update the corresponding information in the `README`.

done - https://github.com/ably/ably-cocoa/pull/1851/commits/695e8f61567930aef5263aaf24bd5253ac512511

lawrence-forooghian (approved):

PR #1869, opened 2024-02-05
[SDK-4076] Removed unnecessary checks for RTP1.

### LINKED ISSUE #1868: Better RTP1 complience.
About code
```
    else if ([self.presenceMap.members count] > 0 || [self.presenceMap.localMembers count] > 0) {
        if (!message.resumed) {
            // When an ATTACHED message is received without a HAS_PRESENCE flag and PresenceMap has existing members
            [self.presenceMap startSync];
            [self.presenceMap endSync];
            ARTLogDebug(self.logger, @"R:%p C:%p (%@) PresenceMap has been reset", _realtime, self, self.name);
        }
    }
```
We don't need check `if (!message.resumed)`

_Originally posted by @sacOO7 in https://github.com/ably/ably-cocoa/pull/1714#discussion_r1478085245_

maratal:
Closes #1868

based on [RTP1](https://sdk.ably.com/builds/ably/specification/main/features/#RTP1) we shouldn't check on message's `resume` flag when resetting presence map. Also members count check is not necessary either.

### DIFF
--- a/Source/ARTRealtimeChannel.m
+++ b/Source/ARTRealtimeChannel.m
@@ -731,13 +731,11 @@ dispatch_sync(_queue, ^{
     if (message.hasPresence) {
         [self.presenceMap startSync];
     }
-    else if ([self.presenceMap.members count] > 0 || [self.presenceMap.localMembers count] > 0) {
-        if (!message.resumed) {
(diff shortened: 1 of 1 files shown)


sacOO7 (approved):
lgtm

lawrence-forooghian (approved):

PR #1870, opened 2024-02-08
[SDK-4009] Refactoring presence

### LINKED ISSUE #1848: Refactor presence implementation and re-enable skipped presence tests.
Also, I would love someone to go through the current presence implementation and check if it's compatible with spec https://sdk.ably.com/builds/ably/specification/main/features/#RTP1.
You can refer to https://github.com/ably/ably-go/blob/14cd5a206e18571f1012f992397e1092c535919e/ably/realtime_presence.go file for more refined implementation in accordance with spec.

_Originally posted by @sacOO7 in https://github.com/ably/ably-cocoa/pull/1714#discussion_r1423840044_

### LINKED ISSUE #1875: Refactor presence implementation
See umbrella [issue](https://github.com/ably/ably-cocoa/issues/1848) for details.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4229) by [Unito](https://www.unito.io)

maratal:
Closes #1875

See commit messages on what goes where.


sacOO7 (changes requested):
I feel most of the presence specific code is still defined under `Source/ARTRealtimeChannel.m`. It should be moved under `Source/ARTRealtimePresence.m`. Some of those methods are `sync`, `onsync` and `onPresence`. Also, part of the code under `setAttached` should be moved to `Source/ARTRealtimePresence.m` as an individual method.

Source/ARTRealtimePresence.m
```diff
@@ -153,6 +154,7 @@ @implementation ARTRealtimePresenceInternal {
     __weak ARTRealtimeChannelInternal *_channel; // weak because channel owns self
     dispatch_queue_t _userQueue;
     NSMutableArray<ARTQueuedMessage *> *_pendingPresence;
+    ARTEventEmitter<ARTEvent *, ARTPresenceMessage *> *_presenceEventEmitter;
```
lawrence-forooghian, line 157: Is the word "presence" needed here still?

Test/Tests/RealtimeClientPresenceTests.swift
```diff
@@ -531,9 +531,9 @@ class RealtimeClientPresenceTests: XCTestCase {
         let channel = client.channels.get(test.uniqueChannelName())
 
         let listener = channel.presence.subscribe { _ in }!
-        XCTAssertEqual(channel.internal.presenceEventEmitter.anyListeners.count, 1)
+        XCTAssertEqual(channel.internal.presence.presenceEventEmitter.anyListeners.count, 1)
```
lawrence-forooghian, line 534: This change seems to be in the wrong commit.

Source/PrivateHeaders/Ably/ARTRealtimePresence+Private.h
```diff
@@ -1,9 +1,12 @@
+#import <Ably/ARTPresenceMap.h>
 #import <Ably/ARTRealtimePresence.h>
 #import <Ably/ARTRealtimeChannel+Private.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface ARTRealtimePresenceInternal : NSObject <ARTRealtimePresenceProtocol>
+@interface ARTRealtimePresenceInternal : NSObject <ARTRealtimePresenceProtocol, ARTPresenceMapDelegate>
+
+@property (readonly, nonatomic) ARTEventEmitter<ARTEvent *, ARTPresenceMessage *> *presenceEventEmitter;
```
lawrence-forooghian, line 10: This change seems to be in the wrong commit.

lawrence-forooghian (commented):
These changes alone don't seem to resolve #1848; the PR doesn’t "re-enable presence tests", nor “go through the current presence implementation and check if it's compatible with spec” (although I'm not really sure what @sacOO7 had in mind when he wrote the latter; was there some reason to believe that we are not conforming to the spec)?

Source/ARTRealtimeChannel.m
```diff
@@ -1220,20 +1217,10 @@ - (void)reenterLocalMembers {
     [self.presenceMap cleanUpAbsentMembers];
 }
 
-#pragma mark - ARTPresenceMapDelegate
-
 - (NSString *)connectionId {
```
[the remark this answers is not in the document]
maratal: yes, I've redeclared it [here](https://github.com/ably/ably-cocoa/blob/ac1152d2d9aefa798eb5f0a9b25455bc636922f4/Source/PrivateHeaders/Ably/ARTRealtimeChannel%2BPrivate.h#L31), bc it is used in tests.

Source/PrivateHeaders/Ably/ARTRealtimePresence+Private.h
maratal: 646f555a1e037aaa2a43ebe440343e3ff1396075

Source/ARTRealtimePresence.m
maratal: 646f555a1e037aaa2a43ebe440343e3ff1396075

maratal:
> These changes alone don't seem to resolve #1848; the PR doesn’t "re-enable presence tests", nor “go through the current presence implementation and check if it's compatible with spec” (although I'm not really sure what @sacOO7 had in mind when he wrote the latter; was there some reason to believe that we are not conforming to the spec)?

I've changed the issue to https://github.com/ably/ably-cocoa/issues/1875.

lawrence-forooghian:
Changes look good to me — will let @sacOO7 say what other changes he'd like to see as part of this refactor.

maratal:
> Changes look good to me — will let @sacOO7 say what other changes he'd like to see as part of this refactor.

I'm working on it - move all code with `self.presence...` into presence object.

Source/ARTRealtimePresence.m
```diff
@@ -622,4 +658,295 @@ - (void)failPendingPresence:(ARTStatus *)status {
...
+    else {
+        ARTLogDebug(self.logger, @"RT:%p C:%p (%@) PresenceMap sync is in progress", _realtime, _channel, _channel.name);
+    }
+
+    for (int i = 0; i < [message.presence count]; i++) {
+        ARTPresenceMessage *presence = [message.presence objectAtIndex:i];
+        if ([self add:presence]) {
+            [self broadcast:presence];
+        }
+    }
```
sacOO7, line 770:
```suggestion
    [self onMessage:message];
```

Source/ARTRealtimePresence.m
```diff
@@ -622,4 +658,295 @@ - (void)failPendingPresence:(ARTStatus *)status {
...
+
+        if ([self add:presence]) {
+            [self broadcast:presence];
+        }
+
+        ++i;
+    }
+}
+
+- (void)onSync:(ARTProtocolMessage *)message {
```
sacOO7, line 754: Can you please refer to https://github.com/ably/ably-go/blob/dd8236fc301eb078803e83efee7615f3ba133341/ably/realtime_presence.go#L276 to check if we are not missing anything here.
We have a separate method to calculate `syncserial` https://github.com/ably/ably-go/blob/dd8236fc301eb078803e83efee7615f3ba133341/ably/realtime_presence.go#L135.

sacOO7 (changes requested):

Source/ARTRealtimeChannel.m
```diff
@@ -1196,34 +1084,25 @@ - (void)startDecodeFailureRecoveryWithErrorInfo:(ARTErrorInfo *)error {
...
-    presence.action = ARTPresenceLeave;
-    presence.id = nil;
-    presence.timestamp = [NSDate date];
-    [self broadcastPresence:presence];
-    ARTLogDebug(self.logger, @"RT:%p C:%p (%@) member \"%@\" no longer present", _realtime, self, self.name, presence.memberKey);
-}
-
-- (void)map:(ARTPresenceMap *)map shouldReenterLocalMember:(ARTPresenceMessage *)presence {
-    [self.presence enterWithPresenceMessageId:presence.id clientId:presence.clientId data:presence.data callback:^(ARTErrorInfo *error) {
+- (void)reenterLocalMembers {
```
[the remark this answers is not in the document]
maratal: I need `^(ARTPresenceMessage *member, ARTErrorInfo *error)` callback in the channel class to emit channel events. `reenterLocalMembersWithMemberCallback` is already inside presence object.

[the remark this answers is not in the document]
sacOO7: You can pass `channel` instance to `presence` and then call error method from there, like we do it =>
https://github.com/ably/ably-go/blob/dd8236fc301eb078803e83efee7615f3ba133341/ably/realtime_presence.go#L158
This feels way more clean WDYT?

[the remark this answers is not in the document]
sacOO7: Also, code calling `reenterLocalMembers` will be already inside `presence` itself.  So, anyways we have to move this method there

Source/ARTRealtimePresence.m
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/1664da5c4d85fc6dce4e853f6081793acdd1966f

Source/ARTRealtimeChannel.m
```diff
@@ -733,15 +674,13 @@ - (void)setAttached:(ARTProtocolMessage *)message {
     }
```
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/1664da5c4d85fc6dce4e853f6081793acdd1966f

Source/ARTRealtimeChannel.m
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/1664da5c4d85fc6dce4e853f6081793acdd1966f

Source/ARTRealtimePresence.m
maratal: Yes, it's similar - you check serials[1] on emptiness, it's done [here](https://github.com/ably/ably-cocoa/blob/913659d50c46a064ef28fbde5e72ea45d09e1d00/Source/ARTRealtimeChannel.m#L676).

Source/ARTRealtimePresence.m
```diff
@@ -622,4 +660,325 @@ - (void)failPendingPresence:(ARTStatus *)status {
...
+    else {
+        // RTP1 - when an ATTACHED message is received without a HAS_PRESENCE flag, reset PresenceMap
+        [self startSync];
+        [self endSync];
+        ARTLogDebug(self.logger, @"R:%p C:%p (%@) PresenceMap has been reset", _realtime, self, _channel.name);
+    }
+}
+
+- (void)afterAttached:(ARTProtocolMessage *)message fromState:(ARTRealtimeChannelState)state {
+    if (state == ARTRealtimeChannelAttached) {
```
sacOO7, line 729: Is this a duplicate code here. We are already handling this before calling `beforeAttached` right?
Also, `if (state == ARTRealtimeChannelAttached) {` will always return `true`.
Can we please simplify this and remove `beforeAttached` and `afterAttached`. I don't think we need two separate methods.
We can keep just one `onAttached` like we do here 
https://github.com/ably/ably-go/blob/dd8236fc301eb078803e83efee7615f3ba133341/ably/realtime_presence.go#L164

Source/PrivateHeaders/Ably/ARTRealtimePresence+Private.h
```diff
@@ -5,17 +5,24 @@ NS_ASSUME_NONNULL_BEGIN
...
+- (void)sync:(nullable ARTCallback)callback;
 
-/*
- * Helper method for enforcing RTP17g, where in addition to clientId and data, message id is required.
- */
-- (void)enterWithPresenceMessageId:(NSString *)messageId clientId:(NSString *)clientId data:(id)data callback:(ARTCallback)cb;
+- (void)onMessage:(ARTProtocolMessage *)message;
+- (void)onSync:(ARTProtocolMessage *)message;
+
+- (void)beforeAttached:(ARTProtocolMessage *)message;
```
sacOO7, line 24: We don't need two methods. `afterAttached` can be removed. Can we just keep one method `onAttach` ->https://github.com/ably/ably-go/blob/dd8236fc301eb078803e83efee7615f3ba133341/ably/realtime_presence.go#L164

sacOO7 (changes requested):

Source/PrivateHeaders/Ably/ARTRealtimePresence+Private.h
maratal: you are right re two methods, but I'm still gonna need the previous state value, since checking the new one in `onAttached` is pointless (since it is known that it's ATTACHED already as per method name) - https://github.com/ably/ably-cocoa/pull/1870/commits/a8d40623b9b6e1d8afba752f91084c0e1f24201f

Source/ARTRealtimePresence.m
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/a8d40623b9b6e1d8afba752f91084c0e1f24201f

Source/PrivateHeaders/Ably/ARTRealtimePresence+Private.h
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/bee8c1eb61cf094eb8cd88491c70c713776ca9f8

Source/ARTRealtimePresence.m
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/bee8c1eb61cf094eb8cd88491c70c713776ca9f8

Source/ARTRealtimePresence.m
```diff
@@ -622,4 +660,315 @@ - (void)failPendingPresence:(ARTStatus *)status {
...
 
+- (void)broadcast:(ARTPresenceMessage *)pm {
+    [_eventEmitter emit:[ARTEvent newWithPresenceAction:pm.action] with:pm];
+}
+
+- (void)sync {
+    [self sync:nil];
+}
+
+- (void)sync:(ARTCallback)callback {
```
sacOO7, line 671: We don't have a sync method on `realtime presence` as per [class Presence => protocol 2.0](https://sdk.ably.com/builds/ably/specification/main/features/#TO3l10:~:text=class%20RealtimePresence). Also, `_syncChannelSerial` as a value is not valid anymore. I have removed implementation for `sync` in ably-java. You can check https://github.com/ably/ably-java/blob/0af4d4e465d777462e92845078f0cd966049541b/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L371

Source/ARTRealtimePresence.m
```diff
@@ -622,4 +660,315 @@ - (void)failPendingPresence:(ARTStatus *)status {
...
+    }
+    else {
+        // RTP1 - when an ATTACHED message is received without a HAS_PRESENCE flag, reset PresenceMap
+        [self startSync];
+        [self endSync];
+        ARTLogDebug(self.logger, @"R:%p C:%p (%@) PresenceMap has been reset", _realtime, self, _channel.name);
+    }
+    [self sendPendingPresence];
+    [self reenterLocalMembers]; // RTP17i
+}
```
sacOO7, line 728:
```suggestion
- (void)onAttached:(ARTProtocolMessage *)message {
    [self startSync];
    if !(message.hasPresence) {
        // RTP1 - when an ATTACHED message is received without a HAS_PRESENCE flag, reset PresenceMap
        [self endSync];
        ARTLogDebug(self.logger, @"R:%p C:%p (%@) PresenceMap has been reset", _realtime, self, _channel.name);
    }
    [self sendPendingPresence];
    [self reenterLocalMembers]; // RTP17i
}
```

sacOO7 (changes requested):

Source/ARTRealtimePresence.m
```diff
@@ -622,4 +660,315 @@ - (void)failPendingPresence:(ARTStatus *)status {
     }
 }
 
+- (void)broadcast:(ARTPresenceMessage *)pm {
+    [_eventEmitter emit:[ARTEvent newWithPresenceAction:pm.action] with:pm];
+}
+
+- (void)sync {
```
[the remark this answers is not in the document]
maratal: It's still used in [RTP19](https://sdk.ably.com/builds/ably/specification/main/features/#RTP19):
"..._send a SYNC message with the channel attribute populated with the current channel which will trigger a server initiated SYNC_."

Also is used in a test for `RTP17b` which is skipped though.

Source/ARTRealtimePresence.m
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/e55050718b5e97815abe14db279f561571a9dcdb

Source/ARTRealtimePresence.m
maratal: https://github.com/ably/ably-cocoa/pull/1870#discussion_r1510398105

Source/ARTRealtimePresence.m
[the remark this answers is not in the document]
sacOO7: I don't think it says what you think it says. It's telling to write a local test where we mock server sent `SYNC` message and inject it client side as a part of test.

Source/ARTRealtimePresence.m
sacOO7: I don't think it says what you think it says. It's telling to write a local test where we mock server sent `SYNC` message and inject it client side as a part of test.

Source/ARTRealtimePresence.m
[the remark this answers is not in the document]
sacOO7: That said, I need to check with @SimonWoolf regarding this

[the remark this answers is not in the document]
sacOO7: Adding @SimonWoolf as a reviewer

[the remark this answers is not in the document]
sacOO7: Okay, got a reply, This is not a public method and should only be used for testing. Can we not make it part of the test suite instead of the codebase. I strongly feel, this is going to be misused if we keep it part of the codebase.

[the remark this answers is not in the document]
sacOO7: Seems we can do this using https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html.
You can move this code to the test suite, since it;s not a part of the standard spec.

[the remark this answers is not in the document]
sacOO7: If this is a public method, you can add a warning message. I have done the same here
https://github.com/ably/ably-java/pull/989/commits/23513c643b6bd90fd852daeb238f8f948d41c887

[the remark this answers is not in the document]
sacOO7: If this is a public method, add deprecation notice `sync() method is intended only for internal testing purpose as per RTP19`. If not, then move it under test-suite.

Source/ARTRealtimePresence.m
maratal: Duplicate of https://github.com/ably/ably-cocoa/pull/1870/files#r1509076164

Source/ARTRealtimePresence.m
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/35f43649a5aba361a3fdc7c8cc2dbc35c2247931

Source/ARTRealtimePresence.m
```diff
@@ -622,4 +660,263 @@ - (void)failPendingPresence:(ARTStatus *)status {
...
+    if (!self.syncInProgress) {
+        [self startSync];
+    }
+    else {
+        ARTLogDebug(self.logger, @"RT:%p C:%p (%@) PresenceMap sync is in progress", _realtime, _channel, _channel.name);
+    }
+
+    [self onMessage:message];
+
+    if ([_channel isLastChannelSerial:message.channelSerial]) {
```
sacOO7, line 720: Can you move `isLastChannelSerial` method to `Source/ARTRealtimePresence.m`. It doesn't make sense to keep it as a part of `realtimechannel.m` file.

sacOO7: Also, `isLastChannelSerial` should be redefined as
```
- (bool)isLastChannelSerial:(NSString *)channelSerial {
    if ([channelSerial sEqualToString:@""]) {
        return true;
    }
    NSArray * a = [channelSerial componentsSeparatedByString:@":"];
    if([a count] >1 && ![[a objectAtIndex:1] isEqualToString:@""] ) {
        return false;
    }
    return true;
}
```

Source/ARTRealtimePresence.m
```diff
@@ -149,20 +153,54 @@ @interface ARTRealtimePresenceInternal ()
...
 @implementation ARTRealtimePresenceInternal {
     __weak ARTRealtimeChannelInternal *_channel; // weak because channel owns self
+    __weak ARTRealtimeInternal *_realtime;
     dispatch_queue_t _userQueue;
     NSMutableArray<ARTQueuedMessage *> *_pendingPresence;
+    ARTEventEmitter<ARTEvent *, ARTPresenceMessage *> *_eventEmitter;
+    ARTDataEncoder *_dataEncoder;
+    
+    int64_t _syncMsgSerial;
+    NSString *_syncChannelSerial;
```
sacOO7, line 179:
```suggestion
```
Not used anywhere, so should be removed. 
`_syncChannelSerial` is only used for `sync` method moved to test utilities, so can be removed safely.

sacOO7 (changes requested):
Few more refactorings to be done

Source/ARTRealtimePresence.m
sacOO7: This is as per spec RTP18b, RTP18c

sacOO7: You should check https://github.com/ably/ably-go/blob/dd8236fc301eb078803e83efee7615f3ba133341/ably/realtime_presence.go#L286

Source/ARTRealtimePresence.m
```diff
@@ -622,4 +660,263 @@ - (void)failPendingPresence:(ARTStatus *)status {
...
+                
+                ARTLogWarn(self.logger, @"RT:%p C:%p (%@) Re-entering member \"%@\" is failed with code %ld (%@)", self->_realtime, self->_channel, self->_channel.name, member.memberKey, (long)error.code, error.message);
+            }
+            else {
+                ARTLogDebug(self.logger, @"RT:%p C:%p (%@) re-entered local member \"%@\"", self->_realtime, self->_channel, self->_channel.name, member.memberKey);
+            }
+        }];
+        ARTLogDebug(self.logger, @"RT:%p C:%p (%@) re-entering local member \"%@\"", _realtime, _channel, _channel.name, member.memberKey);
+    }
+    [self cleanUpAbsentMembers];
```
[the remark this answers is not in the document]
sacOO7: Also, as per spec, cleaning up should only happen at the end of the sync.

Source/ARTRealtimePresence.m
```diff
@@ -622,4 +660,263 @@ - (void)failPendingPresence:(ARTStatus *)status {
...
+        }
+
+        if (!presence.timestamp) {
+            presence.timestamp = message.timestamp;
+        }
+
+        if (!presence.id) {
+            presence.id = [NSString stringWithFormat:@"%@:%d", message.id, i];
+        }
+
```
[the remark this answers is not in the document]
sacOO7: This is as per sepc TM2a, TM2c, TM2f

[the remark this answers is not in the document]
sacOO7: You also need to add the same check right below [ARTRealtimeChannel.m](https://github.com/ably/ably-cocoa/blob/5971d7c434305e42d04697669f58ca549fb65256/Source/ARTRealtimeChannel.m#L870-L875) -> https://github.com/ably/ably-cocoa/blob/5971d7c434305e42d04697669f58ca549fb65256/Source/ARTRealtimeChannel.m#L870-L875

sacOO7 (changes requested):

Source/ARTRealtimePresence.m
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/1d2bec63b7bd25a6c9fbf890b1b8e0512711b686

Source/ARTRealtimePresence.m
```diff
@@ -622,4 +660,263 @@ - (void)failPendingPresence:(ARTStatus *)status {
...
+            [self broadcast:presence];
+        }
+
+        ++i;
+    }
+}
+
+- (void)onSync:(ARTProtocolMessage *)message {
+    _syncMsgSerial = [message.msgSerial longLongValue];
+    _syncChannelSerial = message.channelSerial;
```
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/1d2bec63b7bd25a6c9fbf890b1b8e0512711b686

Source/ARTRealtimePresence.m
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/51a00d2dcaef813c06ac9cfe6d3c005285a70ed7

Source/ARTRealtimePresence.m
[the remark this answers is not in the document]
maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/a5a7679bf9e2d16777c1197dfc8cd8367b5e2fc9

Source/ARTRealtimePresence.m
```diff
@@ -415,7 +451,7 @@ - (BOOL)syncComplete {
 }
 
 - (BOOL)syncComplete_nosync {
-    return _channel.presenceMap.syncComplete;
+    return !(_syncState == ARTPresenceSyncInitialized || _syncState == ARTPresenceSyncStarted);
```
sacOO7, line 454: should be simplified as 
```suggestion
    return _syncState == ARTPresenceSyncEnded && _syncState != ARTPresenceSyncStarted);
```

sacOO7: This is as per https://github.com/ably/ably-dotnet/blob/f7b18f4902e42e90697e7fd35021025574edf650/src/IO.Ably.Shared/Realtime/Presence.cs#L44C50-L44C63 and
https://github.com/ably/ably-go/blob/dd8236fc301eb078803e83efee7615f3ba133341/ably/realtime_presence.go#L192

sacOO7 (changes requested):

Source/ARTRealtimePresence.m
maratal: I don't think this expression makes sense, because it's a single variable which if it's equal to `ARTPresenceSyncEnded` it's obviously already not `ARTPresenceSyncStarted`. For ably-go it's two different variables and in that case it does make sense to compare like that.

sacOO7: Then let's make it 
```
return _syncState == ARTPresenceSyncEnded
```

sacOO7: I think current expression evaluates to 
```
return _syncState == ARTPresenceSyncEnded || _syncState == ARTPresenceSyncFailed
```
We don't have `ARTPresenceSyncFailed` state in other SDKs though

maratal: Is it ok that in this case the `ARTPresenceSyncFailed` state will be ignored? I mean `syncComplete` is true when in failed state for the old logic. It will be false after this proposal.

maratal: https://github.com/ably/ably-cocoa/pull/1870/commits/cb04f96f05d4ac5205a7b243a96583d3a2f11f75

sacOO7 (approved):
LGTM

PR #1872, opened 2024-02-11
Run test__047 in a loop

maratal:


maratal:
See https://github.com/ably/ably-cocoa/pull/1867

PR #1877, opened 2024-02-18
Run test__047 in a loop (4)

maratal:
Polling instead of a timeout.

[Results](https://test-observability.herokuapp.com/repos/ably/ably-cocoa/uploads?branches[]=1859-test__047-in-loop-poll)


maratal:
see https://github.com/ably/ably-cocoa/pull/1870

PR #1879, opened 2024-02-22
[ECO-4234] Updates Push Example to include Channels

umair-ably:
https://github.com/ably/ably-cocoa/assets/157486912/3d63fc98-a030-43c0-8e77-488091a5d576

- Users can now subscribe to 2 predefined example channels to test Channels based push. 
- Updated the UX to be more in line with Ably's Design System.


maratal:
Looks really cool, thanks @umair-ably 
A couple of suggestions:
- "Subscribe" buttons might want to have more friendly captions.
- "Close" icon on a subscription button should be in a different state (for unsubscribe).
- After subscription fails it should show an error. I think it's better to introduce a dedicated control for this (a TextView in the bottom for example), where all the diagnostics info is displayed. Alerts are kinda headache in the SwiftUI (you can do this in a different PR later).

Simulator Screenshot - iPhone 15 Pro Max - 2024-03-07 at 18 20 59

Examples/AblyPush/AblyPushExample/ContentView.swift
```diff
@@ -62,11 +91,78 @@ struct ContentView: View {
...
+                .fixedSize(horizontal: false, vertical: true)
+                Text("Channels Push")
+                    .fontWeight(.bold)
+                    .padding(.top)
+
+                HStack {
+                    PushButton(
+                        isButtonEnabled: ablyHelper.isPushActivated,
+                        systemImage: ablyHelper.isSubscribedToExampleChannel1
+                        ? "checkmark.circle.fill"
```
maratal, line 114: Should be "xmark.circle.fill" in case of subscribe.

Examples/AblyPush/AblyPushExample/Buttons/PushButton.swift
```diff
@@ -0,0 +1,52 @@
...
+    
+    let activeColor = Color(red: 1.00, green: 0.33, blue: 0.09)
+    let inActiveColor = Color(red: 0.00, green: 0.56, blue: 0.02)
+    
+    init(
+        isButtonEnabled: Bool,
+        systemImage: String,
+        activeTitle: String,
+        inactiveTitle: String,
+        isActive: Binding<Bool>,
```
maratal, line 19: Is binding needed here? `isButtonEnabled` works in similar way but without binding.

maratal:
I think sending by `clientId` also should be added to this example. It's the [same code](https://github.com/ably/ably-cocoa/blob/5971d7c434305e42d04697669f58ca549fb65256/Source/include/Ably/ARTPushAdmin.h#L23), but you set `clientId` instead of `deviceId` in the recipient dictionary.

Examples/AblyPush/AblyPushExample/Buttons/PushButton.swift
umair-ably: It's required as it will only update the state of the button once the `Published` property has changed from the caller class. This ensures the button doesn't just toggle states without the action actually succeeding.

umair-ably:
> I think sending by `clientId` also should be added to this example. It's the [same code](https://github.com/ably/ably-cocoa/blob/5971d7c434305e42d04697669f58ca549fb65256/Source/include/Ably/ARTPushAdmin.h#L23), but you set `clientId` instead of `deviceId` in the recipient dictionary.

The readme already covers this use-case using the Push Inspector. When running on a single device/simulator, push via client or device behave the same - you only really notice the difference if you had several simulators running. Could be something worth adding as a future improvement, but not a pressing improvement for the key things this example showcases imo

umair-ably:
> Looks really cool, thanks @umair-ably A couple of suggestions:
> 
> * "Subscribe" buttons might want to have more friendly captions.
> * "Close" icon on a subscription button should be in a different state (for unsubscribe).
> * After subscription fails it should show an error. I think it's better to introduce a dedicated control for this (a TextView in the bottom for example), where all the diagnostics info is displayed. Alerts are kinda headache in the SwiftUI (you can do this in a different PR later).
> 
> 

- I left the Subscribe captions more "code-based" as that is the names the readme walks the user through when setting up the channel capabilities in the dashboard
- Good spot! I've updated this. 
- Agreed, error handling in general is pretty naff in this. Will look to update in a future PR.

jamienewcomb:
Hey team , unless there is anything generally wrong this is a vast improvement on what we had before so lets work on this idea and get it landed.

any further enhancements can be implemented as required

cc @lawrence-forooghian @umair-ably @maratal

Examples/AblyPush/AblyPushExample/ContentView.swift
maratal: Usually we put a comment with the commit link under the thread and the one who asked should resolve.

https://github.com/ably/ably-cocoa/pull/1879/commits/6a68d68f43de37cc0285fd4881e2910ba1818c39

Examples/AblyPush/AblyPushExample/Buttons/PushButton.swift
maratal: Couldn't you elaborate? I've checked one more time and it works exactly the same without binding. I think we shouldn't confuse users with a sample code. If it works without this complication then it should be removed.

umair-ably: Ahh my bad - I misunderstood what you were trying to say. I assumed you were for the removal of `isActive` as opposed to changing the type of it to just `Bool` instead of `Binding<Bool>` 😅 

All sorted in [240e927](https://github.com/ably/ably-cocoa/pull/1879/commits/240e927c223ed28d866cb273bfbf838560cecee2)

maratal (approved):
LGTM

PR #1894, opened 2024-03-26
[ECO-4715] Addresses push notification edge cases

umair-ably:
- clear partial state when CalledDeactivate come to NotActive state (RSH3a1c & RSH3a1d)
- ignore errors with 401 status code and 40005 code (invalid credentials) - (RSH3d2c1)

Will implement RSH3d2b (use `deviceIdentityToken` to perform request) later. The issue for Bank Jago is fixed with the implementation of the spec items above. RSH3d2b just ensures there are no stale/out of sync states between client and server, but the actual push activation/deactivation states are not affected.


ttypic (approved):
LGTM

PR #1898, opened 2024-03-31
Addressing workaround for RTN20c

### LINKED ISSUE #1713: Address workaround for RTN20c
The workaround was needed, because alternative solutions to the [RTN20c](https://sdk.ably.com/builds/ably/specification/main/features/#RTN20c) spec were causing a massive disruption of the whole test suit and risks related to significant changes in the library core for seemingly minor addition to functionality.

Also, there is no certain way to detect whether reachability callback was called because you've just set it up, or because it really has detected reachability of the host. 

By introducing a small delay before deciding whether a reconnection attempt should be performed we address both of these problems. Chances that the reachability callback will be called within that small delay are negligible (in this case library will behave the old way), hence the workaround performs what specs describe with a high accuracy.

To remove this workaround the actual code for establishing the connection to the Realtime should be moved to the reachability callback itself. But taking into consideration that behavior of this system API might change, this is probably not the best idea either. So the implementation strategy for RTN20c remains open.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/ECO-4541) by [Unito](https://www.unito.io)

maratal:
Closes #1713 

This PR reverts the work from [this](https://github.com/ably/ably-cocoa/pull/1703) PR and the fix for it from [this](https://github.com/ably/ably-cocoa/pull/1883) PR.

The idea is to ignore the first report of reachability with status "reachable" since it's not needed* - if the host is reachable, then the connection will proceed as expected. If the first call reports the host as "unreachable", then the realtime will fall into disconnected/suspended. And when the system will again report the host as reachable, then this call will not be ignored since the previous one was "unreachable" (in oppose to the first call where the previous one is "unknown").

* - the system always calls reachability callback immediately after the callback is set by the user (to initialize the user's state).

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -300,7 +300,7 @@ const NSTimeInterval _reachabilityReconnectionAttemptThreshold = 0.1;
         ARTLogDebug(self.logger, @"RS:%p halt current connection and reconnect with %@", self.rest, tokenDetails);
         [self abortAndReleaseTransport:[ARTStatus state:ARTStateOk]];
         [self setTransportWithResumeKey:self->_transport.resumeKey];
-        [self connectTransportWithToken:tokenDetails.token orKey:nil];
+        [self->_transport connectWithToken:tokenDetails.token];
         [self cancelAllPendingAuthorizations];
         waitForResponse();
     };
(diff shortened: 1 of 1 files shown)


umair-ably (approved):
LGTM

PR #1900, opened 2024-04-02
[ECO-4723] Adds required privacy manifest prior to May deadline

umair-ably:
Jira [ticket](https://ably.atlassian.net/browse/ECO-4723) holds majority of the information including the urgency of this change.

From a review of the Apple guidelines around this privacy manifest, I believe we need to make clear we collect or use the following (which this PR covers):

- NSUserDefaults (which we use to write and read data solely from this app - `CA92.1` and to provide a wrapper in our SDK which in turn uses NSUserDefaults - `C56D.1`)
- User Data and Device ID (we can and do identify individual clients e.g. by company name within Snowflake for analytics. I've marked we do this for tracking and it is linked to the user)
- ~~Crash Data (this is marked as not being used for tracking (as User Data is more useful for this) but it is still linked to the user)~~
- User Data and Device Data are both marked as core to app functionality which fits Apple's definition here:

> Such as to authenticate the user, enable features, prevent fraud, implement security measures, ensure server up-time, minimize app crashes, improve scalability and performance, or perform customer support.

Full list of possible additions to the Privacy Manifest can be found [here](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files). Dig into `NSPrivacyCollectedDataTypes` and `NSPrivacyAccessedAPITypes` to see the required level of detail.

Please comment if you think additional categories need adding to our manifest.

### DIFF
--- a/Ably.xcodeproj/project.pbxproj
+++ b/Ably.xcodeproj/project.pbxproj
@@ -281,6 +281,10 @@
 		56190954238C3D3200A862A6 /* CryptoTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 56190953238C3D3200A862A6 /* CryptoTest.m */; };
 		56190955238C3D3200A862A6 /* CryptoTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 56190953238C3D3200A862A6 /* CryptoTest.m */; };
 		56190956238C3D3200A862A6 /* CryptoTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 56190953238C3D3200A862A6 /* CryptoTest.m */; };
+		80E519C72BBCBE21006545B4 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 80E519C62BBCBE20006545B4 /* PrivacyInfo.xcprivacy */; };
+		80E519C82BBCBE21006545B4 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 80E519C62BBCBE20006545B4 /* PrivacyInfo.xcprivacy */; };
+		80E519C92BBCBE21006545B4 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 80E519C62BBCBE20006545B4 /* PrivacyInfo.xcprivacy */; };
+		80E519CA2BBCBE21006545B4 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 80E519C62BBCBE20006545B4 /* PrivacyInfo.xcprivacy */; };
 		841134782722205400CFA837 /* ARTArchiveTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 841134772722205400CFA837 /* ARTArchiveTests.m */; };
 		8412FDE72661AC37001FE9E6 /* AblyDeltaCodec.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 8412FDE12661AC37001FE9E6 /* AblyDeltaCodec.xcframework */; };
 		8412FDED2661AC37001FE9E6 /* msgpack.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 8412FDE32661AC37001FE9E6 /* msgpack.xcframework */; };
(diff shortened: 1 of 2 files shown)


Ably.xcodeproj/project.pbxproj
```diff
@@ -1135,7 +1139,6 @@
 		21113B5829DCA4C700652C86 /* DataGatherer.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DataGatherer.swift; sourceTree = "<group>"; };
 		21113B5E29DDDDD000652C86 /* LogAdapterTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LogAdapterTests.swift; sourceTree = "<group>"; };
 		21113B6229DDF7E800652C86 /* ARTInternalLogTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = ARTInternalLogTests.m; sourceTree = "<group>"; };
-		21113B5829DCA4C700652C86 /* DataGatherer.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DataGatherer.swift; sourceTree = "<group>"; };
```
maratal, line 1138: Is it by accident here?

maratal (commented):
1. AFAIK `ably-cocoa` doesn't use any crash reporting tool (it [used](https://github.com/ably/ably-cocoa/pull/1073) some in the past tho).
2. `C56D.1` is very confusing, so I suggest double check it before inclusion:
"_Information accessed for this reason, or any derived information, may not be used for your third-party SDK’s own purposes or sent off-device by your third-party SDK_."
We DO use generated device's `id` and `secret` and for sure are sending those "off-device" for "your third-party SDK’s own purposes".
3. I would replace `User-ID` with `Device-ID` as it's literally what we are using.

umair-ably:
> 1. AFAIK `ably-cocoa` doesn't use any crash reporting tool (it [used](https://github.com/ably/ably-cocoa/pull/1073) some in the past tho).
> 2. `C56D.1` is very confusing, so I suggest double check it before inclusion:
>    "_Information accessed for this reason, or any derived information, may not be used for your third-party SDK’s own purposes or sent off-device by your third-party SDK_."
>    We DO use generated device's `id` and `secret` and for sure are sending those "off-device" for "your third-party SDK’s own purposes".
> 3. I would replace `User-ID` with `Device-ID` as it's literally what we are using.



> 1. AFAIK `ably-cocoa` doesn't use any crash reporting tool (it [used](https://github.com/ably/ably-cocoa/pull/1073) some in the past tho).
> 2. `C56D.1` is very confusing, so I suggest double check it before inclusion:
>    "_Information accessed for this reason, or any derived information, may not be used for your third-party SDK’s own purposes or sent off-device by your third-party SDK_."
>    We DO use generated device's `id` and `secret` and for sure are sending those "off-device" for "your third-party SDK’s own purposes".
> 3. I would replace `User-ID` with `Device-ID` as it's literally what we are using.

1. Cool, happy to remove this and add again once we reintroduce crash reporting at some point.
2. `C56D.1` is strictly in reference to `NSUserDefaults` and not other Device/User tracking we do. We have `setObject` and `objectForKey` functions which are solely wrappers on `NSUserDefaults` which wholly meets the definition of `C56D.1`. Our data usage would be covered by Device-ID/User-ID tracking.
3. I'd argue this should be in addition to, instead of replacing. We also track User's in the way of business/client names. I assume we use `Device-ID` for more granular tracking. What do you think?

Ably.xcodeproj/project.pbxproj
```diff
@@ -281,6 +281,10 @@
 		56190954238C3D3200A862A6 /* CryptoTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 56190953238C3D3200A862A6 /* CryptoTest.m */; };
 		56190955238C3D3200A862A6 /* CryptoTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 56190953238C3D3200A862A6 /* CryptoTest.m */; };
 		56190956238C3D3200A862A6 /* CryptoTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 56190953238C3D3200A862A6 /* CryptoTest.m */; };
+		80E519C72BBCBE21006545B4 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 80E519C62BBCBE20006545B4 /* PrivacyInfo.xcprivacy */; };
+		80E519C82BBCBE21006545B4 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 80E519C62BBCBE20006545B4 /* PrivacyInfo.xcprivacy */; };
+		80E519C92BBCBE21006545B4 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 80E519C62BBCBE20006545B4 /* PrivacyInfo.xcprivacy */; };
+		80E519CA2BBCBE21006545B4 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 80E519C62BBCBE20006545B4 /* PrivacyInfo.xcprivacy */; };
```
maratal, line 287: I think one of these is by accident.

umair-ably: one is for the ably-soaktest app, not sure if it's used but i don't think there's any harm in leaving that target assigned

maratal:
>We also track User's in the way of business/client names

By default requests contain [this](https://github.com/ably/ably-cocoa/blob/35427f3120e571f5af922f68da3f99d65e668612/Test/Tests/ClientInformationTests.swift#L28) information. Users might add something else with `agentIdentifierWithAdditionalAgents:` method, but it marked explicitly as "This method should only be used by Ably-authored SDKs". `ClientOptions.clientId` may contain anything and I'm not sure whether we should be attributed for this.

Ably.xcodeproj/project.pbxproj
maratal: It will not harm, but why not to keep things clean, right?

umair-ably:
> > We also track User's in the way of business/client names
> 
> By default requests contain [this](https://github.com/ably/ably-cocoa/blob/35427f3120e571f5af922f68da3f99d65e668612/Test/Tests/ClientInformationTests.swift#L28) information. Users might add something else with `agentIdentifierWithAdditionalAgents:` method, but it marked explicitly as "This method should only be used by Ably-authored SDKs". `ClientOptions.clientId` may contain anything and I'm not sure whether we should be attributed for this.

Can we remove User-ID then?

maratal:
> > > We also track User's in the way of business/client names
> > 
> > 
> > By default requests contain [this](https://github.com/ably/ably-cocoa/blob/35427f3120e571f5af922f68da3f99d65e668612/Test/Tests/ClientInformationTests.swift#L28) information. Users might add something else with `agentIdentifierWithAdditionalAgents:` method, but it marked explicitly as "This method should only be used by Ably-authored SDKs". `ClientOptions.clientId` may contain anything and I'm not sure whether we should be attributed for this.
> 
> Can we remove User-ID then?

I think yes.

umair-ably:
> > > > We also track User's in the way of business/client names
> > > 
> > > 
> > > By default requests contain [this](https://github.com/ably/ably-cocoa/blob/35427f3120e571f5af922f68da3f99d65e668612/Test/Tests/ClientInformationTests.swift#L28) information. Users might add something else with `agentIdentifierWithAdditionalAgents:` method, but it marked explicitly as "This method should only be used by Ably-authored SDKs". `ClientOptions.clientId` may contain anything and I'm not sure whether we should be attributed for this.
> > 
> > 
> > Can we remove User-ID then?
> 
> I think yes.

Sure thing. Done in [#0fae06](https://github.com/ably/ably-cocoa/pull/1900/commits/0fae0618101eadcb31198e9347e7037598166d06)

Ably.xcodeproj/project.pbxproj
umair-ably: Done in [#0fae06](https://github.com/ably/ably-cocoa/pull/1900/commits/0fae0618101eadcb31198e9347e7037598166d06)

maratal (approved):
LGTM

PR #1906, opened 2024-04-15
Release V1.2.30 

umair-ably:


maratal (approved):
Should go in two commits, but I think it's debatable. Also I can see this release on the cocoapods (received an email), is that by accident?

umair-ably:
> Should go in two commits, but I think it's debatable. Also I can see this release on the cocoapods (received an email), is that by accident?

We should be releasing to cocoapods, right?

PR #1928, opened 2024-06-02
[ECO-4816] Refactor for RTP2f test.

### LINKED ISSUE #1927: test__050__Presence__PresenceMap__if_a_SYNC_is_in_progress... fails a lot
test__050__Presence__PresenceMap__if_a_SYNC_is_in_progress__then_when_a_presence_message_with_an_action_of_LEAVE_arrives__it_should_be_stored_in_the_presence_map_with_the_action_set_to_ABSENT doesn't always have `syncInProgress` true after attach and fails.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4816) by [Unito](https://www.unito.io)

maratal:
Closes #1927 

This test was failing because of race condition - `channel.internal.presence.syncInProgress` is not thread safe. So instead of checking it unreliably I've moved injecting code into the hook for `endSync` to guarantee injecting before the sync ends and checking whether it was properly added right after.

### DIFF
--- a/Test/Tests/RealtimeClientPresenceTests.swift
+++ b/Test/Tests/RealtimeClientPresenceTests.swift
@@ -1833,35 +1833,37 @@ class RealtimeClientPresenceTests: XCTestCase {
             fail("TestProxyTransport is not set"); return
         }
         
+        var leaveMessage: ARTProtocolMessage!
+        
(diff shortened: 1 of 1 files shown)


maratal:
Related issue - https://github.com/ably/ably-cocoa/issues/1929

umair-ably (approved):

Issue #1938, opened 2024-06-29
Invalid RTL4i usage


maratal:
`performTransitionToState:` calls `_attach` and referring `RTL4i` for it. This is wrong, `RTL4i` is talking about the continuation of the explicit attach, not implicit upon `CONNECTED`.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4857) by [Unito](https://www.unito.io)

PR #1941, opened 2024-06-29
[ECO-4858] TO3g compliance

### LINKED ISSUE #1936: `shouldQueueEvents` should not check `options.queueMessages`
> Why is all of this still gated under a condition that fails if `options.queueMessages` is false? This is [RTL3](https://sdk.ably.com/builds/ably/specification/main/features/#RTL3), right? That spec item doesn't mention anything about queueMessages, what do they have to do with each other?

_Originally posted by @SimonWoolf in https://github.com/ably/ably-cocoa/pull/1005#discussion_r999661537_

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4839) by [Unito](https://www.unito.io)

### LINKED ISSUE #1940: Violation of TO3g
[TO3g](https://sdk.ably.com/builds/ably/specification/main/features/#TO3g) says to queue publish and presence messages only. Current implementation queues everything. See also https://github.com/ably/specification/pull/194

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4858) by [Unito](https://www.unito.io)

maratal:
Closes #1936, #1938 and #1940

These three bugs were cancelling each other, so it's impossible to address them in different PRs since to fix all the tests failures you need to fix other two bugs. Please see commit messages and review per commit basis.

See also https://github.com/ably/specification/pull/194

Some additional info for commits messages.

d60269af "Proceeding with detach on `CONNECTED`":

What happens here is that in addition to re-attach we need re-detach as well to continue the process after connection becomes `CONNECTED`. But in oppose to `attach` there were no internal method for doing so, that's why I've splitted  `_detach` into `internalDetach:` which doesn't do state check before proceeding with detaching. Also because detach message shouldn't be queued I've removed `shouldQueueEvents` check from the `detachAfterChecks`.

305d8db3 "Sending queued messages should be after reattach":

If you send message before sending `ATTACH`, realtime won't respond and tests will timeout. You can check it with `test__065__Channel__publish__Message_connectionId_should_match_the_current_Connection_id_for_all_published_messages` test.

16b38064 "Only queue publish and presence messages":

The check `else if (msg.ackRequired)` you should read as "if message can be queued", since `ACK` required with the same condition as queuing - message is either `MESSAGE` or `PRESENCE`. Also I've added log line if message sending was ignored.

### DIFF
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -743,9 +743,8 @@ const NSTimeInterval _immediateReconnectionDelay = 0.1;
     
     if ([self shouldSendEvents]) {
         for (ARTRealtimeChannelInternal *channel in channels) {
-            // Reattach channel regardless resume success - RTN15c6, RTN15c7
             ARTAttachRequestParams *const params = [[ARTAttachRequestParams alloc] initWithReason:stateChange.reason];
-            [channel reattachWithParams:params];
+            [channel proceedAttachDetachWithParams:params];
         }
         [self sendQueuedMessages];
     }
--- a/Source/ARTRealtime.m
+++ b/Source/ARTRealtime.m
@@ -1275,22 +1275,34 @@ const NSTimeInterval _immediateReconnectionDelay = 0.1;
     if ([self shouldSendEvents]) {
         [self sendImpl:msg reuseMsgSerial:reuseMsgSerial sentCallback:sentCallback ackCallback:ackCallback];
     }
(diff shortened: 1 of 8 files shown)


maratal:
Closed in favor of https://github.com/ably/ably-cocoa/pull/1941

maratal:
Closed by mistake

umair-ably (approved):
LGTM

PR #2004, opened 2025-01-07
Remove nullability of `-subscribe*` return value

lawrence-forooghian:
There’s no reason for these to be nullable (it just makes it harder to call the `-unsubscribe` methods, which take a nonnull listener), and the return type of the underlying methods is nonnull.


maratal (approved):

lawrence-forooghian:
As much as I'd like to merge this, I think that it might actually be a backwards compatibility breaking change (Swift will give you a compile error if you try to unwrap a non-optional).

lawrence-forooghian:
Gonna close this; let's do in our next major release (created https://github.com/ably/ably-cocoa/issues/2026).

Issue #2009, opened 2025-01-09
Realtime connection does not use fallback hosts in custom environment


lawrence-forooghian:
For example, change the test `test__091__Connection__Host_Fallback__every_connection_is_first_attempted_to_the_primary_host_realtime_ably_io` so that it uses the sandbox environment (i.e. write `options.environment = "sandbox"`). You will see that the client does not attempt to connect to any fallback hosts. I think that, per RTN17c and RSC15g2, it should do.

Fallback hosts do seem to be working correctly for custom environments for REST connections; see the tests for RSC15g2.

PR #2017, opened 2025-01-23
[ECO-5190] Fallbacks in custom environment

maratal:
Closes https://github.com/ably/ably-cocoa/issues/2009


Source/ARTDefault.m
lawrence-forooghian, line 1: Not sure I understand this change, couple of questions:

- Why would it be a breaking change to mark `ARTClientOptions.environment` as nullable? Its default value is already `nil`, right? So if we changed the property declaration to nullable we'd be correcting it, wouldn't we?
- What do you mean that `nil` "translates to an empty string in Swift"?

Test/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -4042,7 +4042,7 @@ class RealtimeClientConnectionTests: XCTestCase {
         try testMovesToDisconnectedWithNetworkingError(NSError(domain: "NSPOSIXErrorDomain", code: 50, userInfo: [NSLocalizedDescriptionKey: "shouldn't matter"]), for: test)
     }
 
-    func test__102__Connection__Host_Fallback__should_move_to_disconnected_when_there_s_no_internet__with_any_kCFErrorDomainCFNetwork() throws {
+    func skipped_test__102__Connection__Host_Fallback__should_move_to_disconnected_when_there_s_no_internet__with_any_kCFErrorDomainCFNetwork() throws {
```
lawrence-forooghian, line 4045: I'm curious why this test was failing? Shouldn't the connection become DISCONNECTED per [RTN14d](https://sdk.ably.com/builds/ably/specification/main/features/#RTN14d)?

Source/ARTRealtime.m
```diff
@@ -1637,7 +1637,7 @@ - (void)realtimeTransportFailed:(id<ARTRealtimeTransport>)transport withError:(A
     
     if ([self shouldRetryWithFallback:transportError]) {
         ARTLogDebug(self.logger, @"R:%p host is down; can retry with fallback host", self);
-        if (!_fallbacks && [transportError.url.host isEqualToString:[ARTDefault realtimeHost]]) {
+        if (!_fallbacks && [transportError.url.host hasSuffix:[ARTDefault realtimeHost]]) { // this checks if host is a default one regardless environment used, RTN17b
```
lawrence-forooghian, line 1640: Do you not think that instead of this check, it might be better for us to actually implement the check that's described in RTN17b? For example, this check doesn't take into account whether the user has changed the port, or whether they've provided an array of `fallbackHosts`, right? (Meaning that we probably aren't satisfying the sub-points of RTN17b, and we should fix and test that too.)

Test/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -3838,7 +3838,7 @@ class RealtimeClientConnectionTests: XCTestCase {
     func test__086__Connection__Host_Fallback__failing_connections_with_custom_endpoint_should_result_in_an_error_immediately() {
         let test = Test()
         let options = ARTClientOptions(key: "xxxx:xxxx")
-        options.environment = "test" // do not use the default endpoint
```
lawrence-forooghian, line 3841: Why are these changes separate from 88611490120f3ad93565deca454903bf4a1ffbcb? Aren't they fixing test failures introduced by a change in that commit?

Test/Tests/RealtimeClientConnectionTests.swift
```diff
@@ -4139,16 +4141,28 @@ class RealtimeClientConnectionTests: XCTestCase {
         }
 
         XCTAssertEqual(urlConnections.count, 3)
-        XCTAssertTrue(NSRegularExpression.match(urlConnections.at(0)?.absoluteString, pattern: "//realtime.ably.io"))
-        XCTAssertTrue(NSRegularExpression.match(urlConnections.at(1)?.absoluteString, pattern: "//[a-e].ably-realtime.com"))
```
lawrence-forooghian, line 4143: So we're no longer testing that the fallback URLs are correctly formed?

Test/Tests/RealtimeClientConnectionTests.swift
lawrence-forooghian: e.g.

image

vs

image

Source/ARTDefault.m
maratal: >If they're expecting it to return a String it will, yeah

I've messed up here, currently `options.environment` can't be nil, so no one should have a check like that. Sorry, fixed - https://github.com/ably/ably-cocoa/pull/2017/commits/e58e13a7d2bee948767c4101bb256549c70deee3

>plus with some logic about trimming whitespace

I agree that it's some unexpected string manipulation here, so I've removed it. But I don't have an idea of why it wasn't just `nil` only check, I think since `ARTClientOptions.hasEnvironment` already has the same check, why remove it from here then? wdyt

Source/ARTRealtime.m
maratal: done https://github.com/ably/ably-cocoa/pull/2017/commits/2a9b845dad248acaa9cfc2584bd3a52481749f63

Test/Tests/RealtimeClientConnectionTests.swift
maratal: https://github.com/ably/ably-cocoa/pull/2017/commits/2a9b845dad248acaa9cfc2584bd3a52481749f63

Test/Tests/RealtimeClientConnectionTests.swift
maratal: Yeah, I missed that lines when came up with a proper check somewhere else - https://github.com/ably/ably-cocoa/pull/2017/commits/2a9b845dad248acaa9cfc2584bd3a52481749f63

lawrence-forooghian (approved):

Issue #2020, opened 2025-01-29
Implement wrapper SDK proxy


lawrence-forooghian:
As described in [ADR-117: Tracking wrapper SDK usage, continued](https://ably.atlassian.net/wiki/spaces/ENG/pages/3413016589/ADR-117+Tracking+wrapper+SDK+usage+continued).

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5204) by [Unito](https://www.unito.io)

PR #2021, opened 2025-01-29
[ECO-5204] Groundwork for wrapper SDK proxy

lawrence-forooghian:
Groundwork for #2020. See commit messages for more details.


Source/include/Ably/ARTRealtime.h
```diff
@@ -25,9 +25,9 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /**
- The protocol upon which the top level object `ARTRealtime` is implemented.
+ This protocol contains the non-initializer instance methods provided by the `ARTRealtime` client class.
  */
-@protocol ARTRealtimeProtocol <NSObject>
+@protocol ARTRealtimeInstanceMethodsProtocol <NSObject>
```
maratal, line 30: Could also be `ARTRealtimeInstanceProtocol` or `ARTRealtimePrincipalProtocol`, just a minor suggestion.

maratal (approved):
LGTM

Source/include/Ably/ARTRealtime.h
lawrence-forooghian: Thanks for the suggestions! I don't think that any one of the three is loads better than the others, so will keep as is 🙂

Issue #2045, opened 2025-04-10
Investigate supporting visionOS


lawrence-forooghian:
A customer has asked if we support visionOS. I did a very brief investigation into it, here's what I found:

- It already runs on "Apple Vision Pro (Designed for iPad)" — see AblyPushExample app
- Briefly tried getting it working for native apps too
  - tried adding to `Package.swift` and adding to the AblyPushExample app, but refused to compile and wasn't immediately obvious why
  - I think that we need to sort out the conditional compilation and add `#if TARGET_OS_VISIONOS`
- we probably should run the tests for it, and do the same for watchOS whilst we're at it, this would mean creating new Xcode targets
- and when you change the Package.swift to the tools version needed for visionOS, it says the other platforms we're trying to build for are deprecated
- there are other deprecation warnings that you get when you try and build for visionOS (probably because even the earliest visionOS is a lot newer than the oldest platforms we support)

The aim of this task is to get a better idea of how much work it would be to add visionOS support. If there are any useful changes that come out of this investigation, please push them to a branch so that we can pick them up if / when we decide to proceed with adding visionOS support.

maratal:
WIP branch 2045-visionOS-research

lawrence-forooghian:
Not sure I understand why you talk about "visionOS targets", what's the significance of "target" there? From the second screenshot you posted it seems that visionOS is supported by Nimble, no?

And then, if we _weren't_ able to build our visionOS test dependencies using Carthage, is there some other mechanism by which our tests could get hold of the dependencies? For example, by telling the Xcode project to get its dependencies from SPM instead of from Carthage?

maratal:
> And then, if we _weren't_ able to build our visionOS test dependencies using Carthage, is there some other mechanism by which our tests could get hold of the dependencies? For example, by telling the Xcode project to get its dependencies from SPM instead of from Carthage?

Will look into https://github.com/ably/ably-cocoa/issues/2034

lawrence-forooghian:
> Will look into https://github.com/ably/ably-cocoa/issues/2034

Switching to using SPM for running tests is, I think, a big task. My question was whether we could continue to use the Xcode project but tell it to get its _dependencies_ using SPM (this seems like less work).

lawrence-forooghian:
@maratal I did an example of this for you: https://github.com/ably/ably-cocoa/tree/xcode-proj-dependencies-from-SPM

[...]

PR #2050, opened 2025-04-21
Vision os research

maratal:
Adresses #2045


.github/workflows/integration-test-xros.yaml
lawrence-forooghian, line 1: Why are you using the name "xrOS" here? Isn't this old terminology?

maratal: It still called like that in the SDK - https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md

PR #2199, opened 2026-03-17
Add `onAttach` callback variants to annotations subscribe

lawrence-forooghian:
Commit d7c8b45 introduced `RealtimeAnnotations#subscribe` but omitted the `onAttach` callback variants that `RealtimeChannel#subscribe` provides. Per RTAN4d, annotations subscribe must have the same return value and callback semantics as channel subscribe (RTL7g), which includes the optional attach callback.

This adds `subscribeWithAttachCallback:callback:` and `subscribe:onAttach:callback:` to `ARTRealtimeAnnotations`, mirroring the existing channel API. The internal `_subscribe:onAttach:callback:` method already accepted an `onAttach` parameter but the public API was not wiring it through.

Also fixes a pre-existing bug in `_subscribe:onAttach:callback:` where the `onAttach` callback was not being wrapped to dispatch to the user queue, unlike the channel's equivalent. The channel wraps `onAttach` to dispatch to the user queue so that it is called on the correct queue per the public API contract.

Tests are copied from the RTL7g/RTL7h channel subscribe tests in `RealtimeClientChannelTests.swift`, adapted to use annotations subscribe.

## Summary by CodeRabbit

* **New Features**
  * Added attach-callback support to annotation subscriptions, including per-type subscription variants.
  * Attach callbacks are now invoked on the user's execution queue for correct asynchronous behavior.

* **Tests**
  * Added comprehensive tests covering annotation subscription behavior, channel attach scenarios, and error/state handling during subscribe operations.

### DIFF
--- a/Source/ARTRealtimeAnnotations.m
+++ b/Source/ARTRealtimeAnnotations.m
@@ -59,8 +59,16 @@
     return [_internal subscribe:callback];
 }
 
+- (ARTEventListener *)subscribeWithAttachCallback:(ARTCallback)onAttach callback:(ARTAnnotationCallback)callback {
+    return [_internal subscribeWithAttachCallback:onAttach callback:callback];
+}
+
 - (ARTEventListener *)subscribe:(NSString *)type callback:(ARTAnnotationCallback)callback {
     return [_internal subscribe:type callback:callback];
+}
+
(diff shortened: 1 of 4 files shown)


maratal (approved):
LGTM (two small formatting issues)

PR #2207, opened 2026-05-17
[ECO-5729] Store all persisted data atomically in a single file with None protection

### LINKED ISSUE #2206: Store all persisted data atomically and with `None` protection class
Currently:

- most of the persisted data is stored in `NSUserDefaults`, which is not guaranteed to be always-available (https://mjtsai.com/blog/2024/10/07/launching-before-userdefaults-is-available/)
- the `LocalDevice` secret is stored in the keychain with after-first-unlock protection class

We have had a long-standing belief that this (i.e. the fact that data might not be always available, and the fact that some pieces of data may be available when others are not) may be causing various kinds of errors for our users. The most recent investigation of such issues can be found in #2204. We have not been able to reproduce the issue but we believe that it's worth making a speculative fix anyway.

Things to be done:

- do not use `UserDefaults` or keychain as a storage mechanism for any data
- the new storage mechanism should store data such that it is always available regardless of whether the device is unlocked — concretely this probably means a file with [`None` protection class](https://support.apple.com/guide/security/data-protection-classes-secb010e978a/web)
- migrate existing persisted data to this new storage mechanism
- make sure that this new storage mechanism stores `LocalDevice` data atomically i.e. no way to end up with a `deviceIdentityToken` that doesn't belong to that device ID

We have previously avoided making this change because we weren't sure whether it was OK to store the device secret in an always-available fashion, what with it being labelled a "secret". But after discussing this with @paddybyers we are happy to do this, and it's what Android does.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5729) by [Unito](https://www.unito.io)

maratal:
Closes #2206.

## Summary

- Replaces `NSUserDefaults` + keychain with a single property-list file (`Application Support/Ably/LocalDevice.plist`) written atomically and, on iOS, with `NSFileProtectionNone`, so the data is readable regardless of device lock state.
- Adds `performBatchUpdate:` to `<ARTDeviceStorage>`: a block-scoped API that batches related writes into a single flush, so the `(id, secret)` tuple can be updated atomically.
- Implements **RSH8a1** from spec PR ably/specification#474 in `+[ARTLocalDevice deviceWithStorage:logger:]`: on load failure (`id` or `deviceSecret` missing) all persisted LocalDevice attributes and all persisted Activation State Machine data are discarded, then a fresh `(id, secret)` pair is regenerated as part of LocalDevice initialisation (the eager-generation behaviour noted by RSH8k2).
- One-time migration of legacy `UserDefaults` + keychain data on first init; legacy entries are deleted after migration. If the keychain secret can't be loaded for any reason (realistically: the device hasn't been unlocked since reboot), the legacy data is discarded entirely and the device-fetch path starts clean — `push.activate()` will re-register the device on next call.

## Manual test (cross-branch migration)

1. Check out `main`, build and run the `AblyPushExample` app on a device.
2. Register the device for push, then subscribe to a channel. Find the device by id in the Ably dashboard (Devices) or the Ably CLI (`ably devices list`).
3. Without uninstalling or resetting app state, switch to this PR's branch (`fix/2206-atomic-file-storage`) and rebuild.
4. Launch the app with the SDK log level set to info or higher and watch for:
   ```
   ARTLocalDeviceStorage: migrated N legacy fields
   ```
5. Verify the same device id is still registered. Unsubscribing the previously-subscribed channel from this branch should succeed — the dashboard should show the subscription gone — confirming the migrated `deviceIdentityToken` is still bound to the migrated device id.

If migration runs while the keychain is inaccessible (e.g. the app launches in the background before first unlock), the legacy data is discarded and a fresh device id is generated. The dashboard then shows a new device id and the old registration becomes orphaned.

## Summary by CodeRabbit

* **New Features**
  * Atomic file-backed device storage with thread-safe reads/writes and optional legacy migration
  * Atomic batching API for grouped device identity and push-state updates
  * Example helper now reports identity-token availability from persisted device details

* **Bug Fixes**
  * Improved reliability when loading, persisting, and migrating device identity and push tokens
  * Safer handling of missing or inconsistent persisted device data

* **Refactor**
  * Replaced direct legacy defaults/keychain access with coordinated atomic storage

* **Tests**
  * Added tests for migration, atomicity, and persistence scenarios

### DIFF
--- a/Examples/AblyPush/AblyPushExample/AblyHelper.swift
+++ b/Examples/AblyPush/AblyPushExample/AblyHelper.swift
@@ -61,10 +61,15 @@ extension AblyHelper {
     }
 
     func printIdentityToken() {
-        if UserDefaults.standard.value(forKey: "ARTDeviceIdentityToken") != nil {
-            print("IDENTITY TOKEN: exists")
+        let device = realtime.device
+        // identityTokenDetails may be nil if the device hasn't been activated for push
+        if let details = device.identityTokenDetails {
+            // The token is a credential; print only a short prefix so debug
+            // output / device logs don't leak it in full.
+            let preview = details.token.prefix(8)
+            print("Identity token exists for device id \(device.id): \(preview)…")
         } else {
-            print("IDENTITY TOKEN: doesn't exist")
+            print("Identity token doesn't exist")
         }
     }
 
--- a/Source/ARTAtomicFileStorage.m
+++ b/Source/ARTAtomicFileStorage.m
@@ -1,0 +1,115 @@
+#import "ARTAtomicFileStorage.h"
+#import "ARTInternalLog.h"
+
+@implementation ARTAtomicFileStorage {
+    ARTInternalLog *_logger;
+    NSLock *_lock;
+}
+
+- (instancetype)initWithFileURL:(NSURL *)fileURL logger:(nullable ARTInternalLog *)logger {
+    if (self = [super init]) {
+        _fileURL = [fileURL copy];
+        _logger = logger;
+        _lock = [[NSLock alloc] init];
+        _lock.name = @"io.ably.ARTAtomicFileStorage";
+    }
+    return self;
+}
+
+- (BOOL)fileExists {
+    [_lock lock];
+    BOOL exists = [[NSFileManager defaultManager] fileExistsAtPath:_fileURL.path];
+    [_lock unlock];
+    return exists;
+}
+
+- (NSDictionary<NSString *, id> *)load {
+    [_lock lock];
+    NSDictionary *result = [self load_nosync];
+    [_lock unlock];
+    return result;
+}
+
+- (NSDictionary<NSString *, id> *)load_nosync {
+    NSError *readError = nil;
+    NSData *data = [NSData dataWithContentsOfURL:_fileURL options:0 error:&readError];
+    if (data == nil) {
+        if (readError.code != NSFileReadNoSuchFileError) {
+            ARTLogWarn(_logger, @"ARTAtomicFileStorage: could not read %@: %@", _fileURL.lastPathComponent, readError.localizedDescription);
+        }
+        return @{};
+    }
+
+    NSError *parseError = nil;
+    id plist = [NSPropertyListSerialization propertyListWithData:data options:NSPropertyListImmutable format:NULL error:&parseError];
+    if (![plist isKindOfClass:[NSDictionary class]]) {
+        ARTLogError(_logger, @"ARTAtomicFileStorage: %@ is not a dictionary plist (%@); treating as empty", _fileURL.lastPathComponent, parseError.localizedDescription);
+        return @{};
+    }
+    return plist;
+}
+
+- (BOOL)save:(NSDictionary<NSString *, id> *)dictionary error:(NSError * _Nullable * _Nullable)outError {
+    [_lock lock];
+    @try {
+        NSError *serializeError = nil;
+        NSData *data = [NSPropertyListSerialization dataWithPropertyList:dictionary
+                                                                  format:NSPropertyListBinaryFormat_v1_0
+                                                                 options:0
+                                                                   error:&serializeError];
+        if (data == nil) {
+            if (outError) *outError = serializeError;
+            ARTLogError(_logger, @"ARTAtomicFileStorage: failed to serialise %@: %@", _fileURL.lastPathComponent, serializeError.localizedDescription);
+            return NO;
+        }
+
+        NSError *dirError = nil;
+        if (![[NSFileManager defaultManager] createDirectoryAtURL:[_fileURL URLByDeletingLastPathComponent]
+                                      withIntermediateDirectories:YES
+                                                       attributes:nil
+                                                            error:&dirError]) {
+            if (outError) *outError = dirError;
+            ARTLogError(_logger, @"ARTAtomicFileStorage: failed to create parent directory for %@: %@", _fileURL.lastPathComponent, dirError.localizedDescription);
+            return NO;
+        }
+
+        NSDataWritingOptions options = NSDataWritingAtomic;
+        // `NSDataWritingFileProtectionNone` is only available on macOS 11+; on
+        // iOS/tvOS it's always available (use of `*` below). Older macOS has no
+        // data-protection concept, so skipping it there is fine.
+        if (@available(macOS 11.0, *)) {
+          // IMPORTANT; DO NOT REMOVE: We use NSDataWritingFileProtectionNone
+          // because the file must be always available to load. This is our
+          // solution to issues that we previously had when storing data as
+          // "protected until first user authentication"; there are scenarios in
+          // which iOS applications get launched before a first unlock and the
+          // LocalDevice data needs to still be loadable in these scenarios.
+          options |= NSDataWritingFileProtectionNone;
+        }
+
+        NSError *writeError = nil;
+        BOOL ok = [data writeToURL:_fileURL options:options error:&writeError];
+        if (!ok) {
+            if (outError) *outError = writeError;
+            ARTLogError(_logger, @"ARTAtomicFileStorage: failed to write %@: %@", _fileURL.lastPathComponent, writeError.localizedDescription);
+            return NO;
+        }
+
+        // Make sure the file is excluded from iCloud backup — the device
+        // secret was previously protected with `ThisDeviceOnly` keychain
+        // semantics, and we want to preserve the "doesn't leave this device"
+        // property even though we've moved away from the keychain.
+        NSURL *mutableURL = [_fileURL copy];
+        NSError *attrError = nil;
+        if (![mutableURL setResourceValue:@YES forKey:NSURLIsExcludedFromBackupKey error:&attrError]) {
+            ARTLogWarn(_logger, @"ARTAtomicFileStorage: could not mark %@ as excluded from backup: %@", _fileURL.lastPathComponent, attrError.localizedDescription);
+        }
+
+        return YES;
+    }
+    @finally {
+        [_lock unlock];
+    }
+}
+
+@end
--- a/Source/ARTLocalDevice.m
+++ b/Source/ARTLocalDevice.m
@@ -8,6 +8,7 @@
 #import "ARTCrypto+Private.h"
 #import "ARTInternalLog.h"
 #import "ARTTypes+Private.h"
+#import "ARTPushActivationStateMachine+Private.h"
 
 NSString *const ARTDevicePlatform = @"ios";
 
@@ -57,12 +58,23 @@ NSString* ARTAPNSDeviceTokenKeyOfType(NSString *tokenType) {
     return self;
 }
 
-- (void)generateAndPersistPairOfDeviceIdAndSecret {
-    self.id = [self.class generateId];
-    self.secret = [self.class generateSecret];
-
-    [_storage setObject:self.id forKey:ARTDeviceIdKey];
-    [_storage setSecret:self.secret forDevice:self.id];
+// RSH8b helper. Only generates a new (id, deviceSecret) pair and persists it.
+// Decision about when generation should happen and what else to clear belongs to the caller.
+// If `writer` is omitted, device's storage is used (for standalone calls).
+- (void)generateAndPersistPairOfDeviceIdAndSecretForStorage:(id<ARTDeviceStorage>)writer {
+    NSString *newId = [self.class generateId];
+    NSString *newSecret = [self.class generateSecret];
+    self.id = newId;
+    self.secret = newSecret;
+    // The inner batch makes the helper safe to call standalone: id and
+    // secret reach disk together rather than as two separate flushes. When
+    // a caller is already inside its own batch (e.g. `resetDetails`) the
+    // nested batch is just a counter bump — the outer batch still owns the
+    // single flush.
+    [(writer ?: _storage) performBatchUpdate:^(id<ARTDeviceStorage> writer) {
+        [writer setObject:newId forKey:ARTDeviceIdKey];
+        [writer setObject:newSecret forKey:ARTDeviceSecretKey];
+    }];
 }
 
 + (instancetype)deviceWithStorage:(id<ARTDeviceStorage>)storage logger:(nullable ARTInternalLog *)logger {
@@ -73,72 +85,61 @@ NSString* ARTAPNSDeviceTokenKeyOfType(NSString *tokenType) {
     device.push.recipient[@"transportType"] = ARTDevicePushTransportType;
 
     NSString *deviceId = [storage objectForKey:ARTDeviceIdKey];
-    NSString *deviceSecret = deviceId == nil ? nil : [storage secretForDevice:deviceId];
+    NSString *deviceSecret = [storage objectForKey:ARTDeviceSecretKey];
 
-    if (deviceId == nil || deviceSecret == nil) {
-        if (deviceId == nil) {
-            ARTLogDebug(logger, @"Generating device ID and secret: no stored device ID");
-        }
-        else {
-            ARTLogError(logger, @"Regenerating device ID and secret: stored device ID %@ has no secret", deviceId);
-        }
-        [device generateAndPersistPairOfDeviceIdAndSecret]; // Should be removed later once spec issue #180 resolved.
-    }
-    else {
+    // RSH8a
+    if (deviceId && deviceSecret) {
         device.id = deviceId;
         device.secret = deviceSecret;
+
+        ARTDeviceIdentityTokenDetails *identityTokenDetails = [ARTDeviceIdentityTokenDetails art_unarchiveFromStorage:storage
+                                                                                                                  key:ARTDeviceIdentityTokenKey
+                                                                                                           withLogger:logger];
+        device->_identityTokenDetails = identityTokenDetails;
+
+        NSString *clientId = [storage objectForKey:ARTClientIdKey];
+        if (clientId == nil && identityTokenDetails.clientId != nil) {
+            clientId = identityTokenDetails.clientId; // Older versions of the SDK did not persist clientId, so as a fallback when loading data persisted by these versions we use the clientId of the stored identity token
+            [storage setObject:clientId forKey:ARTClientIdKey];
+        }
+        device.clientId = clientId;
+
+        NSArray *supportedTokenTypes = @[
+            ARTAPNSDeviceDefaultTokenType,
+            ARTAPNSDeviceLocationTokenType
+        ];
+
+        for (NSString *tokenType in supportedTokenTypes) {
+            NSString *token = [ARTLocalDevice apnsDeviceTokenOfType:tokenType fromStorage:storage];
+            [device setAPNSDeviceToken:token tokenType:tokenType];
+        }
     }
-
-    ARTDeviceIdentityTokenDetails *identityTokenDetails = [ARTDeviceIdentityTokenDetails art_unarchiveFromStorage:storage
-                                                                                                              key:ARTDeviceIdentityTokenKey
-                                                                                                       withLogger:logger];
-    device->_identityTokenDetails = identityTokenDetails;
-
-    NSString *clientId = [storage objectForKey:ARTClientIdKey];
-    if (clientId == nil && identityTokenDetails.clientId != nil) {
-        clientId = identityTokenDetails.clientId; // Older versions of the SDK did not persist clientId, so as a fallback when loading data persisted by these versions we use the clientId of the stored identity token
-        [storage setObject:clientId forKey:ARTClientIdKey];
-    }
-    device.clientId = clientId;
-
-    NSArray *supportedTokenTypes = @[
-        ARTAPNSDeviceDefaultTokenType,
-        ARTAPNSDeviceLocationTokenType
-    ];
-
-    for (NSString *tokenType in supportedTokenTypes) {
-        NSString *token = [ARTLocalDevice apnsDeviceTokenOfType:tokenType fromStorage:storage];
-        [device setAPNSDeviceToken:token tokenType:tokenType];
+    else {
+        // RSH8b, RSH8k2
+        ARTLogDebug(logger, @"LocalDevice: generating a fresh id+secret pair");
+        [device generateAndPersistPairOfDeviceIdAndSecretForStorage:nil];
     }
     return device;
 }
 
 - (void)setupDetailsWithClientId:(NSString *)clientId {
-    NSString *deviceId = self.id;
-    NSString *deviceSecret = self.secret;
-
-    if (deviceId == nil || deviceSecret == nil) {
-        [self generateAndPersistPairOfDeviceIdAndSecret];
-    }
-
     self.clientId = clientId;
     [_storage setObject:clientId forKey:ARTClientIdKey];
 }
 
 - (void)resetDetails {
-    // Should be replaced later to resetting device's id/secret once spec issue #180 resolved.
-    [self generateAndPersistPairOfDeviceIdAndSecret];
+    [_storage performBatchUpdate:^(id<ARTDeviceStorage> writer) {
+        // Wipe everything persisted, then regenerate — all in the same batch,
+        // so the id+secret regeneration lands together with the wipe.
+        [writer removeAll];
+        [self generateAndPersistPairOfDeviceIdAndSecretForStorage:writer];
 
-    self.clientId = nil;
-    [_storage setObject:nil forKey:ARTClientIdKey];
-    [self setAndPersistIdentityTokenDetails:nil];
-    NSArray *supportedTokenTypes = @[
-        ARTAPNSDeviceDefaultTokenType,
-        ARTAPNSDeviceLocationTokenType
-    ];
-    for (NSString *tokenType in supportedTokenTypes) {
-        [self setAndPersistAPNSDeviceToken:nil tokenType:tokenType];
-    }
+        // Mirror the persistent clear in the in-memory device state.
+        self.clientId = nil;
+        _identityTokenDetails = nil;
+        [self setAPNSDeviceToken:nil tokenType:ARTAPNSDeviceDefaultTokenType];
+        [self setAPNSDeviceToken:nil tokenType:ARTAPNSDeviceLocationTokenType];
+    }];
 }
 
 + (NSString *)generateId {
@@ -180,13 +181,18 @@ NSString* ARTAPNSDeviceTokenKeyOfType(NSString *tokenType) {
 }
 
 - (void)setAndPersistIdentityTokenDetails:(ARTDeviceIdentityTokenDetails *)tokenDetails {
-    [self.storage setObject:[tokenDetails art_archiveWithLogger:self.logger]
-                     forKey:ARTDeviceIdentityTokenKey];
+    NSData *tokenData = [tokenDetails art_archiveWithLogger:self.logger];
+    BOOL adoptClientId = (self.clientId == nil && tokenDetails.clientId != nil);
     _identityTokenDetails = tokenDetails;
-    if (self.clientId == nil) {
+    if (adoptClientId) {
         self.clientId = tokenDetails.clientId;
-        [self.storage setObject:tokenDetails.clientId forKey:ARTClientIdKey];
     }
+    [self.storage performBatchUpdate:^(id<ARTDeviceStorage> writer) {
+        [writer setObject:tokenData forKey:ARTDeviceIdentityTokenKey];
+        if (adoptClientId) {
+            [writer setObject:tokenDetails.clientId forKey:ARTClientIdKey];
+        }
+    }];
 }
 
 - (BOOL)isRegistered {
--- a/Source/ARTLocalDeviceStorage.m
+++ b/Source/ARTLocalDeviceStorage.m
@@ -1,23 +1,95 @@
 #import "ARTLocalDeviceStorage.h"
+#import <Security/Security.h>
+#import "ARTAtomicFileStorage.h"
 #import "ARTInternalLog.h"
 #import "ARTLocalDevice+Private.h"
+#import "ARTPushActivationStateMachine+Private.h"
+
+static NSString *const ARTLocalDeviceStorageDirectoryName = @"Ably";
+static NSString *const ARTLocalDeviceStorageFileName = @"LocalDevice.plist";
+
+static NSString *const ARTLegacyAPNSDeviceTokenKey = @"ARTAPNSDeviceToken";
+
+NSString *const ARTMigratedFromLegacyStorageKey = @"ARTMigratedFromLegacyStorage";
 
 @implementation ARTLocalDeviceStorage {
     ARTInternalLog *_logger;
     BOOL _logValues;
+
+    ARTAtomicFileStorage *_store;
+    NSMutableDictionary<NSString *, id> *_cache;
+
+    NSRecursiveLock *_lock;
+    NSInteger _batchDepth;
+    BOOL _isModified;
+
+    ARTLegacyKeychainSecretReader _legacyKeychainReader;
 }
 
-- (instancetype)initWithLogger:(ARTInternalLog *)logger logValues:(BOOL)logValues {
+#pragma mark - Init
+
+- (instancetype)initWithLogger:(nullable ARTInternalLog *)logger logValues:(BOOL)logValues {
+    return [self initWithBaseDirectoryURL:[ARTLocalDeviceStorage defaultBaseDirectoryURL]
+                                   logger:logger
+                                logValues:logValues];
+}
+
++ (instancetype)newWithLogger:(nullable ARTInternalLog *)logger logValues:(BOOL)logValues {
+    return [[self alloc] initWithLogger:logger logValues:logValues];
+}
+
+- (instancetype)initWithBaseDirectoryURL:(NSURL *)baseDirectoryURL
+                                  logger:(nullable ARTInternalLog *)logger
+                               logValues:(BOOL)logValues {
+    return [self initWithBaseDirectoryURL:baseDirectoryURL
+                                   logger:logger
+                                logValues:logValues
+                     legacyKeychainReader:nil];
+}
+
+- (instancetype)initWithBaseDirectoryURL:(NSURL *)baseDirectoryURL
+                                  logger:(nullable ARTInternalLog *)logger
+                               logValues:(BOOL)logValues
+                    legacyKeychainReader:(nullable ARTLegacyKeychainSecretReader)legacyKeychainReader {
     if (self = [super init]) {
         _logger = logger;
         _logValues = logValues;
+        _lock = [[NSRecursiveLock alloc] init];
+        _lock.name = @"io.ably.ARTLocalDeviceStorage";
+        _legacyKeychainReader = [legacyKeychainReader copy];
+
+        NSURL *fileURL = [baseDirectoryURL URLByAppendingPathComponent:ARTLocalDeviceStorageFileName];
+        _store = [[ARTAtomicFileStorage alloc] initWithFileURL:fileURL logger:logger];
+        _cache = [[_store load] mutableCopy];
+
+        // Technically the RSH8a1 check-and-discard should happen lazily when the
+        // LocalDevice is first required; for simplicity we are instead
+        // performing it eagerly at storage init (and thus at client init).
+        // We'll address this later if it turns out to cause problems; tracking
+        // in https://github.com/ably/ably-cocoa/issues/2214.
+        [self migrateLegacyDataIfNeeded];
     }
     return self;
 }
 
-+ (instancetype)newWithLogger:(ARTInternalLog *)logger logValues:(BOOL)logValues {
-    return [[self alloc] initWithLogger:logger logValues:logValues];
++ (NSURL *)defaultBaseDirectoryURL {
+    NSFileManager *fm = [NSFileManager defaultManager];
+    NSError *error = nil;
+    NSURL *appSupport = [fm URLForDirectory:NSApplicationSupportDirectory
+                                   inDomain:NSUserDomainMask
+                          appropriateForURL:nil
+                                     create:YES
+                                      error:&error];
+    if (appSupport == nil) {
+        // Fall back to the temp directory if Application Support is somehow
+        // unavailable — the SDK won't persist anything across launches in that
+        // case, but at least won't crash.
+        appSupport = [NSURL fileURLWithPath:NSTemporaryDirectory()];
+    }
+    return [appSupport URLByAppendingPathComponent:ARTLocalDeviceStorageDirectoryName];
 }
+
+#pragma mark - Logging helpers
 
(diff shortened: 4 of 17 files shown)


Source/ARTLocalDeviceStorage.m
```diff
@@ -1,24 +1,88 @@
 #import "ARTLocalDeviceStorage.h"
+#import "ARTAtomicFileStorage.h"
 #import "ARTInternalLog.h"
 #import "ARTLocalDevice+Private.h"
+#import "ARTPushActivationStateMachine+Private.h"
```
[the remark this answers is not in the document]
maratal: Added explicit `#import <Security/Security.h>` in b441793b. The symbols were resolving transitively but the import belongs here.

Source/ARTLocalDeviceStorage.m
```diff
@@ -32,193 +96,213 @@ - (id)valueToLogForValue:(id)value {
...
+- (void)performBatchUpdate:(NS_NOESCAPE void (^)(id<ARTDeviceStorage>))block {
+    [_lock lock];
+    @try {
+        _batchDepth++;
+        block(self);
+        _batchDepth--;
+        if (_batchDepth == 0) {
+            [self flushIfModified];
+        }
     }
```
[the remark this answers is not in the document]
maratal: Fixed in b441793b: increment now runs before `@try` and decrement + flush are in `@finally`, so a throwing block can't leave the counter stale.

Source/ARTLocalDevice.m
```diff
@@ -73,72 +99,60 @@ + (instancetype)deviceWithStorage:(id<ARTDeviceStorage>)storage logger:(nullable
...
     [_storage setObject:clientId forKey:ARTClientIdKey];
 }
 
 - (void)resetDetails {
-    // Should be replaced later to resetting device's id/secret once spec issue #180 resolved.
-    [self generateAndPersistPairOfDeviceIdAndSecret];
+    [_storage performBatchUpdate:^(id<ARTDeviceStorage> writer) {
+        [self cleanupStorage];
+        // Same batch — id+secret regeneration lands together with the wipe.
+        [self generateAndPersistPairOfDeviceIdAndSecret];
```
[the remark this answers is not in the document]
maratal: Deliberate. `ARTLocalDeviceStorage` passes `self` into the block as the writer, so `writer == self == _storage` by construction — going through `_storage` and going through `writer` are the same call. The protocol leaves room for a separate buffered writer in a future implementation, but we considered that earlier and chose the current shape; nested `performBatchUpdate:` calls just bump the depth counter, they don't reach outside the active batch.

Source/ARTPushActivationStateMachine.m
```diff
@@ -155,13 +158,17 @@ - (void)handleEvent:(nonnull ARTPushActivationEvent *)event {
...
-                              forKey:ARTPushActivationCurrentStateKey];
+        stateData = [_current art_archiveWithLogger:_logger];
     }
-    [self.rest.storage setObject:[_pendingEvents art_archiveWithLogger:_logger]
-                          forKey:ARTPushActivationPendingEventsKey];
+    NSData *eventsData = [_pendingEvents art_archiveWithLogger:_logger];
+    [self.rest.storage performBatchUpdate:^(id<ARTDeviceStorage> writer) {
+        if (stateData != nil) {
+            [writer setObject:stateData forKey:ARTPushActivationCurrentStateKey];
+        }
```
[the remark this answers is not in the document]
maratal: Deliberate. The non-persistent states (`WaitingForDeviceRegistration`, `WaitingForRegistrationSync`, `WaitingForDeregistration`) are transient — preserving the last persistent state is intentional so that, if the process is killed mid-transition, we resume from a stable point and retry the operation. The case you'd worry about (entering `NotActivated` with a stale persisted state) doesn't apply: `NotActivated` *is* an `ARTPushActivationPersistentState`, and every path that takes us there in this SDK (RSH3a1d / RSH3g2c / RSH8a1) goes through `resetDetails`, which explicitly clears the state-machine keys.

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
```diff
@@ -0,0 +1,284 @@
+#if os(iOS)
+import Ably
+import Ably.Private
+import XCTest
```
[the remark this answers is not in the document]
maratal: Added `import Security` in b441793b. The constants were resolving via a transitive Foundation/XCTest include, but the import is correct here.

Examples/AblyPush/AblyPushExample/AblyHelper.swift
```diff
@@ -61,10 +61,12 @@ extension AblyHelper {
...
     func printIdentityToken() {
-        if UserDefaults.standard.value(forKey: "ARTDeviceIdentityToken") != nil {
-            print("IDENTITY TOKEN: exists")
+        let device = realtime.device
+        // identityTokenDetails may be nil if the device hasn't been activated for push
+        if let details = device.identityTokenDetails {
+            print("Identity token for device id \(device.id): \(details.token)")
         } else {
-            print("IDENTITY TOKEN: doesn't exist")
+            print("Identity token doesn't exist")
```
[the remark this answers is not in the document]
maratal: Truncated to an 8-char prefix in b441793b — same fix as CodeRabbit's comment on the next line.

Examples/AblyPush/AblyPushExample/AblyHelper.swift
```diff
@@ -61,10 +61,12 @@ extension AblyHelper {
...
         }
```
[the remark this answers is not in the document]
maratal: Truncated to an 8-char prefix in b441793b — single fix covers Copilot's matching comment too.

Source/ARTLocalDeviceStorage.m
```diff
@@ -32,193 +96,213 @@ - (id)valueToLogForValue:(id)value {
...
-    else if (error) {
-        ARTLogError(_logger, @"Device Secret couldn't be read for device %@ (%@)", deviceId, [error localizedDescription]);
+    @finally {
+        [_lock unlock];
+    }
+}
```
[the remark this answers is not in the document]
maratal: Fixed in b441793b — same change addresses Copilot's identical comment. Increment is now before `@try`, decrement and flush are inside `@finally`.

Source/ARTLocalDeviceStorage.m
```diff
@@ -32,193 +96,213 @@ - (id)valueToLogForValue:(id)value {
...
-        if (error) {
-            *error = [self keychainErrorWithCode:status];
-        }
+        migrated[ARTLegacyAPNSDeviceTokenKey] = legacyApnsLegacy;
+        migrated[ARTAPNSDeviceTokenKeyOfType(ARTAPNSDeviceDefaultTokenType)] = legacyApnsDefault;
+        migrated[ARTAPNSDeviceTokenKeyOfType(ARTAPNSDeviceLocationTokenType)] = legacyApnsLocation;
+
+        migrated[ARTPushActivationCurrentStateKey] = legacyState;
+        migrated[ARTPushActivationPendingEventsKey] = legacyPendingEvents;
     }
```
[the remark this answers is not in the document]
maratal: False positive. The code uses the keyed-subscript form (`dict[key] = value`), which calls `-setObject:forKeyedSubscript:` — that selector is nil-tolerant and just removes the entry. The throwing selector is `-setObject:forKey:`, which we don't use. The two migration tests exercise this branch with nil legacy fields and pass; no `NSInvalidArgumentException` is raised.

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
```diff
@@ -0,0 +1,284 @@
...
+            writer.setObject("client-1", forKey: ARTClientIdKey)
+        }
+
+        // All four writes must land in the new file content; a follow-up load
+        // sees the full tuple.
+        let reloaded = makeStorage()
+        XCTAssertEqual(reloaded.object(forKey: ARTDeviceIdKey) as? String, "device-1")
+        XCTAssertEqual(reloaded.object(forKey: ARTDeviceSecretKey) as? String, "secret-1")
+        XCTAssertEqual(reloaded.object(forKey: ARTClientIdKey) as? String, "client-1")
+    }
```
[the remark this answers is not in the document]
maratal: Added the missing `ARTDeviceIdentityTokenKey` assertion in b441793b.

Source/PrivateHeaders/Ably/ARTAtomicFileStorage.h
```diff
@@ -0,0 +1,33 @@
+#import <Foundation/Foundation.h>
+
+@class ARTInternalLog;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// Reads and writes a single property-list-backed file atomically. On iOS the
```
lawrence-forooghian: Is there anything special about iOS? Seems easier to reason about if we just always use `None` file protection.

Source/PrivateHeaders/Ably/ARTLocalDeviceStorage.h
```diff
@@ -5,11 +5,48 @@
...
+///
+/// Writes are atomic: every persisted value lives in the same file and is
+/// written with a write-to-temp-and-rename, so it is impossible for, e.g., a
+/// `deviceIdentityToken` to be paired with a `deviceId` that it does not
+/// belong to. On iOS the file is written with `NSFileProtectionNone` so the
+/// data is readable regardless of device lock state.
+///
+/// On first init, data persisted by older SDK versions (in `NSUserDefaults`
+/// and the keychain) is migrated into the new file, and the old entries are
+/// removed.
```
lawrence-forooghian: What was your thinking behind deleting the old data? It seems like it might be good to have it around as a backup in case there's a bug in our migration code. You're already bypassing the old data load if the new file exists so is there a need to delete?

Source/ARTLocalDeviceStorage.m
```diff
@@ -32,193 +97,213 @@ - (id)valueToLogForValue:(id)value {
...
+        migrated[ARTDeviceIdKey] = legacyDeviceId;
+        migrated[ARTDeviceSecretKey] = legacySecret;
+        migrated[ARTDeviceIdentityTokenKey] = legacyIdentityToken;
+        migrated[ARTClientIdKey] = legacyClientId;
 
-    if (status != errSecSuccess) {
-        if (error) {
-            *error = [self keychainErrorWithCode:status];
-        }
+        migrated[ARTLegacyAPNSDeviceTokenKey] = legacyApnsLegacy;
```
lawrence-forooghian: Why are we writing to the new data with a key that has "legacy" in its name?

Source/ARTLocalDeviceStorage.m
```diff
@@ -32,193 +97,213 @@ - (id)valueToLogForValue:(id)value {
...
+    NSMutableDictionary *migrated = [NSMutableDictionary dictionary];
 
-    [query setObject:@YES forKey:(__bridge id)kSecReturnData];
-    [query setObject:(__bridge id)kSecMatchLimitOne forKey:(__bridge id)kSecMatchLimit];
-    CFTypeRef result = NULL;
-    OSStatus status = SecItemCopyMatching((__bridge CFDictionaryRef)query, &result);
+    // RSH8a1: the legacy data crosses into the new file only if id and
+    // secret are both loadable. Otherwise we drop everything and let the
+    // device-fetch path start clean.
+    if (legacyDeviceId != nil && legacySecret != nil) {
```
lawrence-forooghian, line 251: Do you think it's worth adding some sort of flag to the migrated data that indicates that it's been migrated? Given that we don't 100% trust the legacy data (e.g. we have reason to believe that the legacy data _possibly_ contains a device identity token that's out of sync with the device ID) then it feels like it might be worth having a flag for future usage that says "this data was migrated, maybe at some point if we're still having out-of-sync issues we should consider validating it"?

Source/PrivateHeaders/Ably/ARTAtomicFileStorage.h
```diff
@@ -0,0 +1,33 @@
...
+- (NSDictionary<NSString *, id> *)load;
+
+/// Writes `dictionary` to the file. The write is atomic at the file-system
+/// level (write-to-temp-and-rename). Returns `YES` on success.
+- (BOOL)save:(NSDictionary<NSString *, id> *)dictionary error:(NSError * _Nullable * _Nullable)error;
+
+- (BOOL)fileExists;
+
+/// Returns `YES` if the file existed before the call.
+- (BOOL)removeFile;
```
lawrence-forooghian: This is unused

Source/ARTLocalDeviceStorage.m
```diff
@@ -32,193 +97,213 @@ - (id)valueToLogForValue:(id)value {
...
-    NSError *error = nil;
-    NSString *value = [self keychainGetPasswordForService:ARTDeviceSecretKey account:(NSString *)deviceId error:&error];
-
-    if ([error code] == errSecItemNotFound) {
-        ARTLogDebug(_logger, @"Device Secret read miss for device %@", deviceId);
+- (void)performBatchUpdate:(NS_NOESCAPE void (^)(id<ARTDeviceStorage>))block {
+    [_lock lock];
+    _batchDepth++;
+    @try {
+        block(self);
```
lawrence-forooghian, line 149: what is the purpose of the writer that gets passed to the block if it's just always `self`?

Test/AblyTests/Tests/PushTests.swift
lawrence-forooghian, line 1: I don't see any mentions of RSH3h in the codebase; what would be the best way of making sure this spec point is explicitly addressed? Perhaps an explicit test somewhere of RSH3h, i.e. `LocalDevice` failure cascades to you getting a `NotActivated` state machine?

Source/ARTLocalDeviceStorage.m
```diff
@@ -32,193 +97,213 @@ - (id)valueToLogForValue:(id)value {
...
+}
 
-    return value;
+- (void)flushIfNotBatching {
+    if (_batchDepth > 0) return;
+    [self flushIfModified];
 }
 
-- (void)setSecret:(NSString *)value forDevice:(ARTDeviceId *)deviceId {
+- (void)flushIfModified {
```
lawrence-forooghian, line 179: Worth a debug log when we flush?

Source/ARTLocalDeviceStorage.m
```diff
@@ -32,193 +97,213 @@ - (id)valueToLogForValue:(id)value {
...
 
-    return nil;
+    NSError *saveError = nil;
+    if (![_store save:migrated error:&saveError]) {
+        ARTLogError(_logger, @"ARTLocalDeviceStorage: failed to write migrated file: %@", saveError.localizedDescription);
+        return;
+    }
+    _cache = [migrated mutableCopy];
+    [self clearLegacyEntriesForDeviceId:legacyDeviceId];
+    ARTLogInfo(_logger, @"ARTLocalDeviceStorage: migrated %lu legacy fields", (unsigned long)migrated.count);
```
lawrence-forooghian: Do you think it might be useful to log the data that was migrated?

Source/ARTLocalDevice.m
```diff
@@ -57,12 +58,37 @@ - (instancetype)initWithStorage:(id<ARTDeviceStorage>)storage logger:(nullable A
     return self;
 }
 
-- (void)generateAndPersistPairOfDeviceIdAndSecret {
-    self.id = [self.class generateId];
-    self.secret = [self.class generateSecret];
+- (void)cleanupStorage {
+    [_storage setObject:nil forKey:ARTDeviceIdKey];
```
lawrence-forooghian: Is there a reason we're doing this key-by-key? It feels a bit error-prone; would it not be better for the storage to expose a single method that clears it?

Source/ARTLocalDeviceStorage.m
```diff
@@ -32,193 +97,213 @@ - (id)valueToLogForValue:(id)value {
...
+- (void)migrateLegacyDataIfNeeded {
+    if ([_store fileExists]) {
+        return; // already migrated (or new install with the storage already populated)
     }
+    [self migrateLegacyData];
 }
 
-#pragma mark - Keychain
+/// Reads any legacy data from `NSUserDefaults` + the keychain into the new
+/// file. This is where RSH8a1 is enforced against the legacy data: a
```
lawrence-forooghian, line 200: I'm curious why we do RSH8a1 in _both_ the migration path and the general-purpose `ARTLocalDevice` load path?

More generally, the code is not very explicit about how we achieve the order of events in RSH3h — i.e. ensuring that `LocalDevice` data validation happens as part of initialising the push state machine. Maybe it would be easier to reason about if `ARTLocalDeviceStorage` initialisation were the chokepoint through which we handled the loading of all persisted data; i.e. inside `ARTLocalDeviceStorage` initialisation:

1. if there's data to migrate, migrate it
2. validate the loaded data per RSH8a1
3. if invalid per RSH8a1, clear the data per RSH8a1

so now, _before exposing any storage contents to any other components_, your persisted data either is OK or both the `LocalDevice` and push activation state machine data is cleared, regardless of who initiated the data load. What do you think? (For the avoidance of doubt, I'm _not_ suggesting moving the id/secret regeneration code into the `ARTLocalDeviceStorage`.)

(I'm open to other suggestions for how to make the RSH3h ordering explicit in the code.)

Source/ARTLocalDevice.m
```diff
@@ -57,12 +58,37 @@ - (instancetype)initWithStorage:(id<ARTDeviceStorage>)storage logger:(nullable A
...
+}
 
-    [_storage setObject:self.id forKey:ARTDeviceIdKey];
-    [_storage setSecret:self.secret forDevice:self.id];
+// RSH8b helper. Generates a new (id, deviceSecret) pair and persists it —
+// nothing more. Decisions about when generation should happen and what else
+// to clear alongside it belong to the caller; today that's `resetDetails`,
+// which is the single recovery/reset path used by both RSH8a1 (load failure)
+// and the state machine's deactivation transitions.
+- (void)generateAndPersistPairOfDeviceIdAndSecret {
```
lawrence-forooghian: see question elsewhere about the purpose of the `writer` that `performBatchUpdate:` passes to its block — if that writer were actually meaningful then presumably it should be passed as an argument here in the nested case?

lawrence-forooghian:
I'd like to understand a bit more about the manual testing that you performed:

> If migration runs while the keychain is inaccessible (e.g. the app launches in the background before first unlock), the legacy data is discarded and a fresh device id is generated.

How did you test this using `AblyPushExample` i.e. how did you make it run before first unlock? And on a similar note how did you verify that the data persisted by `ARTAtomicFileStorage` is available before first unlock? It would be great to have a list of step-by-step instructions so that I can try reproducing.

(You might find the test app in https://github.com/ably/ably-cocoa/pull/2204 helpful for launching the app before first unlock and also for seeing all of the log events that the library emits in such a scenario — it publishes them to an Ably channel so that you can use the CLI to subscribe and see them.)

Source/PrivateHeaders/Ably/ARTAtomicFileStorage.h
maratal: 0f594640 — now uses `NSFileProtectionNone` on all platforms, not just iOS.

Source/PrivateHeaders/Ably/ARTLocalDeviceStorage.h
maratal: 0f594640 — agreed; no longer deletes the legacy data after migration, keeping it as a fallback (cleanup commented out pending #1257).

Source/ARTLocalDeviceStorage.m
maratal: a6d3c3be — removed; we no longer write the legacy APNs token key into the new file.

Source/ARTLocalDeviceStorage.m
maratal: e8aac6ef — added `ARTMigratedFromLegacyStorageKey`, written into the migrated data.

Source/PrivateHeaders/Ably/ARTAtomicFileStorage.h
maratal: 4e27494e — removed the unused method.

Source/ARTLocalDeviceStorage.m
maratal: b5cf4d12 — the writer is now meaningful: `resetDetails` regenerates the id+secret through the passed writer (nil falls back to the device's own storage).

Test/AblyTests/Tests/PushTests.swift
maratal: 0f594640 — made RSH3h explicit (the state machine initialises the `LocalDevice` before reading its persisted state) and added RSH3h tests in `LocalDeviceStorageTests`.

Source/ARTLocalDeviceStorage.m
maratal: 8acd0e73 — added a debug log on each successful flush (keys + file name).

Source/ARTLocalDeviceStorage.m
maratal: 8acd0e73 — migration now logs the migrated fields (values redacted unless value-logging is enabled).

Source/ARTLocalDevice.m
maratal: ee62ce8f — added `-[ARTDeviceStorage removeAll]`; `resetDetails` now clears via that single call instead of key-by-key.

Source/ARTLocalDeviceStorage.m
maratal: 0f594640 — RSH3h ordering made explicit: the state machine initialises the `LocalDevice` (RSH8a/RSH8a1) before reading persisted state. RSH8a1's legacy discard stays in migration; the atomic single-file format keeps id+secret together, so there is no separate load-path discard to duplicate.

Source/ARTLocalDevice.m
maratal: b5cf4d12 — the nested case now passes the writer through (`generateAndPersistPairOfDeviceIdAndSecretForStorage:`).

Source/ARTLocalDevice.m
```diff
@@ -73,72 +85,61 @@ + (instancetype)deviceWithStorage:(id<ARTDeviceStorage>)storage logger:(nullable
...
-        [device setAPNSDeviceToken:token tokenType:tokenType];
+        for (NSString *tokenType in supportedTokenTypes) {
+            NSString *token = [ARTLocalDevice apnsDeviceTokenOfType:tokenType fromStorage:storage];
+            [device setAPNSDeviceToken:token tokenType:tokenType];
+        }
+    }
+    else {
+        // RSH8b, RSH8k2
+        ARTLogDebug(logger, @"LocalDevice: generating a fresh id+secret pair");
+        [device generateAndPersistPairOfDeviceIdAndSecret];
```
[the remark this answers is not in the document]
maratal: this method always gets storage in a consisted state

Source/ARTLocalDeviceStorage.m
```diff
@@ -1,24 +1,91 @@
...
 #import "ARTInternalLog.h"
 #import "ARTLocalDevice+Private.h"
+#import "ARTPushActivationStateMachine+Private.h"
+
+static NSString *const ARTLocalDeviceStorageDirectoryName = @"Ably";
+static NSString *const ARTLocalDeviceStorageFileName = @"LocalDevice.plist";
+
+static NSString *const ARTLegacyAPNSDeviceTokenKey = @"ARTAPNSDeviceToken";
+
+NSString *const ARTMigratedFromLegacyStorageKey = @"ARTMigratedFromLegacyStorage";
```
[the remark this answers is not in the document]
maratal: new file-backed store doesn't need it

Source/ARTLocalDeviceStorage.m
```diff
@@ -32,193 +99,233 @@ - (id)valueToLogForValue:(id)value {
...
-    return nil;
-}
+    NSError *saveError = nil;
+    if (![_store save:migrated error:&saveError]) {
+        ARTLogError(_logger, @"ARTLocalDeviceStorage: failed to write migrated file: %@", saveError.localizedDescription);
+        return;
+    }
+    _cache = [migrated mutableCopy];
+    // TODO: Uncomment once issue #1257 is resolved
+    // [self clearLegacyEntriesForDeviceId:legacyDeviceId];
```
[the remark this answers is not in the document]
maratal: on purpose

maratal:
> How did you test this using AblyPushExample i.e. how did you make it run before first unlock?

I've just simulated secret loading failure (return `nil` manually). 

> And on a similar note how did you verify that the data persisted by ARTAtomicFileStorage is available before first unlock?

Wasn't `NSDataWritingFileProtectionNone` [made](https://developer.apple.com/documentation/foundation/nsdata/writingoptions/nofileprotection?language=objc) for this?

Anyways, worth checking through your app, will do.

lawrence-forooghian:
> Wasn't NSDataWritingFileProtectionNone [made](https://developer.apple.com/documentation/foundation/nsdata/writingoptions/nofileprotection?language=objc) for this?

Sure but I still think we want to test in real life.

maratal:
> Sure but I still think we want to test in real life.

Here is a new instruction:

0. Run the app from your branch (`AIT-569-investigating-push-registration-failures`) and register device. Find `pushActivateResult` event in your ably CLI session and remember device `id` and `secret`.
1. Switch to my branch (`AIT-569-investigating-push-registration-failures-with-fix`, merge of your branch and the fix branch).
2. Run updated app with breakpoint set in `migrateLegacyDataIfNeeded`, once hit, terminate the app to avoid actual migration (we just need an updated app on this step).
3. Restart your device with cellular available upon reboot.
4. Send voip push and ensure the device received a call.
5. Look for the last `pushActivateResult` event in your ably CLI session, device `id` and `secret` should be different from previous launch (due to unavailable keychain before first unlock).
6. Unlock the device, launch the app and ensure that this time the device `id/secret` are the same as in step 5.
7. Repeat steps 3 and 4.
8. Look again for the last `pushActivateResult` event, this time device `id` and `secret` should be the same as in previous launch (in step 6).

@lawrence-forooghian

Source/ARTAtomicFileStorage.m
```diff
@@ -0,0 +1,109 @@
...
+                                                            error:&dirError]) {
+            if (outError) *outError = dirError;
+            ARTLogError(_logger, @"ARTAtomicFileStorage: failed to create parent directory for %@: %@", _fileURL.lastPathComponent, dirError.localizedDescription);
+            return NO;
+        }
+
+        NSDataWritingOptions options = NSDataWritingAtomic;
+        // `NSDataWritingFileProtectionNone` is only available on macOS 11+; on
+        // iOS/tvOS it's always available (the `*` branch). Older macOS has no
+        // data-protection concept, so skipping it there is fine. (issue #1257)
```
lawrence-forooghian: Please can you check through all of the references to #1257 that you've introduced in your latest batch of changes? At a glance they don't seem relevant or appropriate.

Source/PrivateHeaders/Ably/ARTLocalDeviceStorage.h
```diff
@@ -5,11 +5,55 @@
...
+/// migration's keychain-locked / unreachable branches can be exercised.
+typedef NSString * _Nullable (^ARTLegacyKeychainSecretReader)(NSString *deviceId, OSStatus * _Nullable outStatus);
+
+/// Persists `LocalDevice` data and push-activation-state-machine data to a
+/// single always-available file in the app's Application Support directory.
+///
+/// Writes are atomic: every persisted value lives in the same file and is
+/// written with a write-to-temp-and-rename, so it is impossible for, e.g., a
+/// `deviceIdentityToken` to be paired with a `deviceId` that it does not
+/// belong to. On iOS the file is written with `NSFileProtectionNone` so the
```
lawrence-forooghian: This comment remains outdated

Source/PrivateHeaders/Ably/ARTLocalDeviceStorage.h
lawrence-forooghian: What does it have to do with #1257? and "pending" in what sense? Is it because you think we _should_ start clearing the data at some point (the fact that you've just commented out the call to `-clearLegacyEntriesForDeviceId:` suggests so)? If so what point?

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
```diff
@@ -0,0 +1,358 @@
...
+        XCTAssertEqual(storage.object(forKey: ARTDeviceIdKey) as? String, "legacy-device")
+        XCTAssertEqual(storage.object(forKey: ARTDeviceSecretKey) as? String, "legacy-secret")
+        XCTAssertEqual(storage.object(forKey: ARTDeviceIdentityTokenKey) as? Data, identityTokenData)
+        XCTAssertEqual(storage.object(forKey: ARTClientIdKey) as? String, "legacy-client")
+        XCTAssertEqual(storage.object(forKey: ARTPushActivationCurrentStateKey) as? Data, stateData)
+        XCTAssertEqual(storage.object(forKey: ARTPushActivationPendingEventsKey) as? Data, eventsData)
+        XCTAssertEqual(storage.object(forKey: "ARTAPNSDeviceToken-default") as? String, "apns-default-token")
+        XCTAssertEqual(storage.object(forKey: "ARTAPNSDeviceToken-location") as? String, "apns-location-token")
+
+        // The migrated data is flagged as such (see #2207), so future versions
```
lawrence-forooghian: Ditto the references to #2207 aren't very helpful (someone lands on this PR and… then what?)

Source/ARTLocalDeviceStorage.m
lawrence-forooghian: I think we'd benefit from a comment explaining why RSH8a1 is only done on the migration path and not in the general storage load path.

lawrence-forooghian: non-blocking: this technically doesn't actually conform to RSH8a1, which, in combination with RSH8a's "The LocalDevice is initialised when first required", means that the check-and-discard should happen lazily when the `LocalDevice` is first required; we are instead performing it eagerly at storage init and thus at client init.

It's hard to think of many concrete negative consequences of this though; I originally wrote this but then realised that we do `LocalDevice` access also whenever the client becomes identified (to set its `clientID`) so really my speculated workaround (that might not be happening in practice anyway) is only useful for nonidentified clients.

> It's plausible that some users have previously noticed that they were running into before-first-unlock id/secret regeneration issues and thus decided to work around this by not trying to access any push-related features until they have themselves verified that `protectedDataAvailable` is returning `YES` in order to avoid a device ID regeneration. By performing the migration eagerly we'd be breaking these workarounds.

Will leave it to you to consider and see if you think it's worth doing anything. Worst case behaviourally is a one-off device ID regeneration for hypothetical users who had been trying to avoid one. Also technically there's now a synchronous disk read on the caller thread (possibly main) on client init.

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
```diff
@@ -0,0 +1,358 @@
...
+    func test_storageFileIsCreatedWithFileProtectionNone() throws {
+        let storage = makeStorage()
+        storage.setObject("device-1", forKey: ARTDeviceIdKey)
+        storage.setObject("secret-1", forKey: ARTDeviceSecretKey)
+
+        let fileURL = tempDirectory.appendingPathComponent("LocalDevice.plist")
+        XCTAssertTrue(FileManager.default.fileExists(atPath: fileURL.path))
+
+        let attrs = try FileManager.default.attributesOfItem(atPath: fileURL.path)
+        // On a tmp directory the protection key may be absent (it inherits the
```
lawrence-forooghian: Could you please help me understand this comment? Given that we set this key explicitly when writing the file why would it not be set when we read it back?

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
```diff
@@ -0,0 +1,358 @@
...
+        // All four writes must land in the new file content; a follow-up load
+        // sees the full tuple.
+        let reloaded = makeStorage()
+        XCTAssertEqual(reloaded.object(forKey: ARTDeviceIdKey) as? String, "device-1")
+        XCTAssertEqual(reloaded.object(forKey: ARTDeviceSecretKey) as? String, "secret-1")
+        XCTAssertEqual(reloaded.object(forKey: ARTDeviceIdentityTokenKey) as? Data, Data("token".utf8))
+        XCTAssertEqual(reloaded.object(forKey: ARTClientIdKey) as? String, "client-1")
+    }
+
+    // MARK: RSH3h ordering
```
lawrence-forooghian: these RSH3h tests have been inserted in between two tests for the atomic data writing

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
```diff
@@ -0,0 +1,358 @@
...
+
+    // In this test the device data is *complete* (legacy id + keychain secret migrate
+    // successfully), so the persisted Activation State Machine state survives
+    // RSH3h step (1) and the machine constructed in step (2) resumes it.
+    func test_RSH3h_persistedActivationStateResumesWhenDeviceDataIsComplete() {
```
lawrence-forooghian: Given that both of these RSH3h tests explicitly load the device (`let device = rest.device`), it's not clear to me how they're testing RSH3h, which states that loading the push activation state machine should load the device.

Test/AblyTests/Tests/PushActivationStateMachineTests.swift
```diff
@@ -857,6 +857,11 @@ class PushActivationStateMachineTests: XCTestCase {
         options.clientId = "client1"
         let rest = ARTRest(options: options)
         rest.internal.storage = storage
+        // The device is a process-wide singleton; `setUp`'s `initialStateMachine`
```
lawrence-forooghian: All these changes to `resetDeviceSingleton()` that have been scattered throughout the existing test files seem like a bit of a code smell. I don't have a great suggestion right now for what we can do to address whatever problem they're trying to work around in a more holistic way. Do you? If not I think we need an issue to track it at least, with comments against each.

(I think that we want to solve this by ending up in a situation where the tests don't have to _undo_ some work that was done in their setup; i.e. if the setup is not flexible enough then I think it needs to be parameterised, but I don't have a sense of how much work that would be and if it's reasonable to do it within this PR, which has already been open for quite some time and has a lot of complexity)

Source/PrivateHeaders/Ably/ARTLocalDeviceStorage.h
maratal: >If so what point?

There is no strong "point" on this tbh. Just some local data housekeeping without deep thoughts about it. You are right, it doesn't take much space and is absent on new installations anyways, so I'll just remove this method.

Source/ARTLocalDeviceStorage.m
maratal: I've created an issue, I would like to avoid making this PR any more complicated - https://github.com/ably/ably-cocoa/issues/2214

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
maratal: Removed this useless check since there is no such thing as protection attribute under simulator. created `test_storageFileIsCreated` instead. Let me know if it worth it to create `FileManager` mock to check setting of this attribute.

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
maratal: moved

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
maratal: Added dedicated RSH3h test exercising loading device together with the state machine. Although it's a rare usecase for unidentified clients (mostly clients have clientId which forces loading of the device before the state machine initialization).

Source/ARTAtomicFileStorage.m
maratal: removed

Source/PrivateHeaders/Ably/ARTLocalDeviceStorage.h
maratal: fixed

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
maratal: removed

Test/AblyTests/Tests/PushActivationStateMachineTests.swift
maratal: fixed here - c160fcb8e2ca9c697a7ab555d6fb1a2c35549da9

There is also a change in `- [ARTPushActivationStateMachine initWithRest:..`, but it only affects tests which were fixed.

Source/ARTAtomicFileStorage.m
lawrence-forooghian: There are still a couple of incorrect references to #1257 in `LocalDeviceStorageTests.swift`

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
lawrence-forooghian: I think just add a comment inside `test_storageFileIsCreated` to explain why there's no assertion for `.protectionKey`

Source/ARTLocalDeviceStorage.m
lawrence-forooghian: Looks like my comment re "why RSH8a1 is only done on the migration path and not in the general storage load path" got lost in this; I've pushed 6917b501e252814d5b0611aacdc28523fe83a3a9 to address

Source/ARTAtomicFileStorage.m
lawrence-forooghian: Pushed c9326120648a053ff9ab8f62d0187daa3e233935 to fix

Test/AblyTests/Tests/LocalDeviceStorageTests.swift
lawrence-forooghian: Pushed 29236bcbd3ca4c3cb66ce0ae54de298ac0866bd0 with comments

lawrence-forooghian (approved):

maratal:
@lawrence-forooghian I've tested again and it works as before per instruction posted above. The branch is actual release branch with your app merged - `fix/release-1-2-61-test`. One thing to notice is that I've encountered missed `pushActivateResult` and `pushSubscribe*` events a few times in ably CLI (even with both startup settings `on` and independently of the device locked/unlocked state) which I couldn't reproduce on the `AIT-569-investigating-push-registration-failures-with-fix` branch (nor on your branch). Although storage-wise everything works as intended - when I see `pushActivateResult` it contains values I expect. Once I've introduced channel's logger (`ChannelEventLog` - logs into file), I wasn't able to reproduce this on the `fix/release-1-2-61-test` branch either. At this point it needs a second pair of eyes I think.

PR #2209, opened 2026-06-04
[AIT-846] ably-cocoa liveobjects UTS ground work

maratal:
Adds a **Universal Test Suite (UTS)** target deriving tests from the language-neutral pseudo-code specs in [`ably/specification`](https://github.com/ably/specification) under `uts/`.

## What's here
- New standalone `UTS` XCTest target (`Test/UTS`), separate from `AblyTests`, no Nimble.
- Mocking harness (`Test/UTS/Harness`):
  - **MockWebSocket** — a real `ARTWebSocketTransport` over a faked `ARTWebSocket`, so URL/query-param building stays real; the provider drives the simulated server (`respondWithSuccess`/`sendToClient`/`simulateDisconnect`).
  - **MockHTTP** — a fake `ARTHTTPExecutor` to intercept/observe requests and inject responses.
  - **FakeTimeProvider** — deterministic clock, opt-in via `enableFakeTimers()`; self-contained `advanceTime`.
  - **UTSTestCase** — `installMock` / `makeRealtime` / `makeRest`, `AWAIT_STATE` polling, `CapturingLog`, `NoOpReachability`.
- Tests: `ConnectionRecoveryTests` (RTN16g/g1, g2, f, f1, j, k) and `TimeTests` (RSC16 ×5).

Test names/IDs and fake-timer usage mirror the ably-java UTS PR ([ably/ably-java#1209](https://github.com/ably/ably-java/pull/1209)); requests/connection attempts are captured into local arrays per the spec. `deviations.md` is currently empty — no SDK non-compliance found for these specs.

## Builds on
This branch also contains the `ARTTimeProvider` groundwork (injectable clock/scheduler) that the fake-timer harness relies on.

## ⚠️ Not mergeable as-is
Includes a temporary branch dependency on `ably-cocoa-plugin-support` (`Package.swift`) — must be unpinned before merge.

## Test plan
`swift test --filter UTS` → 11 tests, all passing.


Test/UTS/Tests/ConnectionRecoveryTests.swift
```diff
@@ -0,0 +1,399 @@
+import XCTest
```
lawrence-forooghian, line 1: Is there a reason we're not using Swift Testing?

Package.swift
```diff
@@ -19,7 +19,10 @@ let package = Package(
         .package(name: "msgpack", url: "https://github.com/rvi/msgpack-objective-C", from: "0.4.0"),
         .package(name: "AblyDeltaCodec", url: "https://github.com/ably/delta-codec-cocoa", from: "1.3.5"),
         .package(name: "Nimble", url: "https://github.com/quick/nimble", from: "11.2.2"),
-        .package(name: "ably-cocoa-plugin-support", url: "https://github.com/ably/ably-cocoa-plugin-support.git", from: "2.0.0")
+        // TODO: Unpin before release. Points at the unreleased `mocking-for-uts` branch of
+        // ably-cocoa-plugin-support (adds APContinuousClockInstant / APSchedulerHandle); switch back
+        // to a versioned reference once a release containing them is available.
+        .package(name: "ably-cocoa-plugin-support", url: "https://github.com/ably/ably-cocoa-plugin-support.git", .branch("mocking-for-uts"))
```
lawrence-forooghian, line 25: from Slack:

> btw I see that your PR introduces a new version of ably-cocoa-plugin-support . i think that this is unnecessary for the first PR (that'll only be needed once we want to get this working with LiveObjects, and we don't yet know if we're going to merge the repos or not). we can just include those new types (SchedulerHandle and whatever else) inside ably-cooca for now

Test/UTS/Tests/TimeTests.swift
```diff
@@ -0,0 +1,182 @@
...
+final class TimeTests: UTSTestCase {
+
+    // UTS: rest/unit/RSC16/returns-server-time-0
+    func test_RSC16_time_returns_server_time() throws {
+        let serverTimeMs = 1_704_067_200_000 // 2024-01-01 00:00:00 UTC
+        var capturedRequests: [PendingHTTPRequest] = []
+        let mockHTTP = MockHTTP(
+            onConnectionAttempt: { connection in connection.respondWithSuccess() },
+            onRequest: { request in
+                capturedRequests.append(request)
```
lawrence-forooghian, line 16: I'm pretty sure that if we were compiling in the Swift 6 language mode (i.e. strict concurrency checking) then this wouldn't compile (the compiler can't reason about whether this `append()` is safe). It would be great if this new test suite could use the latest Swift language so that we get all these compiler checks.

lawrence-forooghian: (I would have loved for our existing test suite to also use strict concurrency checking, but porting it all would be a lot of work. But I think we should be doing it for all new targets.)

maratal:
Superseded by #2210, which rebases this work onto a clean base and drops the unreleased `ably-cocoa-plugin-support` dependency (the new boundary protocols are vendored locally under `ART` names). Closing in favour of #2210.

PR #2210, opened 2026-06-04
[AIT-846] Time mocking (ARTTimeProvider abstraction)

maratal:
## Summary

First of two PRs split out from the original combined PR. This one is the **SDK-side time abstraction** only: a unified `ARTTimeProvider` (wall clock, continuous clock, scheduler) that all time-dependent internal code routes through, so tests can install a deterministic clock via `ClientOptions.testOptions.timeProvider`. Default behaviour is unchanged (`ARTSystemTimeProvider`).

These abstractions aren't yet exposed to plugins — that's handled later in AIT-781.

### Stacked PR
The Universal Test Suite (UTS) groundwork that builds on this abstraction is in **#2213** (based on this branch). Review/merge this PR first.

### Review
Addresses the comments from review [4483891555](https://github.com/ably/ably-cocoa/pull/2210#pullrequestreview-4483891555); per-comment commit links are posted as replies below.

## Release Notes

* **Refactor**
  * Reworked internal timeouts, timestamps, and delayed scheduling to use an injectable time provider with cancelable scheduler handles, unifying wall-clock and monotonic timing behavior across realtime/auth/events.
  * Removed the legacy scheduled-dispatch helper and deprecated continuous-clock implementation details in favor of the new instant-based abstraction.

* **Tests**
  * Updated and expanded the suite to validate system time provider behavior, continuous-clock instant semantics, and delayed execution/cancellation.
  * Removed outdated continuous-clock test coverage.

* **Documentation**
  * Added coding standards requiring all new time-dependent logic to use the injected time provider abstraction.


Source/ARTContinuousClock.m
```diff
@@ -41,8 +25,14 @@ - (instancetype)initWithTime:(uint64_t)time {
     return self;
 }
 
-- (BOOL)isAfter:(ARTContinuousClockInstant *)other {
-    return self.timeInNanosecondsSinceClockReferenceInstant > other.timeInNanosecondsSinceClockReferenceInstant;
+- (BOOL)isAfter:(id<ARTContinuousClockInstantProtocol>)other {
+    ARTContinuousClockInstant *const concreteOther = (ARTContinuousClockInstant *)other;
+    return self.timeInNanosecondsSinceClockReferenceInstant > concreteOther.timeInNanosecondsSinceClockReferenceInstant;
+}
```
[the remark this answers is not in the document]
maratal: Fixed in 4d5777ec. `-isAfter:` now checks `isKindOfClass:` and raises `NSInvalidArgumentException` on a type mismatch instead of force-casting. In practice both the system and fake time providers return the same concrete `ARTContinuousClockInstant`, but the guard makes a cross-provider mix fail fast rather than read invalid memory.

CONTRIBUTING.md
```diff
@@ -71,6 +71,20 @@ ably-cocoa allows users to pass in Ably-authored plugins via the `ARTClientOptio
...
 
+### Time-related operations
+
+All time-dependent code in ably-cocoa goes through the injectable `ARTTimeProvider` abstraction (declared in `Source/PrivateHeaders/Ably/ARTTimeProvider.h`). This indirection allows the Universal Test Suite to install a single fake-time implementation that controls every clock-dependent code path across both ably-cocoa and any Ably-authored plugins.
+
+New code MUST obtain time and scheduling primitives from an injected `id<ARTTimeProvider>` rather than calling system primitives directly. In particular:
+
+- Don't call `[NSDate date]` (or `Date()` from Swift) directly. Use `[timeProvider wallClockNow]`.
+- Don't call `clock_gettime_nsec_np`, `mach_continuous_time`, or similar continuous-clock primitives directly. Use `[timeProvider continuousClockNow]`, which returns an `id<APContinuousClockInstant>` that supports `isAfter:` and `addingDuration:`.
+- Don't call `dispatch_after`, `NSTimer`, or `dispatch_source_set_timer` directly to run something after a delay. Use `[timeProvider scheduleAfter:queue:block:]`, which returns an `id<APSchedulerHandle>` that supports `cancel`.
```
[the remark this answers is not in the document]
maratal: Fixed in 4d5777ec. The bullets now reference `id<ARTContinuousClockInstantProtocol>` and `id<ARTSchedulerHandle>` (the actual types) — this was leftover from the protocol rename.

Package.swift
```diff
@@ -66,6 +66,21 @@ let package = Package(
                 .copy("ably-common")
             ]
         ),
+        // Universal Test Suite (UTS)
+        // A standalone XCTest suite derived from the language-neutral specs in the
+        // `ably/specification` repo (uts/). Deliberately does not depend on Nimble.
+        .testTarget(
```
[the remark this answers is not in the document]
maratal: Fixed in 4d5777ec. The comment now describes it as a Swift Testing suite (`import Testing`), not an XCTest suite.

Test/UTS/Harness/MockHTTP.swift
```diff
@@ -0,0 +1,147 @@
...
+            withLock { connectionError = firstConnection.resolvedError }
+        }
+
+        // Request phase — deliver to onRequest unless the connection failed
+        if let failure = withLock({ connectionError }) {
+            callback?(nil, nil, failure)
+            return NoopCancellable()
+        }
+        onRequest?(PendingHTTPRequest(request: request, completion: callback))
+        return NoopCancellable()
```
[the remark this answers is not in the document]
maratal: Good catch — fixed in 4d5777ec. `execute(_:completion:)` now fails fast when no `onRequest` handler is installed (an `assertionFailure` plus an explicit error passed to the completion), so a request no longer hangs with its callback never invoked.

Test/UTS/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+Test/UTS/
+  Harness/
+    UTSTestCase.swift        # base case: installMock, makeRealtime, awaitConnectionState/awaitChannelState, advanceTime, message builders
+    MockWebSocket.swift      # MockWebSocketProvider (the spec's mock_ws) + MockWebSocket (fake socket) + factories
+    FakeTimeProvider.swift   # deterministic ARTTimeProvider (enable_fake_timers / ADVANCE_TIME)
+    NoOpReachability.swift   # disables OS network monitoring in unit tests
+  Tests/
+    ConnectionRecoveryTests.swift   # RTN16 (pilot: 3 tests)
+  deviations.md              # SDK deviations found during evaluation
```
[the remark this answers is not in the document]
maratal: Already addressed in the preceding commit (fb7352a8): the layout now lists `MockHTTP.swift`, `Captured.swift`, `CapturingLog.swift`, and `ARTProtocolMessage+UTS.swift`, includes `TimeTests.swift`, and the 'pilot: 3 tests' note is gone.

maratal:
@lawrence-forooghian Re two claude generated messages above, I propose to discuss if it's worth addressing in this PR.

maratal:
## Why tests marked as `@Suite(.serialized)`?                                                                                                                                                                

A few reasons, but the main one is concrete: `Swift Testing` runs tests in parallel by default (unlike `XCTest`, which runs a class's tests serially). `.serialized` opts each suite back into sequential execution.

That matters here because these aren't pure-logic unit tests — each one drives a real `ARTRealtime/ARTWebSocketTransport` over real GCD queues, and the harness helpers block the calling thread to wait for it:

  - `poll / awaitConnectionState / awaitChannelState` spin with `Thread.sleep(0.0005)` until the state arrives.
  - `FakeTimeProvider.advanceTime` → `drainAllQueues()` does `queue.sync {}` barriers.

Swift Testing executes tests on a cooperative thread pool. Running several of these block-and-wait tests concurrently risks starving that pool (blocked threads can't service the very queues the tests are waiting on) and makes timing nondeterministic. Serializing sidesteps that, and it matches the serial semantics the suite had under XCTest. The cost is ~nothing — the whole 11-test suite runs in ~15 ms.
  
Two clarifications:
  - `.serialized` only serializes tests within a suite. `ConnectionRecoveryTests` and `TimeTests` still run in parallel with each other (you can see both suites "start" before either finishes in the run output) — and that's fine because each test owns its own client, mocks, and per-client queues, so there's no shared mutable state across suites.
  - So it's a conservative, cheap safety choice, not a strict correctness requirement.

Source/ARTLog.m
```diff
@@ -87,13 +89,14 @@ - (instancetype)initCapturingOutput:(BOOL)capturing historyLines:(NSUInteger)his
         _history = [[NSMutableArray alloc] init];
         _historyLines = historyLines;
         _queue = dispatch_queue_create("io.ably.log", DISPATCH_QUEUE_SERIAL);
+        _timeProvider = [[ARTSystemTimeProvider alloc] init];
     }
     return self;
```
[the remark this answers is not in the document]
maratal: ARTLog needs real timestamp always.

Source/ARTPluginAPI.m
lawrence-forooghian, line 1: The changes to this file are unnecessary; they aren't implementing methods from `APPluginAPIProtocol` and we haven't yet decided how we'll do the UTS for LiveObjects (AIT-781).

Source/PrivateHeaders/Ably/ARTContinuousClockInstantProtocol.h
```diff
@@ -0,0 +1,26 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ An instant on a continuous clock; that is, a clock which keeps incrementing while the system is asleep.
+
+ An instant is an opaque value type: code compares instants and derives new instants from them by
+ adding durations, but does not depend on the underlying representation.
+
+ - Note: This protocol is named with a `Protocol` suffix to distinguish it from the concrete
```
lawrence-forooghian: Can't the protocol be called `ARTContinuousClockInstant` and the implementation `ARTSystemContinuousClockInstant`?

Source/PrivateHeaders/Ably/ARTContinuousClock.h
```diff
@@ -1,39 +1,33 @@
...
+ The concrete implementation of `ARTContinuousClockInstantProtocol` used by `ARTSystemTimeProvider`.
 
- - Note: We don't give this one an `NS_SWIFT_NAME`, to avoid confusion with the Swift standard library type also named `ContinuousClock`.
- */
-@interface ARTContinuousClock: NSObject
+ Represents an instant on a continuous clock — a clock that keeps incrementing while the system is asleep.
 
-/**
- Returns the current instant in time.
+ - Note: We don't give this an `NS_SWIFT_NAME`, to avoid confusion with the Swift standard library type also named `ContinuousClock`.
```
lawrence-forooghian: This comment doesn't belong on this class

Source/PrivateHeaders/Ably/ARTContinuousClock.h
lawrence-forooghian, line 1: The name of this file doesn't match its contents. Also, why do we need to expose the `ARTContinuousClockInstant` class to the rest of the codebase? The initializer wasn't exposed before and the `ARTContinuousClockInstant` class is now abstracted behind a protocol anyway; can't its interface and implementation live inside `ARTSystemTimeProvider.m`?

CONTRIBUTING.md
```diff
@@ -71,6 +71,20 @@ ably-cocoa allows users to pass in Ably-authored plugins via the `ARTClientOptio
...
+
+New code MUST obtain time and scheduling primitives from an injected `id<ARTTimeProvider>` rather than calling system primitives directly. In particular:
+
+- Don't call `[NSDate date]` (or `Date()` from Swift) directly. Use `[timeProvider wallClockNow]`.
+- Don't call `clock_gettime_nsec_np`, `mach_continuous_time`, or similar continuous-clock primitives directly. Use `[timeProvider continuousClockNow]`, which returns an `id<ARTContinuousClockInstantProtocol>` that supports `isAfter:` and `addingDuration:`.
+- Don't call `dispatch_after`, `NSTimer`, or `dispatch_source_set_timer` directly to run something after a delay. Use `[timeProvider scheduleAfter:queue:block:]`, which returns an `id<ARTSchedulerHandle>` that supports `cancel`.
+
+Internal classes obtain their `ARTTimeProvider` by reading `options.testOptions.timeProvider` in their `-init` (the default value is an `ARTSystemTimeProvider`, backed by real system primitives) and stash it as an ivar, following the same threading pattern used for `logger` and the internal dispatch queue. Downstream consumers (e.g. `ARTEventEmitter`) receive the provider as an init parameter from their owning class.
```
lawrence-forooghian: I assume the word "threading" here refers to "threading it through" but that's not what people think when they see the word "threading"

CONTRIBUTING.md
```diff
@@ -71,6 +71,20 @@ ably-cocoa allows users to pass in Ably-authored plugins via the `ARTClientOptio
...
+
+The vendored SocketRocket implementation in `Source/SocketRocket/` is exempt from this rule for now — its time-related code will be addressed when the WebSocket layer becomes test-injectable.
```
lawrence-forooghian: Is this comment true? I don't see any changes to SocketRocket as part of the WebSocket mockability. I assume we have no intention of touching the SocketRocket code ever. Worth checking if there's any time-related code in SocketRocket that we depend on (e.g. whether any of the `ClientOptions` timeouts end up being translated into some option that we pass to SocketRocket and thus a UTS test testing these client options would want to be able to provoke these timeouts by manipulating time; I hope not).

Test/AblyTests/Tests/ContinuousClockTests.swift
```diff
@@ -1,30 +0,0 @@
-import XCTest
-import Ably.Private
-
-class ContinuousClockTests: XCTestCase {
```
lawrence-forooghian, line 4: Can these not become `SystemTimeProvider` tests?

Test/AblyTests/Tests/GCDTests.swift
```diff
@@ -12,11 +12,11 @@ class GCDTests: XCTestCase {
         weak var weakObject = object
 
         // prepare schedule block
-        var scheduledBlock: ARTScheduledBlockHandle? = artDispatchScheduled(0, .main) { [object] in
+        var scheduledBlock: ARTScheduledBlockHandle? = ARTScheduledBlockHandle(delay: 0, queue: .main, block: { [object] in
```
lawrence-forooghian: Given that we want all time-related ops to go via the `TimeProvider`, I'd suggest that we remove the internal `ARTScheduledBlockHandle` interface and move `ARTScheduledBlockHandle`'s implementation into `ARTSystemProvider.m` and move this tests into the tests for that class.

Source/ARTGCD.m
```diff
@@ -51,25 +51,6 @@ - (void)dealloc {
 
 @end
 
-ARTScheduledBlockHandle *artDispatchScheduled(NSTimeInterval seconds, dispatch_queue_t queue, dispatch_block_t block) {
-    // We don't pass the block directly; instead, we put it in a property, and
```
lawrence-forooghian, line 55: This comment has been lost; feels worth preserving?

Source/ARTJsonLikeEncoder.m
```diff
@@ -64,6 +70,7 @@ - (instancetype)initWithLogger:(ARTInternalLog *)logger delegate:(id<ARTJsonLike
         _rest = nil;
         _logger = logger;
         _delegate = delegate;
+        _timeProvider = [[ARTSystemTimeProvider alloc] init];
```
lawrence-forooghian: Why aren't we injecting this? (Feels like it would be a good moment to look into why this class has so many initializers, and to find out how they're all used and whether they can be consolidated. Maybe a piece of groundwork before the time work?)

Similar point for `ARTLog.m` which also doesn't inject the `_timeProvider`.

Source/PrivateHeaders/Ably/ARTLog+Private.h
```diff
@@ -19,6 +21,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (readonly) NSArray<ARTLogLine *> *captured;
 @property (readonly) NSArray<ARTLogLine *> *history;
 
+/**
+ The time provider used to stamp `ARTLogLine` entries. Defaults to an `ARTSystemTimeProvider`. Settable so that internal code (or tests) can install an alternative implementation, e.g. to make log line timestamps follow the fake-time clock used elsewhere in the SDK.
```
lawrence-forooghian: > internal code (…) can install an alternative implementation, e.g. to make log line timestamps follow the fake-time clock used elsewhere in the SDK.

As far as I can tell we're not setting this anywhere, meaning that when you create a client it doesn't use the injected `timeProvider`. Have I missed it somewhere?

Also I'm not sure it needs to be mutable; I think we can update the internal initializers to accept it. And just add a comment in the `-init` initializer (which for reasons unclear to me is public) explain that this initializer is public API and we can't change its signature but that internal components should inject

Source/PrivateHeaders/Ably/ARTRealtime+Private.h
```diff
@@ -59,6 +60,11 @@ NS_ASSUME_NONNULL_BEGIN
 /// Provides the implementation for `-[ARTPluginAPI nosync_latestConnectionDetailsForClient:]`. See documentation for that method in `APPluginAPIProtocol`.
 @property (nullable, readonly, nonatomic) ARTConnectionDetails *latestConnectionDetails;
 
+/// The time provider used by this client; sourced from `options.testOptions.timeProvider` at `-init` time.
+///
+/// Provides the implementation for `-[ARTPluginAPI wallClockNowForClient:]`, `-[ARTPluginAPI continuousClockNowForClient:]`, and `-[ARTPluginAPI scheduleForClient:after:queue:block:]`. See documentation for those methods in `APPluginAPIProtocol`.
+@property (nonatomic, readonly) id<ARTTimeProvider> timeProvider;
```
lawrence-forooghian: This isn't needed currently

Source/PrivateHeaders/Ably/ARTTestClientOptions.h
```diff
@@ -43,6 +44,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic) id<ARTJitterCoefficientGenerator> jitterCoefficientGenerator;
 
+/**
+ The unified time-provider abstraction consumed by all of ably-cocoa's time-dependent internal code. Tests install a fake-time implementation here to control wall-clock, continuous-clock, and scheduler behaviour. Initial value is an instance of `ARTSystemTimeProvider`.
```
lawrence-forooghian:
```suggestion
 The unified time-provider abstraction consumed by all of ably-cocoa's time-dependent internal code. Tests (currently only the UTS tests) install a fake-time implementation here to control wall-clock, continuous-clock, and scheduler behaviour. Initial value is an instance of `ARTSystemTimeProvider`.
```

Source/ARTContinuousClock.m
```diff
@@ -41,8 +25,21 @@ - (instancetype)initWithTime:(uint64_t)time {
     return self;
 }
 
-- (BOOL)isAfter:(ARTContinuousClockInstant *)other {
-    return self.timeInNanosecondsSinceClockReferenceInstant > other.timeInNanosecondsSinceClockReferenceInstant;
+- (BOOL)isAfter:(id<ARTContinuousClockInstantProtocol>)other {
+    // `other` is only ever an instant produced by the same `ARTContinuousClockInstant` (a single
```
lawrence-forooghian: it _should_ be but the whole reason we have a guard is that it technically might not be

```
    // `other` should in practice only ever be an instant produced by the same `ARTTimeProvider` 
    // (a single client has a single time provider, and instants aren't mixed across providers).
```

lawrence-forooghian (commented):
Have looked at the time stuff and left some comments. Also:

- I'd suggest rewriting the commit message for 2ea2f5d6e271bb86a3189410b94dbd99c8e50a03 to make sense in isolation without referring to WIP work that people aren't going to see. And the only way in which we need to mention plugin-api is just that these abstractions aren't yet exposed to plugins, and we'll figure out how to let them participate as part of AIT-781
- the later fixes to the time stuff should be squashed into it
- the time work is large enough that I think it can be in its own PR (btw the name for this branch is only about the time stuff anyway)

Will take a look at the UTS stuff now.

.claude/skills/uts-to-swift/SKILL.md
```diff
@@ -0,0 +1,420 @@
...
+## Step 2 — Determine output path and class name
+
+Map the spec path to a test path:
+
+| Spec location | Test location |
+|---|---|
+| `.../uts/rest/unit/<name>.md` | `Test/UTS/Tests/<Name>Tests.swift` |
+| `.../uts/realtime/unit/<sub>/<name>.md` | `Test/UTS/Tests/<Name>Tests.swift` |
+
+Class name: take the file name, strip a trailing `_test`, convert `snake_case` → `PascalCase`, append `Tests`. Example: `connection_recovery_test.md` → `ConnectionRecoveryTests`. All test classes are flat under `Test/UTS/Tests/` (no per-category subfolders); each is a **Swift Testing** suite — `@Suite(.serialized) final class <Name>Tests: UTSTestCase` (the suite framework is Swift Testing / `import Testing`, **not** XCTest; `.serialized` runs the suite's tests in order, matching the harness's per-test isolation).
```
lawrence-forooghian: > no per-category subfolders

Is there any reason not to mirror the directory structure of the spec `uts/` directory? Else we're going to have clashes e.g. `uts/realtime/integration/auth.md` and `uts/rest/integration/auth.md` will both map to `Tests/UTS/AuthTests.swift`.

.claude/skills/uts-to-swift/SKILL.md
```diff
@@ -0,0 +1,420 @@
...
+
+Apply the translation rules below, then write the file.
+
+### Client construction
+
+The mock is **installed first** (`installMock(_:)`), then the client is built with `makeRealtime`/`makeRest`. There is no default key — set `options.key` in the `configure` block. The harness already sets `autoConnect = false`, `useBinaryProtocol = false`, and dedicated internal/callback queues.
+
+| Pseudocode | Swift |
+|---|---|
+| `Rest(options: ClientOptions(key: "..."))` | `installMock(mockHTTP)` then `let rest = makeRest { $0.key = "..." }` |
```
lawrence-forooghian: I don't really understand this translation; `install_mock` is a separate step in the pseudo-code tests but this makes it sounds like we should indiscriminately include it when we translate `Rest(options: ClientOptions(key: "..."))`

.claude/skills/uts-to-swift/SKILL.md
```diff
@@ -0,0 +1,420 @@
...
+| `Realtime(options: ClientOptions(key: "...", autoConnect: false))` | `installMock(wsProvider)` then `let client = makeRealtime { $0.key = "..." }` |
+| `recover: "..."` / `disconnectedRetryTimeout: 0.5` | set on `options` inside the `configure` block |
+| `enable_fake_timers()` | `enableFakeTimers()` — **before** `makeRealtime`/`makeRest` (the clock is injected at construction) |
+
+### Mock setup — WebSocket
+
+Use the **handler pattern**: configure the simulated server in `onConnectionAttempt`. `respondWithSuccess()` opens the socket; `sendToClient(...)` delivers a message (two separate calls, unlike ably-java's combined `respondWithSuccess(msg)`).
```
lawrence-forooghian: Why are we talking about ably-java here?

.claude/skills/uts-to-swift/SKILL.md
```diff
@@ -0,0 +1,420 @@
...
+
+```swift
```
lawrence-forooghian: Given that this file is about how to translate UTS tests would it not be helpful, for such examples (here and elsewhere in this file), to show what is the pseudocode example that this is a translation of?

.claude/skills/uts-to-swift/SKILL.md
```diff
@@ -0,0 +1,420 @@
...
+        request.respondWith(status: 200, body: [1234567890000])
+    }
+)
+installMock(mockHTTP)
+let rest = makeRest { $0.key = "app.key:secret" }
+```
+
+### Capturing connection attempts / requests (the spec's local-array pattern)
+
+Per the spec ("Common Mistakes": use a **local** `captured_*` collector, not a mock property). The handler closures run on the SDK's queues, so the target is built in the Swift 6 language mode and the closures are `@Sendable` — a plain `var array` captured into them is a compile error (a data race). Use the harness's thread-safe `Captured<T>` instead: it's still test-local, but lock-guarded. Read it only after the operation has settled (after `awaitConnectionState` / the awaited completion):
```
lawrence-forooghian: > "The handler closures run on the SDK's queues, so the target is built in the Swift 6 language mode"

why "so"?

.claude/skills/uts-to-swift/SKILL.md
```diff
@@ -0,0 +1,420 @@
...
+| `ASSERT x IS NOT null` | `let x = try #require(optional)` (or `#expect(x != nil)`) |
+| `ASSERT x IS null` | `#expect(x == nil)` |
+| `ASSERT "k" IN map` / `NOT IN` | `#expect(map["k"] != nil)` / `#expect(map["k"] == nil)` |
+| `ASSERT list.length == N` | `#expect(list.count == N)` |
+| `ASSERT x == y (with message)` | `#expect(x == y, "message")` |
+| `AWAIT op FAILS WITH error` | capture the error (see async below) and `#expect(error.code == ...)` / `#expect(error.statusCode == ...)` |
+
+`#expect(...)` records a failure and continues; `try #require(...)` unwraps/asserts and stops the test on failure (use it when later lines depend on the value, like XCTUnwrap did).
+
+Async REST calls (e.g. `rest.time { ... }`) — make the `@Test` `async throws` and bridge the completion handler with a continuation rather than `XCTestExpectation`:
```
lawrence-forooghian: for this "rather than" and the next (XCTFail) do you think that it would try and use those anyway given that it says emphatically to use Swift Testing? it feels like these are artifacts of updating a document that previously said to use these constructs

Test/UTS/Harness/UTSTestCase.swift
```diff
@@ -0,0 +1,183 @@
...
+
+        if let timeProvider {
+            options.testOptions.timeProvider = timeProvider
+        }
+        options.testOptions.transportFactory = MockWebSocketTransportFactory(wsProvider: wsProvider)
+
+        configure(options)
+
+        let client = ARTRealtime(options: options)
+        client.internal.setReachabilityClass(NoOpReachability.self)
```
lawrence-forooghian: I would prefer that we not do this; `ARTTestClientOptions` is the place to inject things like this. Can you do this as a separate change please?

Test/UTS/Harness/FakeTimeProvider.swift
```diff
@@ -0,0 +1,174 @@
...
+///
+/// Both the wall clock and the continuous clock start at a fixed origin and only ever move when the test calls
+/// `advanceTime(byMilliseconds:)` — nothing fires on its own, so tests are fully in control
+/// (the UTS `enable_fake_timers()` / `ADVANCE_TIME(ms)` primitives).
+///
+/// Every block scheduled through the SDK's `scheduleAfter:queue:block:` is recorded here, which
+/// means this provider doubles as the timer-leak safety net the UTS derived-tests guide asks for:
+/// `cancelAllScheduled()` in teardown cancels every surviving timer, including ones the SDK may
+/// have orphaned and which are no longer reachable through its public API.
+final class FakeTimeProvider: NSObject, TimeProvider, @unchecked Sendable {
```
lawrence-forooghian: nit: <code><strong>Mock</strong>HTTP</code> and <code><strong>Mock</strong>WebSocket</code> but <code><strong>Fake</strong>TimeProvider</code>

Test/UTS/Harness/UTSTestCase.swift
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    // MARK: Install mock
+
+    /// Registers `wsProvider` to be used by the next `makeRealtime()` call (UTS `install_mock`).
+    func installMock(_ wsProvider: MockWebSocketProvider) {
+        installedWebSocketProvider = wsProvider
+    }
+
+    /// Registers `mockHTTP` to be used by the next `makeRest()` call (UTS `install_mock`).
```
lawrence-forooghian: Why only `makeRest()`? We have a bunch of Realtime specs that use `install_mock(mock_http)`:

```
uts/realtime/unit/auth/auth_callback_errors_test.md:install_mock(mock_http)
uts/realtime/unit/connection/connection_failures_test.md:install_mock(mock_http
)
uts/realtime/unit/connection/connection_failures_test.md:install_mock(mock_http
)
uts/realtime/unit/connection/connection_failures_test.md:install_mock(mock_http
)
uts/realtime/unit/connection/connection_failures_test.md:install_mock(mock_http
)
uts/realtime/unit/connection/connection_open_failures_test.md:install_mock(mock
_http)
uts/realtime/unit/connection/fallback_hosts_test.md:install_mock(mock_http)
uts/realtime/unit/connection/fallback_hosts_test.md:install_mock(mock_http)
```

Test/UTS/Harness/UTSTestCase.swift
```diff
@@ -0,0 +1,183 @@
...
+    /// shared `FakeTimeProvider`. `autoConnect` defaults to `false`, matching the UTS specs. A
+    /// provider must be installed first via `installMock(_:)`.
+    func makeRealtime(configure: (ARTClientOptions) -> Void = { _ in }, sourceLocation: SourceLocation = #_sourceLocation) -> ARTRealtime {
+        guard let wsProvider = installedWebSocketProvider else {
+            Issue.record("No MockWebSocketProvider installed — call installMock(_:) before makeRealtime()", sourceLocation: sourceLocation)
+            fatalError("No MockWebSocketProvider installed")
+        }
+
+        let options = ARTClientOptions()
+        options.autoConnect = false
```
lawrence-forooghian: Why do we set these client options by default? What do the UTS specs say about this option?

Test/UTS/Harness/UTSTestCase.swift
```diff
@@ -0,0 +1,183 @@
...
+    /// Builds an `ARTRest` whose HTTP layer is the currently installed `MockHTTP` (so requests are
+    /// intercepted, not sent over the network). A mock must be installed first via `installMock(_:)`.
+    func makeRest(configure: (ARTClientOptions) -> Void = { _ in }, sourceLocation: SourceLocation = #_sourceLocation) -> ARTRest {
+        guard let mockHTTP = installedMockHTTP else {
+            Issue.record("No MockHTTP installed — call installMock(_:) before makeRest()", sourceLocation: sourceLocation)
+            fatalError("No MockHTTP installed")
+        }
+
+        let options = ARTClientOptions()
+        options.useBinaryProtocol = false
```
lawrence-forooghian: Why do we set these client options by default? What do the UTS specs say about this option?

Test/UTS/Harness/UTSTestCase.swift
```diff
@@ -0,0 +1,183 @@
...
+        options.dispatchQueue = DispatchQueue(label: "io.ably.uts.callback.\(suffix)")
+
+        if let timeProvider {
+            options.testOptions.timeProvider = timeProvider
+        }
+
+        configure(options)
+
+        let rest = ARTRest(options: options)
+        rest.internal.httpExecutor = mockHTTP
```
lawrence-forooghian: I would prefer that we not do this; `ARTTestClientOptions` is the place to inject things like this (same as `transportFactory`). Can you do this as a separate change please?

Test/UTS/Harness/MockHTTP.swift
```diff
@@ -0,0 +1,157 @@
...
+/// The UTS `MockHttpClient` — a fake `ARTHTTPExecutor` that intercepts the SDK's outgoing HTTP
+/// requests so tests can observe them and inject responses, with no real network. Installed via
+/// `rest.internal.httpExecutor` (the cocoa mapping of the spec's `install_mock`).
+///
+/// Mirrors `uts/rest/unit/helpers/mock_http.md`. The cocoa HTTP seam is **request-level**
+/// (`executeRequest:completion:`), so the connection phase is modelled lightly: `onConnectionAttempt`
+/// fires once, before the first request, derived from that request's URL.
+/// `respond_with_refused`, `timeout`, `dns_error` there make every request fail with the corresponding `NSError`;
+/// a successful (or absent) connection handler lets requests through to `onRequest`.
+final class MockHTTP: NSObject, ARTHTTPExecutor, @unchecked Sendable {
```
lawrence-forooghian: might as well call it `MockHTTPClient` like in UTS

Test/UTS/Harness/MockHTTP.swift
```diff
@@ -0,0 +1,157 @@
...
+    typealias ConnectionHandler = @Sendable (PendingHTTPConnection) -> Void
+    typealias RequestHandler = @Sendable (PendingHTTPRequest) -> Void
+
+    private let onConnectionAttempt: ConnectionHandler?
+    private let onRequest: RequestHandler?
+
+    private let lock = NSLock()
+    private var connectionResolved = false
```
lawrence-forooghian: It's not clear to me why this class needs any mutable state; isn't each request a standalone thing?

Test/UTS/Harness/MockHTTP.swift
```diff
@@ -0,0 +1,157 @@
...
+
+        // Request phase — deliver to onRequest unless the connection failed
+        if let failure = withLock({ connectionError }) {
+            callback?(nil, nil, failure)
+            return NoopCancellable()
+        }
+        guard let onRequest else {
+            // No request handler installed: fail fast rather than leaving the SDK (and the test
+            // awaiting this request) hanging forever with the callback never invoked.
+            assertionFailure("MockHTTP received a request but no onRequest handler is installed")
```
lawrence-forooghian: Under what circumstances would test code proceed beyond this assertion failure to the callback code below? Feels like a `fatalError()` here would suffice if this is indeed a programmer error.

Test/UTS/Harness/MockHTTP.swift
```diff
@@ -0,0 +1,157 @@
...
+            return NoopCancellable()
+        }
+        onRequest(PendingHTTPRequest(request: request, completion: callback))
+        return NoopCancellable()
+    }
+}
+
+/// A connection attempt (UTS `PendingConnection`). The cocoa HTTP seam doesn't expose real TCP, so
+/// this is derived from the first request's URL and its outcome is applied to all requests.
+final class PendingHTTPConnection: @unchecked Sendable {
```
lawrence-forooghian: This class doesn't seem very `Sendable` to me (`resolvedError`).

Test/UTS/Harness/MockHTTP.swift
```diff
@@ -0,0 +1,157 @@
...
+    let host: String
+    let port: Int
+    let tls: Bool
+
+    fileprivate private(set) var resolvedError: NSError?
+
+    init(request: URLRequest) {
+        let url = request.url
+        self.tls = (url?.scheme?.lowercased() == "https")
+        self.host = url?.host ?? ""
```
lawrence-forooghian: Would you ever expect to receive a URL without a host? feels like something has gone wrong if we do

Test/UTS/Harness/MockHTTP.swift
```diff
@@ -0,0 +1,157 @@
...
+/// this is derived from the first request's URL and its outcome is applied to all requests.
+final class PendingHTTPConnection: @unchecked Sendable {
+    let host: String
+    let port: Int
+    let tls: Bool
+
+    fileprivate private(set) var resolvedError: NSError?
+
+    init(request: URLRequest) {
+        let url = request.url
```
lawrence-forooghian: would you ever expect to receive a request without a `url`? feels like something has gone wrong if you do

Test/UTS/Harness/MockHTTP.swift
```diff
@@ -0,0 +1,157 @@
...
+
+    func execute(_ request: URLRequest, completion callback: ((HTTPURLResponse?, Data?, Error?) -> Void)? = nil) -> (ARTCancellable & NSObjectProtocol)? {
+        // Connection phase — fire onConnectionAttempt once, before the first request
+        let firstConnection: PendingHTTPConnection? = withLock {
+            guard !connectionResolved else { return nil }
+            connectionResolved = true
+            return PendingHTTPConnection(request: request)
+        }
+        if let firstConnection, let onConnectionAttempt {
+            onConnectionAttempt(firstConnection)
```
lawrence-forooghian: We seem to have two different models here; you have to respond to `onConnectionAttempt()` inside the callback (which, by the way, if this is the case, should be enforced i.e. calling it after the callback should be a programmer error), whereas `onRequest()` lets you escape the `PendingHTTPRequest` from the callback and call its `respond*` methods later. In the UTS tests, is there a requirement to be able to escape a `PendingHTTPConnection` from the callback and respond to it later or is the fact that you can do this for `PendingHTTPRequest` just incidental?

lawrence-forooghian (commented):
I've started going through this but there's a lot here. Things I've not yet looked at:

- `MockWebSocket`, in any detail
- any of the translated tests
- `MockTimeProvider`

I've tried to focus on the overall infrastructure that you've set up (the test target, the skill, the test case, the interfaces provided by the mocks), hoping that even if the implementation is dodgy (and some of it seems hard to reason about particularly when it comes to thread-safety given that it uses a lot of locking instead of making more use of the SDK's existing synchronisation mechanism) we can sort that out later.

I don't want to spend _loads_ more time reviewing this so if I look at it more it'll probably be focussed on the translated tests to get an idea of how they've been translated. Is there anything you'd like particular feedback on? In general, it seems like the skill and the `UTSTestCase` provide a good initial foundation, and if there are issues I hope they'll come up as we try to translate further tests. Is there anything outstanding that you think it'd be worth spending further upfront time on now?

Test/UTS/Tests/TimeTests.swift
```diff
@@ -0,0 +1,183 @@
...
+
+/// Time API (RSC16)
+/// Derived from https://github.com/ably/specification/blob/main/uts/rest/unit/time.md
+@Suite(.serialized)
+final class TimeTests: UTSTestCase {
+
+    // UTS: rest/unit/RSC16/returns-server-time-0
+    @Test
+    func test_RSC16_time_returns_server_time() async throws {
+        let serverTimeMs = 1_704_067_200_000 // 2024-01-01 00:00:00 UTC
```
lawrence-forooghian: I'd find it useful, when comparing this to the markdown spec at a glance, to have comments for the subheadings; "Setup", "Test Steps", "Assertions" etc.

Test/UTS/Tests/TimeTests.swift
```diff
@@ -0,0 +1,183 @@
...
+        let error = try #require(await awaitTimeError(rest))
+        #expect(error.statusCode == 500)
+        #expect(error.code == 50000)
+    }
+}
+
+/// Test case helpers
+extension TimeTests {
+
+    /// Awaits `rest.time(...)`, returning the result (records an issue if it errors).
```
lawrence-forooghian: It'd be useful for the comments on the test case helpers to give an example of what is the UTS psuedocode that they provide a translation of.

Test/UTS/Tests/TimeTests.swift
```diff
@@ -0,0 +1,183 @@
...
+    private func awaitTime(_ rest: ARTRest, sourceLocation: SourceLocation = #_sourceLocation) async -> Date? {
```
lawrence-forooghian: Feels like neither this nor `awaitTimeError` should return an optional.

Source/ARTPluginAPI.m
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/fe47839183ff95f97044d0dd1ccb1b3ebb7bcbd0 — Remove unnecessary ARTPluginAPI time forwarders

Source/PrivateHeaders/Ably/ARTContinuousClockInstantProtocol.h
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/e2307f12367ca36bd3d2cbf80b809b598dc0cac4 — Rename the continuous-clock instant protocol and hide its impl

Source/PrivateHeaders/Ably/ARTContinuousClock.h
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/e7bd1ce282d68a317e7d4dd06d06d4ebda16ccaa — Remove misplaced NS_SWIFT_NAME note from ARTContinuousClockInstant

Source/PrivateHeaders/Ably/ARTContinuousClock.h
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/e2307f12367ca36bd3d2cbf80b809b598dc0cac4 — Rename the continuous-clock instant protocol and hide its impl

CONTRIBUTING.md
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/245154902d8a30f6b3895f71046ef9ca881af669 — Reword CONTRIBUTING time-provider convention to avoid 'threading'

Test/AblyTests/Tests/ContinuousClockTests.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/890d6ba9c04c8ebd28997195387dd0967f5c11ee — Port ContinuousClockTests to ARTSystemTimeProvider

Test/AblyTests/Tests/GCDTests.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/96fa51702a1200de901681a044b0b47965dd92f9 — Move ARTScheduledBlockHandle into ARTSystemTimeProvider.m

Source/ARTGCD.m
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/77447962a0e70ed48057fbcadf981792d86399ce — Restore lost ARTScheduledBlockHandle block-retention comment

Source/ARTJsonLikeEncoder.m
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/120817945bb4b556bb2f0c81ef6dd40c78e9ef51 — Remove unused `ARTJsonLikeEncoder initWithLogger:delegate:`

The only place where `- [ARTJsonLikeEncoder initWithDelegate:...]` is used is in the static `+ [ARTMessage fromEncoded:...]`. I'm not sure if `_timeProvider` is needed there since timestamps in the message are being created via `+ [NSDate artDateFromNumberMs:]`. Replace them as well?

Source/PrivateHeaders/Ably/ARTLog+Private.h
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/28214a819b9d6eadab9e5c558be391495cd91ae2 — Remove ARTLog's settable timeProvider property

I'm not sure about mocking timestamps in log lines. Should we? IMHO log timestamps have utilitarian function, for analysis mocks should be embed into log lines body if needed, wdyt?

Source/PrivateHeaders/Ably/ARTRealtime+Private.h
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/16a6cc66119597e0f854820105eaa897db881c92 — Move ARTRealtimeInternal.timeProvider out of the private header

Source/PrivateHeaders/Ably/ARTTestClientOptions.h
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/6613af0f8e641cdcec2329e6a8f2eee288d96dff — Note in ARTTestClientOptions doc that only UTS tests use timeProvider

Source/ARTContinuousClock.m
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/5fd3442d2ae7f40759133496d38e53390fa98bff — Reword ARTContinuousClock isAfter guard comment

.claude/skills/uts-to-swift/SKILL.md
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/bd37359621ad99ebb44d88c9977ba39f48a0ed95 — Skill: mirror the spec uts/ directory layout for test paths

.claude/skills/uts-to-swift/SKILL.md
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/fe9fbff3de079e831abe86763e0454007239a21a — Skill: drop ably-java reference in the WebSocket handler note

.claude/skills/uts-to-swift/SKILL.md
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/4431cb57c9aeeaff3f1e9eeeedb577dfb2635d45 — Skill: drop the 'rather than XCTestExpectation/XCTFail' framing

Test/UTS/Harness/UTSTestCase.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/eb3f40e64305b5cfcc57499ee6225d9068b6fa53 — Add ARTTestClientOptions.reachabilityClass injection point

Test/UTS/Harness/FakeTimeProvider.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/edf9d2d5e522b53dc32f3b787668d05485cf0498 — Rename FakeTimeProvider to MockTimeProvider

Test/UTS/Harness/UTSTestCase.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/16ed533f9633a1e0ff1fd43e54a919f1f1055221 — Wire installed MockHTTPClient into makeRealtime's REST layer

Test/UTS/Harness/UTSTestCase.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/324508cfd592550df3e93b754aa5c81d2a1e9fa4 — Don't default autoConnect=false in the UTS harness

Test/UTS/Harness/UTSTestCase.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/472e733e629fc96ef03d70e56333b999a423c699 — Inject UTS HTTP executor via ARTTestClientOptions

Test/UTS/Harness/MockHTTP.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/27cb96b937af9cc920775c4ba042a6aea6daafa8 — Rename MockHTTP to MockHTTPClient

Test/UTS/Harness/MockHTTP.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/5eb88258697e30a52a3519bf6f5d51db2a1c3b84 — Model the HTTP mock per-request, with no shared mutable state

Test/UTS/Harness/MockHTTP.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/2ee65fc8ba373395c2a857bc98b9b2f0123e451c — MockHTTPClient: fatalError when no onRequest handler is installed

Test/UTS/Harness/MockHTTP.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/5eb88258697e30a52a3519bf6f5d51db2a1c3b84 — Model the HTTP mock per-request, with no shared mutable state

Test/UTS/Harness/MockHTTP.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/79c676c0f74a12a42c4cc1f8d591cfbf6a15bef9 — Fail fast on a connection attempt without a URL host

Test/UTS/Harness/MockHTTP.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/b322c1d4a51634edfbb58921cd0e8a80051412ea — MockHTTPClient: fail fast on a request without a URL

Test/UTS/Harness/UTSTestCase.swift
maratal: Resolved (no code change) — `useBinaryProtocol = false` is required because the mocks return JSON while cocoa defaults to msgpack (per the spec's binary-vs-JSON note), so it stays a harness-level default - https://github.com/ably/specification/blob/main/uts/docs/writing-derived-tests.md#required-options-vary-by-sdk

Test/UTS/Harness/MockHTTP.swift
maratal: Partially addressed in https://github.com/ably/ably-cocoa/commit/5eb88258697e30a52a3519bf6f5d51db2a1c3b84 — Model the HTTP mock per-request, with no shared mutable state

Claude response:
```
MockHTTP exposed two inconsistent interaction models for the two handlers:

  - onConnectionAttempt(PendingHTTPConnection) — In the original code you had to respond synchronously, inside the callback. execute read the result back the instant the handler returned (withLock { 
  connectionError = firstConnection.resolvedError }). So if you stashed the PendingHTTPConnection and called respondWithRefused() later, it would do nothing — the outcome had already been read. His
  point: if synchronous response is required, that should be enforced — calling respondWith* after the callback returns ought to be a programmer error, not silently ignored.
  - onRequest(PendingHTTPRequest) — Here you can escape the object from the callback: PendingHTTPRequest holds the SDK's completion closure, so you can capture it and call respondWith(...) later (async).

  So one handler is "respond now or never" and the other is "respond whenever." His actual question: in the UTS tests, is there a real need to escape a PendingHTTPConnection and respond to it later — or
  is PendingHTTPRequest's escapability just incidental? In other words, the two should probably be one consistent model.

  How it stands now (after the per-request rewrite, 5eb88258)

  The footgun he flagged is gone, because the connection model changed shape:

  - onConnectionAttempt is now @Sendable (PendingHTTPConnection) -> NSError? — you respond by returning the error (or nil). PendingHTTPConnection.respondWith* are now pure functions returning NSError?;
  they have no side effects. You can still capture the struct, but calling respondWithRefused() later just produces a value that goes nowhere — there's no silent "I thought I rejected the connection but
  it was ignored" trap, because the only way to communicate an outcome is the return value. So the synchronous contract is now enforced by construction, which is exactly what he asked for.
  - onRequest is unchanged: PendingHTTPRequest still holds the completion closure and is escapable.

  So the asymmetry is now intentional and safe, rather than accidental and footgun-prone:
  - connection outcome = decided synchronously (return value),
  - request response = can be deferred (escapable).
  
  Is that the right split? (the remaining WDYT)

  Grounded in actual usage:
  - Connection: no test escapes it; the outcome (success/refuse/timeout/dns) is always known up front. There's no requirement to defer a connection response — so the synchronous return model fits.
  - Request: the escapability is used — e.g. TimeTests does capturedRequests.append(request) and asserts request.method/request.url after the call settles. (The response there happens inline, but the
  object is genuinely escaped for later inspection.) Deferred responses are also the natural HTTP-mock pattern (inspect/assert, advance time, then respond).
  
  So the honest answer to Lawrence is: there's no requirement to escape a PendingHTTPConnection; its outcome is always synchronous, and the code now enforces that via the return value. PendingHTTPRequest
  is deliberately escapable because request inspection/response genuinely benefits from it (and the tests use that). The two models differ on purpose, matching how connections vs. requests are actually
  used.
```

Test/UTS/Tests/TimeTests.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/b81f97b5d47520e123622932cda69d1f128e30d0 — Add spec-section comments (Setup / Test Steps / Assertions) and align inline comments with the spec wording.

Test/UTS/Tests/TimeTests.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/287957775b0ecd9cae5603ef4fd177012111b1be — Document on each helper the UTS step it translates (result = AWAIT client.time() / AWAIT client.time() FAILS WITH error).

Test/UTS/Tests/TimeTests.swift
maratal: Addressed in https://github.com/ably/ably-cocoa/commit/eef84f8e1113e992a65aa98457d0f6d0a8fd9cb6 — Make awaitTime / awaitTimeError non-optional — they #require internally and return Date / ARTErrorInfo.

Source/PrivateHeaders/Ably/ARTLog+Private.h
lawrence-forooghian: I don't see a compelling reason to mock timestamps in log lines, no. But I'd add a comment alongside the `_timeProvider` setting in the initializer:

```objective-c
        // Note that we don't provide the ability to mock the time provider used for log timestamps; doing so is more likely to cause confusion to those reading log messages than it is to be useful.
        _timeProvider = [[ARTSystemTimeProvider alloc] init];
```

Source/ARTJsonLikeEncoder.m
lawrence-forooghian: I think that `ARTJsonLikeEncoder` will be easier to reason about if it _always_ has a valid `_timeProvider`, regardless of whether it ends up being used on a given code path.

I'd suggest:

- mark `-[ARTJsonLikeEncoder init]` as unavailable and get rid of its implementation
- update `-initWithDelegate` to accept a `timeProvider` arg
- in the two class methods that call `initWithDelegate:`, create and pass an `ARTSystemTimeProvider` explicitly

>  I'm not sure if _timeProvider is needed there since timestamps in the message are being created via + [NSDate artDateFromNumberMs:]. Replace them as well?

You mean replace the calls to `artDateFromNumberMs:` with one to `[_timeProvider wallClockNow]`? I don't think that makes sense.

maratal: Done in https://github.com/ably/ably-cocoa/commit/e4fc9cbaff1059a323f85bab69a41a6996a54ca2 — `-init` is now `NS_UNAVAILABLE`, `initWithDelegate:` takes a `timeProvider:`, and the two static `ARTMessage` decoders pass an explicit `ARTSystemTimeProvider`; the rest-backed init still sources its provider from `testOptions`. Re the `artDateFromNumberMs:` calls — agreed, left as-is.

maratal: fixed - e4fc9cba

Source/PrivateHeaders/Ably/ARTLog+Private.h
maratal: added - 427bda0f

CONTRIBUTING.md
maratal: removed - 253e362c

lawrence-forooghian (approved):

PR #2215, opened 2026-06-17
Remove `logExceptionReportingUrl` client option

### LINKED ISSUE #2025: Remove not used var
Looks like `_logExceptionReportingUrl` was forgotten in declarations and is not needed anymore.

lawrence-forooghian:
Setting this option has been a no-op since 4ccc7b5. Technically we should wait until a major release to remove this option, but we've had a customer report ([Slack thread](https://ably-real-time.slack.com/archives/CURL4U2FP/p1781515415287729)) that the default value has triggered an automated security check due to using the `scheme://user:password@host` URL format. I think we can just remove the option; it seems unlikely anybody was _ever_ setting it to a non-default value, and even if they were and still are, a compilation failure will make it clear to them that doing so is achieving nothing.

Resolves #2025.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed the deprecated exception reporting URL configuration option from client options
  * Applications currently using this configuration setting will require updates

### DIFF
--- a/Source/ARTClientOptions.m
+++ b/Source/ARTClientOptions.m
@@ -56,7 +56,6 @@ NSString *ARTDefaultEnvironment = nil;
     _httpMaxRetryCount = 3;
     _fallbackHosts = nil;
     _fallbackHostsUseDefault = false;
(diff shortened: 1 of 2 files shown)


maratal (approved):
LGTM

PR #2218, opened 2026-07-02
[AIT-1023] Initial liveobjects path-based public API surface

maratal:
This is preview for https://ably.atlassian.net/browse/AIT-1023 and not intended to be merged yet.

New public path-based API (this was used as a Claude prompt sanitized from my mumbling):

1. Analyze changes in the public API surface in these two PRs:

Main changes - https://github.com/ably/specification/pull/485/changes
Renaming - https://github.com/ably/specification/pull/489/changes

Docs:
https://ably.com/docs/liveobjects
https://ably.com/docs/liveobjects/concepts/path-object
https://ably.com/docs/liveobjects/concepts/instance

Note that we don't do Type inference yet.

2. Here is an idea of how user-facing interfaces should look like:

```
class PathObject: // RTPO*
  path() -> String // RTPO4
  instance() -> Instance? // RTPO8
  compactJson() -> JsonElement/Any? // RTPO14
  subscribe((PathObjectSubscriptionEvent) -> listener, PathObjectSubscriptionOptions? options) -> Subscription // RTPO19

LiveMapPathObject extends PathObject
  get(String key) -> PathObject // RTPO5
  at(String path) -> PathObject // RTPO6
  entries() -> [String, PathObject][] // RTPO9
  keys() -> String[] // RTPO10
  values() -> PathObject[] // RTPO11
  set(String key, (Boolean | Binary | Number | String | JsonArray | JsonObject | LiveCounterValueType | LiveMapValueType) value) => io // RTPO15
  remove(String key) => io // RTPO16
  size() -> Number? // RTPO12

LiveCounterPathObject extends PathObject
  increment(Number amount?) => io // RTPO17
  decrement(Number amount?) => io // RTPO18
  value() -> Number

class Instance:
  compactJson() -> JsonElement/Any?

LiveMapInstance extends Instance
  id: String // RTINS3
  get(String key) -> Instance? // RTINS5
  entries() -> [String, Instance][] // RTINS6
  keys() -> String[] // RTINS7
  values() -> Instance[] // RTINS8
  set(String key, (Boolean | Binary | Number | String | JsonArray | JsonObject | LiveCounterValueType | LiveMapValueType) value) => io // RTINS12
  remove(String key) => io // RTINS13
  size() -> Number? // RTINS9
  subscribe((InstanceSubscriptionEvent) -> listener) -> Subscription // RTINS16, RTINS16c

LiveCounterInstance extends Instance
  id: String // RTINS3
  increment(Number amount?) => io // RTINS14
  decrement(Number amount?) => io // RTINS15
  value() -> Number
  subscribe((InstanceSubscriptionEvent) -> listener) -> Subscription // RTINS16, RTINS16c

public enum Primitive: Sendable {
    case string(String)
    case number(Double)
    case bool(Bool)
    case data(Data)
    case jsonArray([JSONValue])
    case jsonObject([String: JSONValue])

    /// If this `Primitive` has case `string`, this returns the associated value. Else, it returns `nil`.
    public var stringValue: String? {
        if case let .string(value) = self {
            return value
        }
        return nil
    }
... etc ...
}
```

3. Here is current public API that you should update:

- `/../PublicTypes.swift` (file from `ably-liveobjects-swift-plugin` repo)
- `/../JSONValue.swift` (file from `ably-liveobjects-swift-plugin` repo)

This API becomes internal, user only should access to path objects or instances derived from them.


ably-liveobjects-swift/Public/Instance.swift
```diff
@@ -0,0 +1,120 @@
...
+/// Unlike ``PathObject``, which is path-addressed and re-resolves on each call, an `Instance` is
+/// identity-addressed: it follows the specific object it was created with, regardless of where that
+/// object sits in the graph.
+///
+/// An `Instance` is obtained from ``PathObject/instance()``. `Instance` is loosely typed; use
+/// ``asLiveMap()``, ``asLiveCounter()`` or ``asPrimitive()`` to obtain a view with the methods
+/// applicable to a particular expected type.
+///
+/// Spec: `RTINS1`.
+public protocol Instance: Sendable {
```
sacOO7, line 16: AFAIK, `Instance` is a `enum` representing union of Instance sub-types for swift.
So we don't need `asLiveMap`, `asLiveCounter` and `asPrimitive` methods on it, 
performing `switch (instance)` should give all possible types.
This is only applicable for `Instance` since it's fixed and underlying object type doesn't change dynamically unlike `PathObject`

ably-liveobjects-swift/Public/PathObject.swift
```diff
@@ -0,0 +1,151 @@
...
+/// navigating with ``LiveMapPathObject/get(key:)`` or ``LiveMapPathObject/at(path:)``.
+///
+/// `PathObject` is loosely typed. To obtain a view with the methods applicable to a particular
+/// expected type, use ``asLiveMap()``, ``asLiveCounter()`` or ``asPrimitive()``. These do not
+/// guarantee that the value actually at the path has that type; that can only be determined when the
+/// value is evaluated (e.g. via `value()`), at which point `nil` is returned if the actual type
+/// differs from the one requested.
+///
+/// Spec: `RTPO1`.
+public protocol PathObject: Sendable {
```
sacOO7, line 23: We also have `type` property on `PathObject`, based on the result we call `asLiveMap`, `asLiveCounter`, `asPrimitive` methods on it

sacOO7: Check `Typed-SDK variant` at https://github.com/ably/specification/pull/491/changes

ably-liveobjects-swift/Public/Instance.swift
sacOO7: This is specific to `swift` since we can't represent union types in `java`/ `kotlin` effectively

sacOO7: So, `compactJson` method can go into specific `subtype` returning type specific `JsonType`, `Instance` will become just a marker class

ably-liveobjects-swift/Default/DefaultInstance.swift
lawrence-forooghian, line 1: Is there anything that you wished to demonstrate with these `Default*` types? Or are they just placeholders? If the latter probably can just remove them and make the single entrypoint (`channel.object`) just `fatalError`?

maratal: Only `DefaultRealtimeObject` is a stub for the entry point. Everything else will be filled with an implementation, f.e. `DefaultInstance` is a base class for `DefaultLiveCounterInstance`.

lawrence-forooghian: Yeah but none of it is necessary in a PR for reviewing the public API really, I can just ignore when reviewing?

maratal: done - https://github.com/ably/ably-cocoa/pull/2218/commits/5cbe0ff06677689e4c838ea366546346f7c79b34

lawrence-forooghian:
I've got Claude to do an initial review of this, and there's also a document where I wrote my thoughts on each of the points it raised (see the `LAWRENCE:` sections in the note)

https://github.com/ably/ably-cocoa/tree/lawrence-review-path-based-api/review

maratal:
See https://github.com/ably/ably-liveobjects-swift-plugin/pull/134

PR #2219, opened 2026-07-06
Add API for registering Live Activity push-to-start token

ttypic:
### Changes Introduced

`registerPushToStartToken`: Allows the addition of a token to an already-activated device and syncs the updated recipient to Ably using the existing PATCH path.

### Example App

https://github.com/ttypic/liveactivity-example

### Docs

https://github.com/ably/docs/pull/3464

## Summary by CodeRabbit

* **New Features**
  * Added support for registering and syncing push-to-start tokens for Live Activities on iOS.
  * Existing device storage now recognizes this token type and restores it when available.
* **Bug Fixes**
  * Resetting push device details now clears push-to-start token state along with other APNS token types.
* **Tests**
  * Added coverage for saving, loading, and sending push-to-start tokens during device registration updates.

### DIFF
--- a/Source/include/Ably/ARTPush.h
+++ b/Source/include/Ably/ARTPush.h
@@ -92,6 +92,16 @@ NS_ASSUME_NONNULL_BEGIN
 /// Call this method if you got an error calling `CLLocationManager.startMonitoringLocationPushes(completion:)`.
 + (void)didFailToRegisterForLocationNotificationsWithError:(NSError *)error realtime:(ARTRealtime *)realtime;
 
+// Live Activity push-to-start token
+
+/**
+ * Registers a Live Activity push-to-start token (obtained from `Activity.pushToStartTokenUpdates`) with Ably, adding it to the existing device registration. Use this when the device has already been activated (i.e. `-activate` has completed). The token is stored in `ARTLocalDevice`'s push recipient and synced to Ably.
+ * You should implement `-[ARTPushRegistererDelegate didUpdateAblyPush:]` to handle success or failure of this operation.
+ *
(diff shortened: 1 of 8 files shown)


maratal:
From the app description:

```
iOS app in this example does not integrate the Ably SDK
```

What do you mean? It calls `push.activate()`, right? @ttypic

Source/include/Ably/ARTPush.h
```diff
@@ -92,6 +92,16 @@ NS_ASSUME_NONNULL_BEGIN
 /// Call this method if you got an error calling `CLLocationManager.startMonitoringLocationPushes(completion:)`.
 + (void)didFailToRegisterForLocationNotificationsWithError:(NSError *)error realtime:(ARTRealtime *)realtime;
 
+// Live Activity push-to-start token
+
+/**
+ * Registers a Live Activity push-to-start token (obtained from `Activity.pushToStartTokenUpdates`) with Ably, adding it to the existing device registration. Use this when the device has already been activated (i.e. `-activate` has completed). The token is stored in `ARTLocalDevice`'s push recipient and synced to Ably.
```
lawrence-forooghian, line 98: > Use this when the device has already been activated (i.e. `-activate` has completed)

I think that this is an unfortunate requirement and pushes complexity onto callers. However, having gotten Claude to look into it, it seems like the same is true for location pushes, and at least here we're being honest about the limitation. I've created https://github.com/ably/ably-cocoa/issues/2220 for tracking this and I think it would be good to remove this limitation for all supplementary push types in the future.

lawrence-forooghian (approved):

ttypic:
> iOS app in this example does not integrate the Ably SDK

@maratal I just forgot to update readme, original example was without ably integration, I then replaced it

PR #2222, opened 2026-07-10
Merge LiveObjects plugin and plugin-support into ably-cocoa

lawrence-forooghian:
Resolves [AIT-1137](https://ably.atlassian.net/browse/AIT-1137) (split from [AIT-1027](https://ably.atlassian.net/browse/AIT-1027)): merges the contents of [ably-liveobjects-swift-plugin](https://github.com/ably/ably-liveobjects-swift-plugin) (at `a3bf63d`) and [ably-cocoa-plugin-support](https://github.com/ably/ably-cocoa-plugin-support) (at `a290b89`) into this repository, so that ably-cocoa and its plugins live in a single repo (as with ably-js and ably-java).

## Background

[ADR-128](https://ably.atlassian.net/wiki/spaces/ENG/pages/3838574593/ADR-128+Plugins+for+ably-cocoa+SDK) placed the LiveObjects plugin in a separate repository on the belief that, because SPM platform requirements are package-wide, a plugin needing modern OS versions could not live in the same package as ably-cocoa and its much older platform floor. That belief turned out to be wrong: platform requirements set only the *floor*, and per-declaration `@available` annotations let individual components of a package require newer OS versions (with the compiler enforcing that consumers on older OS versions guard their use). That removes the reason for the separate repos, whose cost — every cross-cutting change (e.g. the UTS groundwork, AIT-1027) rippling across three repos and three PRs — motivated this merge.

## Scope

The plugin is integrated as a largely self-contained directory (`LiveObjects/`) rather than being unified with the core SDK's tooling — the two keep their separate Xcode workspaces, lint setups (SwiftLint/SwiftFormat vs EditorConfig), CI workflows, test harnesses, example apps and dev docs. That's deliberate: unifying any of these would balloon the scope of an already-large change and destroy the verbatim-import provenance that makes it reviewable. The plugin-support layer (`_AblyPluginSupportPrivate`) is likewise kept exactly as it was, including machinery that arguably no longer needs to exist — e.g. the handshake by which the plugin and the core SDK check that they agree on the Realtime protocol version, which cannot diverge now that both sides live in the same repo and version. Note that one seemingly obvious simplification — dropping plugin-support entirely and having the plugin `import Ably.Private` (the Clang submodule that the test suite uses) — is a non-direction: as noted in ADR-128's "Alternatives considered", when this was tried, the private API appeared to be transitively exposed to the plugin's own users (`internal import` did not hide it). The repo merge is the enabler, not the end state — we can integrate further incrementally as pain points show up in the development workflow (candidates: a single workspace, shared lint tooling, folding the LiveObjects tests into a unified test setup, a combined example app).

**Review commit by commit.** The sequence is: commits 1–5 internalise plugin-support as an in-repo target (staying on tools 5.3); commits 6 onwards raise swift-tools-version to 6.1 and integrate LiveObjects. The two import commits are verbatim, script-generated copies of the source repos at a stated SHA (`Scripts/import-repo-contents.sh`); every deviation from verbatim is its own commit with the reasoning in its message. The `@available` annotation commit is entirely script-generated (`Scripts/annotate-liveobjects-availability.py`).

## Key decisions

Full rationale is in the relevant commit messages:

- ably-cocoa's declared platforms are **unchanged**; the LiveObjects module declares its newer OS requirements via `@available` on every top-level declaration (script-enforced in CI).
- SPM consumers will require Xcode 16.3+ (tools 6.1). CocoaPods/Carthage consumers unaffected. See the tools-bump commit message for the consumer-impact analysis and the documented escape hatch (a version-specific-manifest retrofit) if this bites anyone.
- `_AblyPluginSupportPrivate` becomes an internal target, deliberately not vended as a product.
- LiveObjects is documented as experimental and exempt from semver while the `Ably` product keeps full guarantees.
- One ably-common submodule (the plugin's is dropped; its pin was an ancestor of ours).

The branch is 19 commits, including the LiveObjects CI workflow (ported from the plugin repo, plus new availability-annotation enforcement jobs). Local verification of the full branch: all 290 LiveObjects tests pass (including sandbox integration tests) via both SPM and BuildTool/xcodebuild test plans, on macOS and tvOS simulator; example app builds; full lint passes; docs build via `xcodebuild docbuild`; `-warnings-as-errors` build clean.

One structural discovery from the workflow's first CI runs, fixed in the final two commits: xcodebuild compiles the test target at (near) the package's platform floor, the test sources cannot carry `@available` annotations (the swift-testing macros reject them), so BuildTool now raises the deployment target of test builds via build-setting overrides — see those commits' messages for the full story.

## After merging

1. Reopen the plugin repo's pending PRs against this repo: [ably-liveobjects-swift-plugin#131](https://github.com/ably/ably-liveobjects-swift-plugin/pull/131) ([ECO-5658] Switch to something simpler for subscription storage) and [ably-liveobjects-swift-plugin#133](https://github.com/ably/ably-liveobjects-swift-plugin/pull/133) (Implement a couple of recent spec-clarity changes).
2. Open [ably-liveobjects-swift-plugin#134](https://github.com/ably/ably-liveobjects-swift-plugin/pull/134) ([AIT-1023] Path-based public API) as a new PR here, into a new `feature/path-based-liveobjects` integration branch that is branched from `integration/liveobjects`.

## Release coordination

To be done before/alongside the first ably-cocoa release containing this:

1. Cut a final release of the standalone plugin whose manifest caps its ably-cocoa dependency with an upper bound (`..< <first-integrated-version>`) and whose README carries a deprecation notice pointing here. Number it 0.4.1, so that users on `from: "0.4.0"` pick up the cap on a routine update, after which SwiftPM keeps resolving the last working pair (and, if they later deliberately bump ably-cocoa, they get a clear version conflict that names the plugin as the blocker).
2. Quote the following in the migration notes (ably-cocoa's CHANGELOG entry and the archived plugin repo's README), so that searching for it finds the fix. It is what users who bump ably-cocoa without ever taking the capped plugin release will see — an immediate error at package-graph load (verified empirically):
   ```
   error: multiple packages ('ably-cocoa', 'ably-cocoa-plugin-support') declare targets with a conflicting name: '_AblyPluginSupportPrivate'; target names need to be unique across the package graph
   error: multiple packages ('ably-cocoa', 'ably-liveobjects-swift-plugin') declare targets with a conflicting name: 'AblyLiveObjects'; target names need to be unique across the package graph
   ```
   The fix being: replace the plugin package dependency with this package's `AblyLiveObjects` product; module name and API are unchanged.
3. Archive ably-liveobjects-swift-plugin and ably-cocoa-plugin-support (transferring any open issues here). Their published releases remain resolvable indefinitely.
4. Once this work lands on `main`, add this repo's LiveObjects workflow `all-checks-completed` job to `main`'s branch-protection required status checks.
5. Update the ably.com docs (both in the docs repo):
   - the [LiveObjects Swift quickstart](https://ably.com/docs/liveobjects/quickstart/swift) (`src/pages/docs/liveobjects/quickstart/swift.mdx`) installs the standalone ably-liveobjects-swift-plugin package; it should instead install ably-cocoa and select the `AblyLiveObjects` product;
   - the LiveObjects nav's Swift API-reference link (`src/data/nav/liveobjects.ts`) points at the generated docs' old location, `https://sdk.ably.com/builds/ably/ably-liveobjects-swift-plugin/main/AblyLiveObjects/documentation/ablyliveobjects/`; this workflow now publishes them at `https://sdk.ably.com/builds/ably/ably-cocoa/main/AblyLiveObjects/documentation/ablyliveobjects/`.

## Release timing

This PR targets the `integration/liveobjects` branch rather than `main`. Releasing the repo move on its own would mean two disruptive changes for users of the current plugin (first switching to getting it from ably-cocoa, then adopting the path-based API later), so instead the repo move and the path-based API work will accumulate on the integration branch and be released to `main` together as a single release with a simpler narrative: "LiveObjects has a new API and now lives in ably-cocoa". Meanwhile `main` remains freely releasable for the core SDK. (This also means the docs updates above happen once, as part of documenting the new API.)

[AIT-1027]: https://ably.atlassian.net/browse/AIT-1027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AIT-1137]: https://ably.atlassian.net/browse/AIT-1137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ECO-5658]: https://ably.atlassian.net/browse/ECO-5658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lawrence-forooghian:
@coderabbitai review

Docs/plugins.md
```diff
@@ -2,18 +2,18 @@
...
 
-I will expand on this documentation once this mechanism is more mature; currently [ADR-128: Plugins for ably-cocoa SDK](https://ably.atlassian.net/wiki/spaces/ENG/pages/3838574593/ADR-128+Plugins+for+ably-cocoa+SDK) provides a decent description of our approach. The key divergence between ADR-128 and our current approach is that `_AblyPluginSupportPrivate` (called `AblyPlugin` in the ADR) is no longer part of ably-cocoa and instead lives in a separate repository. I will update that ADR or create a new one later on as part of clearing technical debt caused by this last-minute change of approach.
+I will expand on this documentation once this mechanism is more mature; currently [ADR-128: Plugins for ably-cocoa SDK](https://ably.atlassian.net/wiki/spaces/ENG/pages/3838574593/ADR-128+Plugins+for+ably-cocoa+SDK) provides a decent description of our approach. The key divergence between ADR-128 and our current approach is that `_AblyPluginSupportPrivate` (called `AblyPlugin` in the ADR) is a standalone library that does not depend on ably-cocoa (for a while it also lived in a separate repository, before being merged back into this one).
```
lawrence-forooghian: I think it's probably no longer entirely true to say that ADR-128 provides a decent description of our approach since it's also centered around having the plugin live in a separate repo for reasons that have turned out to be wrong

LiveObjects/CONTRIBUTING.md
```diff
@@ -147,21 +149,8 @@ Some of our integration tests require access to ably-cocoa internals that are no
 
 ## Developing ably-cocoa alongside this plugin
 
-The quickest way to edit ably-cocoa is to use `swift package edit ably-cocoa --path ably-cocoa`, which will give you a Git checkout of ably-cocoa at `ably-cocoa`. To edit ably-cocoa using Xcode, you will then need to open `ably-cocoa/Package.swift` _separately_ (making sure to close any other LiveObjects Xcode windows, else Xcode will not let you edit it).
-
-If you use edit mode, Xcode will not let you edit ably-cocoa from _within_ `./Package.swift` or `AblyLiveObjects.xcworkspace` (it does not let you edit SPM dependencies even if they're in edit mode). If you wish to edit ably-cocoa in one of these environments, consider temporarily instead pulling ably-cocoa in as a submodule. See commit [`29c6aa8`](https://github.com/ably/ably-liveobjects-swift-plugin/commit/29c6aa8) as an example.
+The plugin lives in the same repository and package as ably-cocoa, so there is nothing special to do: edit the ably-cocoa sources directly.
```
lawrence-forooghian: I think we can just remove this section of the doc then

README.md
```diff
@@ -183,6 +183,13 @@ realtime.connection.on { stateChange in
 
 ---
 
+## LiveObjects
+
+This repository also contains the Ably LiveObjects plugin, which enables real-time, collaborative data synchronisation on top of the core SDK. It is available via Swift Package Manager only: add this package as above and additionally select the `AblyLiveObjects` product for your target. See the [LiveObjects README](LiveObjects/README.md) and the [LiveObjects documentation](https://ably.com/docs/liveobjects) to get started.
+
+> [!WARNING]
+> LiveObjects is currently experimental. Breaking changes to the `AblyLiveObjects` API may be made in minor or patch releases of ably-cocoa, without a major version bump; ably-cocoa's semantic versioning guarantees apply only to the `Ably` product. If you use LiveObjects, consider pinning an exact version of this package.
```
lawrence-forooghian: I don't think we need the "consider pinning" instruction; users can decide how to handle this

Docs/plugins.md
lawrence-forooghian: Reworded in 94c13526 — ADR-128 is now described as the original design, with the separate-repo premise called out as having turned out to be wrong (per-declaration availability removes the constraint).

LiveObjects/CONTRIBUTING.md
lawrence-forooghian: Removed in 94c13526.

README.md
lawrence-forooghian: Removed from both READMEs in 94c13526.

.github/workflows/liveobjects.yaml
```diff
@@ -0,0 +1,394 @@
...
+          persist-credentials: false
+          submodules: true
+
+      # This step can be removed once the runners' default version of Xcode is 16.4 or above
+      - uses: maxim-lobanov/setup-xcode@ed7a3b1fda3918c0306d1b724322adc0b8cc0a90 # v1
+        with:
+          xcode-version: 16.4
+
+      - id: generation-step
+        run: swift run --package-path BuildTool BuildTool generate-matrices >> $GITHUB_OUTPUT
```
maratal, line 118: Here and elsewhere: `BuildTool BuildTool` looks not very convenient, why can't stay just one `BuildTool`?

maratal (commented):
Now we have root `/Source` (core SDK), root `/Sources` for `_AblyPluginSupportPrivate` and `LiveObjects/Sources/AblyLiveObjects` for plugin. It should be unified as root `_AblyPluginSupportPrivate/Source` and `Plugins/AblyLiveObjects/Source` (and maybe replace "Source" with "Sources" everywhere).

<strike>Now I'm thinking that keeping all the plugins in the main repo is probably not the best idea, we could leave `_AblyPluginSupportPrivate` in the main sdk repo and make plugin repo depending on it and importing test harness from it as well. WDYT @lawrence-forooghian @ttypic @sacOO7 (Lawrence, I believe you had this variant in mind too, why you still decided to merge everything into one?)</strike>

.github/workflows/liveobjects.yaml
sacOO7: @maratal if you think you can make a PR to fix some of those inconsistencies, can you accordingly make the PR on top of current one to fix it?

sacOO7: Ideally, we want this PR to get merged by end of this week

maratal: In the plugin repo (ably-liveobjects-swift-plugin) there was only **one** SwiftPM package — the root one. `BuildTool` was just an `.executableTarget` of that root package (sources at `Sources/BuildTool/`, no separate `Package.swift`). CI ran from the repo root, so `swift run BuildTool lint` resolved the root package implicitly and ran its `BuildTool` target — no `--package-path` needed, because the package it wanted *was* the current directory.

During the merge into ably-cocoa, BuildTool was **extracted into its own standalone package** at `LiveObjects/BuildTool/` (its own `Package.swift`), deliberately, so its dependencies (swift-argument-parser, async-algorithms, Table) aren't resolved by consumers of ably-cocoa. Now that it's a separate package in a subdirectory, `swift run` has to be pointed at it with `--package-path BuildTool`, while the trailing `BuildTool` is still the executable-target name.

So the second `BuildTool` isn't new and nothing was renamed — the executable was always called `BuildTool`. What's new is the `--package-path BuildTool` in front of it, forced by the target→separate-package split.

maratal (approved):
LGTM with follow up PR - https://github.com/ably/ably-cocoa/pull/2228 @sacOO7

sacOO7:
> LGTM with follow up PR - #2228 @sacOO7

Great, I will merge this PR, since it's approved 👍

lawrence-forooghian:
> Now I'm thinking that keeping all the plugins in the main repo is probably not the best idea, we could leave _AblyPluginSupportPrivate in the main sdk repo and make plugin repo depending on it

this is how things used to be — see 9e172dd3e744c6d8c2bb0be09a22853aa702005b for why we couldn't keep plugin-support in ably-cocoa

Issue #134, opened 2016-01-15
RealtimeChannel: channel state changes is inconsistent with the API


ricardopereira:
**Problem:**
`channel.subscribeToStateChanges` should be `channel.on`.

mattheworiordan:
This should follow the `EventEmitter` pattern, see http://docs.ably.io/client-lib-development-guide/features/#eventemitter

PR #2225, opened 2026-07-17
Add path-based LiveObjects public API, ported from the plugin repo

sacOO7:
Part of [AIT-1023](https://ably.atlassian.net/browse/AIT-1023).

## 🎯 What this PR does

Replaces the LiveObjects public API — the `channel.objects` proxy objects and `PublicTypes` surface — with the new **path-based / instance API**, by porting the [`feature/path-based-liveobjects`](https://github.com/ably/ably-liveobjects-swift-plugin/tree/feature/path-based-liveobjects) branch of the (now-merged-in) plugin repo into `LiveObjects/`.

This executes item 2 of #2222's "After merging" plan: bringing [ably-liveobjects-swift-plugin#134](https://github.com/ably/ably-liveobjects-swift-plugin/pull/134) into this repo, targeting the `integration/liveobjects` line so the repo move and the new API are released together ("LiveObjects has a new API and now lives in ably-cocoa").

**This ports the API *shape*, not its behaviour**: the `Default*` implementations behind the new public protocols currently trap via `notImplemented()`. The implementation follows (driven by the UTS), which is also why the previously ported JS integration tests (~6,300 lines) are removed rather than rewritten.

## 🧭 The new API at a glance

- **Entry point** — `ARTRealtimeChannel.object` returns a `RealtimeObject`; `get()` waits for sync and returns the root map as a path object (`RTL27`, `RTO23`).
- **`PathObject`** (`RTPO`) — a *lazy, path-addressed* reference: it stores path segments from the root and re-resolves on every call, so it **survives object replacement**. Typed views via `asLiveMap()` / `asLiveCounter()` / `asPrimitive()` (`RTTS5`); maps navigate with `get(key:)` / `at(path:)` and mutate with `set` / `remove`; counters `increment` / `decrement`.
- **`Instance`** (`RTINS`) — an *identity-addressed* view obtained from `PathObject.instance()`, modelled as a Swift **enum** (`.liveMap` / `.liveCounter` / `.primitive`) so discrimination is compile-time-exhaustive — the Swift-specific decision recorded on AIT-1023, in place of the spec's base-type + casts model (`RTTS9`).
- **Value types** — `Primitive` (a deliberate consolidation of the spec's six primitive sub-types, `RTTS6c`/`RTTS10c`), plus declarative creation: setting a `LiveMap.create()` / `LiveCounter.create()` value into the graph creates the object (`RTLMV`/`RTLCV`) — no separate `createMap()`/`createCounter()` calls.
- **Public wire types** — `ObjectMessage`, `ObjectOperation`, `ObjectData` etc. now user-facing with their plain names, enabled by namespacing the internal namesakes under `ProtocolTypes` (first commit).
- Subscriptions come in callback and `AsyncStream` (`events()`) flavours; all public types are `Sendable` and use typed `throws(ARTErrorInfo)`.

## 📦 Review commit by commit

| Commit | What | Provenance |
|---|---|---|
| `liveobjects: namespace internal ObjectMessage types under ProtocolTypes` | Mechanical groundwork: moves the 13 internal wire types into `internal enum ProtocolTypes`, freeing their plain names for the public API (28 files) | **Verbatim** port of plugin commit [`ab36a32`](https://github.com/ably/ably-liveobjects-swift-plugin/commit/ab36a329308b1b87e714d44eeeb923e8c14a5592) |
| `liveobjects: add path-based public API` | The new public surface + `Default*` skeletons; removes the old proxy objects, `PublicTypes`, and the ported JS integration tests (29 files, +1.6k/−7.3k) | **Verbatim** port of plugin commit [`c7ede2f`](https://github.com/ably/ably-liveobjects-swift-plugin/commit/c7ede2ff4d7f6ea17738fc877b1be74af708b35d) (= plugin PR #134) |
| `liveobjects: adapt example app to the path-based API` | Rewrites the two example view models to the new API | **Cocoa-side deviation** (see below) — the only hand-written code in this PR |

"Verbatim" is machine-checked, not aspirational: for the two port commits, every file is the byte-exact upstream content with `@available` annotations re-inserted by `Scripts/annotate-liveobjects-availability.py` (the same script CI enforces). The resulting tree was diffed against a pre-computed *annotate(upstream tree)* golden copy — **byte-identical**, including file sets.

## 📱 Why the example-app commit deviates from upstream

Plugin PR #134 merged with its example app still on the removed API — its three example-app CI jobs were red. This repo has the same `check-example-app` job gating `all-checks-completed`, so importing that state verbatim would permanently block this PR. The adaptation is deliberately minimal: view models hold path objects (killing the old re-fetch-on-replacement logic), create objects declaratively, and reload on subscription events (`PathObjectSubscriptionEvent` carries no per-key change set). A prominent comment in both files notes the app compiles but traps at runtime until the API is implemented.

## ✅ Verification

- **Byte-parity**: `LiveObjects/Sources/AblyLiveObjects` identical to annotate(upstream feature tree); `Tests/` identical except the two pre-existing cocoa deltas (`Sandbox.swift` fixture path, no `ably-common` submodule).
- **Builds**: `AblyLiveObjects` + `AblyLiveObjectsTests` targets build green at *both* intermediate commit states (bisectability) and at HEAD.
- **Tests**: LiveObjects unit suite (`BuildTool test-library --only-unit-tests`) passes at both states.
- **Lint**: full `BuildTool lint` (SwiftFormat + SwiftLint + prettier) green.
- **Example app**: builds on macOS locally; iOS/tvOS lanes couldn't run locally (no iOS/tvOS 26.2 simulator runtimes installed — environmental) and are covered by this repo's CI matrix.
- **Availability annotations**: annotation script is a no-op on the final tree (CI-equivalent check).
- **Scope**: nothing outside `LiveObjects/` is touched.

## 📝 Notes for follow-up (not in this PR)

- **Upstream doc-fix candidates** carried over verbatim: `Plugin.swift`'s public doc example still shows the removed `channel.objects.getRoot()`, and `InternalDefault{LiveMap,LiveCounter,RealtimeObjects}.swift` / `CONTRIBUTING.md` still reference the deleted `PublicDefault*` types. Left untouched here to keep the port verbatim; worth a small cleanup PR.
- The sandbox-hitting LiveObjects integration tests (e.g. `ObjectLifetimesTests`) run in CI; the full local sandbox pass is tracked separately.

[AIT-1023]: https://ably.atlassian.net/browse/AIT-1023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Introduced a path-based LiveObjects API for navigating maps, counters, primitives, subscriptions, events, and object messages.
  * Added `channel.object` as the new entry point for realtime objects.
  * Added support for direct sandbox and proxy-based integration testing.

* **Documentation**
  * Expanded UTS testing and translation guidance, including setup, tiers, execution, and auditing.

* **Tests**
  * Added coverage for REST time, channel history, token requests, and authentication reauthorization.

* **Refactor**
  * Reorganized LiveObjects protocol messaging and replaced the previous proxy-based API surface.


maratal:
@sacOO7 Why section "Review commit by commit" only contains 3 commits related to the PR title, but PR itself contains other 9 commits with all that "infra"/integration/proxy and skill work. Shouldn't these 9 commits be in a different PR prepending this PR (there were even separated tasks for this)?

maratal (commented):
"liveobjects:*" commits are fine, but I deliberately avoided modifying example app because you can't immediately check if these modifications lead to a correct behavior. But it's ok and easily can be fixed in case it doesn't work as expected.

I'm waiting this PR splitted according to tasks implemented @sacOO7

sacOO7:
Closing in favor of https://github.com/ably/ably-cocoa/pull/2227

PR #2231, opened 2026-07-29
Read the channel's data encoder live when encoding annotations

SimonWoolf:
`ARTRestAnnotationsInternal` and `ARTRealtimeAnnotationsInternal` each captured `_channel.dataEncoder` into an ivar at construction time. But `ARTChannel` *replaces* its `_dataEncoder` with a brand-new instance whenever `setOptions` is called (`recreateDataEncoderWith:`, `ARTChannel.m:63`), and there is no hook to refresh the annotations objects' copies.

So after a `setOptions` that added or changed cipher params, annotation data was encoded and decoded with the encoder built at channel-creation time — typically one with no cipher at all — while messages correctly picked up the new cipher, because `encodeMessageIfNeeded:` reads the `dataEncoder` property live (`ARTChannel.m:252`). `ARTRealtimePresence`'s publish path also reads it live (`ARTRealtimePresence.m:596`), deliberately bypassing its own cached ivar.

Fix: read `_channel.dataEncoder` at each encode/decode site rather than caching it. Four sites; the REST decode path (`ARTRestAnnotations.m:268`) already read it live, which was itself an inconsistency within that file.

## Scope

This is **not** the same bug as the one being fixed in ably-js ([ably-js#2277](https://github.com/ably/ably-js/pull/2277)), where annotation data was never encrypted at all. ably-cocoa is correct on the normal path: a channel constructed with cipher options encrypts annotation data properly, because construction order guarantees the captured encoder is already cipher-bearing.

This is the narrower latent case — a channel reconfigured via `setOptions` after creation. It also fails *symmetrically* (annotations stay internally consistent, just encrypted under the wrong policy — in practice not encrypted), so unlike the ably-js bug it wouldn't corrupt a round trip within a single client. It surfaces as annotation data sitting in plaintext on the service, and as garbling against a client that constructed its channel with the cipher up front.

Found while auditing the other SDKs for the ably-js bug. Java is unaffected; Python has the ably-js bug and worse (its decode path is broken too).

## Test

`test__annotation_data_uses_a_cipher_added_by_setOptions_after_channel_creation` in `RealtimeAnnotationsTests.swift` creates a channel with no cipher, adds one via `setOptions`, then publishes and receives an annotation with a data payload and asserts it round-trips.

Before the fix the annotation is published as plaintext and arrives as ciphertext, so the assertion fails.

Worth noting there was **no existing test coverage for annotations on encrypted channels at all** — neither `RealtimeAnnotationsTests.swift` nor `RestAnnotationsTests.swift` contained a single reference to `cipher` or `encrypt`. That gap is what let this sit unnoticed.

## CI status

The new test passes on all three platforms (macOS, iOS, tvOS), and the SDK compiles cleanly everywhere. EditorConfig passes.

The three `check` jobs are nonetheless red, on a **pre-existing flaky failure unrelated to this change**: `test__annotations_subscribe_should_implicitly_attach_the_channel_if_options_attachOnSubscribe_is_true`, failing with `Error 80017 - connection broken before receiving publishing acknowledgment`. That test never publishes an annotation and never configures a cipher — it only exercises attach/detach lifecycle — so it cannot be affected by this change. It also passed on an earlier run of this same branch. tvOS additionally failed a presence test (`test__050__Presence__PresenceMap__...`).

For reference, the last three `Integration Test` runs on `main` each failed a *different* test (`Presence.test__038`, `Presence.test__039`, `RestClientStats.test__002`) — none repeating, which is the signature of sandbox/network flakiness rather than a regression.

Note the REST path has no equivalent test; the realtime one covers the shared mechanism.

An earlier revision of this test used `reaction:distinct.v1` and failed with `Error 40000 - Aggregation method not allowed for anonymous client publish; allowed methods are: [multiple.v1 total.v1]`. Now uses `reaction:multiple.v1`, matching the other tests in the file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Annotations now use the channel’s current encryption settings when published or received.
  * Cipher changes made after channel creation are correctly applied to realtime and REST annotation data.

* **Tests**
  * Added coverage verifying encrypted annotations are transmitted securely and decoded back to their original content.

### DIFF
--- a/Source/ARTRealtimeAnnotations.m
+++ b/Source/ARTRealtimeAnnotations.m
@@ -98,7 +98,6 @@
     __weak ARTRealtimeInternal *_realtime;
     dispatch_queue_t _userQueue;
     ARTEventEmitter<ARTEvent *, ARTAnnotation *> *_eventEmitter;
-    ARTDataEncoder *_dataEncoder;
 }
 
 - (instancetype)initWithChannel:(ARTRealtimeChannelInternal *)channel logger:(ARTInternalLog *)logger {
@@ -109,7 +108,6 @@
         _queue = _realtime.rest.queue;
         _logger = logger;
         _eventEmitter = [[ARTInternalEventEmitter alloc] initWithQueue:_queue timeProvider:_realtime.rest.timeProvider];
-        _dataEncoder = _channel.dataEncoder;
     }
     return self;
 }
@@ -150,7 +148,8 @@
                                                            extras:outboundAnnotation.extras];
 art_dispatch_sync(_queue, ^{
     NSError *error = nil;
-    ARTAnnotation *annotationToPublish = _dataEncoder ? [annotation encodeDataWithEncoder:_dataEncoder error:&error] : annotation; // RSAN1c3
+    ARTDataEncoder *dataEncoder = self->_channel.dataEncoder;
+    ARTAnnotation *annotationToPublish = dataEncoder ? [annotation encodeDataWithEncoder:dataEncoder error:&error] : annotation; // RSAN1c3
     if (error) {
         if (callback) {
             callback([ARTErrorInfo createFromNSError:error]);
@@ -301,11 +300,12 @@ art_dispatch_sync(_queue, ^{
 }
 
 - (void)onMessage:(ARTProtocolMessage *)message {
+    ARTDataEncoder *dataEncoder = _channel.dataEncoder;
     for (ARTAnnotation *a in message.annotations) {
         ARTAnnotation *annotation = a;
-        if (annotation.data && _dataEncoder) {
+        if (annotation.data && dataEncoder) {
             NSError *decodeError = nil;
-            annotation = [a decodeDataWithEncoder:_dataEncoder error:&decodeError];
+            annotation = [a decodeDataWithEncoder:dataEncoder error:&decodeError];
             if (decodeError != nil) {
                 ARTErrorInfo *errorInfo = [ARTErrorInfo wrap:[ARTErrorInfo createWithCode:ARTErrorUnableToDecodeMessage message:decodeError.localizedFailureReason] prepend:@"Failed to decode data: "];
                 ARTLogError(self.logger, @"RT:%p C:%p (%@) %@", _realtime, _channel, _channel.name, errorInfo.message);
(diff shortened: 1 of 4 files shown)


maratal:
Had a look through this. The production change looks right to me; I have one substantive question about the test.

### The fix itself

Checked the surrounding code and it holds up:

- All three converted sites run on the internal queue (`art_dispatch_sync`/`art_dispatch_async` on `_queue`), and `setOptions` mutates `_dataEncoder` on that same queue (`ARTChannel.m:52-61`). So reading the property live is actually *more* correct than caching it, not just fresher — no new race is introduced.
- `ARTRealtimeAnnotationsInternal` is constructed eagerly in the channel's init (`ARTRealtimeChannel.m:319`), not lazily on first `channel.annotations` access, so the stale capture is a real bug rather than a hypothetical one.
- Three sites converted plus the already-live REST decode at `ARTRestAnnotations.m:268` accounts for all four; no other `dataEncoder` reads remain in either file.

### The test may not fail without the fix

> Before the fix the annotation is published as plaintext and arrives as ciphertext, so the assertion fails.

I don't think that follows. Publish (`ARTRealtimeAnnotations.m:151`) and receive (`onMessage`, line 303) go through the *same* annotations object and, pre-fix, the *same* cached cipher-less encoder. The annotation goes out as a plaintext `ARTProtocolMessageAnnotation` and comes back plaintext, decoded by that same cipher-less encoder — the service doesn't encrypt on the client's behalf. So `receivedAnnotation.data as? String == annotationData` should hold either way.

That's precisely the "fails symmetrically" property the Scope section describes, and a symmetric round-trip within a single client is the one test shape that can't observe a symmetric failure. Catching it needs an asymmetric assertion — either check that the published annotation's `encoding` contains `cipher+aes-256-cbc`, or receive on a second client whose channel was constructed with the cipher up front (which is how the description says the bug actually surfaces).

Did you run the test against unpatched source and see it fail? If so my reading is wrong and I'd like to understand where.

### Smaller notes

- No REST-side test. The argument that the realtime test covers the shared mechanism is a little thin given the code is duplicated per file rather than genuinely shared, though the risk is low.
- `var receivedAnnotation: ARTAnnotation!` followed by `guard let receivedAnnotation else { XCTFail(...) }` shadows an IUO — the declaration could just be optional.
- The CI status section is now stale: everything is green except the five integration `check` jobs, which are pending on the run from today's rebase rather than failing. The pre-existing flakes it discusses should also be less disruptive now that #2249 landed runner-level retries.

Net: I'd take the fix. The test is the part worth another look before this leaves draft.

maratal (approved):
Looks good - I've fixed the realtime test so it properly fail without the fix and also added the REST test.

PR #2236, opened 2026-08-02
docs: document LiveObjects versioning, distribution and test targets

maratal:
## Summary

Documents what the LiveObjects monorepo merge changed for contributors and releasers. No code changes — `CONTRIBUTING.md` and `LiveObjects/CONTRIBUTING.md` only.

Three questions prompted this: how LiveObjects tests run in CI, how the plugin is versioned and released now that it shares a repository with the core SDK, and what the recommended minimum OS versions are.

## What's documented

**Test targets and CI.** A table of the package's four test targets and which workflow runs each. The key fact is that every target selects its files by directory — SwiftPM globs the target's `path`, and the test plans name whole targets, never individual tests — so new tests are picked up with no manifest, test-plan or workflow change. Also records that the LiveObjects UTS ports live inside `AblyLiveObjectsTests`, not the `UTS` target, and that `BuildTool test-library` uses the `AllTests` plan while the coverage job uses `UnitTests` (which skips `.integration`). That asymmetry is currently harmless but would silently diverge the moment a UTS port carries that tag.

**Supported OS versions.** macOS 11 / iOS 14 / tvOS 14 for LiveObjects against the core SDK's macOS 10.11 / iOS 9 / tvOS 10, how per-declaration `@available` lets one package host both, and why test code can't use the same mechanism (swift-testing's `@Suite` macro rejects `@available`, so `LiveObjects/BuildTool` passes deployment-target overrides to `xcodebuild` instead).

**Distribution.** `AblyLiveObjects` is SPM-only. This was already asserted in `README.md`, but nowhere recorded as a decision, so the section gives the channel/product matrix and the four independent reasons — podspec/pbxproj don't ship the sources, `ABLY_SUPPORTS_PLUGINS` is defined only in `Package.swift`, `_AblyPluginSupportPrivate` is a target-not-product with no CocoaPods/Carthage equivalent, and the podspec's Swift/OS floors are too low. It also states that OS version requirements are *not* among the reasons, since ADR-128's superseded premise makes that the natural wrong guess.

**Versioning.** One version number for everything the repository publishes; the standalone plugin's 0.4.0 line does not continue; a tag doesn't mean the same thing on every channel; `AblyLiveObjects` is exempt from semver while experimental.

**Drive-by fix.** The release checklist told releasers to check whether ably-cocoa-plugin-support needed its own release. It's been the in-repo `_AblyPluginSupportPrivate` target since `cbe4a692` and is no longer an external dependency.

## Two things needing a decision before merge

**1. The minor-version policy sentence isn't sourced.** The draft says minor versions aren't bumped ahead of 2.0.0. That comes from a Slack conversation, not from any commit, ADR or doc in the repo. Confirm or correct it.

**2. ADR-114 has a gap that this PR documents around rather than fixes.**

[ADR-114](https://ably.atlassian.net/wiki/spaces/ENG/pages/3199500291/ADR-114+Increase+Cocoa+SDK+minimum+supported+version+to+iOS+14) is the source of the iOS 14 floor, and the LiveObjects sources sit at it today via `@available`. Two problems with treating it as the authority:

- **It is iOS-only.** The ADR and its [RFC](https://ably.atlassian.net/wiki/spaces/SDKs/pages/2986147844/RFC+Deprecate+iOS+13+support+for+ably-cocoa) contain no adoption data, competitor comparison or API justification for macOS or tvOS. `.macOS(.v11)` and `.tvOS(.v14)` are same-year equivalents (all autumn 2020) paired by convention, never argued for. So there is no defensible answer to "what should the minimum macOS/tvOS version be?" — only an inherited one.

- **It was never applied to the core SDK.** All four ADR action items are still `incomplete`, and `Package.swift` remains `.macOS(.v10_11), .iOS(.v9), .tvOS(.v10)`. The one commit that raised the whole package (`176dd4a7`, "Raise minimum OS requirements — Matches Chat and LiveObjects") lives on an unmerged Swift-migration branch.

The provenance is also thinner than the ADR reference suggests. The floor reached LiveObjects by copying: the plugin's initial commit (`b38ce55f`) took it from ably-chat-swift, whose own initial commit (`646c220c`) cites Umair's guidance and ADR-114. No one has re-evaluated it for LiveObjects specifically, and the RFC itself concedes that iOS 13 is where nearly all the technical value is — Swift Concurrency, native WebSockets and `ObservableObject` all arrive in 13. The step to 14 rests mainly on `Logger`, plus the argument that breaking compatibility twice costs more than breaking it once.

This has a concrete cost today: the split floor is why `Scripts/annotate-liveobjects-availability.py` and the BuildTool deployment-target overrides exist. Closing it — either by raising the package floor or by writing a successor ADR that covers all three platforms — would let both disappear.

I've deliberately kept this out of `CONTRIBUTING.md`, since it's an org-level decision rather than contributor guidance. Raising it here so it isn't lost.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,9 +63,71 @@ To run tests, do any of the following:
 
 The `make test_*` commands are used by CI and expect you to have a simulator device of a specific model and OS version. See [`Fastfile`](./fastlane/Fastfile) for these values. If you don't have a matching simulator, you can create one using `simctl`. For example, `xcrun simctl create "iPhone 12 (14.4)" "iPhone 12" "com.apple.CoreSimulator.SimRuntime.iOS-14-4"`.
 
+### Test targets
+
+`swift test` builds and runs every test target in the package:
+
+| Target | Path | Contents |
+| --- | --- | --- |
+| `AblyTests` | `Test/AblyTests` | Swift tests for the core SDK |
+| `AblyTestsObjC` | `Test/AblyTestsObjC` | Objective-C tests for the core SDK |
+| `UTS` | `Test/UTS` | Universal Test Suite, derived from the language-neutral specs in the [`specification`](https://github.com/ably/specification) repository |
+| `AblyLiveObjectsTests` | `LiveObjects/Tests/AblyLiveObjectsTests` | LiveObjects tests, including the ported UTS `objects` unit specs under `UTS/` |
+
+To run just one of them, filter by module name — for example `swift test --filter 'AblyLiveObjectsTests\.'`.
+
+Every one of these targets selects its files by directory: SwiftPM globs the target's `path`, and the test plans name whole targets rather than individual tests. A new test file is therefore picked up automatically, with no change to `Package.swift`, to any test plan, or to any CI workflow. Note in particular that the LiveObjects UTS ports live inside `AblyLiveObjectsTests`, so they are run by that target's CI jobs and not by the `UTS` target's.
+
+In CI:
+
+- [`integration-test.yaml`](.github/workflows/integration-test.yaml) runs the Fastlane lanes, which build the `ably-cocoa` scheme against [`Test/Ably.xctestplan`](Test/Ably.xctestplan) — `AblyTests`, `AblyTestsObjC` and `UTS`.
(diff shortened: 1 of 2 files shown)


maratal:
Superseded by #2237, which rebases this onto #2235 (branch `docs/contribution-update`) and squashes the six commits into one.

PR #2249, opened 2026-08-20
[AIT-1307] Fix remaining flaky CI: runner-level retries, test race fixes, UTS harness + spec alignment

sacOO7:
- Fixes https://ably.atlassian.net/browse/AIT-1307

## Related PR

- Companion spec PR (the UTS spec/docs changes these ports align to): ably/specification#517

## Motivation

Follow-up to #2241 (sandbox provisioning retries). With provisioning failures gone, the latest `main` Integration Test run ([32277853593](https://github.com/ably/ably-cocoa/actions/runs/32277853593)) still failed on all three lanes — each on a **different, intermittent** test:

| Lane | Failure | Class |
|---|---|---|
| macOS | UTS `test_RSP4b1` (both protocol variants) | Clock-skew flake — **second live firing** (tvOS on 08-15, macOS on 08-19): the test bracketed a server-timestamped history window with the runner's local clock |
| tvOS | `RealtimeClientPresenceTests.test__019` | Test-code race: presence-map emptiness asserted inside the `.failed` callback, racing the internal-queue map clear (the CI log shows the still-present member) |
| iOS | `AuthTests.test__002` — publish `waitUntil` > 20s | Environmental round-trip stall, same family as the earlier `test__038/039/043/044` presence timeouts |

Each class gets a fix at its own level: the environmental family gets a runner-level retry; the races get code fixes; the clock-skew flake gets a spec fix (companion PR) with the port aligned here.

### 1. CI level
- **`fastlane/Fastfile`** — `number_of_retries: 2` on all SDK test lanes: maps to `xcodebuild -retry-tests-on-failure`, re-running **only failed tests** in place. Absorbs single-occurrence environmental stalls; a genuine regression still fails every attempt. (Parity with ably-java's Gradle test-retry plugin.)
- **`Gemfile`** — literal `eval_gemfile("fastlane/Pluginfile")`: Dependabot rejects interpolated arguments, which had every Dependabot run red.

### 2. Legacy test race fixes
- **`test__019`** — emptiness assertions moved out of the `.failed` state-change callback to `toEventually(beEmpty())` after it: the map is cleared on the internal queue and the event can be delivered to the user queue first; eventual emptiness is RTP5a's observable contract.
- **`test__039`** — one-shot off-queue precondition read (`members.count == 0`) converted to the same `toEventually` pattern the file already uses.

### 3. UTS harness (`Test/UTS/infra`)
- **`pollUntil` is now the spec's value-returning `poll_until`** (generic overload; a Bool overload remains for latch/monotonic conditions). Tests assert on the page the poll settled on — a refetch after a Bool poll can hit a less-replicated frontend and under-return, failing exact-count assertions against a healthy backend. All three sibling corpora already implement this contract (spec pseudocode, ably-js's generic `pollUntil<T>`, ably-java via capture-inside-the-condition); the old cocoa Bool-then-refetch shape was the outlier.
- **`awaitState`/`awaitChannelState` short-circuit on FAILED** (when it isn't the awaited state), reporting the actual `errorReason` instead of burning the 15s timeout plus a cascading second timeout.
- **Cleanup** (`withRealtimeClient`): close-await on the 15s harness default (matching the specs' normalized cleanup template), skipped entirely when the connection is FAILED — `close()` from FAILED is a no-op (RTN12), so awaiting CLOSED there recorded a spurious teardown failure after a passing body.

### 4. UTS ports aligned to ably/specification#517
- **`PresenceTests` RSP4b1** — time-range boundaries derived from the events' server-assigned timestamps (±1s), per the corrected spec.
- **Five poll-then-refetch sites** (`HistoryTests` RSL2a, `PublishTests` RSL1k5, `PresenceTests` RSP4/RSP4b1/RSP5) converted to settled-value polls, matching the spec's value-assigned `poll_until`.
- **`AuthReauthTests`** — AUTH-frame poll in the spec's value form (asserting on the frames the poll observed), with the spec's 500ms interval.
- **Connect-waits**: port-invented ones (spec relies on auto-connect) use the harness default 15s; `ChannelHistoryTests`' two awaits keep `timeout: 10` because that spec times them explicitly (ably-java honours the same 10s).

### 5. Docs
- **`Test/UTS/README.md`** — `pollUntil` row documents both forms.
- **`.claude/skills/uts-to-swift/SKILL.md`** — *slimmed*: translation semantics (poll forms, deliberate `WAIT`s, untimed-step timeouts, `random_id`) now live in the spec repo's translator-facing `writing-derived-tests.md` (companion PR), so every SDK inherits them; the skill keeps only harness mechanics and fixes a misleading pre-existing timeout hint.

## Validation

- Every touched suite passes against the live sandbox/proxy: `HistoryTests`, `PublishTests`, `PresenceTests` (33 tests incl. RSP4b1 both variants), `ObjectsFaultsTests`, `AuthReauthTests`, `ChannelHistoryTests`, plus `test__019`/`test__039`/`AuthTests.test__002` individually.
- `swift build --build-tests`, `make lint`, LiveObjects BuildTool lint all clean.
- Spec↔port alignment machine-verified: a three-corpus audit (spec ↔ ably-js ↔ cocoa) plus an independent adversarial verification pass over every staged hunk; the one genuine mismatch it caught (ChannelHistory's spec-timed 10s awaits) is fixed as described above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asynchronous connection, presence, history, and publishing test reliability.
  * Failures and errors are now reported immediately instead of being hidden or delayed by timeouts.
  * Preserved results obtained during polling to avoid inconsistent follow-up checks.
  * Added retries for failed test runs, up to two attempts.

* **Documentation**
  * Clarified polling behavior, timeout handling, default waits, and supported result types.

### DIFF
--- a/Test/UTS/README.md
+++ b/Test/UTS/README.md
@@ -705,13 +705,14 @@ callbacks — ably-java's `AuthReauthTest` does the same).
 **The shared async helpers** both integration walkthroughs below lean on (from
 `infra/Utils.swift` — java's `infra/Utils.kt`, see §6.9). All are wall-clock and poll-based
 (never a fixed sleep); on timeout they record a Swift Testing `Issue` (and return `false`)
-rather than throwing:
+rather than throwing — though `pollUntil` re-throws an error thrown by its condition, aborting
+the poll (a thrown error is a genuine failure; "not ready yet" is signalled by returning nil):
 
 | Helper | Shape | Purpose |
 |--------|-------|---------|
 | `awaitState` | `await awaitState(client, .connected, timeout: 15)` | suspend until `connection.state == target` (or already there) |
 | `awaitChannelState` | `await awaitChannelState(channel, .attached, timeout: 15)` | same, for a channel's state |
-| `pollUntil` | `let page = await pollUntil("…") { cond ? value : nil }` or `await pollUntil("…") { condition }` | the spec's `poll_until`: the generic form returns the settled value (assert on it — a refetch can under-return); the Bool form is for latch/monotonic conditions with no page to keep |
+| `pollUntil` | `let page = await pollUntil("…") { cond ? value : nil }` or `await pollUntil("…") { condition }` | the spec's `poll_until`: the generic form returns the settled value (assert on it — a refetch can return fewer items); the Bool form is for latch/monotonic conditions with no page to keep |
 
 ### 11.4 Walkthrough: the direct-sandbox test (`ChannelHistoryTests`)
 
--- a/Test/UTS/README.md
+++ b/Test/UTS/README.md
@@ -711,7 +711,7 @@ rather than throwing:
 |--------|-------|---------|
 | `awaitState` | `await awaitState(client, .connected, timeout: 15)` | suspend until `connection.state == target` (or already there) |
 | `awaitChannelState` | `await awaitChannelState(channel, .attached, timeout: 15)` | same, for a channel's state |
-| `pollUntil` | `await pollUntil("what you wait for", timeout: 15, interval: 0.1) { condition }` | suspend until an arbitrary predicate holds — e.g. `pollUntil("re-auth") { count.count > original }` |
+| `pollUntil` | `let page = await pollUntil("…") { cond ? value : nil }` or `await pollUntil("…") { condition }` | the spec's `poll_until`: the generic form returns the settled value (assert on it — a refetch can under-return); the Bool form is for latch/monotonic conditions with no page to keep |
 
 ### 11.4 Walkthrough: the direct-sandbox test (`ChannelHistoryTests`)
 
--- a/Test/UTS/infra/Utils.swift
+++ b/Test/UTS/infra/Utils.swift
@@ -16,17 +16,22 @@ import Ably.Private
 /// same loop shape as ably-java's `Utils.pollUntil`): polls `condition` every `interval` seconds
 /// until it yields a value and returns that settled value, or records a test failure once
 /// `timeout` (wall-clock seconds) elapses. Callers assert on the returned page rather than
-/// refetching — a refetch can hit a less-replicated frontend and under-return, reintroducing the
-/// eventual-consistency race the poll just absorbed.
+/// refetching — a refetch can hit a less-replicated frontend and return fewer items than the poll
+/// just observed, reintroducing the eventual-consistency race the poll just absorbed.
+///
+/// `condition` is `rethrows`: if it throws, the poll aborts immediately and the error propagates
+/// to the caller (failing the test) — no timeout `Issue` is recorded. This matches the spec's
+/// plain `poll_until`: a thrown error is a genuine failure, whereas a not-ready state is
+/// signalled by returning nil (e.g. an under-replicated item count), not by throwing.
 @discardableResult
 func pollUntil<T>(_ description: String,
                   timeout: TimeInterval = 15,
                   interval: TimeInterval = 0.1,
                   sourceLocation: SourceLocation = #_sourceLocation,
-                  _ condition: () async -> T?) async -> T? {
+                  _ condition: () async throws -> T?) async rethrows -> T? {
     let deadline = Date().addingTimeInterval(timeout)
     while Date() < deadline, !Task.isCancelled {
-        if let value = await condition() {
+        if let value = try await condition() {
             return value
         }
         try? await Task.sleep(nanoseconds: UInt64(interval * 1_000_000_000))
@@ -44,9 +49,9 @@ func pollUntil(_ description: String,
                timeout: TimeInterval = 15,
                interval: TimeInterval = 0.1,
                sourceLocation: SourceLocation = #_sourceLocation,
-               _ condition: () async -> Bool) async -> Bool {
-    await pollUntil(description, timeout: timeout, interval: interval, sourceLocation: sourceLocation) {
-        await condition() ? true : nil
+               _ condition: () async throws -> Bool) async rethrows -> Bool {
+    try await pollUntil(description, timeout: timeout, interval: interval, sourceLocation: sourceLocation) {
+        try await condition() ? true : nil
     } != nil
 }
 
--- a/Test/UTS/infra/Utils.swift
+++ b/Test/UTS/infra/Utils.swift
@@ -12,32 +12,42 @@ import Ably.Private
 /// anti-flake rule: no fixed blind waits — poll on observable state, with a short interval
 /// between checks).
 
-/// Suspends until `condition` returns `true`, polling every `interval` seconds, or records a test
-/// failure once `timeout` (wall-clock seconds) elapses. Returns whether the condition was met.
-///
-/// The integration analogue of `UTSTestCase.poll` — e.g.
-/// `await pollUntil("auth callback re-invoked") { authCallbackCount.count > original }`.
+/// The integration tier's `poll_until` (the spec's `RETURN result IF … ELSE null` contract, the
+/// same loop shape as ably-java's `Utils.pollUntil`): polls `condition` every `interval` seconds
+/// until it yields a value and returns that settled value, or records a test failure once
+/// `timeout` (wall-clock seconds) elapses. Callers assert on the returned page rather than
+/// refetching — a refetch can hit a less-replicated frontend and under-return, reintroducing the
+/// eventual-consistency race the poll just absorbed.
+@discardableResult
+func pollUntil<T>(_ description: String,
+                  timeout: TimeInterval = 15,
+                  interval: TimeInterval = 0.1,
+                  sourceLocation: SourceLocation = #_sourceLocation,
+                  _ condition: () async -> T?) async -> T? {
+    let deadline = Date().addingTimeInterval(timeout)
+    while Date() < deadline, !Task.isCancelled {
+        if let value = await condition() {
+            return value
+        }
+        try? await Task.sleep(nanoseconds: UInt64(interval * 1_000_000_000))
+    }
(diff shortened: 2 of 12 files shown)


Test/UTS/integration/standard/realtime/ChannelHistoryTests.swift
```diff
@@ -40,6 +40,7 @@ final class ChannelHistoryTests: IntegrationTestCase {
                     publisher.connect()
                     subscriber.connect()
 
+                    // The spec times these in-body connect-awaits explicitly (WITH timeout: 10 seconds).
```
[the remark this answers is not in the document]
sacOO7: Confirmed and fixed in bef9bdf4 — the spec (`channel_history_test.md:87`) is indeed value-assigned, and the port now asserts on the page the poll settled on (the now-unused reporting history helper is removed). Cross-validating this against every value-assigned `poll_until` in the ported specs also surfaced one more site of the same class that neither this review nor our earlier audits flagged: **RSL2b3** (`history.md:218` → `HistoryTests.swift`), where the refetched page additionally drove the time-boundary computation — fixed in the same commit. Both suites re-validated against the live sandbox (all protocol variants).

Test/UTS/infra/integration/IntegrationTestCase.swift
```diff
@@ -35,8 +35,13 @@ class IntegrationTestCase {
...
         try await runThenCleanUp(client, body: body) { client in
-            client.close()
-            await awaitState(client, .closed, timeout: 10)
+            // Per the specs' common cleanup: only close from a closable state. close() from
+            // FAILED is a no-op (the connection is already terminal), so awaiting CLOSED there
+            // would burn the timeout and record a spurious failure after the body passed.
+            if client.connection.state != .failed {
+                client.close()
+                await awaitState(client, .closed)
+            }
```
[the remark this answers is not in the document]
sacOO7: Confirmed and fixed in bef9bdf4 — `ARTRealtime _close` (Source/ARTRealtime.m:491-496) returns without a transition from INITIALIZED exactly as described, so the cleanup would have burned the 15s wait and recorded a spurious teardown failure for any body exiting before `connect()`. The cleanup now skips close-and-await from both INITIALIZED and FAILED — the two states verified against the full `_close` switch as never reaching CLOSED (the others are safe: CLOSED satisfies the wait immediately, CLOSING's in-flight close completes, CONNECTING re-closes on its next transition, DISCONNECTED/SUSPENDED transition straight to CLOSED). This also matches the UTS specs' cleanup guard, which deliberately excludes INITIALIZED.

fastlane/Fastfile
```diff
@@ -13,6 +13,10 @@ class LaneConfig < Struct.new(:name, :devices)
       scheme: "ably-cocoa",
       derived_data_path: "derived_data",
       test_without_building: false,
+      # Retry only the tests that failed, in place (xcodebuild -retry-tests-on-failure).
+      # Absorbs single-occurrence environmental timeouts on the CI runners; a genuine
+      # regression still fails CI because it fails every attempt.
+      number_of_retries: 2,
```
sacOO7, line 19: Super helpful: this retries **only the tests that failed** (maps to xcodebuild `-retry-tests-on-failure -test-iterations 3`), not the whole suite — a flaky failure costs seconds instead of a full lane re-run, while a genuine regression still fails every attempt. This is the standard industry mechanism for absorbing flaky tests; we follow the same approach in ably-java via the Gradle test-retry plugin (`maxRetries = 3`).

Test/UTS/infra/Utils.swift
```diff
@@ -12,81 +12,120 @@ import Ably.Private
...
-/// Suspends until `condition` returns `true`, polling every `interval` seconds, or records a test
-/// failure once `timeout` (wall-clock seconds) elapses. Returns whether the condition was met.
-///
-/// The integration analogue of `UTSTestCase.poll` — e.g.
-/// `await pollUntil("auth callback re-invoked") { authCallbackCount.count > original }`.
+/// The integration tier's `poll_until` (the spec's `RETURN result IF … ELSE null` contract, the
+/// same loop shape as ably-java's `Utils.pollUntil`): polls `condition` every `interval` seconds
+/// until it yields a value and returns that settled value, or records a test failure once
+/// `timeout` (wall-clock seconds) elapses. Callers assert on the returned page rather than
+/// refetching — a refetch can hit a less-replicated frontend and under-return, reintroducing the
```
maratal: Is it even a valid term - "under-return"?

Test/UTS/README.md
```diff
@@ -711,7 +711,7 @@ rather than throwing:
 |--------|-------|---------|
 | `awaitState` | `await awaitState(client, .connected, timeout: 15)` | suspend until `connection.state == target` (or already there) |
 | `awaitChannelState` | `await awaitChannelState(channel, .attached, timeout: 15)` | same, for a channel's state |
-| `pollUntil` | `await pollUntil("what you wait for", timeout: 15, interval: 0.1) { condition }` | suspend until an arbitrary predicate holds — e.g. `pollUntil("re-auth") { count.count > original }` |
+| `pollUntil` | `let page = await pollUntil("…") { cond ? value : nil }` or `await pollUntil("…") { condition }` | the spec's `poll_until`: the generic form returns the settled value (assert on it — a refetch can under-return); the Bool form is for latch/monotonic conditions with no page to keep |
```
maratal: "under-return"?

Test/UTS/integration/standard/realtime/ChannelHistoryTests.swift
```diff
@@ -103,22 +105,9 @@ extension ChannelHistoryTests {
...
-                continuation.resume(returning: result?.items ?? [])
-            }
-        }
-    }
-
-    /// Non-reporting variant of `historyItems` for use inside `pollUntil` predicates: a transient
-    /// `history()` error must surface as a retry — and, at worst, as the poll's own timeout issue —
-    /// not fail the test. Returns an empty list on error.
+    /// swallowing errors, for use inside `pollUntil` predicates: a transient `history()` error
+    /// must surface as a retry — and, at worst, as the poll's own timeout issue — not fail the
```
maratal: Why in case of history request failure test should survive?

fastlane/Fastfile
maratal, line 1: finally 👍

maratal (commented):
A few questions

Test/UTS/integration/standard/realtime/ChannelHistoryTests.swift
sacOO7: Good catch — it shouldn't, and I've reworked this in 30158d84.

The intent was only to absorb the eventual-consistency race, but that race shows up as an *under-count* (history() returns fewer than N items), not as an error — so the poll's `items.count == N` check already handles it. Swallowing history() *errors* on top of that was wrong: it hides a genuine 5xx/network failure behind a bland poll timeout.

The spec's plain `poll_until` — which the history/publish/presence specs all use — aborts on the first error raised inside the condition (reference definition in `writing-test-specs.md`: "an error raised here aborts the poll immediately"). The error-tolerant `poll_until_success` variant is deliberately reserved for reads that raise *by design* when data isn't ready yet (`getMessage`/`getMessageVersions`/`annotations.get` on the eventually-consistent store, and objects reads across a DETACHED recovery window); history() isn't one of those. Both reference SDKs match this: ably-js polls history() with plain `pollUntil` (no catch), and ably-java with a bare `subChannel.history(null)` inside `pollUntil` (no `runCatching`) — a history() error propagates out and fails the test.

So I've dropped the `*Quietly` helpers and let history()/presence.history() errors propagate out of the poll, aborting it and surfacing the real error — matching js/java and the plain `poll_until` reference semantics.

Test/UTS/infra/Utils.swift
sacOO7: Fair — it was an invented shorthand, not an established term. Replaced with plain English in 30158d84 at every site it had spread to (8 in this PR, and the matching 2 in the companion spec-docs PR ably/specification#517, so the vocabulary stays in sync): "a refetch can hit a less-replicated frontend and return fewer items than the poll just observed".

Test/UTS/README.md
sacOO7: Same as the Utils.swift thread — invented shorthand, now replaced with plain English in 30158d84 ("a refetch can return fewer items"), here and at every other site in this PR and the companion spec PR.

Test/UTS/integration/proxy/realtime/AuthReauthTests.swift
```diff
@@ -84,9 +84,10 @@ final class AuthReauthTests: ProxyTestCase {
                 // TokenRequest that the SDK still exchanges for a token (a REST round-trip through
                 // the proxy) before sending AUTH — so also wait for the AUTH frame to reach the
                 // proxy log before asserting on it.
-                guard await pollUntil("client-to-server AUTH frame recorded in the proxy log", timeout: 15, {
+                guard let clientAuthFrames = await pollUntil("client-to-server AUTH frame recorded in the proxy log", timeout: 15, interval: 0.5, {
                     let log = (try? await session.getLog()) ?? []
-                    return !self.clientToServerAuthFrames(in: log).isEmpty
+                    let frames = self.clientToServerAuthFrames(in: log)
+                    return frames.isEmpty ? nil : frames
```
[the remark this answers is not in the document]
sacOO7: Okay, we already do `try? await session.getLog` right?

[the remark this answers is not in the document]
sacOO7: But we want to keep it optional right? So, even if we dont get log, test should work right

maratal (approved):
LGTM

PR #2250, opened 2026-08-20
[AIT-1308][LiveObjects] Example app: rely on object.get()'s implicit attach; remove the channel async helper

sacOO7:
- Fixes https://ably.atlassian.net/browse/AIT-1308

## Problem statement

The LiveObjects example app carried a helper, `Helpers/ARTRealtimeChannel+Async.swift`, exposing continuation-bridged `attachAsync()` / `detachAsync()` wrappers. Both view models called `channel.attachAsync()` immediately before `channel.object.get()`. That explicit attach is redundant — the SDK's `get()` entry point already performs the spec's *ensure-active-channel* procedure, which attaches implicitly — and `detachAsync()` was never called at all. The example was therefore modelling an outdated usage pattern: readers copying it would write an unnecessary attach into their own apps.

## What this PR does

- **Removes the explicit `attachAsync()` call from both view models** (`LiveCounterViewModel.initializeCounters()`, `TaskBoardViewModel.initializeTasks()`), replacing the "Attach channel first" comment with a note that `object.get()` performs the ensure-active-channel procedure (RTO23e/RTL33) and attaches implicitly.
- **Deletes `Helpers/ARTRealtimeChannel+Async.swift`** — with the attach calls gone, `attachAsync()` is unreferenced, and `detachAsync()` had zero call sites to begin with (the demo's view models are created once and live for the app's lifetime; the app deliberately never detaches). No `project.pbxproj` edit is needed: the example project uses an Xcode 16 synchronized file group.
- **Drops the dead `private var realtime` stored property from both view models** — it was only written in `init` and never read again; the `init` parameter still feeds `channels.get(...)` directly.

## Why this is safe

- **Implicit attach is the specified behaviour.** `PublicDefaultRealtimeObject.get()` runs `ChannelConfigGuards.ensureActiveChannel`, the SDK's implementation of RTO23e → RTL33: RTL33b implicitly attaches from `INITIALIZED`/`DETACHED`/`DETACHING`/`ATTACHING`. ably-js documents exactly this pattern for its LiveObjects examples ("no explicit attach needed — `.get()` handles it").
- **No call-order hazard.** In both view models every other objects operation (`set`, `subscribe`, `increment`, `entries`, …) runs after `get()` resolves: the views render their controls only once loading completes, and all mutations route through the `root` reference, which is nil until `get()` returns — so no channel-state guard (RTO25/RTO26) can ever fire on an unattached channel, and a user cannot race the initial load.
- **Same error surface.** Attach failures now surface from `get()` in the same `do/catch` that previously caught them from `attachAsync()`; the user-visible error message is unchanged.
- **Connection lifetime unchanged by the property removal.** The view models still retain the `ARTRealtimeChannel` wrapper, which co-owns the internal client and the shared dealloc gate — the removed `realtime` property was a redundant additional reference, not the retaining one.

## Validation

- Example app builds: `BuildTool build-example-app --platform macOS` → BUILD SUCCEEDED (same command as the `check-example-app` CI job).
- `swift build --build-tests` and `make lint` clean.
- Grep gate: no `attachAsync`/`detachAsync` references remain anywhere under `LiveObjects/Example` (the test targets' own separate async helpers are unrelated and untouched).
- Implementation independently reviewed step-by-step against the written refactoring plan, including a re-verification of the retention chain behind the property removal.


maratal (approved):
LGTM, merge this after https://github.com/ably/ably-cocoa/pull/2249 and then rebase on main before merging

PR #2257, opened 2026-08-23
[AIT-1329] Render JSONValue as compact JSON text

### LINKED ISSUE #2245: LiveObjects: conform JSONValue to CustomStringConvertible for readable, deterministic JSON output
## Problem statement

`JSONValue` is the type LiveObjects users meet first: it's what `compactJson()` returns, and printing it is the very first thing a quickstart reader does after fetching an object's state. But `JSONValue` declares no `CustomStringConvertible` conformance, so `print(json)` and string interpolation fall back to Swift's reflection-based rendering:

```swift
let json = try reactionsMap.compactJson() // logically {"likes": 10, "hearts": 0}
print("Reactions updated: \(json)")
```

prints:

```text
Reactions updated: object(["hearts": AblyLiveObjects.JSONValue.number(0.0), "likes": AblyLiveObjects.JSONValue.number(10.0)])
```

This has three problems:

1. **It's noise.** The type-name prefixes and enum-case wrappers bury the actual data. Nobody reading a console wants `AblyLiveObjects.JSONValue.number(10.0)` where `10` would do.
2. **It's nondeterministic.** Dictionary iteration order makes the printed key order vary run to run, so the same state prints differently each time.
3. **It's inconsistent with the other Ably SDKs.** In ably-java, `compactJson()` hands back a gson `JsonObject` whose `toString()` prints compact JSON text; in ably-js, `console.log` shows the plain object. Swift is the only SDK where printing the compact JSON doesn't give you something that looks like JSON — and the only one where the documentation's `// console:` example outputs can't literally match what the user sees.

There's also no public escape hatch: the `JSONValue` → Foundation bridge (`toJSONSerializationInput`) is `internal`, so a user who wants readable output has to hand-write a recursive renderer over the enum.

## Proposed solution

Conform `JSONValue` to `CustomStringConvertible`, rendering compact JSON text, with object keys sorted for deterministic output:

```swift
extension JSONValue: CustomStringConvertible {
    /// Compact JSON text, with object keys sorted for deterministic output.
    public var description: String { ... }
}
```

With that in place, the example above prints:

```text
Reactions updated: {"hearts":0,"likes":10}
```

Notes on the shape of the change:

- **Additive and non-breaking**: a conformance on a type this SDK owns, no signature changes.
- The implementation can reuse the existing internal serialization machinery, or a small recursive renderer over the six cases; sorted keys keep the output stable.
- Optionally also conform `CustomDebugStringConvertible` so LLDB's `po` gives the same readable form.

## Why this helps beyond nicer printing

- **Debugging**: `po json` in LLDB and `print` in the console immediately show the state you actually care about, instead of enum plumbing. That matters most in exactly the situations where LiveObjects users reach for `compactJson()` — inspecting what the synchronized state currently is.
- **Logging**: interpolating a `JSONValue` into any log line produces compact, greppable JSON.
- **Testing**: deterministic output (sorted keys) means test assertions and snapshot comparisons against printed state are reproducible.
- **Documentation accuracy**: the quickstart and reference examples show outputs like `// console: {"likes":10.0,"hearts":0.0}`. Today those are a logical representation rather than what Swift actually prints; with this conformance they become literally true, matching the behavior ably-java and ably-js users already get.

## References

- Swift LiveObjects docs PR where this surfaced (quickstart step 7 map subscription output): ably/website#9502
- Related API-ergonomics discussion: #2242

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5743) by [Unito](https://www.unito.io)

maratal:
Fixes #2245.

`JSONValue` declared no `CustomStringConvertible` conformance, so printing one — the first thing a reader does with the result of `compactJson()` — fell back to Swift's reflection form:

```text
Reactions updated: object(["hearts": AblyLiveObjects.JSONValue.number(0.0), "likes": AblyLiveObjects.JSONValue.number(10.0)])
```

It now prints:

```text
Reactions updated: {"hearts":0,"likes":10}
```

Conforms `JSONValue` to `CustomStringConvertible` and `CustomDebugStringConvertible`, rendering compact JSON text with object keys sorted so the output is stable between runs. Additive — no signature changes, and no change to any value the SDK computes.

> **Scope note.** This PR previously also carried a fix to how numbers are *read back* from `JSONSerialization` (`art_exactDoubleValue`). @sacOO7 asked for that to be looked at separately, which was right: it touches counters, maps and protocol conversion, and unlike this PR it changes computed values. It is now **#2261**. Both target `main`, and they do not conflict, so they can land in either order.

## Why the text is built directly rather than by `JSONSerialization`

`JSONSerialization` was the obvious implementation, and it is unsuitable in three ways, none of which it offers an option to change:

- **Numbers get a fixed 17 significant digits** rather than the shortest representation that round-trips, so a value that is not exactly representable in binary floating point prints its full expansion: `0.1` → `0.10000000000000001`, `0.3` → `0.29999999999999999`. This is the case a reader is most likely to meet, since a `LiveCounter` holding a fractional amount produces it.
- **The forward slash is escaped**, which JSON permits but does not require, so URLs — a common thing for a `LiveMap` to hold — print as `https:\/\/example.com\/a\/b`.
- **Non-finite numbers raise an uncatchable Objective-C exception.** `JSONValue.number` wraps a `Double`, so `.number(.nan)` is constructible through the public API, and `JSONSerialization` raises rather than throwing, which `try?` cannot catch — printing one would terminate the process. They are written as `null` instead, as `JSON.stringify` does.

That reasoning is recorded as a comment on the implementation rather than in the public doc comment: it explains a choice a maintainer might otherwise redo, and is not something a caller of `description` needs. What stays on the public API is the usage example and two `> Note:` asides that are contractual — the stability caveat and the non-finite behaviour.

Numbers come from `Double`'s own description, which is already shortest-round-trip, with two pieces of its punctuation dropped: the zeros trailing a fractional part, so `10.0` prints as `10` — the complaint the issue opens with — and the zeros padding an exponent, so `1e-07` prints as `1e-7`.

Where `Double`'s description chooses exponent notation is left alone, and that is not always where `JSON.stringify` chooses it: `448542840031168640` prints here as `4.4854284003116864e+17`. The digits are identical in every such case; only the threshold differs, and both forms are valid JSON that parse back to the same value. Fuzzed against `node` over 54,514 doubles, that threshold is the sole remaining difference — the digits never disagree.

Sorting the keys here also gives a plain lexicographic order, where `JSONSerialization`'s `sortedKeys` compares digits numerically and treats case as secondary.

## `JSONObjectOrArray`

Given the same two conformances, deferring to `JSONValue.description`. There was already a consumer: `ProtocolTypes.ObjectData.debugDescription` interpolates its `json` field, so every debug rendering of an object message — including Swift Testing's failure output — was printing reflection mid-message.

## Tests

`JSONValueDescriptionTests` — 22 tests covering scalars, integral and fractional numbers, the numeric edge cases (`-0`, `1e18`, `1e21`, `2^53`, `greatestFiniteMagnitude`, `leastNonzeroMagnitude`), non-finite values, both punctuation strips, every string escape including the `\u00xx` form, non-ASCII passthrough, empty/nested containers, key sorting and its determinism across three different insertion orders, escaped keys, empty keys, array order, and that interpolation, `String(describing:)` and `String(reflecting:)` all route through the conformances.

Each edge case where the output differs from `JSON.stringify` carries a `// JSON.stringify: …` comment giving the other value, so the divergence is recorded at the assertion rather than left implicit.

Two are worth calling out:

- `descriptionAgreesWithJSONSerializationBridge` checks that `description` and the internal `JSONSerialization` bridge parse back to the same value. They are independent implementations and the bridge is what actually goes on the wire, so they must not drift. Both sides go through the same bridge, so the check is agreement, not absolute fidelity — the fidelity question is #2261's.
- `0.1 + 0.2` is asserted as `0.30000000000000004`, which is correct: that genuinely is a different `Double` from `0.3`, so the long form there is not a formatting artifact.

## Verification

- `JSONValueDescriptionTests`: 22 tests passing; LiveObjects unit suite: 404 tests in 105 suites passing.
- SwiftLint (0.59.1, as pinned in `Mintfile`), EditorConfig, availability annotations and `xcodebuild docbuild` all clean.
- Fuzzed against `node` over 54,514 doubles: digits identical in all of them; the only differences are the exponent-notation threshold described above.

## Note for the docs

The quickstart and reference examples in ably/website#9502 show outputs like `// console: {"likes":10.0,"hearts":0.0}`. The actual output is `{"hearts":0,"likes":10}` — integral values print without the `.0`, and keys are sorted. Those examples need updating to match.

### DIFF
--- a/LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
+++ b/LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
@@ -333,3 +333,154 @@ internal extension JSONObjectOrArray {
         return string
     }
 }
+
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+extension JSONValue: CustomStringConvertible {
+    /// Compact JSON text, with object keys sorted for deterministic output.
+    ///
+    /// ```swift
+    /// let jsonValue: JSONValue = ["likes": 10, "hearts": 0]
+    /// print("Reactions updated: \(jsonValue)") // Reactions updated: {"hearts":0,"likes":10}
+    /// ```
+    ///
+    /// > Note: JSON has no representation for the non-finite `Double` values, so a `number` whose value
+    /// > is NaN or infinite is written as `null`.
+    ///
+    public var description: String {
+        // The text is built here rather than by `JSONSerialization`, which offers no option to change
+        // any of the three things that make its output unsuitable: it writes numbers to a fixed 17
+        // significant digits rather than the shortest form that round-trips, so `0.1` prints as
+        // `0.10000000000000001`; it escapes the forward slash, so URLs print as
+        // `https:\/\/example.com`; and it raises an Objective-C exception, uncatchable from Swift, when
+        // asked to write a non-finite number, which `number(_:)` makes constructible.
+        //
+        // Sorting the keys also gives a plain lexicographic order, where `JSONSerialization`'s
+        // `sortedKeys` compares digits numerically and treats case as secondary.
+        switch self {
+        case let .object(object):
+            let members = object
+                .sorted { $0.key < $1.key }
+                .map { key, value in "\(Self.jsonText(forString: key)):\(value.description)" }
+            return "{\(members.joined(separator: ","))}"
+        case let .array(array):
+            return "[\(array.map(\.description).joined(separator: ","))]"
+        case let .string(string):
+            return Self.jsonText(forString: string)
+        case let .number(number):
+            return Self.jsonText(forNumber: number)
+        case let .bool(bool):
+            return bool ? "true" : "false"
+        case .null:
+            return "null"
+        }
+    }
+}
+
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+extension JSONValue: CustomDebugStringConvertible {
+    /// Compact JSON text, with object keys sorted for deterministic output; the same as ``description``,
+    /// so that `String(reflecting:)` and LLDB's `po` also show the value rather than the enum's structure.
+    public var debugDescription: String {
+        description
+    }
+}
+
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+private extension JSONValue {
+    /// The JSON text for a string, including the surrounding quotation marks.
+    ///
+    /// Only the escapes that JSON requires are applied. In particular the forward slash, which JSON
+    /// permits escaping but does not require it to be escaped, is left alone so that URLs stay readable.
+    static func jsonText(forString string: String) -> String {
+        var result = "\""
+
+        for scalar in string.unicodeScalars {
+            switch scalar {
+            case "\"":
+                result += #"\""#
+            case "\\":
+                result += #"\\"#
+            case "\n":
+                result += #"\n"#
+            case "\r":
+                result += #"\r"#
+            case "\t":
+                result += #"\t"#
+            case "\u{08}":
+                result += #"\b"#
+            case "\u{0C}":
+                result += #"\f"#
+            case _ where scalar.value < 0x20:
+                result += String(format: #"\u%04x"#, scalar.value)
+            default:
+                result.unicodeScalars.append(scalar)
+            }
+        }
+
+        return result + "\""
+    }
+
+    /// The JSON text for a number.
+    ///
+    /// `Double`'s own description is the shortest representation that round-trips, so it supplies the
+    /// digits. Two pieces of its punctuation are dropped: the zeros that trail a fractional part, so
+    /// that `10` is not written as `10.0`, and the zeros that pad an exponent, so that `1e-7` is not
+    /// written as `1e-07`.
+    ///
+    /// Where `Double`'s description chooses exponent notation is left as it is, which is not always
+    /// where `JSON.stringify` chooses it: `4.4854284003116864e+17` here is `448542840031168640` there.
+    /// Both are valid JSON carrying the same digits, and both parse back to this value.
+    static func jsonText(forNumber number: Double) -> String {
+        guard number.isFinite else {
+            // NaN and the infinities have no JSON representation.
+            return "null" // .nan, .infinity -> null
+        }
+
+        let text = String(number)
+
+        guard let exponentIndex = text.firstIndex(where: { $0 == "e" || $0 == "E" }) else {
+            return withoutTrailingFractionZeros(text) // 10.0 -> 10; 0.1 -> 0.1; -0.0 -> -0
+        }
+
+        let mantissa = withoutTrailingFractionZeros(String(text[..<exponentIndex]))
+        let exponent = withoutExponentPadding(String(text[text.index(after: exponentIndex)...]))
+
+        return "\(mantissa)e\(exponent)" // 1e-07 -> 1e-7; 1.7976931348623157e+308 unchanged
+    }
+
+    /// Drops the zeros trailing a fractional part, and the decimal point too if they were all of it.
+    static func withoutTrailingFractionZeros(_ text: String) -> String {
+        // Without a point the trailing zeros are significant: 100 must not become 1.
+        guard text.contains(".") else {
+            return text
+        }
+
+        var result = text
+        while result.hasSuffix("0") {
+            result.removeLast()
+        }
+        if result.hasSuffix(".") {
+            result.removeLast()
+        }
+
+        return result // 10.0 -> 10; 1.10 -> 1.1; 0.0 -> 0; 123.456 -> 123.456
+    }
+
+    /// Drops the zeros `Double`'s description pads a single-digit exponent with.
+    static func withoutExponentPadding(_ text: String) -> String {
+        var sign = ""
+        var digits = Substring(text)
+
+        if let first = digits.first, first == "+" || first == "-" {
+            sign = String(first)
+            digits = digits.dropFirst()
+        }
+
+        // Keep the last digit even if it is a zero, so that an exponent of 0 survives.
+        while digits.count > 1, digits.first == "0" {
+            digits = digits.dropFirst()
+        }
+
+        return sign + digits // -07 -> -7; +308 -> +308
+    }
+}
--- a/LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
+++ b/LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
@@ -166,6 +166,11 @@ internal extension JSONValue {
 // MARK: - JSON objects and arrays
 
 /// A subset of ``JSONValue`` that has only `object` or `array` cases.
+///
+/// > Note: These are the two cases a JSON *text* was restricted to under RFC 4627, so `JSON` would be a
+/// > natural name for this type. It is spelled out instead: `JSON` is broad enough to be mistaken for
+/// > `JSONValue`, and current JSON (RFC 8259, and the json.org grammar linked above) allows any value at
+/// > the top level, so the short name would no longer say what the restriction is.
 @available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
 internal enum JSONObjectOrArray: Equatable {
     case object([String: JSONValue])
@@ -343,9 +348,11 @@ extension JSONValue: CustomStringConvertible {
     /// print("Reactions updated: \(jsonValue)") // Reactions updated: {"hearts":0,"likes":10}
     /// ```
     ///
+    /// > Note: This is a display and debugging aid, not a guaranteed-stable serialization format; the
+    /// > exact bytes — for example where exponent notation is used — may change.
+    ///
     /// > Note: JSON has no representation for the non-finite `Double` values, so a `number` whose value
     /// > is NaN or infinite is written as `null`.
-    ///
     public var description: String {
         // The text is built here rather than by `JSONSerialization`, which offers no option to change
         // any of the three things that make its output unsuitable: it writes numbers to a fixed 17
@@ -482,5 +489,31 @@ private extension JSONValue {
         }
 
         return sign + digits // -07 -> -7; +308 -> +308
+    }
+}
+
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+extension JSONObjectOrArray: CustomStringConvertible {
+    /// Compact JSON text, with object keys sorted for deterministic output.
+    ///
+    /// ```swift
+    /// let jsonObject: JSONObjectOrArray = ["likes": 10, "hearts": 0]
+    /// print("Reactions updated: \(jsonObject)") // Reactions updated: {"hearts":0,"likes":10}
+    /// ```
+    ///
+    /// Since this type is the `object`/`array` subset of ``JSONValue``, it defers to that type's
+    /// ``JSONValue/description`` rather than repeating the rendering; see there for how numbers and
+    /// strings are written.
+    internal var description: String {
+        toJSONValue.description
+    }
+}
+
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+extension JSONObjectOrArray: CustomDebugStringConvertible {
+    /// Compact JSON text, with object keys sorted for deterministic output; the same as ``description``,
+    /// so that `String(reflecting:)` and LLDB's `po` also show the value rather than the enum's structure.
+    internal var debugDescription: String {
+        description
     }
 }
--- a/LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
+++ b/LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
@@ -92,11 +92,18 @@ struct JSONValueDescriptionTests {
         #expect(JSONValue.number(value).description == expected)
     }
 
-    // Whatever layout is chosen, the digits have to identify the original value exactly.
+    // Whatever layout is chosen, the digits have to identify the original value exactly. `Double(_:)`
+    // is the oracle because it is correctly rounded, which `JSONSerialization`'s number parser is not;
+    // this is where numeric fidelity is guaranteed, rather than in the container round-trip below.
     @Test(arguments: [
         0.1,
         0.1 + 0.2,
         123.456,
+        // Awkward fractional values — the motivating case, a `LiveCounter` holding a fractional
+        // amount. `JSONSerialization` re-parses both of these one ULP out, so they would not survive
+        // that parser; they do survive a correctly-rounded one, which is what this asserts.
+        0.015034388851090229,
+        0.023942638935083197,
         1e-7,
         1e-21,
         1_000_000_000_000_000_128,
@@ -283,7 +290,13 @@ struct JSONValueDescriptionTests {
 
     // MARK: - Output is valid JSON
 
-    // Whatever is printed has to parse back to the value it came from.
+    // Whatever is printed has to be valid JSON that parses back to the same structure — the same keys,
+    // the same nesting, the same cases.
+    //
+    // This deliberately does not claim numeric fidelity. The oracle is `JSONSerialization`, whose
+    // number parser is not correctly rounded, so it cannot witness that a fractional value survives
+    // exactly; the arguments below stay clear of values it would misparse. `numbersRoundTrip` covers
+    // fidelity instead, through the correctly-rounded `Double(_:)`.
     @Test(arguments: [
         JSONValue.null,
         true,
--- a/LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
+++ b/LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
@@ -1,0 +1,337 @@
+@testable import AblyLiveObjects
+import Foundation
+import Testing
+
+/// Tests for `JSONValue`'s `CustomStringConvertible` and `CustomDebugStringConvertible` conformances
+///
+/// The conformances render compact JSON text directly rather than going through
+/// `JSONSerialization`, which writes numbers to 17 significant digits rather than the shortest form
+/// that round-trips. Numbers come from `Double`'s own description, so where that chooses exponent
+/// notation differently from JS `JSON.stringify` the output differs too; the tests below record
+/// those cases.
+struct JSONValueDescriptionTests {
+    // MARK: - Scalars
+
+    @Test(arguments: [
+        (value: JSONValue.null, expected: "null"),
+        (value: true, expected: "true"),
+        (value: false, expected: "false"),
+        (value: "someString", expected: #""someString""#),
+        (value: "", expected: #""""#),
+    ] as [(value: JSONValue, expected: String)])
+    func scalars(value: JSONValue, expected: String) {
+        #expect(value.description == expected)
+    }
+
+    // MARK: - Numbers
+
+    @Test(arguments: [
+        // The point of the issue: an integral value is written without the fractional part that
+        // `Double`'s own description would give it.
+        (value: 0.0, expected: "0"),
+        (value: 1, expected: "1"),
+        (value: 10, expected: "10"),
+        (value: -42, expected: "-42"),
+        (value: 123.456, expected: "123.456"),
+        (value: -0.5, expected: "-0.5"),
+        (value: 1e15, expected: "1000000000000000"),
+    ] as [(value: Double, expected: String)])
+    func numbers(value: Double, expected: String) {
+        #expect(JSONValue.number(value).description == expected)
+    }
+
+    // A value that is not exactly representable in binary floating point is written with the fewest
+    // digits that identify it uniquely, rather than its full 17-significant-digit expansion. This is
+    // the case a LiveObjects user is most likely to hit — a counter holding a fractional amount.
+    @Test(arguments: [
+        (value: 0.1, expected: "0.1"),
+        (value: 0.3, expected: "0.3"),
+        (value: 1.1, expected: "1.1"),
+        (value: 0.1 + 0.2, expected: "0.30000000000000004"), // genuinely a different Double from 0.3
+    ] as [(value: Double, expected: String)])
+    func fractionalNumbersUseShortestRoundTrip(value: Double, expected: String) {
+        #expect(JSONValue.number(value).description == expected)
+    }
+
+    // Where `Double`'s description switches to exponent notation is inherited rather than chosen, so it
+    // does not always agree with the other SDKs: the trailing comments give `JSON.stringify`'s output
+    // where it differs.
+    @Test(arguments: [
+        (value: -0.0, expected: "-0"), // JSON.stringify: 0
+        (value: 1e18, expected: "1e+18"), // JSON.stringify: 1000000000000000000
+        (value: 1e19, expected: "1e+19"), // JSON.stringify: 10000000000000000000
+        (value: 1e21, expected: "1e+21"),
+        // 2^53, the last integer whose neighbours are all exactly representable.
+        (value: 0x1p53, expected: "9007199254740992"),
+        // Above 2^53 the digits are still the shortest that round-trip, so no exact expansion appears.
+        (value: 1_000_000_000_000_000_128, expected: "1.0000000000000001e+18"), // JSON.stringify: 1000000000000000100
+        (value: 1.2345678901234567e19, expected: "1.2345678901234567e+19"), // JSON.stringify: 12345678901234567000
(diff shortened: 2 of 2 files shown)


sacOO7:
@coderabbitai review

LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
```diff
@@ -333,3 +333,119 @@ internal extension JSONObjectOrArray {
...
+
+        // Also normalises -0, which would otherwise be written as "-0".
+        if number == 0 {
+            return "0"
+        }
+
+        // An integral value is written without a fractional part (10, not 10.0), up to the magnitude at
+        // which exponent notation becomes the conventional form.
+        if number == number.rounded(), abs(number) < 1e21 {
+            return String(format: "%.0f", number)
```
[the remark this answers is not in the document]
maratal: Confirmed and fixed. The `%.0f` branch is gone — digits now come from `String(number)`, which is shortest-round-trip, so `1000000000000000128` renders as `1.0000000000000001e+18`.

Same significant digits as `JSON.stringify`'s `1000000000000000100`; only the exponent threshold differs, and both parse back to the same `Double`. That case is in `numberEdgeCases`, with the `JSON.stringify` output recorded alongside it.

sacOO7:
@coderabbitai review

sacOO7:
@coderabbitai full review

LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
```diff
@@ -333,3 +333,162 @@ internal extension JSONObjectOrArray {
...
+}
+
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+private extension JSONValue {
+    /// The JSON text for a string, including the surrounding quotation marks.
+    ///
+    /// Only the escapes that JSON requires are applied. In particular the forward slash, which JSON
+    /// permits escaping but does not require it to be escaped, is left alone so that URLs stay readable.
+    static func jsonText(forString string: String) -> String {
+        var result = "\""
```
sacOO7, line 402: Picking this up from CodeRabbit's nitpick — since the escape cases right below already use raw string literals (`#"\""#`, `#"\\"#`, …), this initializer could use one too for consistency:

```suggestion
        var result = #"""#
```

wdyt? I think it makes sense.

LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
```diff
@@ -0,0 +1,337 @@
...
+    @Test
+    func forwardSlashesAreNotEscaped() {
+        let value = JSONValue.string("https://example.com/a/b")
+
+        #expect(value.description == #""https://example.com/a/b""#)
+    }
+
+    @Test(arguments: ["héllo", "日本語", "👍🏽", "a\u{0301}"])
+    func nonASCIICharactersArePreserved(value: String) {
+        #expect(JSONValue.string(value).description == "\"\(value)\"")
```
sacOO7, line 164: Same nitpick from CodeRabbit here — the expected string could be a raw literal to avoid the escaped quotes, matching the style used elsewhere in this file (e.g. the `forwardSlashesAreNotEscaped` test):

```suggestion
        #expect(JSONValue.string(value).description == #""\#(value)""#)
```

wdyt? I think it makes sense.

LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
maratal: I think `#` notation makes sense only for complex cases to not pollute escaped string. In cases where it's just a single `\` it's ok to use it, bc a pair of `#` pollute actually more.

LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
maratal: same reason as above

sacOO7 (commented):
Nice PR — I validated the shipped printer independently and it is correct: I confirmed the 1-ULP JSONSerialization mis-parse the number path deliberately avoids, checked the empty-key and non-finite paths, and the design decisions in the doc comments all hold up. I have no blocking concerns; everything below is test-quality and polish on an already-good change. Most of my comments are inline; one is here because it touches a file outside this diff.

**Optional, and out of this PR's stated scope: `JSONObjectOrArray` still prints reflection on a real debug path.** This PR gives `JSONValue` a clean `CustomStringConvertible`, but its sibling `JSONObjectOrArray` doesn't get the same treatment, and there's a live interpolation of one. `ProtocolTypes.ObjectData.debugDescription` (`Protocol/ObjectMessage.swift:912`) does `parts.append("json: \(json)")`, where `json` is a `JSONObjectOrArray?` (declared at `ObjectMessage.swift:70`). Since `JSONObjectOrArray` has no `CustomStringConvertible`, that field renders as Swift reflection — the exact ugliness this PR removes for `JSONValue`. (The parallel `WireObjectData` site at `Protocol/WireObjectMessage.swift:761` interpolates a `String`, so it's unaffected.) Because `JSONObjectOrArray` already exposes `toJSONValue` (`JSONValue.swift:209`), the fix is a one-liner you could drop in right next to the new extension:

```swift
@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
extension JSONObjectOrArray: CustomStringConvertible {
    public var description: String {
        toJSONValue.description
    }
}
```

Strictly outside this PR's scope, so entirely your call whether to fold it in here or leave it for a follow-up — flagging it only because this PR is the natural moment for it. I've also left an inline note on the new `JSONValue` extension pointing at this.

LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
```diff
@@ -0,0 +1,337 @@
...
+        "https://example.com/a/b",
+        "日本語 👍🏽",
+        [:],
+        [],
+        ["someKey": "someValue", "nested": ["a": [1, 2, .null]], "n": 10],
+        ["someElement", 1, true, .null, ["k": "v"]],
+    ] as [JSONValue])
+    func descriptionRoundTrips(value: JSONValue) throws {
+        let data = try #require(value.description.data(using: .utf8))
+        let deserialized = try JSONSerialization.jsonObject(with: data, options: [.fragmentsAllowed])
```
sacOO7, line 322: _Part of my review pass on this PR — overall assessment (and one suggestion touching a file outside this diff) in [the review summary](https://github.com/ably/ably-cocoa/pull/2257#pullrequestreview-5041189005)._

The printer itself is correct here — my concern is only the oracle this test validates against. `descriptionRoundTrips` parses `value.description` with `JSONSerialization.jsonObject` and compares back via `Equatable`, and the comment above ("Whatever is printed has to parse back to the value it came from") reads as a numeric-fidelity guarantee. But Foundation's `JSONSerialization` number parser is not correctly rounded, so it can't actually keep that promise for fractional doubles.

I measured this directly. For `0.015034388851090229` the PR emits exactly `0.015034388851090229` (the correct shortest form), but `JSONSerialization.jsonObject` parses that text back to `0.015034388851090227` — bitPattern `0x3f8eca5983723c8c` becoming `0x3f8eca5983723c8b`, one ULP off. `Double("0.015034388851090229")` and `JSONDecoder` both return the exact original. Fuzzing ~200k random finite doubles, about 0.26% re-parse to a different bit pattern this way. The test only stays green because its two fractional arguments (`0.1`, `123.456`) happen to be values `JSONSerialization` parses exactly; adding an arbitrary fractional value — precisely the motivating case, a `LiveCounter` holding a fractional amount — would make it fail on the validator, not because the printed text is wrong.

I'd suggest validating through a correctly-rounded parser (the way `numbersRoundTrip` at line 109 already does with `Double(text)`), and then adding a genuinely awkward fractional value so the container round-trip actually guarantees numeric fidelity. For a container it's easiest to decode into `JSONValue` directly, e.g.:

```swift
func descriptionRoundTrips(value: JSONValue) throws {
    let data = try #require(value.description.data(using: .utf8))
    let decoded = try JSONDecoder().decode(JSONValue.self, from: data)
    #expect(decoded == value)
}
```

(`JSONValue` isn't `Decodable` today, so this needs a small test-only `Decodable` conformance — or keep the container cases as exact-string assertions and use `Double(text)` for the numeric leaves, which is `numbersRoundTrip`'s approach.) Then a hard value like `0.015034388851090229` in the arguments would exercise the very case the feature exists for. Worth softening the comment to match too — as written it claims more than the current validator can deliver.

LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
```diff
@@ -0,0 +1,337 @@
...
+    // The rendering has to agree with the `JSONSerialization` bridge that is used to put a `JSONValue`
+    // on the wire, otherwise what a user is shown is not what gets published.
+    @Test(arguments: [
+        JSONValue.null,
+        true,
+        0,
+        10,
+        -42,
+        123.456,
+        0.1,
```
sacOO7, line 336: _Part of my review pass on this PR — overall assessment (and one suggestion touching a file outside this diff) in [the review summary](https://github.com/ably/ably-cocoa/pull/2257#pullrequestreview-5041189005)._

This test is sound as-is — it parses both `value.description` and the bridge output through the *same* `JSONSerialization`, so any shared parser imprecision cancels and it genuinely checks for drift between the printer and the wire bridge rather than for absolute numeric fidelity. The only thing I'd add is a harder number: the sole fractional argument here is `0.1`. Dropping in something awkward like `0.015034388851090229` would pass (both sides absorb the identical 1-ULP imprecision) while actually exercising the drift check on a difficult value rather than an easy one. Purely additive.

LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
```diff
@@ -0,0 +1,337 @@
...
+        let renderings = insertionOrders.map { pairs in
+            JSONValue.object(.init(uniqueKeysWithValues: pairs)).description
+        }
+
+        #expect(Set(renderings).count == 1)
+        #expect(renderings[0] == #"{"claps":3,"hearts":0,"likes":10}"#)
+    }
+
+    @Test
+    func objectKeysAreEscaped() {
```
sacOO7, line 226: _Part of my review pass on this PR — overall assessment (and one suggestion touching a file outside this diff) in [the review summary](https://github.com/ably/ably-cocoa/pull/2257#pullrequestreview-5041189005)._

One small edge that isn't covered anywhere: an empty-string object key. `emptyContainers` handles `{}`/`[]` and this test handles a key needing escaping, but nothing pins `["": 1]` → `{"":1}`. An empty string is a legal JSON member name and a classic edge; I checked the implementation and it does render it correctly as `{"":1}`, so this is just locking in that behaviour. Cheap to add right alongside this one:

```swift
@Test
func emptyStringObjectKey() {
    let value = JSONValue.object(["": 1])

    #expect(value.description == #"{"":1}"#)
}
```

LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
```diff
@@ -333,3 +333,162 @@ internal extension JSONObjectOrArray {
         return string
     }
 }
+
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+extension JSONValue: CustomStringConvertible {
```
sacOO7, line 343: _Part of my review pass on this PR — overall assessment (and one suggestion touching a file outside this diff) in [the review summary](https://github.com/ably/ably-cocoa/pull/2257#pullrequestreview-5041189005)._

Not for this line specifically, but this is the natural place to raise it: `JSONObjectOrArray` doesn't get the same `CustomStringConvertible` treatment, and there's a live path that still prints reflection because of it. `ProtocolTypes.ObjectData.debugDescription` (`Protocol/ObjectMessage.swift:912`) interpolates `json` where `json` is a `JSONObjectOrArray?` (`ObjectMessage.swift:70`), so that field renders as Swift reflection — the exact ugliness this extension removes for `JSONValue`. (The parallel `WireObjectData` site at `WireObjectMessage.swift:761` interpolates a `String`, so it's fine.) Since `JSONObjectOrArray` already has `toJSONValue` (line 209), it's a one-liner you could add right here:

```swift
@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
extension JSONObjectOrArray: CustomStringConvertible {
    public var description: String {
        toJSONValue.description
    }
}
```

Strictly outside this PR's stated scope, so entirely your call — but this seems like the natural moment for it.

LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
```diff
@@ -333,3 +333,162 @@ internal extension JSONObjectOrArray {
...
+    /// Compact JSON text, with object keys sorted for deterministic output.
```
sacOO7, line 344: _Part of my review pass on this PR — overall assessment (and one suggestion touching a file outside this diff) in [the review summary](https://github.com/ably/ably-cocoa/pull/2257#pullrequestreview-5041189005)._

Making a public enum `CustomStringConvertible` and documenting an exact expected string (the `{"hearts":0,"likes":10}` example just below) effectively invites users to match on or parse this text. But as you note further down on `jsonText(forNumber:)` (lines 438-440), the exponent-notation threshold deliberately differs from `JSON.stringify`, so the exact bytes aren't a stable serialization format — they're a display/debug rendering. That caveat lives only on the number helper, which a consumer reading this public doc won't necessarily see. I'd add a one-line caution here so nobody builds on the exact output, e.g.:

```swift
/// > Note: This is a display and debugging aid, not a guaranteed-stable serialization format;
/// > the exact bytes (for example where exponent notation is used) may change. Use the
/// > `JSONSerialization` bridge if you need a stable on-the-wire encoding.
```

LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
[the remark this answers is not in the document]
maratal: Housekeeping: the branch was rebased, so the hashes in this thread's earlier history no longer resolve. The number-rendering work described above is now in `29393902`, with the surrounding test and documentation changes in `0c1ed2e8`.

Behaviour is unchanged from my previous reply — `Double`'s description supplies the digits, with trailing fractional zeros and exponent padding stripped, and the exponent-notation threshold is inherited rather than chosen and documented as such.

LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
maratal: Confirmed and fixed in `29393902` — though the fix landed a layer deeper than suggested.

I reproduced it exactly: `0.015034388851090229` re-parses to bit pattern `0x3f8eca5983723c8b` where the correct value is `0x…8c`, and fuzzing 199,913 random finite doubles gave 530 wrong (0.265%) against your ~0.26%, with `JSONDecoder` and `Double(_:)` both clean.

Digging into where the digit goes, it is not the parser. `JSONSerialization` returns an `NSDecimalNumber` whose `stringValue` and `decimalValue` are both exact — all 17 digits survive. The loss is in `NSDecimalNumber.doubleValue`, i.e. in this SDK's own `createNumberValue: { $0.doubleValue }`. Every Foundation bridge shares that conversion and is wrong on identically the same values: `Double(truncating:)`, `Double(exactly:)`, `as? Double`. `Double(exactly:)` is the notable one — it never returns `nil` for these, so the API whose contract is "nil if not exactly representable" reports success while rounding.

So rather than change the oracle, `NSNumber.art_exactDoubleValue` now reads the digits back as text for the `NSDecimalNumber` case and lets `Double(_: String)` convert. Fuzzed over 200,000 doubles: 0 wrong, against 513–566 for `doubleValue`. The type test is load-bearing — an unconditional `stringValue` read is wrong for ~45% of values, because a plain `NSNumber`'s `stringValue` rounds to 16 significant digits and turns `greatestFiniteMagnitude` into infinity.

That removes the need for a test-only `Decodable` conformance, and it means `descriptionRoundTrips` now genuinely delivers the numeric-fidelity guarantee its comment claimed, so `0.015034388851090229` and `0.023942638935083197` are in its arguments as you asked rather than the comment being softened.

It applies to production too: the same conversion sits on the inbound path via `JSONObjectOrArray.init(jsonString:)`, so real `ObjectData.json` payloads were losing a ULP. Ten wire-facing sites now use it, including the COUNTER_INC amount.

LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
maratal: Added — `0.015034388851090229` and `0.023942638935083197` are in this test's arguments.

Your reading is the one that holds: both sides go through the same bridge, so the shared imprecision cancels and this exercises the drift check on a hard value rather than an easy one. Absolute numeric fidelity is asserted separately, in `numbersRoundTrip` against the correctly-rounded `Double(_: String)`, and the inbound rounding defect itself is #2261.

LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
maratal: Added in `d09b6b4b`, as `emptyKeyIsValid`. Confirmed it renders `{"":1}` as you found. Cheap and a classic edge — thanks.

LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
maratal: Added in `d09b6b4b`, with `CustomDebugStringConvertible` alongside it so `String(reflecting:)` and `po` agree with `description`.

Your diagnosis of the live path was the reason to do it rather than symmetry: `ProtocolTypes.ObjectData.debugDescription` interpolates `json`, so every debug rendering of an object message was printing reflection mid-message. I confirmed the `WireObjectMessage.swift:761` site interpolates a `String` and needed nothing.

One deviation from your snippet: the members are `internal`, not `public`, since the type is internal.

LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
maratal: Added in `d09b6b4b`, placed ahead of the usage example so it is read before the output that invites the dependency.

I dropped your second sentence. The `JSONSerialization` bridge is not reachable from outside the module, so pointing users at it would point them at an API they cannot call. #2245 raised the same gap, so if we do expose it the sentence can come back.

LiveObjects/Sources/AblyLiveObjects/Utility/JSONValue.swift
```diff
@@ -333,3 +338,191 @@ internal extension JSONObjectOrArray {
...
+}
+
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+extension JSONObjectOrArray: CustomDebugStringConvertible {
+    /// Compact JSON text, with object keys sorted for deterministic output; the same as ``description``,
+    /// so that `String(reflecting:)` and LLDB's `po` also show the value rather than the enum's structure.
+    internal var debugDescription: String {
+        description
+    }
+}
```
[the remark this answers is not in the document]
maratal: Not applicable — the suggested code does not compile:

```
error: 'internal' modifier cannot be used with extensions that declare protocol conformances
```

The convention holds for plain extensions (`internal extension JSONValue` at line 139 follows it), but a conformance extension has no extension-level option, so per-member modifiers are the only way to be explicit — which is what the neighbouring `JSONObjectOrArray` and `JSONValue` conformances already do.

sacOO7 (approved):
LGTM

sacOO7 (changes requested):
Seems, its also affecting other liveobjects source files, let me check again

LiveObjects/Sources/AblyLiveObjects/Utility/NSNumber+Extensions.swift
```diff
@@ -0,0 +1,42 @@
...
+    /// success while handing back a rounded value.
+    ///
+    /// The exact digits survive in the decimal representation, so the way out is to read them back as
+    /// text and let `Double(_: String)`, which is correctly rounded, do the conversion.
+    ///
+    /// The `NSDecimalNumber` test is not defensive. For an ordinary `__NSCFNumber`, which already holds
+    /// a `Double`, `stringValue` rounds to 16 significant digits and loses precision that `doubleValue`
+    /// would have preserved — `greatestFiniteMagnitude` comes back as `1.797693134862316e+308`, which
+    /// re-parses to infinity. Taking the text route unconditionally is wrong for about 45% of values.
+    var art_exactDoubleValue: Double {
```
sacOO7: @coderabbitai tell me more about this extension

maratal:
@sacOO7 — you were right to pull the approval, so I've split it rather than argue the scope.

This PR is now only the printing work: `JSONValue.swift` plus its tests, no behaviour change to anything the SDK computes. That's the part you'd already LGTM'd at the point you flipped.

The `art_exactDoubleValue` work — the ten call sites across `InternalDefaultLiveCounter`, `InternalDefaultLiveMap`, `ObjectMessage` and `WireValue` that you spotted — is now #2261, also targeting `main`. They don't conflict, so they can land in either order.

Two things I checked while splitting that make #2261 worth a proper look rather than a rubber stamp:

- The trigger is the *textual shape*, not the value. `0.000012345678901234567` comes back as an `NSDecimalNumber` and loses a ULP; `1.2345678901234567e-5` — the same `Double` — comes back as a plain `NSNumber` and is fine.
- That makes it reachable from ordinary cross-SDK traffic. We decode JSON with `NSJSONSerialization` (`Source/ARTJsonEncoder.m:18`), and JS's `JSON.stringify` writes fixed-point notation all the way down to `1e-6` — so `JSON.stringify(0.004910368508982831)` is exactly the shape that trips it. A JS client incrementing a counter by that amount leaves the Swift client one ULP off what every other SDK reports.

I've also cut the `art_exactDoubleValue` doc comment down: the fuzzing percentages were session artifacts that nobody will re-run, so the interop case above is the justification instead. Same for the `JSONValue.description` doc comment here — the "why not `JSONSerialization`" rationale moved off the public DocC page and into an implementation comment.

maratal:
Tidied up after the split — force-pushed, so a note on what moved:

- The branch is now two commits: `f8178049` (the printing work) and `d09b6b4b` (`JSONObjectOrArray` + the stability caution). A third commit that only trimmed doc comments has been folded into the two it belonged to, so there is no cleanup commit to read past.
- Trimming that doc comment had broken `lint` — it left the `///` block detached from the declaration (`orphaned_doc_comment`). Fixed by moving the rationale inside the getter body, and verified against SwiftLint 0.59.1 as pinned in `Mintfile`: the previous revision reproduces the violation, this one is clean. All four static checks are green.
- I have updated my replies in the threads above. The commit hashes several of them cited no longer resolve, and two made claims that were true only while `art_exactDoubleValue` was still here — those now carry a marked correction rather than a silent edit.

@sacOO7 no action needed on your side beyond a re-review when convenient; the change request is still on the PR.

LiveObjects/Tests/AblyLiveObjectsTests/Unit/JSONValueDescriptionTests.swift
maratal: Fixed in `91790b78`, taking your second suggestion (no test-only `Decodable` needed):

- `numbersRoundTrip` validates through the correctly-rounded `Double(_: String)`, so it gains `0.015034388851090229` and `0.023942638935083197` — both confirmed to re-parse one ULP out via `JSONSerialization` and exactly via `Double(_:)`.
- `descriptionRoundTrips` keeps a structural role only, and its comment now says so instead of claiming numeric fidelity.

The underlying `JSONSerialization` rounding defect you measured is fixed separately in #2261, which corrects the inbound path rather than just the test oracle.

LiveObjects/Sources/AblyLiveObjects/Utility/NSNumber+Extensions.swift
maratal: see #2261

sacOO7 (approved):
LGTM

PR #2260, opened 2026-08-28
[AIT-1330] Add per-case value accessors to the primitive path object and instance

### LINKED ISSUE #2244: Add `*Value` helper accessors in `PrimitivePathObject` and `PrimitiveInstance`
Add `*Value` helper accessors in `PrimitivePathObject` and `PrimitiveInstance` to avoid long resolution path like:

`let value = try root.get(key: "myKey").asPrimitive().value()?.stringValue`

Should be:

`let value = try root.get(key: "myKey").asPrimitive().stringValue()`.

[https://ably.atlassian.net/browse/AIT-1330](https://ably.atlassian.net/browse/AIT-1330)

_Originally posted by @maratal in_ [_https://github.com/ably/ably-cocoa/pull/2243#discussion_r3804602675_](https://github.com/ably/ably-cocoa/pull/2243#discussion_r3804602675)

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5742) by [Unito](https://www.unito.io)

maratal:
Closes https://github.com/ably/ably-cocoa/issues/2244

Docs PR - https://github.com/ably/website/pull/9704

Reading a single expected primitive type meant unwrapping a `Primitive` first — `try node.asPrimitive().value()?.stringValue` — even though the expected type is usually known at the call site. This lifts `Primitive`'s per-case getters onto both places you hold a primitive from.

`PrimitivePathObject` gains six methods:

```swift
let name = try root.get(key: "name").asPrimitive().stringValue()
```

Methods rather than properties, for the same reason as the rest of `PathObject`'s accessors (`instance()`, `exists()`, `type()`, …): each resolves the path at call time and is therefore O(path length), so the call parentheses are carrying real information. `PrimitiveInstance` gains the same six as throwing properties, matching the shape of the `value` requirement they wrap — an `Instance` is bound to an already-resolved value, so those reads are O(1).

Both sets delegate to `Primitive`'s getters, which move to an `internal extension`. Callers outside the SDK now reach them through these accessors, or by `switch`ing over the enum — which is how `Primitive`'s own doc comment already describes discriminating it. Three external call sites moved across as a result:

- `Test/UTS/integration/standard/objects/helpers/ObjectsIntegrationHelpers.swift` — the only non-`@testable` importer in the test tree; both helpers collapse to one line.
- `LiveObjects/Example/.../TaskBoardViewModel.swift` — a two-clause `if let` becomes one.
- `LiveObjects/README.md` — the 1.3.0 migration example, which would otherwise no longer compile for a reader following it.

## Public API removals

Two, both deliberate:

- A user holding a `Primitive` can no longer write `.stringValue` on it. The getters are reached through the accessors above, or by `switch`ing the enum.
- `LiveMapValue` loses the same six. It is the **write** type — the value you construct and pass to `set` — so asking it which case it holds answers a question its holder settled when they built it. They came from plugin `main`, where `LiveMapValue` doubled as the read type (`map.get(key:) -> LiveMapValue?`), so `?.stringValue` *was* how you read a map entry; [ably-liveobjects-swift-plugin#134](https://github.com/ably/ably-liveobjects-swift-plugin/pull/134) narrowed it to write-only and its review called them vestigial there. `primitiveValue`, `liveMapValue`, `liveCounterValue` and the `ExpressibleBy*Literal` conformances stay — the conformances being the part that review called valuable. One call site: `Test/UTS/unit/objects/ValueTypesTests.swift`, which drops to comparing the whole value through those conformances and so now reads exactly like the UTS line above it (`vt.entries["name"] == "Alice"`), while also rejecting an absent key or a non-primitive entry that the case getter let through.

The rule that leaves: the read types (`PrimitivePathObject`, `PrimitiveInstance`) carry the typed shortcuts, a bare `Primitive` is discriminated by `switch`, and the write type carries neither.

Both getters were `public` at the `1.3.0` tag, so this is source-breaking against a shipped release — permitted by the experimental warning in `LiveObjects/README.md`, but the compiler error (`value of type 'Primitive' has no member 'stringValue'`) names no replacement. `LiveObjects/CHANGELOG.md` is kept only for the standalone plugin's history (`CONTRIBUTING.md`), and the root `CHANGELOG.md` is written at release time from the release's PR list, so here is that entry ready to copy:

> * The per-case getters on `Primitive` (`stringValue`, `numberValue`, `boolValue`, `dataValue`, `jsonArrayValue`, `jsonObjectValue`) and the same six on `LiveMapValue` are no longer public. Read one expected primitive type through the new accessors on `PrimitivePathObject` (`try node.asPrimitive().stringValue()`) or `PrimitiveInstance` (`try instance.stringValue`), or `switch` over `Primitive`'s cases. `LiveMapValue` keeps `primitiveValue`, `liveMapValue`, `liveCounterValue` and its `ExpressibleBy*Literal` conformances.

## Tests

`DefaultPathObjectTests.primitiveConvenienceAccessors` and `DefaultInstanceTests.primitiveConvenienceAccessors`, one per protocol. Each exercises all six accessors positively plus a wrong-case read; the path-object case also covers a path that does not resolve. The path-object test seeds its own root map holding one entry per `Primitive` case, kept separate from the shared `makeFixture()` so the entry counts `mapReads` asserts on do not shift.

`readOnDetachedChannelThrows` on both sides gains one assertion that an accessor propagates the `RTO25` precondition error. The regression that guards is silent: SE-0230 flattens nested optionals, so `try? value.stringValue` has type `String?` rather than `String??` and compiles cleanly in the `String?`-returning throwing getter, turning the error into `nil` — indistinguishable from a wrong-case read. Verified by mutation: with that one-character change the suite is green without the assertion and fails with it.

These accessors are SDK conveniences rather than spec points, so the tests carry no `@spec` tag.

## Checks

- `swift build --build-tests` clean
- 409/409 LiveObjects unit tests
- 380/380 `swift test --filter UTS`
- Example app builds for macOS via `BuildTool build-example-app`
- `Scripts/annotate-liveobjects-availability.py --check` clean

> [!NOTE]
> Stacked on #2257, so the base is `fix/2245-JSONValue-printing` and GitHub will retarget this to `main` once that merges. The two changes are independent — nothing here touches `JSONValue`.

## Summary by CodeRabbit

* **New Features**
  * Added convenient accessors for reading string, number, Boolean, binary, and JSON values from path-based objects and primitive instances.
  * Accessors return no value when the requested type does not match.

* **Documentation**
  * Updated the migration example to reflect the current primitive-value access pattern.

* **Bug Fixes**
  * Updated task-board and integration helpers to use supported primitive accessors.

* **Tests**
  * Added coverage for matching, mismatched, and error-propagating primitive accessors.

### DIFF
--- a/LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/Instance.swift
+++ b/LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/Instance.swift
@@ -171,6 +171,55 @@ public protocol PrimitiveInstance: Sendable {
     func compactJson() throws(ARTErrorInfo) -> JSONValue
 }
 
+/// Convenience accessors for a single expected primitive type.
+///
+/// Each of these reads ``PrimitiveInstance/value`` and then the correspondingly-named getter on the
+/// resulting ``Primitive``, so `nil` means that the wrapped primitive is of a different case.
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+public extension PrimitiveInstance {
+    /// If the wrapped primitive has case `string`, this returns the associated value. Else, it returns `nil`.
+    var stringValue: String? {
+        get throws(ARTErrorInfo) {
+            try value.stringValue
+        }
+    }
+
+    /// If the wrapped primitive has case `number`, this returns the associated value. Else, it returns `nil`.
+    var numberValue: Double? {
+        get throws(ARTErrorInfo) {
+            try value.numberValue
+        }
+    }
+
+    /// If the wrapped primitive has case `bool`, this returns the associated value. Else, it returns `nil`.
+    var boolValue: Bool? {
+        get throws(ARTErrorInfo) {
+            try value.boolValue
+        }
+    }
+
+    /// If the wrapped primitive has case `data`, this returns the associated value. Else, it returns `nil`.
+    var dataValue: Data? {
+        get throws(ARTErrorInfo) {
+            try value.dataValue
+        }
+    }
+
+    /// If the wrapped primitive has case `jsonArray`, this returns the associated value. Else, it returns `nil`.
+    var jsonArrayValue: [JSONValue]? {
+        get throws(ARTErrorInfo) {
+            try value.jsonArrayValue
+        }
+    }
+
+    /// If the wrapped primitive has case `jsonObject`, this returns the associated value. Else, it returns `nil`.
+    var jsonObjectValue: [String: JSONValue]? {
+        get throws(ARTErrorInfo) {
+            try value.jsonObjectValue
+        }
+    }
+}
+
 // MARK: - AsyncSequence subscription variants
 
 /// `AsyncStream`-based subscription for map instances.
--- a/LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/PathObject.swift
+++ b/LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/PathObject.swift
@@ -185,3 +185,44 @@ public protocol PrimitivePathObject: PathObject {
     /// `RTTS6b`.
     func value() throws(ARTErrorInfo) -> Primitive?
 }
+
+/// Convenience accessors for a single expected primitive type.
+///
+/// Each of these resolves the path via ``PrimitivePathObject/value()`` and then reads the
+/// correspondingly-named getter on the resulting ``Primitive``, so `nil` means either that nothing
+/// primitive resolved at the path or that what did resolve was a primitive of a different case.
+///
+/// These are methods, not properties, for the same reason as the rest of ``PathObject``'s
+/// path-resolving accessors: each resolves the path at call time and is therefore O(path length).
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+public extension PrimitivePathObject {
+    /// If the value at this path is a `string` primitive, this returns the associated value. Else, it returns `nil`.
+    func stringValue() throws(ARTErrorInfo) -> String? {
+        try value()?.stringValue
+    }
+
+    /// If the value at this path is a `number` primitive, this returns the associated value. Else, it returns `nil`.
+    func numberValue() throws(ARTErrorInfo) -> Double? {
+        try value()?.numberValue
+    }
+
+    /// If the value at this path is a `bool` primitive, this returns the associated value. Else, it returns `nil`.
+    func boolValue() throws(ARTErrorInfo) -> Bool? {
+        try value()?.boolValue
+    }
+
+    /// If the value at this path is a `data` primitive, this returns the associated value. Else, it returns `nil`.
+    func dataValue() throws(ARTErrorInfo) -> Data? {
+        try value()?.dataValue
+    }
+
+    /// If the value at this path is a `jsonArray` primitive, this returns the associated value. Else, it returns `nil`.
+    func jsonArrayValue() throws(ARTErrorInfo) -> [JSONValue]? {
+        try value()?.jsonArrayValue
+    }
+
+    /// If the value at this path is a `jsonObject` primitive, this returns the associated value. Else, it returns `nil`.
+    func jsonObjectValue() throws(ARTErrorInfo) -> [String: JSONValue]? {
+        try value()?.jsonObjectValue
+    }
+}
--- a/LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/ValueTypes.swift
+++ b/LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/ValueTypes.swift
@@ -46,11 +46,16 @@ public enum Primitive: Sendable, Equatable {
     case data(Data)
     case jsonArray([JSONValue])
     case jsonObject([String: JSONValue])
+}
 
-    // MARK: - Convenience getters for associated values
+// MARK: - Primitive convenience getters for associated values
 
+/// These back the same-named accessors on ``PrimitivePathObject`` and ``PrimitiveInstance``, which is
+/// how callers outside the SDK reach them; on a `Primitive` in hand, `switch` over the cases.
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+internal extension Primitive {
     /// If this `Primitive` has case `string`, this returns the associated value. Else, it returns `nil`.
-    public var stringValue: String? {
+    var stringValue: String? {
         if case let .string(value) = self {
             return value
         }
@@ -58,7 +63,7 @@ public enum Primitive: Sendable, Equatable {
     }
 
     /// If this `Primitive` has case `number`, this returns the associated value. Else, it returns `nil`.
-    public var numberValue: Double? {
+    var numberValue: Double? {
         if case let .number(value) = self {
             return value
         }
@@ -66,7 +71,7 @@ public enum Primitive: Sendable, Equatable {
     }
 
     /// If this `Primitive` has case `bool`, this returns the associated value. Else, it returns `nil`.
-    public var boolValue: Bool? {
+    var boolValue: Bool? {
         if case let .bool(value) = self {
             return value
         }
@@ -74,7 +79,7 @@ public enum Primitive: Sendable, Equatable {
     }
 
     /// If this `Primitive` has case `data`, this returns the associated value. Else, it returns `nil`.
-    public var dataValue: Data? {
+    var dataValue: Data? {
         if case let .data(value) = self {
             return value
         }
@@ -82,7 +87,7 @@ public enum Primitive: Sendable, Equatable {
     }
 
     /// If this `Primitive` has case `jsonArray`, this returns the associated value. Else, it returns `nil`.
-    public var jsonArrayValue: [JSONValue]? {
+    var jsonArrayValue: [JSONValue]? {
         if case let .jsonArray(value) = self {
             return value
         }
@@ -90,7 +95,7 @@ public enum Primitive: Sendable, Equatable {
     }
 
     /// If this `Primitive` has case `jsonObject`, this returns the associated value. Else, it returns `nil`.
-    public var jsonObjectValue: [String: JSONValue]? {
+    var jsonObjectValue: [String: JSONValue]? {
         if case let .jsonObject(value) = self {
             return value
         }
--- a/LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/ValueTypes.swift
+++ b/LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/ValueTypes.swift
@@ -223,24 +223,6 @@ public enum LiveMapValue: Sendable, Equatable {
         }
         return nil
     }
-
-    /// If this `LiveMapValue` wraps a `string` primitive, this returns the associated value. Else, it returns `nil`.
-    public var stringValue: String? { primitiveValue?.stringValue }
-
-    /// If this `LiveMapValue` wraps a `number` primitive, this returns the associated value. Else, it returns `nil`.
-    public var numberValue: Double? { primitiveValue?.numberValue }
-
-    /// If this `LiveMapValue` wraps a `bool` primitive, this returns the associated value. Else, it returns `nil`.
-    public var boolValue: Bool? { primitiveValue?.boolValue }
-
-    /// If this `LiveMapValue` wraps a `data` primitive, this returns the associated value. Else, it returns `nil`.
-    public var dataValue: Data? { primitiveValue?.dataValue }
-
-    /// If this `LiveMapValue` wraps a `jsonArray` primitive, this returns the associated value. Else, it returns `nil`.
-    public var jsonArrayValue: [JSONValue]? { primitiveValue?.jsonArrayValue }
-
-    /// If this `LiveMapValue` wraps a `jsonObject` primitive, this returns the associated value. Else, it returns `nil`.
-    public var jsonObjectValue: [String: JSONValue]? { primitiveValue?.jsonObjectValue }
 }
 
 // MARK: - ExpressibleBy*Literal conformances
--- a/LiveObjects/Tests/AblyLiveObjectsTests/Unit/DefaultInstanceTests.swift
+++ b/LiveObjects/Tests/AblyLiveObjectsTests/Unit/DefaultInstanceTests.swift
@@ -212,6 +212,11 @@ struct DefaultInstanceTests {
         #expect(throws: ARTErrorInfo.self) {
             _ = try primitive.value
         }
+
+        // The convenience accessors delegate to `value`, so they must propagate the same error.
+        #expect(throws: ARTErrorInfo.self) {
+            _ = try primitive.stringValue
+        }
     }
 
     // MARK: - compactJson (RTINS11 -> RTPO13c / RTPO14b)
--- a/LiveObjects/Tests/AblyLiveObjectsTests/Unit/DefaultInstanceTests.swift
+++ b/LiveObjects/Tests/AblyLiveObjectsTests/Unit/DefaultInstanceTests.swift
@@ -163,6 +163,33 @@ struct DefaultInstanceTests {
         #expect(primitive.type == .string)
     }
 
+    @Test
+    func primitiveConvenienceAccessors() throws {
+        let internalQueue = TestFactories.createInternalQueue()
+        let coreSDK = MockCoreSDK(channelState: .attached, internalQueue: internalQueue)
+
+        func primitiveInstance(_ value: InternalLiveMapValue) throws -> any PrimitiveInstance {
+            let instance = Instance.from(internalValue: value, coreSDK: coreSDK, realtimeObjects: MockRealtimeObjects(), internalQueue: internalQueue)
+            let primitive: (any PrimitiveInstance)? = switch instance {
+            case let .primitive(primitive):
+                primitive
+            case .liveMap, .liveCounter:
+                nil
+            }
+            return try #require(primitive)
+        }
+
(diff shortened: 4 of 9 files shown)


LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/ValueTypes.swift
```diff
@@ -46,51 +46,56 @@ public enum Primitive: Sendable, Equatable {
...
-    // MARK: - Convenience getters for associated values
+// MARK: - Primitive convenience getters for associated values
 
+/// These back the same-named accessors on ``PrimitivePathObject`` and ``PrimitiveInstance``, which is
+/// how callers outside the SDK reach them; on a `Primitive` in hand, `switch` over the cases.
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+internal extension Primitive {
     /// If this `Primitive` has case `string`, this returns the associated value. Else, it returns `nil`.
-    public var stringValue: String? {
+    var stringValue: String? {
```
sacOO7, line 58: Okay, so those are only `getter` values, we no longer need them right?
So, only way to access right value from `primitive.value` would be 
i.e. if I want to get `string` from `primitive.value`, I have to do

For pathObject
```
if case .string(let score)? = try rootObject.get(key: "score").asPrimitive.value() {
    print(score)
}
```

For Instance
```
witch try rootObject.get(key: "score").instance() {
case .primitive(let primitive):
    if case .string(let score)? = try primitive.value {
      print(score)
    }
```
right?

sacOO7: As discussed, you update swift `typing` section in web docs

maratal:
```
if case .string(let score)? = try rootObject.get(key: "score").asPrimitive().value() {
    print(score)
}
```

```
if case .primitive(let score) = try root.get(key: "score").instance() {
   print(try score.numberValue ?? 0)
}
```
Almost correct for PO, for Instance I wouldn't use double pattern match. For a single primitive case you should be using new helper instead. This is the whole point of this changes. `.value()` and `.value` are for `switch` through all cases.

maratal: >Okay, so those are only getter values, we no longer need them right?

We need them to call from both PO and instance, because otherwise you need to do unwrapping in both places like:

```
var stringValue: String? {
     if case let .string(value) = self {
         return value
     }
     return nil
}
```
Which is annoying. Also if someone will want to get those back, we just change access modifiers.

umair-ably (commented):
Reviewed the diff plus the surrounding public API on the branch. No correctness bugs found — the `try?` flattening in the UTS helpers is sound (`String??` collapses, so `nil` still means absent / wrong-type / threw, exactly as before), and the separate fixture keeps the `mapReads` counts stable. Questions below, plus two that have no diff line to hang off:

**Changelog for the breaking removal** (`LiveObjects/CHANGELOG.md`, untouched here): these getters shipped publicly in 1.3.0, so this is a source-breaking removal — permitted by the experimental warning in the README, but the compiler error a user hits gives no migration hint. Should there be an entry pointing them at the new accessors or at `switch`?

**Three shapes for one identifier**: `stringValue()` on the path object, `stringValue` on the instance, `stringValue` on `LiveMapValue`. The O(path length) vs O(1) rationale makes sense, but a user's autocomplete now shows the same name in three shapes. Is the docs PR (ably/website#9704) calling that distinction out explicitly?

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/ValueTypes.swift
```diff
@@ -46,51 +46,56 @@ public enum Primitive: Sendable, Equatable {
...
     case jsonObject([String: JSONValue])
+}
 
-    // MARK: - Convenience getters for associated values
+// MARK: - Primitive convenience getters for associated values
 
+/// These back the same-named accessors on ``PrimitivePathObject`` and ``PrimitiveInstance``, which is
+/// how callers outside the SDK reach them; on a `Primitive` in hand, `switch` over the cases.
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+internal extension Primitive {
```
umair-ably, line 56: `LiveMapValue` keeps all six of these same getters public just below (lines 228–243), implemented as `primitiveValue?.stringValue` — and `primitiveValue` (line 204) is public and returns a `Primitive`. So a user who gets a `Primitive` out of `primitiveValue` can no longer read a single case from it, while the type one level up still offers exactly the accessors being withdrawn here.

Is the intent that `LiveMapValue`'s shortcuts stay, or should they be reconsidered in the same change so there's one story? Relatedly: does the new doc comment's advice to `switch` over the cases read right for someone who obtained the `Primitive` from `primitiveValue` rather than from a path object?

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/PathObject.swift
```diff
@@ -185,3 +185,44 @@ public protocol PrimitivePathObject: PathObject {
...
+/// Convenience accessors for a single expected primitive type.
+///
+/// Each of these resolves the path via ``PrimitivePathObject/value()`` and then reads the
+/// correspondingly-named getter on the resulting ``Primitive``, so `nil` means either that nothing
+/// primitive resolved at the path or that what did resolve was a primitive of a different case.
+///
+/// These are methods, not properties, for the same reason as the rest of ``PathObject``'s
+/// path-resolving accessors: each resolves the path at call time and is therefore O(path length).
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+public extension PrimitivePathObject {
```
umair-ably, line 198: These are protocol *extension* members rather than requirements, so they're statically dispatched and a conformer can't specialise them. Fine for pure delegation as written, but `stringValue()` here still materialises a whole `Primitive` — including copying a `[String: JSONValue]` for a `jsonObject` — in order to hand back one case.

Was making them requirements (with these as default implementations) considered, to leave room for a cheaper single-case read later?

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/Instance.swift
```diff
@@ -171,6 +171,55 @@ public protocol PrimitiveInstance: Sendable {
     func compactJson() throws(ARTErrorInfo) -> JSONValue
 }
 
+/// Convenience accessors for a single expected primitive type.
+///
+/// Each of these reads ``PrimitiveInstance/value`` and then the correspondingly-named getter on the
+/// resulting ``Primitive``, so `nil` means that the wrapped primitive is of a different case.
+@available(macOS 11.0, iOS 14.0, tvOS 14.0, *)
+public extension PrimitiveInstance {
```
umair-ably, line 179: Same question as on the `PrimitivePathObject` extension: extension members rather than protocol requirements, so a conformer can't override them. Deliberate, given they're pure delegation to `value`?

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

LiveObjects/Tests/AblyLiveObjectsTests/Unit/DefaultInstanceTests.swift
```diff
@@ -163,6 +163,33 @@ struct DefaultInstanceTests {
...
+        }
+
+        #expect(try primitiveInstance(.string("hi")).stringValue == "hi")
+        #expect(try primitiveInstance(.number(5)).numberValue == 5)
+        #expect(try primitiveInstance(.bool(true)).boolValue == true)
+        #expect(try primitiveInstance(.data(Data([1, 2, 3]))).dataValue == Data([1, 2, 3]))
+        #expect(try primitiveInstance(.jsonArray(["a"])).jsonArrayValue == ["a"])
+        #expect(try primitiveInstance(.jsonObject(["k": "v"])).jsonObjectValue == ["k": "v"])
+
+        // An accessor for a case other than the wrapped one yields nil.
```
umair-ably, line 189: Two coverage questions:

1. One wrong-case read is exercised (`.string` read as `numberValue`). The other five are structurally identical, so this may well be deliberate — intended, or worth a loop over all six?
2. These accessors are the `throws(ARTErrorInfo)` path over `value`. Would one assertion that an accessor propagates the `RTO25` precondition error be worth adding, given `readOnDetachedChannelThrows` right below already has the setup for it?

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

LiveObjects/Tests/AblyLiveObjectsTests/Unit/DefaultPathObjectTests.swift
```diff
@@ -231,6 +231,55 @@ struct DefaultPathObjectTests {
...
+        let root = Self.makePrimitivesRootPathObject()
+
+        #expect(try root.get(key: "string").asPrimitive().stringValue() == "hi")
+        #expect(try root.get(key: "number").asPrimitive().numberValue() == 5)
+        #expect(try root.get(key: "bool").asPrimitive().boolValue() == true)
+        #expect(try root.get(key: "data").asPrimitive().dataValue() == Data([1, 2, 3]))
+        #expect(try root.get(key: "jsonArray").asPrimitive().jsonArrayValue() == ["a"])
+        #expect(try root.get(key: "jsonObject").asPrimitive().jsonObjectValue() == ["k": "v"])
+
+        // An accessor for a case other than the one that resolved yields nil, as does any accessor on
```
umair-ably, line 277: Same as on the instance test: only one of the six wrong-case reads is covered. Deliberate, since they're structurally identical? (The non-resolving-path assertion below it is a good addition.)

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/ValueTypes.swift
maratal: The rule the PR actually establishes is coherent: every public type that holds a primitive exposes both the raw `Primitive` and the six typed shortcuts; the bare `Primitive` enum itself is discriminated by `switch`. `LiveMapValue` already satisfied that rule before this PR — it's `PrimitivePathObject`/`PrimitiveInstance` that were brought into line by it.

But: Lawrence's review doc for [ably-liveobjects-swift-plugin#134](https://github.com/ably/ably-liveobjects-swift-plugin/pull/134) (`review/Review by Claude, 3 Jul 2026 with Lawrence's thoughts.md` on `lawrence-review-path-based-api`) raises it twice:

> §10 — **`LiveMapValue`**: correctly narrowed to a **write-only** union … reads now flow through `PathObject`/`Instance`/`Primitive` instead. Minor: the read-oriented convenience getters (`stringValue`, `numberValue`, …) copied from `main` are vestigial on a write-only type and could be dropped.

> §12 — **`Primitive`/`LiveMapValue` ergonomics**: with `LiveMapValue` now write-only, drop the vestigial read getters (§10) and keep the literal conformances, which are the valuable part.

So `LiveMapValue`'s getters were there from the pre-path-based API, when it was a read/write structure. So the existence of these helpers for read structures like `PrimitivePathObject`/`PrimitiveInstance` is well justified. I'm going to remove those from `LiveMapValue`, because it's a write-only structure.

LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/PathObject.swift
maratal: Considered, and I'd keep them as extension members.

The cost concern doesn't quite land: there's no copy. `Primitive`'s payloads are all COW types or scalars (largest case is 16 bytes, so the enum isn't boxed either), and `DefaultPrimitivePathObject.value()` builds it by re-wrapping case-for-case from `InternalLiveMapValue`, which already holds the identical payload types — `case let .jsonObject(value): return .jsonObject(value)` retains the same buffer rather than copying elements. So `stringValue()` on a path holding a big JSON object costs one retain/release of a dictionary it then discards.

And a specialised implementation would still pay everything that actually costs: `throwIfInvalidAccessApiConfiguration`, a cross-queue `internalQueue.ably_syncNoDeadlock` to read the root, and the O(path length) walk with a `get(key:)` per segment. You can't know whether the value is a string without resolving the path, so the cheaper single-case read has almost nothing to save. On `PrimitiveInstance` (`Instance.swift:179`, same construction) it's clearer still — the instance is bound to an already-resolved value, so `value` is O(1) and there's no walk to skip.

On dispatch: agreed on the mechanics, but this is the line the file already draws — `PathObject.subscribe(listener:)`, `PathObject.events(options:)` and `LiveCounterPathObject.increment()`/`decrement()` are all non-overridable extension conveniences layered over a requirement. Making these six requirements would widen the public contract and let a conformer implement `stringValue()` so it disagrees with `value()` — the equivalence the doc comment above the extension promises, currently guaranteed by construction.

It's also free to reverse: moving these bodies to a default implementation later changes no call syntax and forces no conformer to implement anything, and the package is source-distributed via SPM so there's no library-evolution ABI angle. Happy to do it if a cheaper read ever materialises — nothing here forecloses it.

LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/Instance.swift
maratal: see [above](https://github.com/ably/ably-cocoa/pull/2260#discussion_r3909514066)

LiveObjects/Tests/AblyLiveObjectsTests/Unit/DefaultInstanceTests.swift
maratal: **2.** Added in 31fecb08, on both sides — one line each in the existing `readOnDetachedChannelThrows`, which as you say already has the fixture.

It turned out to be worth more than coverage hygiene. The regression it guards is silent: since SE-0230 flattens nested optionals, `try? value.stringValue` has type `String?` rather than `String??`, so tidying the `try` into a `try?` compiles cleanly in the `String?`-returning throwing getter and turns the RTO25 error into `nil` — indistinguishable to a caller from the wrapped primitive being a different case. I confirmed it: with that one-character change the suite was green before this assertion and fails with it. Nothing else covered the path — `primitiveConvenienceAccessors` builds its mock `.attached`, so the guard never fires there, and `readOnDetachedChannelThrows` only exercised `value` / `exists()` / `type()`.

One accessor per side rather than six: the other five are identical delegation, and what is pinned here is the propagation, not the cases.

**1.** Deliberate — I'd leave it at the one. The five extra wrong-case reads would restate the same `value()?.<case>` delegation against a `Primitive` whose per-case getters are already covered; there's no distinct failure mode for a loop to catch. The asymmetry with (2) is the point: that one closed a real silent-failure path, this one wouldn't.

LiveObjects/Tests/AblyLiveObjectsTests/Unit/DefaultPathObjectTests.swift
maratal: see [above](https://github.com/ably/ably-cocoa/pull/2260#discussion_r3909576079)

maratal:
> **Three shapes for one identifier**: `stringValue()` on the path object, `stringValue` on the instance, `stringValue` on `LiveMapValue`. The O(path length) vs O(1) rationale makes sense, but a user's autocomplete now shows the same name in three shapes. Is the docs PR ([ably/website#9704](https://github.com/ably/website/pull/9704)) calling that distinction out explicitly?

@umair-ably WDYT personally about these helper getters? Does `asPrimitive().value()?.stringValue` really needs `asPrimitive().stringValue()` treatment?

LiveObjects/Sources/AblyLiveObjects/Path Based API/Public/ValueTypes.swift
maratal: Done in 449d8ed7 — `LiveMapValue`'s six typed shortcuts are gone, `primitiveValue`/`liveMapValue`/`liveCounterValue` and the `ExpressibleBy*Literal` conformances stay.

One call site, `Test/UTS/unit/objects/ValueTypesTests.swift`, which came out better for it: through the literal conformances it now compares the whole value, so it reads exactly like the UTS line above it —

```swift
// ASSERT vt.entries["name"] == "Alice"
#expect(vt.entries?["name"] == "Alice")
```

— and rejects an absent key or a non-primitive entry, both of which the case getter silently let through as `nil`.

umair-ably:
> > **Three shapes for one identifier**: `stringValue()` on the path object, `stringValue` on the instance, `stringValue` on `LiveMapValue`. The O(path length) vs O(1) rationale makes sense, but a user's autocomplete now shows the same name in three shapes. Is the docs PR ([ably/website#9704](https://github.com/ably/website/pull/9704)) calling that distinction out explicitly?
> 
> @umair-ably WDYT personally about these helper getters? Does `asPrimitive().value()?.stringValue` really needs `asPrimitive().stringValue()` treatment?

I don't have a strong opinion either way but generally lean towards some level of consistency. If we deviate from consistency then there should at least be some reasoning behind it so we've actively made the decision to draw the line somewhere. What's your thinking here?

maratal:
> consistency

You mean `.stringValue()` in path object vs `.stringValue` in instance? @umair-ably 

If so, then we deliberately already have that as `value()` in PO and `value` in instance to highlight that `value()` has some cost, so I thought that continue that separation is consistent.

umair-ably (approved):
happy with the changes in the getters and the consistency answer

PR #2261, opened 2026-08-28
liveobjects: preserve exact Double values for numbers read over the JSON protocol

maratal:
Jira: [AIT-1339](https://ably.atlassian.net/browse/AIT-1339)

Split out of #2257, which was carrying this alongside an unrelated change.

## What happens today

`NumberPrecisionTests` (new, integration, against the sandbox) writes a `Double` into the graph and reads it back bit-for-bit — once for a map value, once for a counter (`counterCreate.count` then `counterInc.number`), each parametrized over three values and split per protocol.

Against the current reads it fails on every JSON case and passes on every MessagePack one:

```text
✘ mapNumberSurvivesJSONRoundTrip          (actual → 0.015034388851090227)  == (value → 0.015034388851090229)
✘ mapNumberSurvivesJSONRoundTrip          (actual → 0.023942638935083194)  == (value → 0.023942638935083197)
✘ mapNumberSurvivesJSONRoundTrip          (actual → 0.05005333533771126)   == (value → 0.050053335337711245)
✘ counterValueSurvivesJSONRoundTrip       (countAfterCreate → 0.015034388851090227) == (value → 0.015034388851090229)
✘ counterValueSurvivesJSONRoundTrip       (countAfterIncrement → 0.030068777702180454) == (value + value → 0.030068777702180458)
   (9 failures in total: 3 map, 6 counter)
✔ mapNumberSurvivesMessagePackRoundTrip       3 cases
✔ counterValueSurvivesMessagePackRoundTrip    3 cases
```

The counter case is the one that matters most: the amount is accumulated, so the divergence persists.

## Why

`JSONSerialization` does not represent every JSON number as a `Double`. Fixed-point text with enough significant digits comes back as an `NSDecimalNumber`, whose conversion to binary floating point is not correctly rounded:

```text
text          0.015034388851090229
.stringValue  0.015034388851090229   exact
.doubleValue  0.015034388851090227   one ULP low
Double(text)  0.015034388851090229   correctly rounded
```

The parser is fine — every digit survives. The loss is in the `NSNumber` → `Double` conversion, and every bridge Foundation offers shares it: `doubleValue`, `Double(truncating:)`, `Double(exactly:)`, `as? Double`. `Double(exactly:)` never returns `nil` for these, so it reports success while handing back a rounded value.

The trigger is textual shape, not magnitude — `1.2345678901234567e-5` is fine, `0.000012345678901234567` is not — and ably-js writes fixed-point notation for everything down to `1e-6`. So a JS client calling `counter.increment(0.004910368508982831)` leaves the Swift client's counter one ULP from what every other SDK reports. MessagePack is unaffected: a float64 decodes to a plain `NSNumber` already holding the closest `Double`.

## The fix

`NSNumber.art_doubleValue` reads the digits back as text for the `NSDecimalNumber` case and lets `Double(_: String)` convert. The type test is load-bearing, not defensive: an ordinary `NSNumber`'s `stringValue` rounds to 16 significant digits, which would turn `greatestFiniteMagnitude` into infinity.

Applied at the nine sites that read wire-derived numbers, each carrying a one-line note on why it needs the exact read. Two reads stay on `doubleValue`: the MessagePack branch of `ObjectData.toWire`, the only outbound conversion, where the number is always one the SDK built from a Swift `Double`; and the `objectsGCGracePeriod` reads, a duration rather than object data.

`NSNumberExactDoubleValueTests` (unit) covers the property and the `JSONValue` bridge directly, including the ordinary values that guard the type test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numeric precision when processing counters, maps, object data, JSON values, and wire data.
  * Prevented rounding errors for high-precision decimal values received or converted by the SDK.
  * Preserved numeric values accurately across JSON and MessagePack round trips.

* **Tests**
  * Added coverage for high-precision decimal handling and standard numeric values.
  * Added integration coverage for precise map and counter values.

[AIT-1339]: https://ably.atlassian.net/browse/AIT-1339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/LiveObjects/Sources/AblyLiveObjects/Internal/InternalDefaultLiveCounter.swift
+++ b/LiveObjects/Sources/AblyLiveObjects/Internal/InternalDefaultLiveCounter.swift
@@ -332,7 +332,8 @@ internal final class InternalDefaultLiveCounter: Sendable {
             liveObjectMutableState.createOperationIsMerged = false
 
             // RTLC6c: Set data to the value of ObjectState.counter.count, or to 0 if it does not exist
-            data = state.counter?.count?.doubleValue ?? 0
+            // art_doubleValue: this count came off the wire, where JSON decoding can give us an NSDecimalNumber
+            data = state.counter?.count?.art_doubleValue ?? 0
 
             // RTLC6d: If ObjectState.createOp is present, merge the initial value into the LiveCounter as described in RTLC16
             // Discard the LiveCounterUpdate object returned by the merge operation
@@ -353,7 +354,9 @@ internal final class InternalDefaultLiveCounter: Sendable {
             let counterCreate = operation.counterCreate ?? operation.counterCreateWithObjectId?.derivedFrom
 
             // RTLC16a: Add counterCreate.count to data, if it exists
-            if let operationCount = counterCreate?.count?.doubleValue {
+            // art_doubleValue: for a remotely-created counter this count came off the wire (a locally-created
+            // one comes from `derivedFrom`, where it is already a Double, and is unaffected)
+            if let operationCount = counterCreate?.count?.art_doubleValue {
                 data += operationCount
                 // RTLC16c
                 update = .update(DefaultLiveCounterUpdate(amount: operationCount))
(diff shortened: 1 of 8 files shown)


sacOO7 (approved):
LGTM

PR #2264, opened 2026-09-06
[Split 1/N] Restructure: vend the core SDK as the AblyPubSubCore product

maratal:
Part of [DX-1726](https://ably.atlassian.net/browse/DX-1726)

First PR of the [PDR-091b](https://ably.atlassian.net/wiki/spaces/product/pages/5362810886) core/device split stack, targeting the long-lived `integration/v2` branch. Reference implementations: [ably-js#2293](https://github.com/ably/ably-js/pull/2293) (`packages/core` move) and [ably-java#1232](https://github.com/ably/ably-java/pull/1232) (module restructure).

## What this PR does

* **`Ably` → `AblyPubSubCore`**: the SPM product that vends the core SDK is renamed, and documented in the manifest as an internal implementation product for Ably's own packages — it declares no side, so clients built from it are classified by Ably's default rules (PDR-091b §1).
* **Manifest `name` → `ably-pubsub-cocoa`**.
* **Version → 2.0.0** via `make bump_major` (its own commit, as `Scripts/set-version.sh` produces).
* Docs: `README.md` (2.0 note + product references), `CONTRIBUTING.md` (Distribution table, release steps), `CLAUDE.md` (a "Version 2.0" section for agents working on this branch), `Examples/SPM`. `CHANGELOG.md` is deliberately untouched: the 2.0.0 entry is written in one pass at the end of the stack, from the generated release notes, as `CONTRIBUTING.md`'s release process describes.
* Separately on `integration/v2`: every workflow's `push:` trigger now includes `integration/**`, so this branch gets branch-level CI and not just PR CI.

## Verification

`swift build -Xswiftc -warnings-as-errors`, `swift build --build-tests`, `swift test --filter ClientInformationTests` (3 passed), `swift build`/`swift test --package-path Examples/SPM`, and `xcodebuild build-for-testing -workspace Ably.xcworkspace -scheme ably-cocoa -destination platform=macOS` → **TEST BUILD SUCCEEDED**, which compiles `AblyTests`, `AblyTestsObjC` and `UTS` through the exact scheme and test plan the Fastlane lanes use. The full sandbox test run is left to CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Updated the Swift Package Manager product name to `AblyPubSubCore` for the 2.0 release.
  * Existing module and class naming remain unchanged for compatibility.
* **Documentation**
  * Updated installation, contribution, and integration guidance for the new 2.0 product names.
[...]

PR #2265, opened 2026-09-07
[Split 2/N] Add the AblyPubSubDevice product and its factory

maratal:

Second PR in the [PDR-091b](https://ably.atlassian.net/wiki/spaces/product/pages/5362810886) split stack, stacked on #2264 (base: `split/restructure-core`, so the diff shows only this PR's changes). Reference implementations: [ably-js#2293](https://github.com/ably/ably-js/pull/2293) `packages/device`, and [ably-java#1233](https://github.com/ably/ably-java/pull/1233), whose side-stamping contract this ports.

## What this PR does

* **New `AblyPubSubDevice` SPM product** (`PubSubDevice/`, Objective-C, depending on the `Ably` target). `ARTPubSubDevice` is `NS_SWIFT_NAME(PubSubDevice)` and exposes the door agreed in [PDR-091b2](https://ably.atlassian.net/wiki/spaces/product/pages/5348425729):
  * `createClient(options:)` — the door named in the rollout plan;
  * `createClient(key:)` / `createClient(token:)`, mirroring the core's own initialisers.

  All three return an `ARTRealtime`, so the side is the package and factory choice plus the agent declaration, not a type distinction — as in ably-go.

* **The stamping contract**, ported from `packages/shared/side.ts`:
  * the identifier is `ably-pubsub-device`, **versionless** — `ARTClientInformationAgentNotVersioned` is Objective-C's spelling of the `null` value ably-java's `Side.java` puts in its agents map, since an `NSDictionary` cannot hold `nil`. Registered as `"versioned": false` in [ably-common#361](https://github.com/ably/ably-common/pull/361);
  * the entry goes onto `[options copy]`, so the caller's options object **and** their `agents` dictionary are both left untouched;
  * the caller's `agents` entries are preserved, so a layered SDK keeps its attribution, and the device entry is applied **last**, so it wins a collision on its own name;
  * the name's `-device` suffix is load-bearing and carries the warning comment adapted from `side.ts`.

  There is no `Side` abstraction: ably-js and ably-java need one because they ship two identifiers from shared code, and Cocoa ships one.

* **Tests** — 10 Swift (`Test/AblyTests/Tests/PubSubDeviceTests.swift`) and 2 Objective-C (`Test/AblyTestsObjC/ARTPubSubDeviceTests.m`).

## Tests: what billing reads, asserted on the wire

`ARTClientInformation` decides between a bare token and `name/version` by **pointer** comparison against the sentinel. If that identity were ever lost the wire would silently carry `ably-pubsub-device/ARTClientInformationAgentNotVersioned` — a misclassification invisible from the options dictionary. So rather than trust it, the tests assert the actual wire output in both directions: the `agent` query param on the realtime connection (via `TestProxyTransport`) and the `Ably-Agent` header on an HTTP request (via `TestProxyHTTPExecutor`), each checking the token is present **and** carries no `/`.

Also covered: caller options not mutated, caller agents preserved, collision resolution, the key/token doors, unrelated options carried over, and that a client built straight from the core declares nothing.

## Notes for reviewers

* **The tests live in `AblyTests` rather than a new test target.** `TestProxyTransport`, `TestProxyHTTPExecutor` and `commonAppSetup` are all in that target, not in the reusable `AblyTesting` one; a separate target would have meant migrating them. This also keeps `Test/Ably.xctestplan` unchanged, so the new tests run in every existing lane on all three platforms.
* **`ARTClientOptions.copy` preserves `testOptions` and `plugins`** as well as `agents`, so a door-built client works with the test proxy transport and with the LiveObjects plugin unchanged. That is what makes the wire tests possible.
* **`import AblyPubSubDevice` alone reaches the core's types** in both Swift and Objective-C — the module re-export works, verified by `Examples/SPM` and the ObjC test. Only `ARTRealtime.internal` needs `Ably.Private`.
* **`Ably.xcodeproj` is untouched**, so the product is SPM-only, consistent with `AblyLiveObjects`. Which channels carry the 2.0 products is still open; if ably-flutter needs the door through CocoaPods that adds a podspec and hand-maintained framework targets.

## Verification

`swift build -Xswiftc -warnings-as-errors`, `swift build --build-tests`, the 12 new tests green against sandbox, `swift test --package-path Examples/SPM`, editorconfig-checker v4 clean, and `xcodebuild build-for-testing -workspace Ably.xcworkspace -scheme ably-cocoa -destination platform=macOS` → **TEST BUILD SUCCEEDED**, which compiles `AblyTests`, `AblyTestsObjC` and `UTS` through the exact scheme and test plan the Fastlane lanes use.

## Summary by CodeRabbit

* **New Features**
  * Added the `AblyPubSubDevice` Swift Package product for creating clients identified as running on end-user devices.
  * Added support for creating device clients with options, API keys, or tokens.
  * Added an SPM integration example demonstrating device client creation.

* **Documentation**
  * Updated product and contributor documentation to describe `AblyPubSubDevice` and its distribution details.

* **Tests**
  * Added coverage for device identification, authentication methods, option preservation, and network requests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,8 @@ Changes made to dependencies in the [Cartfile](Cartfile) need to be reflected in
 
 ## Adding new Objective-C files to the SDK
 
+The steps below are for the core SDK — the `Ably` target, whose sources live in `Source/`. The `AblyPubSubDevice` target (`PubSubDevice/`) is laid out differently: its public headers go in `PubSubDevice/include/AblyPubSubDevice/` and its implementations directly in `PubSubDevice/`, there is no umbrella header to add an `#import` to, and it is absent from `Ably.xcodeproj`. SwiftPM globs the directory and generates the module map, so a new file there needs no other change.
+
 ### Public header (`.h`) files
 
 These are the header files that form the public interface of the SDK.
@@ -127,19 +129,19 @@ If you add another `xcodebuild`-based path that compiles a test target linking `
 
 ### Distribution
 
-`AblyLiveObjects` is available **via Swift Package Manager only**. CocoaPods and Carthage consumers receive the core SDK alone, so a release tag does not deliver the same set of products to every channel:
+`AblyPubSubDevice` and `AblyLiveObjects` are available **via Swift Package Manager only**. CocoaPods and Carthage consumers receive the core SDK alone, so a release tag does not deliver the same set of products to every channel:
 
-| Channel | `AblyPubSubCore` | `AblyLiveObjects` |
-| --- | --- | --- |
-| Swift Package Manager | yes | yes |
-| CocoaPods | yes (as the `Ably` pod) | no |
-| Carthage | yes | no |
+| Channel | `AblyPubSubCore` | `AblyPubSubDevice` | `AblyLiveObjects` |
+| --- | --- | --- | --- |
+| Swift Package Manager | yes | yes | yes |
+| CocoaPods | yes (as the `Ably` pod) | no | no |
+| Carthage | yes | no | no |
 
 > The `AblyPubSubCore` product's module is named `Ably` and its classes carry the `ART` prefix, so
 > consumers reach the core as `import Ably` / `#import <Ably/…>` whichever channel they install it
 > from — which is why the CocoaPods pod and the Carthage framework are both named `Ably`.
 
-This is a deliberate decision rather than an omission. Four separate things would each have to change to lift it:
+The two are SPM-only for different reasons. Nothing about `AblyPubSubDevice` resists the other channels: it is a plain Objective-C target over the core, with the same platform floor and no plugin machinery, so shipping it as a pod would mean adding a podspec and Carthage targets rather than removing an obstacle. `AblyLiveObjects` is the constrained one, and deliberately so rather than by omission — four separate things would each have to change to lift it:
 
 - [`Ably.podspec`](Ably.podspec)'s `source_files` covers `Source/` only, so neither `LiveObjects/` nor `_AblyPluginSupportPrivate/` ships in the pod; and `Ably.xcodeproj`, which Carthage builds, contains no LiveObjects or plugin-support targets.
 - `ABLY_SUPPORTS_PLUGINS` is defined only in `Package.swift`. Without it the plugin hook points — `ARTClientOptions.plugins` and the plumbing in `ARTRealtimeChannel.m`, `ARTRealtime.m` and `ARTJsonLikeEncoder.m` — are compiled out. It is a compile-time define on the core target, so enabling it would enable it for every CocoaPods and Carthage consumer, in a configuration that has never been built or tested.
@@ -209,4 +211,4 @@ For each release, the following needs to be done:
 * Checkout `main` locally, pulling in changes using `git checkout main && git pull`. Make sure the new tag you need was created on publish
 * Release an update for CocoaPods using `pod trunk push Ably.podspec --allow-warnings`. Details on this command, as well as instructions for adding other contributors as maintainers, are at [Getting setup with Trunk](https://guides.cocoapods.org/making/getting-setup-with-trunk.html) in the [CocoaPods Guides](https://guides.cocoapods.org/). This publishes the core SDK only — there is no LiveObjects pod
 * Test the integration of the library in a Xcode project using Carthage and CocoaPods using the [installation guide](https://github.com/ably/ably-cocoa#installation-guide)
-* Test that Swift Package Manager resolves the new tag, selecting both the `AblyPubSubCore` and `AblyLiveObjects` products
+* Test that Swift Package Manager resolves the new tag, selecting the `AblyPubSubCore`, `AblyPubSubDevice` and `AblyLiveObjects` products
--- a/Examples/SPM/Package.swift
+++ b/Examples/SPM/Package.swift
@@ -19,7 +19,8 @@ let package = Package(
             dependencies: [
                 // `package:` is the dependency's identity, which for a path
                 // dependency is the directory name, not the manifest's `name`.
-                .product(name: "AblyPubSubCore", package: "ably-cocoa")
+                .product(name: "AblyPubSubCore", package: "ably-cocoa"),
+                .product(name: "AblyPubSubDevice", package: "ably-cocoa")
             ],
             swiftSettings: [
                 .unsafeFlags(["-warnings-as-errors"])
--- a/Package.swift
+++ b/Package.swift
@@ -18,6 +18,12 @@ let package = Package(
             name: "AblyPubSubCore",
             targets: ["Ably"]
         ),
+        // The entry point for apps: creates clients that declare they are
+        // running on an end user's device.
+        .library(
+            name: "AblyPubSubDevice",
+            targets: ["AblyPubSubDevice"]
+        ),
         .library(
             name: "AblyLiveObjects",
             targets: ["AblyLiveObjects"]
@@ -84,6 +90,16 @@ let package = Package(
             name: "_AblyPluginSupportPrivate",
             path: "_AblyPluginSupportPrivate"
         ),
+        // Creates clients that declare they are running on an end user's
+        // device, by stamping the agent entry Ably classifies on.
+        .target(
+            name: "AblyPubSubDevice",
+            dependencies: [
+                .target(name: "Ably"),
+            ],
+            path: "PubSubDevice",
+            publicHeadersPath: "include"
+        ),
         // The core SDK. Vended as the AblyPubSubCore product; its module is
         // named `Ably`.
         .target(
@@ -120,6 +136,7 @@ let package = Package(
             name: "AblyTests",
             dependencies: [
                 .byName(name: "Ably"),
+                .target(name: "AblyPubSubDevice"),
                 .byName(name: "AblyTesting"),
                 .byName(name: "AblyTestingObjC"),
                 .product(name: "Nimble", package: "nimble"),
@@ -167,6 +184,7 @@ let package = Package(
             name: "AblyTestsObjC",
             dependencies: [
                 .byName(name: "Ably"),
+                .target(name: "AblyPubSubDevice"),
                 .byName(name: "AblyTesting"),
                 .byName(name: "AblyTestingObjC"),
             ],
--- a/PubSubDevice/include/AblyPubSubDevice/ARTPubSubDevice.h
+++ b/PubSubDevice/include/AblyPubSubDevice/ARTPubSubDevice.h
@@ -1,0 +1,58 @@
+#import <Ably/Ably.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Creates Pub/Sub clients for apps running on an end user's device.
+ *
+ * A client created here tells Ably that it is running on an end user's device. That declaration
+ * determines how the connection is counted: on an account billed by monthly active users, device
+ * traffic counts toward the total, the connection must carry a `clientId`, and that `clientId` is
+ * subject to a concurrency limit.
+ *
+ * The returned object is an `ARTRealtime` and behaves exactly as one built with
+ * `-[ARTRealtime initWithOptions:]`, so all of the Ably Pub/Sub documentation applies to it
+ * unchanged.
+ */
+NS_SWIFT_NAME(PubSubDevice)
+@interface ARTPubSubDevice : NSObject
+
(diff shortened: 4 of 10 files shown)


Test/AblyTests/Tests/PubSubDeviceTests.swift
```diff
@@ -0,0 +1,177 @@
...
+                    guard let transport = client.internal.transport as? TestProxyTransport,
+                          let query = transport.lastUrl?.query else {
+                        XCTFail("MockTransport isn't working")
+                        done()
+                        return
+                    }
+                    expect(query).to(haveParam("agent", hasPrefix: "ably-cocoa/"))
+                    XCTAssertTrue(query.contains(deviceAgentName))
+                    XCTAssertFalse(query.contains("\(deviceAgentName)%2F"))
+                    done()
```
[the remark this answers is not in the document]
maratal: "this test could pass even though a version is present" — is true of line 133 in isolation and false of the suite. A version on the device agent would fail lines 28 and 159 immediately.

Test/AblyTests/Tests/PubSubDeviceTests.swift
```diff
@@ -0,0 +1,177 @@
...
+    func test__010__the_core_constructor__declares_nothing() throws {
+        let test = Test()
+        let options = try AblyTests.commonAppSetup(for: test)
+        options.autoConnect = false
+
+        let client = ARTRealtime(options: options)
+        defer { client.dispose(); client.close() }
+
+        XCTAssertNil(client.internal.options.agents?[deviceAgentName])
+    }
```
sacOO7, line 176: **[MAJOR] LiveObjects `plugins`-survive-copy test is missing**
**Finding:** There is no test proving that `ARTClientOptions.plugins` (the LiveObjects plugin hook) survives the `[options copy]` inside the factory. This file (tests 001–010) and `Test/AblyTestsObjC/ARTPubSubDeviceTests.m` never mention `LiveObjects`, `Plugin`, `plugins`, or `.object`. The PR description states that plugins survive the copy, but nothing in the test suite verifies it.
**Why it matters:** `PubSubDevice.createClient` is meant to become the entry point for applications — including every LiveObjects user. The factory works by copying the caller's options; if a future change to `ARTClientOptions` copy semantics silently dropped the `plugins` dictionary, `channel.object` would break for every LiveObjects app constructing through this factory, with no test to catch it.
**Suggested action:** Add a test that constructs options with `AblyLiveObjects.Plugin`, passes them through `PubSubDevice.createClient(options:)`, and asserts `channel.object` works (or, minimally, that `plugins` survives into the client's internal options). The likely mechanical blocker is that the `AblyTests` target doesn't depend on `AblyLiveObjects` — add that dependency. If the test is being deliberately deferred, please say so explicitly in the PR thread so the gap is a recorded decision rather than an accident.

PubSubDevice/include/AblyPubSubDevice/ARTPubSubDevice.h
```diff
@@ -0,0 +1,58 @@
+#import <Ably/Ably.h>
```
sacOO7, line 1: **[MINOR] Header layout: no `AblyPubSubDevice.h` umbrella, SPM-generated module map — please confirm intentional**
**Finding:** The target ships a single public header, `ARTPubSubDevice.h`, which carries the `#import <Ably/Ably.h>` re-export itself, and `publicHeadersPath: "include"` lets SwiftPM auto-generate the umbrella header and module map. The layout agreed for this step of the PubSub device/server split rollout (PDR-091b2) had a dedicated `AblyPubSubDevice.h` umbrella header plus a hand-written module map instead.
**Why it matters:** Nothing is broken — Swift `import AblyPubSubDevice` and ObjC `@import AblyPubSubDevice;` both reach the core's types, and this is arguably simpler than the agreed layout (a hand-written module map would have been redundant with SPM's generation). But it is a silent divergence from a written rollout step, and unacknowledged divergences make the later steps of the rollout harder to audit.
**Suggested action:** No code change requested — just an explicit acknowledgement here that the simpler layout is intentional, so it stands as a recorded decision.

Examples/SPM/Package.swift
```diff
@@ -19,7 +19,8 @@ let package = Package(
             dependencies: [
                 // `package:` is the dependency's identity, which for a path
                 // dependency is the directory name, not the manifest's `name`.
-                .product(name: "AblyPubSubCore", package: "ably-cocoa")
+                .product(name: "AblyPubSubCore", package: "ably-cocoa"),
+                .product(name: "AblyPubSubDevice", package: "ably-cocoa")
```
sacOO7, line 23: **[MINOR] Example target isn't device-only**
**Finding:** This adds `AblyPubSubDevice` alongside the existing `AblyPubSubCore` dependency on the same `SPMIntegration` target. `main.swift` imports only `AblyPubSubDevice` and uses the factory, so the new product is exercised at source level — but the example does not prove that depending on `AblyPubSubDevice` alone is sufficient, because the core product is still linked into the same target.
**Why it matters:** A key promise of the new product is that `import AblyPubSubDevice` re-exports the core's types, so an app needs exactly one dependency. With `AblyPubSubCore` still on the target, a regression that broke that re-export (e.g. a header or module-map change) would slip past this example unnoticed — the core types would still resolve through the direct dependency.
**Suggested action:** Add a device-only example target, or drop the now-redundant `AblyPubSubCore` product dependency from `SPMIntegration`.

Package.swift
```diff
@@ -120,6 +136,7 @@ let package = Package(
             name: "AblyTests",
             dependencies: [
                 .byName(name: "Ably"),
+                .target(name: "AblyPubSubDevice"),
```
sacOO7, line 139: **[MINOR] No separate `AblyPubSubDeviceTests` target — please confirm intentional**
**Finding:** The new product's tests are folded into the existing `AblyTests` (Swift) and `AblyTestsObjC` (ObjC) targets, as here, rather than getting a dedicated `AblyPubSubDeviceTests` target and xctestplan entry as sketched for this step of the PubSub device/server split rollout (PDR-091b2).
**Why it matters:** This reads as a justified divergence rather than a defect: the helpers these tests need (`TestProxyTransport`, `TestProxyHTTPExecutor`, `commonAppSetup`) live in `AblyTests` rather than the shared `AblyTesting` target, and reusing the existing targets keeps the new tests running in every existing CI lane on all three platforms.
**Suggested action:** No code change requested — just a conscious sign-off here so the divergence stands as a recorded decision.

sacOO7 (commented):
**Lead-developer review — [Split 2/N] AblyPubSubDevice product + factory.** Faithful, well-scoped implementation of this step of the PubSub device/server split rollout (PDR-091b2): the factory surface matches the agreed design exactly (`PubSubDevice.createClient(options:)` plus key/token conveniences), all six side-stamping semantics are implemented and tested (options copied, fresh mutable `agents` dict, versionless `ably-pubsub-device` applied last, caller's object untouched, conveniences stamp too, `nil` passes through to the core's own error), both wire-level directions are covered (WS `agent` query param and HTTP `Ably-Agent` header), and boundaries are clean (`Ably.xcodeproj`, `Ably.podspec`, `Cartfile`, xctestplan, `CHANGELOG.md` and `Test/UTS/` all untouched; the core SDK's own agent identifier is correctly left unchanged for a later PR). One Major (a missing test) and four Minor findings are posted inline. The docstring-coverage CI warning (33%) is non-blocking — the public header is fully documented.

sacOO7:
**[MINOR] `CONTRIBUTING.md:191` no-op-release bullet omits `AblyPubSubDevice`**
**Finding:** The release bullet stating that a release whose only change is LiveObjects is a no-op for CocoaPods/Carthage consumers should now also name `AblyPubSubDevice` — the Distribution table this same PR edits already establishes it as SPM-only. (`CONTRIBUTING.md:191` falls outside the diff hunks, hence a bottom comment rather than inline.)
**Why it matters:** As written, the bullet implies LiveObjects is the only SPM-only product; the pod/Carthage no-op reasoning now applies to `AblyPubSubDevice` too, so a reader could misjudge what a device-only release delivers to each channel. (Echoes the outstanding CodeRabbit nit.)
**Suggested action:** Append `AblyPubSubDevice` to that bullet.
_Ref: split plan §4_

CONTRIBUTING.md
```diff
@@ -127,19 +129,19 @@ If you add another `xcodebuild`-based path that compiles a test target linking `
 
 ### Distribution
 
-`AblyLiveObjects` is available **via Swift Package Manager only**. CocoaPods and Carthage consumers receive the core SDK alone, so a release tag does not deliver the same set of products to every channel:
+`AblyPubSubDevice` and `AblyLiveObjects` are available **via Swift Package Manager only**. CocoaPods and Carthage consumers receive the core SDK alone, so a release tag does not deliver the same set of products to every channel:
```
sacOO7, line 132: **[MINOR] Release-notes bullet later in this file omits `AblyPubSubDevice`**
**Finding:** This section now correctly establishes `AblyPubSubDevice` as SPM-only, but the Versioning section's release-notes bullet (line 191, just outside this diff) still reads "A release whose only change is to LiveObjects is a no-op for them" — naming only LiveObjects when explaining what CocoaPods/Carthage consumers don't receive.
**Why it matters:** With this PR, the pod/Carthage no-op reasoning applies equally to `AblyPubSubDevice`: a release whose only change is to the device product delivers nothing to those channels either. As written, the bullet could lead a release author to write inaccurate release notes for such a release. (This echoes an outstanding CodeRabbit nit.)
**Suggested action:** Extend the line-191 bullet to also name `AblyPubSubDevice`, e.g. "A release whose only change is to `AblyPubSubDevice` or LiveObjects is a no-op for them".

ttypic (approved):

Test/AblyTests/Tests/PubSubDeviceTests.swift
maratal: Both halves are now written, in two PRs.

**#2291, on top of this PR** — the copy-semantics guard, `test__011__createClient__carries_over_the_plugins` in this file. It sits with the five other contract tests for the same call and uses their idiom, `client.internal.options`.

The manifest change you expected was not needed. Rather than adding `AblyLiveObjects` to `AblyTests`, the test reuses `PluginAPITests.MockLiveObjectsPlugin`, which already exists in this target and already conforms to the real `LiveObjectsPluginProtocol` — so the core SDK's test target stays clear of the plugin, and the test needs no network, running in every existing lane on all three platforms.

**#2292, at the top of the stack** — the behavioural claim, that `channel.object` works on a factory-built client. It is there rather than here because `AblyLiveObjectsTests` already depends on `AblyPubSubDevice` at that point; at this commit that dependency is still `Ably`, and #2285 renames the same line when it merges the core into the device target, so adding it here would duplicate on rebase without producing a conflict marker.

Both were checked against the regression they describe.

PubSubDevice/include/AblyPubSubDevice/ARTPubSubDevice.h
maratal: Overtaken rather than acknowledged: #2285 merges the core into the device target and lands on the layout you were comparing against — `Source/include/AblyPubSubDevice/AblyPubSubDevice.h` as the umbrella, plus a hand-written `Source/include/module.modulemap` that also declares the `Private` submodule. So the divergence closes on its own further up the stack.

Examples/SPM/Package.swift
maratal: Also resolved by #2285, which removes the core product entirely: `Examples/SPM/Package.swift` now depends on `AblyPubSubDevice` alone, so the target does prove that one dependency suffices. The example's direct-construction check went too, since those initializers are internal in #2287 — what remains exercises `PubSubDevice.createClient` and names `RealtimeClient` as the return type.

Package.swift
maratal: Intentional — your own reasoning is the reason. The helpers these tests need (`TestProxyTransport`, `TestProxyHTTPExecutor`, `commonAppSetup`) live in `AblyTests` rather than the shared `AblyTesting` target, and reusing the existing targets puts the new tests in every existing lane on all three platforms rather than requiring a new xctestplan entry.

It paid off again in #2291: the plugins test reuses `PluginAPITests`' mock plugin, which is only reachable because the tests share a target.

CONTRIBUTING.md
maratal: Stale now — that bullet is gone. #2282 drops CocoaPods and Carthage from 2.x, so `CONTRIBUTING.md` no longer reasons about what those channels do or do not receive; the Distribution section says they stay on the 1.x line. Nothing left to extend.

PR #2267, opened 2026-09-07
[Split 3/N] Send ably-pubsub-cocoa as the SDK agent identifier

maratal:

> [!IMPORTANT]
> **Blocked on [ably/ably-common#361](https://github.com/ably/ably-common/pull/361).** Draft until that lands. The convention agreed on [PDR-091b2](https://ably.atlassian.net/wiki/spaces/product/pages/5348425729) is registry-first: `ably-pubsub-cocoa` must exist in `protocol/agents.json` before this SDK ships it. Checked against `ably-common` `main` at the time of writing — neither `ably-pubsub-cocoa` nor the versionless `ably-pubsub-device` flag is registered yet. [ably-js#2297](https://github.com/ably/ably-js/pull/2297) is held behind the same gate.
>
> Also outstanding once #361 lands: bump the `Test/AblyTests/ably-common` submodule pin (currently at a September 2025 commit). Nothing in this PR depends on it — only JS and Go files in that submodule read `agents.json`, so no Swift test validates identifiers against the registry — so it is housekeeping rather than a prerequisite for these tests to pass.

Third PR in the [PDR-091b](https://ably.atlassian.net/wiki/spaces/product/pages/5362810886) split stack, stacked on #2265 (base: `split/device-door`). Port of [ably-js#2297](https://github.com/ably/ably-js/pull/2297) commit 2.

## What this PR does

Renames the SDK's agent identifier from `ably-cocoa` to `ably-pubsub-cocoa`, in the new major only. Target wire shape:

```
ably-pubsub-cocoa/2.0.0 ably-pubsub-device macOS/26.2.0
```

* **One production line** — `_libraryName` in `ARTClientInformation.m` — plus the `ARTClientInformation.agents` doc comment, which told users to read the version from the `"ably-cocoa"` key.
* **The eight existing assertions** on `ably-cocoa/<version>` across `ClientInformationTests`, `RealtimeClientConnectionTests`, `RestClientTests` and `WrapperSDKProxyTests`, and the one in `PubSubDeviceTests`.
* **A new whole-identifier test**, built by sorting as the emitter does, asserting the SDK entry carries the version and the device entry sits beside it as a bare flag. `test__010` now also asserts that a client built straight from the core produces an identifier containing no flag at all.
* **`Scripts/set-version.sh` gains `PubSubDeviceTests.swift`**, which now pins `2.0.0` inside an expected agent string and would otherwise rot silently at the next `make bump_*`.

## Why the identifier moves

Because the flip happens exactly at the split and the 1.x maintenance line is never touched, the identifier alone partitions the fleet: `ably-cocoa/*` is legacy-package traffic, `ably-pubsub-cocoa/*` is new-package traffic. Migration tracking and the eventual EOL enforcement become a string match rather than a version-range heuristic. The identifier names the *family*, not any one published product, exactly as `ably-cocoa` always did.

## What this PR deliberately does not do

* **Does not touch the 1.x maintenance line.** Released versions keep sending `ably-cocoa` for the rest of their supported life; that is what makes the partition clean.
* **No repo rename**, and no change to the SPM package identity — the `package:` label in a `.product(name:package:)` reference is still `ably-cocoa`, since it follows the directory and URL rather than this identifier or `Package.swift`'s `name`.

## Verification

All 11 `PubSubDeviceTests`, `ClientInformationTests`, `ARTDefaultTests`, `WrapperSDKProxyTests` (17), and both renamed wire assertions in `RestClientTests` and `RealtimeClientConnectionTests` — green against sandbox. Plus `swift build -Xswiftc -warnings-as-errors`, editorconfig-checker v4 clean, and `xcodebuild build-for-testing -workspace Ably.xcworkspace -scheme ably-cocoa -destination platform=macOS` → **TEST BUILD SUCCEEDED**.

## Summary by CodeRabbit

* **Changes**
  * Updated the reported client library identifier to `ably-pubsub-cocoa`.
  * Connection URLs and REST headers now identify the client with the updated library name.
  * Device Pub/Sub clients now report the expected agent information, including platform details.

* **Documentation**
  * Clarified device-side and server-side SDK terminology.

* **Tests**
  * Updated agent identifier expectations and added coverage for device client reporting.


Source/ARTClientInformation.m
```diff
@@ -7,7 +7,7 @@
 
 NSString *const ARTClientInformationAgentNotVersioned = @"ARTClientInformationAgentNotVersioned";
 NSString *const ARTClientInformation_libraryVersion = @"2.0.0";
-static NSString *const _libraryName = @"ably-cocoa";
+static NSString *const _libraryName = @"ably-pubsub-cocoa";
```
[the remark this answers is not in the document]
maratal: After `ably-common` pr 361 lands

sacOO7, line 10: **[MAJOR — MERGE BLOCKER] Do not merge until the identifier is registered in ably-common and the submodule pin is bumped**

**Finding:** This line flips the emitted SDK agent identifier to `ably-pubsub-cocoa`, but the `Test/AblyTests/ably-common` submodule pin is still `783496f4` — identical to `main` and to this PR's base, i.e. unbumped. The registry the tests build against therefore does not yet contain `ably-pubsub-cocoa` or `ably-pubsub-device`. Per the PubSub device/server split rollout (PDR-091b2), agent identifiers must be registered in ably-common *before* they ship, with the registry entry's `source` pointing at `ably-pubsub-cocoa`. So this PR must not merge until (a) ably/ably-common#361 lands, and (b) the submodule pin here is bumped past that commit. Critically, no Swift test validates against `protocol/agents.json`, so CI will NOT turn red on its own — this gate is purely procedural.

**Why it matters:** The agent identifier is how this client declares itself to Ably, which drives identifier-registry integrity and how traffic is attributed for billing/MAU. Emitting an identifier the registry doesn't yet recognise risks misattributed or unrecognised traffic. This PR was a draft explicitly "Blocked on ably/ably-common#361" when reviewed — but it is now marked ready for review while that gate is still open, and because CI stays green regardless, the only remaining guard against a premature merge is human vigilance. That materially raises the risk here.

**Suggested action:** Re-draft the PR, or apply an explicit do-not-merge label, until ably/ably-common#361 lands and the `Test/AblyTests/ably-common` pin is bumped past it. Note the gate transitively blocks #2275 and the rest of the split stack from merging to `integration/v2`, and any `2.0.0-beta` tag inherits it too.

sacOO7 (commented):
**Lead review — verdict:** A surgical, correct implementation of the identifier flip: `_libraryName` moves to `ably-pubsub-cocoa` with zero stale agent emission left in `Source/`, all seven identifier assertions updated, both wire-shape directions tested (`test__011`/`test__010`), `Scripts/set-version.sh` updated, and the maintenance/1.x line untouched. One hard merge blocker and one informational note are posted as inline comments.

sacOO7:
**[INFO] CHANGELOG identifier note deliberately omitted — fine under the one-pass changelog policy**

**Finding:** The per-PR CHANGELOG note for the identifier change was intentionally left out, consistent with the superseded one-pass changelog policy (no 2.0.0 section exists yet). That is the right call for this PR. Relatedly, `ARTDefaultTests` correctly needed no change — its only assertion is version-based with no `ably-cocoa` identifier literal, so there was nothing to flip.

**Why it matters:** The identifier change is user-visible via `ARTClientInformation.agents` (`ably-cocoa/1.x` → `ably-pubsub-cocoa/2.x ably-pubsub-device`), so it must not fall through the cracks when the changelog is eventually written.

**Suggested action:** Ensure the eventual single-pass 2.0.0 CHANGELOG section records the identifier change so consumers who key off the agent string aren't surprised.

Source/include/Ably/ARTClientInformation.h
```diff
@@ -15,7 +15,7 @@ extern NSString *const ARTClientInformationAgentNotVersioned;
 - (instancetype)init NS_UNAVAILABLE;
 
 /**
- * Returns the default key-value entries that the Ably client library uses to identify itself, and the environment in which it's running, to the Ably service. Its keys are the names of the software components, and its values are their optional versions. The full list of keys that this method might return can be found [here](https://github.com/ably/ably-common/tree/main/protocol#agents). For example, users of the `ably-cocoa` client library can find out the library version by fetching the value for the `"ably-cocoa"` key from the return value of this method.
+ * Returns the default key-value entries that the Ably client library uses to identify itself, and the environment in which it's running, to the Ably service. Its keys are the names of the software components, and its values are their optional versions. The full list of keys that this method might return can be found [here](https://github.com/ably/ably-common/tree/main/protocol#agents). For example, users of this client library can find out the library version by fetching the value for the `"ably-pubsub-cocoa"` key from the return value of this method.
```
sacOO7, line 18: **[INFO] User-visible identifier change — make sure the eventual 2.0.0 CHANGELOG section records it**

**Finding:** This doc comment confirms the change is user-visible: consumers who read `ARTClientInformation.agents` (or key off the agent string on the wire) will see the SDK entry change. The PR deliberately ships no CHANGELOG entry, which is the right call under the current policy of writing the 2.0.0 changelog in a single pass at the end (no 2.0.0 section exists yet). Relatedly, `ARTDefaultTests` correctly needed no change here — its only assertion is version-based, with no `ably-cocoa` identifier literal to flip.

**Why it matters:** Without a per-PR entry, the only record of this user-visible behaviour change is the PR itself — easy to lose by the time the changelog is written.

**Suggested action:** When the single-pass 2.0.0 CHANGELOG section is written, it must record that the agent identifier users see changes from `ably-cocoa/1.x` to `ably-pubsub-cocoa/2.x` plus the versionless `ably-pubsub-device` flag (for clients created via `PubSubDevice.createClient`), visible in `ARTClientInformation.agents` and in the `Ably-Agent` header / `agent` query param.

ttypic (approved):

Source/ARTClientInformation.m
maratal: Agreed, and confirmed: ably/ably-common#361 is still open, and the pin is `783496f4` — identical to `main`, so unbumped. Nothing in `Test`, `Source` or `Scripts` reads `agents.json`, so your "CI will not turn red" is right too.

One thing to add: the stale pin is inherited by all ten PRs above this one — it is still `783496f4` at the tip of #2287 — so this gates the `integration/v2` merge, not just this PR.

Source/include/Ably/ARTClientInformation.h
maratal: Recorded, and the list has grown since you wrote this. The same single-pass 2.0.0 section also needs the `ARTRest`→`ARTHttp` and `ARTRealtime`→`ARTRealtimeClient` renames, the deprecated-member deletions, the four `ARTPush` `rest:` overloads, the HTTP client family and the construction API going internal, and the `ART` prefix dropping from every Swift name (#2287).

So this entry is one line of a much larger one. Your aside about `ARTDefaultTests` is right — its assertion is version-based, with no identifier literal to flip.

PR #2270, opened 2026-09-08
CI: run the UTS in its own workflow

maratal:
The Universal Test Suite ran as part of **Integration Test**, alongside `AblyTests` and `AblyTestsObjC`, because all three are targets of `Test/Ably.xctestplan` and one Fastlane lane per platform ran the whole plan. Two problems with that:

* **It misnames the suite.** The UTS is primarily a *unit* suite — the `objects` unit tier alone is ~145 tests, plus the realtime and rest unit suites — with the sandbox-network and (macOS-only) proxy tiers a minority of it. Filing all of that under "Integration Test" tells you the wrong thing about what failed.
* **The two were welded together.** A red `check (iOS, test_iOS26_2)` could be either suite, and re-running one meant re-running both — around 40 minutes per platform leg.

## What this PR does

Gives the UTS its own workflow, so the checks read as what they are:

| Check | Workflow | Targets | Selection |
| --- | --- | --- | --- |
| `UTS (iOS)` / `UTS (tvOS)` / `UTS (macOS)` | **UTS** (`uts.yaml`) | `UTS` | `-only-testing:UTS` |
| `check (iOS, test_iOS26_2)` … | Integration Test | `AblyTests`, `AblyTestsObjC` | `-skip-testing:UTS` |

Between them they cover the test plan exactly once, so total work is unchanged — it just parallelises differently and reports separately. The UTS can now also be dispatched or re-run on its own.

* **`fastlane/Fastfile`** — `LaneConfig#args_for_run_tests` takes a `suite:` option: `sdk` skips the UTS target, `uts` runs only it, `nil`/`all` runs the whole plan. An unrecognised value raises rather than silently running everything. Lane names are unchanged, and a lane invoked without `suite:` behaves exactly as before, so a local `bundle exec fastlane test_macOS` is unaffected.
* **`.github/workflows/uts.yaml`** — reuses `integration-test.yaml`'s setup verbatim (same Xcode pin, `xcparse` checkout and cache, simulator reset, artifact uploads, test-observability upload); only the workflow name, job name, `suite:` value and observability job name differ.
* **`.github/workflows/integration-test.yaml`** — passes `suite:sdk`; job and check names untouched (see below).
* Docs and cross-references that the split falsified: `CONTRIBUTING.md`'s "only place CI executes the `UTS` target's tests" claim, the pointers in `liveobjects.yaml` and `check-spm.yaml`, and the proxy-tier CI note in `Test/UTS/README.md`.

## One deliberate asymmetry, needing a decision

The UTS jobs are named `UTS (<platform>)`, but **Integration Test's job keeps its existing `check (<platform>, <lane>)` name** rather than becoming `Integration Test (<platform>)`.

Renaming a job renames its check. `main` is protected, and if any of those check names are configured as required status checks, renaming them would stop those checks reporting and block PRs until branch protection was updated. I can't read the protection settings without admin. If someone confirms the required checks (or that there are none), renaming is a one-line follow-up.

## Verification

Fastlane is behind a version manager on my machine, so rather than run the lanes I verified the part this PR changes — the argument construction — by evaluating `LaneConfig` directly:

```
suite="sdk"  only_testing=nil      skip_testing=["UTS"]
suite="uts"  only_testing=["UTS"]  skip_testing=nil
suite=nil    only_testing=nil      skip_testing=nil
suite="all"  only_testing=nil      skip_testing=nil
suite="typo" -> raises: Unknown suite 'typo' — expected 'sdk', 'uts' or 'all'.
```

Plus `ruby -c fastlane/Fastfile` clean, editorconfig clean, and a check that the two workflows' artifact names stay parallel (no conflict, since artifacts scope per workflow run). The real proof is this PR's own CI: the `UTS (…)` and `check (…)` legs should both appear, and between them run every test that ran before.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated CI workflow for running Universal Test Suite tests across iOS, tvOS, and macOS.
  * Added options to run SDK tests, UTS tests, or the full test suite independently.

* **Improvements**
  * Updated integration testing to focus on SDK targets while UTS tests run separately.
  * Updated contributor and UTS documentation to reflect the revised test workflow structure.
  * Added test reports, logs, and analyzer results as CI artifacts.


maratal:
Closing unmerged: this work is now the first commit of #2268 rather than a separate PR against `main`.

PR #2271, opened 2026-09-08
PROBE (do not merge): does UTS_SIDE reach the test process on simulators?

maratal:
Throwaway, to be closed as soon as it has answered its question.

Sets `Test/UTSDevice.xctestplan`'s configuration value to `bogus` and restricts the run to `SideSeamTests`.

`Side.swift` aborts on an unrecognised `UTS_SIDE`, so the two outcomes are unambiguous:

* **device legs crash** with `Fatal error: UTS_SIDE must be 'core' or 'device', got 'bogus'` — the plan delivers the variable on that platform.
* **device legs pass** — the variable never arrived, the leg silently ran as `core`, and the device coverage on that platform is illusory.

The `core` legs are the control and should pass either way.

This is the only way to tell, because a device leg that receives nothing runs as `core` and passes exactly like one that works.

### DIFF
--- a/.github/workflows/uts.yaml
+++ b/.github/workflows/uts.yaml
@@ -117,7 +117,7 @@ jobs:
           bundle install
 
       - name: Run Tests
-        run: bundle exec fastlane ${{ matrix.lane }} suite:uts test_plan:${{ matrix.test_plan }}
+        run: bundle exec fastlane ${{ matrix.lane }} suite:uts test_plan:${{ matrix.test_plan }} only_testing:UTS/SideSeamTests
 
       - name: Check Static Analyzer Output
         id: analyzer-output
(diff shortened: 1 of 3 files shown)


maratal:
Answered: the plan **does** deliver `UTS_SIDE` on simulators.

| Leg | Result |
| --- | --- |
| `iOS … device, UTSDevice` | failed with `Fatal error: UTS_SIDE must be 'core' or 'device', got 'bogus'` |
| `tvOS … device, UTSDevice` | same |
| `macOS … device, UTSDevice` | cancelled by fail-fast; already proven locally |
| all three `core, Ably` | passed, as the control |

So the device legs on #2268 are running genuine device-mode coverage on all three platforms.

This probe was the only way to establish that: a device leg that receives nothing runs as `core` and passes exactly like one that works, so a green leg proves nothing. That is the same trap that made `TEST_RUNNER_` appear to work.

Closing; nothing here is meant to merge.

PR #2274, opened 2026-09-09
CI: make the configured test retries actually absorb a flake

maratal:
Split out of #2268. Stacked on #2273 (base: `fix/uts-issue-recording`), which is what makes flakes retryable in the first place.

Preparation for https://github.com/ably/ably-cocoa/pull/2275

## Why

The Fastlane lanes have set `number_of_retries: 2` for a long time, and the comment beside it claims the retry "absorbs single-occurrence environmental timeouts on the CI runners". That was not true. Two separate reasons, one per commit.

**1. scan judged the lane on every attempt.** `output_remove_retry_attempts` defaults to `false`, so a test that failed once and passed on retry still failed CI. Observed on a UTS leg: three suites failed, retried, and passed —

```
::notice ::Test run with 752 tests passed after 578.122 seconds
Test Succeeded

| Number of tests    | 415 (and 830 retries) |
| Number of failures | 3                     |
[!] Tests have failed
```

xcodebuild said `Test Succeeded`; scan failed the lane on the superseded attempts. So the retry was doing nothing it was configured for.

**2. fastlane 2.227.2 does not strip a failure when the retry *skipped* the test** rather than passing it, so one phantom failure survived even with the option on (3 → 1). That is fastlane [#30081](https://github.com/fastlane/fastlane/pull/30081), fixed in **2.237.0**.

## What

- `output_remove_retry_attempts: true` in the shared lane arguments, with the comment corrected.
- `Gemfile.lock` moved to fastlane 2.237.0 — the earliest release carrying the fix. Deliberately not 2.239.0, which also moves faraday to 2.x; 2.237.0 leaves it on 1.10.6. `Gemfile` is unchanged, so the dependency stays unpinned.

This does not weaken the signal: a genuine regression fails every attempt and still fails the lane. The option only discards attempts a later retry superseded.

It applies to every lane, so `Integration Test` benefits as much as the UTS legs — which is the point, since the flaw was never UTS-specific.

## Verification

The argument mapping was checked by evaluating `LaneConfig` directly (`number_of_retries=2  output_remove_retry_attempts=true`), plus `ruby -c` on the Fastfile. `bundle lock` re-resolves to the same lockfile against the unpinned `Gemfile`, so the pin is not accidental.

One gap worth flagging: `bundle install` fails on my machine for unrelated reasons — `bigdecimal`, `json` and `nkf` cannot build native extensions into a `.sitearchdir` — so I have no local proof the bundle installs. CI's `Install Dependencies` step on a clean runner is the first real test of the lockfile, and it fails early and unmistakably if the resolution is bad.

## Summary by CodeRabbit

* **Tests**
  * Test runs now report results based on the final outcome after retries, reducing noise from superseded attempts.


maratal:
@coderabbitai review

maratal:
@coderabbitai review

ttypic (approved):

PR #2275, opened 2026-09-09
[Split 4/N] Split the UTS into its own CI workflow, and run it through the device package's factory

maratal:

> Replaces #2268, which GitHub would not let me retarget — its base was locked with `Cannot change the base branch because the pull request is part of a stack`, so its diff kept including the two PRs below. Same branch and same commits; only the base differs.

Fourth PR in the split stack, stacked on #2274 (base: `ci/retry-absorbs-flakes`), which in turn sits on #2273 and #2267. Two commits.

## 1. `CI: run the UTS in its own workflow`

The Universal Test Suite ran as part of **Integration Test**, alongside `AblyTests` and `AblyTestsObjC`, because all three are targets of `Test/Ably.xctestplan` and one Fastlane lane per platform ran the whole plan.

That misnames the suite. The UTS is primarily a *unit* suite: the `objects` unit tier alone is ~145 fully-mocked tests, plus the realtime and rest unit suites, with the sandbox-network and macOS-only proxy tiers a minority of it.

It also welded the two together. A red `check (iOS, test_iOS26_2)` could be either suite, and re-running one re-ran both, at ~40 minutes per platform leg.

| Check | Workflow | Targets | Selection |
| --- | --- | --- | --- |
| `UTS / check (iOS, test_iOS26_2)` … | **UTS** (`uts.yaml`) | `UTS` | `-only-testing:UTS` |
| `Integration Test / check (iOS, test_iOS26_2)` … | Integration Test | `AblyTests`, `AblyTestsObjC` | `-skip-testing:UTS` |

Between them they cover the plan exactly once, so total work is unchanged; it parallelises differently and reports separately, and the UTS can now be dispatched or re-run on its own.

* **`fastlane/Fastfile`** — `args_for_run_tests` takes a `suite:` option: `uts` sets `only_testing: ['UTS']`, `sdk` sets `skip_testing: ['UTS']`, `nil`/`all` sets neither. `sdk` is defined by exclusion rather than by naming the two targets, so a fourth test target added to the plan later lands in that leg instead of being silently dropped. An unrecognised value raises. Lane names are unchanged, and a lane invoked **without** `suite:` still runs the whole plan — which is what `make test_[iOS|tvOS|macOS]` does, so local runs are unaffected and the `Makefile` needed no edit.
* **`.github/workflows/uts.yaml`** — reuses Integration Test's setup with the same `check` job name, so the two read as siblings and the workflow name does the distinguishing, as it already does for the other `check` and `build` jobs here.
* **No test-observability upload, and so no `xcparse` checkout or cache** — those steps exist only to feed `upload_test_results.sh`, which needs `xcparse` to extract crash reports. That upload earns its place for the SDK's own long-lived suite, where the value is analysing which tests fail intermittently over time. This suite is new, like the LiveObjects one, which does not upload either. Note this is a deliberate behaviour change: the UTS was uploading before the split.
* Docs and cross-references the split falsified: `CONTRIBUTING.md`'s "only place CI executes the `UTS` target's tests" claim, the pointers in `liveobjects.yaml` and `check-spm.yaml`, and the proxy-tier CI note in `Test/UTS/README.md`.

Deliberately **not** here: splitting the UTS further into `Unit` and `Integration` legs. Worth doing, but a change of its own.

## 2. `Run the UTS through the device package's factory`

Port of [ably-js#2294](https://github.com/ably/ably-js/pull/2294). Makes the suite runnable through either entry point, selected by `UTS_SIDE`:

| Mode | Realtime construction | REST construction |
|---|---|---|
| `core` (default, or unset) | `ARTRealtime(options:)` | `ARTRest(options:)` |
| `device` | `PubSubDevice.createClient(options:)` | `ARTRest(options:)` |

The specs are identical either way. The factory has tests of its own in #2265, but they check only that it stamps the declaring agent and leaves the caller's options untouched; they barely use the client it returns. Running the specs through it is what puts that client to work — connection lifecycle, channels, presence, history, objects, auth — and so shows the factory to be a faithful pass-through and not merely a correct stamp.

* **`Test/UTS/infra/Side.swift`** — `UTSSide` plus `makeRealtimeForSide(options:)`.
* **One seam.** The UTS builds realtime clients in exactly two places — `UTSTestCase.makeRealtime` and `IntegrationTestCase.withRealtimeClient` — and both route through the factory. No spec file changes.
* **REST is deliberately not routed through it.** The device package exposes no HTTP door, so `ARTRest` is the only way to build a stateless client in either mode. Same as ably-js's device mode.
* **`SideSeamTests`**, and `UTS_SIDE` documented in `Test/UTS/README.md` alongside the existing `RUN_DEVIATIONS` and `UTS_PROXY_LOCAL_PATH` overrides.

### Guarding against a silently degraded run

If the seam stopped routing, every spec would still pass — the run would just be a second copy of the core run, reported as coverage. Two things prevent that:

* An unrecognised `UTS_SIDE` calls `fatalError` rather than falling back to `core`, so a mistyped leg aborts instead of quietly passing.
* `SideSeamTests` asserts the declaration is present in `device` and absent in `core`. It builds its client through `makeRealtimeForSide` — the function every suite uses — rather than calling the factory directly, so it checks the *wiring*, not just the factory. Verified by mutation: pointing the `device` branch at the core constructor fails it with *"device mode built a client without the device declaration — the suite is silently repeating the core run"*.

### The device-mode CI leg

`uts.yaml` runs the suite twice per platform, via a `side` dimension:

| Leg | Test plan | Entry point |
| --- | --- | --- |
| `core` | `Ably` (the default) | `ARTRealtime(options:)` |
| `device` | `Test/UTSDevice.xctestplan` | `PubSubDevice.createClient(options:)` |

`UTSDevice` holds the `UTS` target and a configuration setting `UTS_SIDE=device`; the lanes gain a `test_plan:` option that becomes scan's `testplan:`. Plans are declared in `.swiftpm/xcode/xcshareddata/xcschemes/ably-cocoa.xcscheme`, which is checked in.

A test plan is the only way to get an environment variable into an SPM test bundle under `xcodebuild`, which took some finding — the obvious routes are dead ends, and they fail in ways that look like success:

* **A plain exported `UTS_SIDE`** does not reach the test process; xcodebuild isolates its environment.
* **`TEST_RUNNER_UTS_SIDE`** as a build setting does not either. That run *passed*, which proves nothing, because `core` passes too — re-running with the value `bogus`, which should have hit the `fatalError`, is what exposed it.
* **`-only-testing:` cannot name a directory.** It takes a test identifier, so `-only-testing:UTS/unit` matches nothing **and exits green**.

Under `swift test` no plan is involved, because the test binary inherits the environment.

**The plan route is confirmed to work on simulators, not just macOS.** A throwaway probe (#2271) set the plan's value to `bogus`: the iOS and tvOS `device` legs aborted with `Fatal error: UTS_SIDE must be 'core' or 'device', got 'bogus'` while all three `core` controls passed. That check was necessary because a device leg receiving nothing runs as `core` and passes exactly like one that works.

## Scope: only the UTS goes through the door

`AblyTests` and `AblyTestsObjC` stay on the core constructors. They exist to test the core SDK, and since the door returns the same `ARTRealtime`, routing their constructions through it would add churn without a distinct claim. The per-side claim for those suites is carried by `PubSubDeviceTests` in #2265, which asserts the agent on the wire in both directions.

That matches the reference implementation: ably-js#2294 changes only `test/uts/**` plus its own workflow and `test:uts*` scripts. Its existing mocha suites construct through a separate `shared_helper` and are untouched; their per-side assertion is `test/unit/pubsub_side_agent.test.js`, from ably-js#2293.

## Verification

Local, both modes: the realtime and REST unit suites (11 tests × 2), the objects unit tier (145 tests × 2), and `SideSeamTests` — all green, plus the mutation check. `swift build -Xswiftc -warnings-as-errors` and `swift build --build-tests` clean, editorconfig clean, `ruby -c fastlane/Fastfile` clean, and the `suite:`/`test_plan:` mapping checked by evaluating `LaneConfig` directly.

A CodeRabbit finding on #2268 — the `actions/checkout` pin declaring `using: node12` — is addressed in the workflow commit itself, so no commit here carries the old pin. Its severity did not hold up: checkout does not fail, because the runner forces such actions onto a supported Node and only warns.

## Summary by CodeRabbit

- **New Features**
  - Added automated Universal Test Suite coverage across iOS, tvOS, and macOS, including core and device configurations.
  - Added device-side test execution using the PubSub device client.
  - Added selectable test suites and test plans for targeted CI runs.

- **Documentation**
  - Updated contributor and UTS testing guidance to explain the split between SDK and UTS workflows, supported platforms, and device-side testing.

- **Tests**
  - Added validation for test-side selection, fallback behavior, and device test configuration.


maratal:
@coderabbitai review

Test/UTS/README.md
```diff
@@ -583,11 +583,28 @@ swift test --filter UTS.AuthReauthTests
...
+and not merely a correct stamp. Every realtime client in
+every tier is built by `makeRealtimeForSide` in [`infra/Side.swift`](infra/Side.swift), so there is
+one seam to keep honest; `SideSeamTests` asserts the routing directly, because a broken seam would
+otherwise leave the suite quietly repeating the core run and reporting it as coverage. REST clients
+are deliberately not routed through it — the device package exposes no HTTP door, so `ARTRest` is
+the only way to build a stateless client in either mode. An unrecognised value aborts the run
+rather than falling back to `core`.
+
 **Where CI runs them:** there is currently **no UTS-specific CI job** — the UTS target runs as part
 of the full test suite (the `ably-cocoa` scheme driven by the fastlane lanes in
```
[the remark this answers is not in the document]
maratal: good catch, will be fixed in upcoming md files update PR

CONTRIBUTING.md
```diff
@@ -84,8 +84,15 @@ Both `AblyLiveObjectsTests` and the `objects` UTS suites also depend on a shared
...
+
+  Test plans are declared in `.swiftpm/xcode/xcshareddata/xcschemes/ably-cocoa.xcscheme`. A plan is also how an environment variable reaches the test process under `xcodebuild`, where neither a plain export nor a `TEST_RUNNER_`-prefixed build setting does. Under `swift test` no plan is needed, because the test binary inherits the environment.
+
+  The two are separate workflows because the UTS is primarily a *unit* suite, so grouping it under "Integration Test" misnamed it; splitting them also means a UTS failure is distinguishable from an `AblyTests` one without opening a log, and either can be dispatched or re-run without the other's sandbox time. Between them they cover the test plan exactly once. A lane invoked without `suite:` still runs the whole plan, which is what a local `bundle exec fastlane test_macOS` does.
```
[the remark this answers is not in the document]
maratal: good catch, will be fixed in upcoming md files update PR

sacOO7 (commented):
**Lead-developer review — SDK device/server split, [Split 4/N].**

Verdict: a tightly scoped, faithful implementation of the UTS-per-side step of the split. There is a single construction seam in `Test/UTS/infra/Side.swift` with both call sites (`UTSTestCase`, `IntegrationTestCase`) routed through it, the harness self-tests both modes (`SideSeamTests`, plus a `fatalError` on an unknown `UTS_SIDE`), `integration-test.yaml` now passes `suite:sdk` so the UTS never runs twice, and the boundaries are clean — including the correctly deferred tier-leg work.

Worth calling out positively: the device-mode CI mechanism is better than originally anticipated. Delivering `UTS_SIDE=device` via a checked-in `Test/UTSDevice.xctestplan` plus the checked-in `ably-cocoa` scheme runs all three platforms across both modes, rather than a macOS-only `swift test` job or a compile-time flag.

One sequencing note: this PR sits above #2267 (the `ably-pubsub-cocoa` agent-identifier change) in the stack, so it inherits that PR's merge gate — it should not merge to `integration/v2` (nor be cut into any prerelease) until ably/ably-common#361 lands, registering the new identifiers, and the `Test/AblyTests/ably-common` submodule pin is bumped past it.

Individual findings are posted as inline comments (two on `uts.yaml`, one on `SideSeamTests.swift`, one on `IntegrationTestCase.swift`).

.github/workflows/uts.yaml
```diff
@@ -0,0 +1,154 @@
...
+    # export nor a TEST_RUNNER_-prefixed build setting reaches an SPM test
+    # bundle. Under `swift test` no plan is involved, because the test binary
+    # inherits the environment — see Test/UTS/README.md.
+    #
+    # Results are not uploaded to test-observability, and so there is no
+    # xcparse checkout here either (the upload script needs it to extract crash
+    # reports). That upload exists for the SDK's own suite, whose value is in
+    # analysing which tests fail intermittently over time. This suite is new,
+    # like the LiveObjects one, which does not upload either.
+    strategy:
```
sacOO7, line 58: **[INFO] Identical pass counts across modes is structural, not auto-compared**
**Finding:** The core and device legs run as independent matrix entries against the same UTS target, so the expectation that both modes report identical results is achieved structurally, and guarded by `SideSeamTests` plus the `fatalError` on an unknown `UTS_SIDE` — but there is no automated cross-leg comparison of pass counts. **Why it matters:** "Same results either way" (as the workflow comment above says) is only ever confirmed by eye; a leg that silently skipped tests would still be green. This matches how ably-js handled the same step, so it is a note, not a request. **Suggested action:** On the first green run, eyeball both legs' counts to confirm they line up; no code change needed.

.github/workflows/uts.yaml
```diff
@@ -0,0 +1,154 @@
...
+      - name: Check Static Analyzer Output
+        id: analyzer-output
+        run: |
+          if [[ -z $(find ./derived_data -name "report-*.html") ]]; then
+            echo "Static Analyzer found no issues."
+          else
+            echo "Static Analyzer found some issues. HTML report will be available in Artifacts section. Failing build."
+            exit 1
+          fi
+
```
sacOO7, line 134: **[INFO] Trailing whitespace on lines 134 and 141**
**Finding:** Both lines carry trailing whitespace. **Why it matters:** Not a lint gate — `.editorconfig` scopes trimming to `*.{swift,m,h}`, so CI won't catch it — but it's avoidable noise, and CodeRabbit flagged it too. **Suggested action:** Strip it if the file is touched again.

sacOO7:
**[MINOR] Record the §6 item 4 spec-check outcome**
**Finding:** `Test/UTS/deviations.md` is untouched and there is no note in the PR that the ably/specification check for split-related spec points (agent-side flags, factory doors) was carried out. **Why it matters:** Plan §6 item 4 explicitly asks for that check and a recorded result; without an artifact it is unclear whether it happened. This may well be N/A — the per-side claim lives in the unit tests (`PubSubDeviceTests`/`SideSeamTests`), mirroring the ably-js approach. **Suggested action:** Confirm whether the spec gained any split-related points; either record "no new spec points — N/A" in this thread, or derive them and update `deviations.md`.
_Ref: split plan §6 item 4_

sacOO7:
**[MINOR] Stale doc-comment example in `ProxySession.swift`**
**Finding:** The usage example in the doc comment at `Test/UTS/infra/integration/proxy/ProxySession.swift:149` still shows `let client = ARTRealtime(options: options)` rather than the seam. **Why it matters:** The runtime proxy path is correctly routed through the seam via `IntegrationTestCase` inheritance, so only the illustrative comment is stale — but it is a location plan §6 item 1 explicitly named. **Suggested action:** Update the comment to build the client through `makeRealtimeForSide(options:)` when the file is next touched. Cosmetic.
_Ref: split plan §6 item 1_

Test/UTS/infra/SideSeamTests.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+import Testing
+import Ably
+import Ably.Private
+
+/// Self-tests for the harness itself, not derived from any UTS spec.
```
sacOO7, line 6: **[MINOR] Spec-check outcome not recorded (`Test/UTS/deviations.md` untouched)**
**Finding:** The split work included a step to check the ably/specification repo for any new spec points covering agent side flags or factory doors, and to record the outcome in `Test/UTS/deviations.md`. No record of that check is visible on this PR: `deviations.md` is untouched and the description doesn't mention it. This line is why it matters here — these self-tests are explicitly "not derived from any UTS spec", so if the specification has gained points covering per-side behaviour, this is where they would need to be derived. It may well be N/A: keeping the per-side claim in unit tests mirrors how ably-js handled the same step. **Why it matters:** Without a recorded outcome, later readers can't distinguish an intentional "nothing to derive" from an overlooked check. **Suggested action:** Confirm whether ably/specification gained any split-related spec points; either note "no new spec points — N/A" in this thread, or derive them and update `Test/UTS/deviations.md`.

Test/UTS/infra/integration/IntegrationTestCase.swift
```diff
@@ -33,7 +33,7 @@ class IntegrationTestCase {
     /// Builds a real (unmocked) `ARTRealtime` from `options`, runs `body`, then always closes the
     /// client and waits for CLOSED.
     func withRealtimeClient(_ options: ARTClientOptions, _ body: (ARTRealtime) async throws -> Void) async throws {
-        let client = ARTRealtime(options: options)
+        let client = makeRealtimeForSide(options: options)
```
sacOO7, line 36: **[MINOR] Stale doc-comment example in `ProxySession.swift`**
**Finding:** This change correctly routes the proxy tier's runtime path through the seam — proxy-tier clients are built via this method through `IntegrationTestCase` inheritance. But the usage example in the doc comment at `Test/UTS/infra/integration/proxy/ProxySession.swift:149` still shows the old constructor:
```swift
/// let client = ARTRealtime(options: options)
```
**Why it matters:** Behaviour is correct; only the illustrative comment is stale. Still, anyone copying that example writes a client that bypasses the seam and always runs core-side regardless of `UTS_SIDE`. **Suggested action:** Update the example to build the client through `makeRealtimeForSide(options:)` next time the file is touched. Cosmetic.

ttypic (approved):

Test/UTS/README.md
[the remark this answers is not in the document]
maratal: Fixed in #2290 (`Test/UTS/README.md`). The section now describes the workflow rather than the state before it existed: the lanes pass `suite:uts`, which fastlane turns into `-only-testing:UTS`; six legs run iOS, tvOS and macOS twice each, once per `UTS_SIDE` value; and `integration-test.yaml` passes `suite:sdk`, which skips this target.

Worth noting how this got worse in the meantime — the filename in the paragraph was updated to `uts.yaml` while the claim around it was left, so the text ended up naming the dedicated workflow as evidence that no dedicated workflow existed.

CONTRIBUTING.md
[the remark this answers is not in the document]
maratal: Fixed in #2290 (`CONTRIBUTING.md`). "Between them they cover the test plan exactly once" is now "Between them they cover every target in the test plan, with `UTS` running twice per platform, once per `UTS_SIDE` value."

The ownership half of your point was already right in the two bullets above that paragraph, which name the targets each workflow covers; only the execution count was wrong.

.github/workflows/uts.yaml
maratal: Eyeballed, as suggested, on the green run for the current head (`5d3de1e2`, run 34503393241). The counts line up per platform:

| Platform | core | device |
|---|---|---|
| iOS | 752 | 752 |
| tvOS | 376 | 376 |
| macOS | 382 | 382 |

So the structural expectation holds on this run. Your underlying point stands — nothing compares them, so a leg that silently stopped selecting tests would still be green — and it is recorded here rather than acted on, in line with how ably-js left the same step.

.github/workflows/uts.yaml
maratal: Fixed in #2290. Both lines were six spaces; `uts.yaml` now has no trailing whitespace anywhere in it.

Test/UTS/infra/SideSeamTests.swift
maratal: Checked, and the answer is that there is nothing to derive, so `Test/UTS/deviations.md` needs no entry.

I pulled `ably/specification` to `d9a04cac` (2026-09-17), 91 commits on from the checkout I had. Searching the whole repository's Markdown for `pubsub-device`, `pubsub-server`, `ably-pubsub`, "device package", "server package", "side flag", `UTS_SIDE` and `createRealtimeClient` returns nothing, and neither does "device-side" or "server-side SDK" in `features.md`.

Two checks that go further than keyword matching. The agent clauses were not touched: `git diff` over `features.md` for that range has no added or removed line mentioning `RSC7a`-`RSC7e`, so `RSC7d`'s `Agent` rules are as this SDK already implements them. And the 44 clause IDs that appeared or changed in the range are all elsewhere — `DF1a`, `RTL3d`, `RTL4j`, `RTL7i`, `RTL10b`, `RTL15*`, `RTL33*`, `RTN8c/d`, `RTN9c/d`, `RTN14h`, `RTN15g*`, `RTN16g*`, `RTN23b/c*`, `RTN27d`, `RTP11b/e/f*`, `TO3r*`. The one new client option, `TO3r` `strictMode`, is about operations that silently succeed without a granted channel mode. `uts/` gained a lot in the window, all of it the objects tier plus doc moves.

Corroborating from the other direction: ably-js's own split code cites no spec points either. `packages/shared/side.ts`, `packages/device/src/index.ts` and the two `index.d.ts` files carry the reasoning for the side stamp and the factory, and none of them reference a clause — what they point at is the decision record, `ably/ably-common#361`, and the core's own source.

So: no new spec points covering agent side flags or factory doors, verified at `d9a04cac`, and keeping the per-side claim in `SideSeamTests` is the intentional outcome rather than an overlooked check.

Test/UTS/infra/integration/IntegrationTestCase.swift
maratal: Fixed in #2290. `ProxySession.swift:149` now reads `let client = makeRealtimeForSide(options: options)`, so the example builds its client the way the rest of the suite does.

PR #2278, opened 2026-09-10
Drop CocoaPods and Carthage support

maratal:
Against `integration/v2` directly, not stacked on the split PRs — this is a distribution decision rather than part of the package restructure, and touches a mostly disjoint set of files.

## Why

2.x ships no pod. PDR-091b2 names only a Swift Package Manager product for this SDK, and a pod would have to be a second published artifact with its own trunk ownership, publish ordering and podspec to keep in step with `Package.swift`.

Existing pod consumers are not stranded: the `Ably` pod's last release stays on the 1.x line, which is maintenance-only with security and critical fixes for its support window. To go beyond that they move to Swift Package Manager.

Two consequences worth stating, because neither is obvious:

- A pod consumer **cannot declare the device side**, since the declaration lives in `AblyPubSubDevice`, which is SPM-only. Their traffic keeps being classified by Ably's default rules for as long as they stay on 1.x. That is a real cost of this decision, not a detail.
- **ably-flutter must move to Flutter's Swift Package Manager support.** Its iOS plugin depends on `pod 'Ably', '1.2.53'` and has no `Package.swift`, and a podspec's `s.dependency 'Ably'` can only resolve against a published pod — so the pod path cannot reach 2.x at all. Its users face a breaking major regardless (renamed repo, new `ably_pubsub_device` package, new factory), so the build-system change lands inside a migration they are already making. Worth confirming with that owner, since PDR-091b2 marks ably-flutter as a prerequisite to schedule early.

## What

* **`Ably.podspec`** deleted, along with the `pod_lint` Makefile target.
* **`Examples/Tests`** deleted — an example app whose only purpose was proving `pod install` worked, plus its step in `examples.yaml` and its `Podfile.lock` entry in `.gitignore`. The SPM, Carthage and push examples are untouched.
* **`check-pod.yaml` → `check-carthage.yaml`.** Only two of its steps concerned pods (printing the CocoaPods version, and `pod lib lint`); everything else builds `Ably.framework.zip`, which Carthage consumers install and which the release process attaches to the GitHub release. So the workflow stays and loses the lint. Its job is still named `check`, so no check name changes.
* Docs corrected where the change falsified them: the README's install list and CocoaPods section (replaced by a short notice), `CONTRIBUTING.md`'s distribution table, the four reasons LiveObjects is SPM-only — three of which cited the podspec — the release process, `CLAUDE.md`'s distribution sentence, and the parenthetical in `Docs/plugins.md` that named CocoaPods as a non-SPM build.

## Second commit: Carthage goes too

With the pod gone, Swift Package Manager is the only channel for 2.x. The Carthage `Ably.xcframework`'s last release stays on the 1.x line alongside the pod.

Removed: `Cartfile` and `Cartfile.resolved`, `check-carthage.yaml` (which the first commit had kept for exactly this artifact), the two `Scripts/*carthage*` helpers, every `carthage_*` Makefile target, and `Examples/AblyCarthage` with its step in `examples.yaml`. `make update` now only updates submodules.

Three things worth calling out:

- **`docs.yml` no longer resolves Carthage dependencies** before generating docs. It never needed them: `jazzy.sh` and `generate-markdown-api-reference.py` parse headers out of `Source/`, and no public header imports `msgpack` or `AblyDeltaCodec`, so the step was vestigial.
- **The `Carthage/` entry stays in `.gitignore`**, narrowed to the whole directory. Removing it exposed a leftover local checkout to both `git status` and editorconfig-checker — the linter failed on third-party trailing whitespace under `Carthage/Checkouts/Nimble/…` until it was restored.
- **`Ably.xcodeproj` is kept but is now dormant.** It took its dependencies from `Carthage/Build`, so it has no way to resolve them, and nothing in CI builds its schemes. It stays because `Ably.xcworkspace` references it and it hosts the soak-test app; `CONTRIBUTING.md` now says so rather than leaving the next person to discover it. Removing it properly is a separate change.

`CONTRIBUTING.md`'s Distribution section is rewritten rather than patched: with one channel its premise — that a release tag delivers different products to different channels — no longer holds, and three of the four reasons it gave for `AblyLiveObjects` being SPM-only cited the podspec or the Xcode project.

## Verification

`swift build` clean, `make` parses, editorconfig clean, and a grep for `Ably.podspec`, `Examples/Tests`, `check-pod` and `pod_lint` across the tree returns nothing outside `CHANGELOG.md` and `.github/issue_template.md` — both left alone deliberately, since they describe released 1.x versions and the people filing issues about them.


maratal:
Closed in favour of #2282, which is the same two commits rebased onto `split/uts-per-side` so the whole transition is a single stack. GitHub does not allow a pull request that belongs to a tracked stack to change its base, and closing the pull request above it does not release it, so recreating was the only route. Branch `dist/drop-cocoapods` is unchanged as a name — it now carries the rebased commits.

PR #2280, opened 2026-09-10
Remove Ably.xcodeproj and port the soak test to SPM

maratal:
Stacked on #2278. With CocoaPods and Carthage gone, nothing consumes the hand-maintained Xcode project — except the soak test, which lived in it. This moves the soak test into the package and then deletes the project.

Part of the PubSub device/server package split (PDR-091b). Merge #2278 first.

### 1. Port the soak test to an SPM test target

`Ably-SoakTest-AppUITests` becomes `Test/AblySoakTests`, a test target of the package, and the host app it no longer needs is deleted. It was a UI-test bundle that never touched `XCUIApplication`, so nothing was lost in the move.

It had stopped compiling against the SDK's private API — nothing has built it since the Xcode project stopped resolving its dependencies — so the fakes needed bringing up to date:

- `ARTRealtimeTransportFactory` and `ARTWebSocketTransport` no longer take a connection serial, and `ARTProtocolMessage` no longer has one.
- `ARTConnectionDetails` gained `objectsGCGracePeriod` and `siteCode`.
- `ARTJsonLikeEncoder` gained a `timeProvider` parameter.
- `ARTReachability` takes an `ARTInternalLog` rather than an `ARTLog`.

One fidelity gap needed fixing too: the fake server never stamped a connection id on the protocol messages it sent, and `ARTRealtimePresenceInternal` copies both `connectionId` and `id` onto each presence member it decodes. `ARTPresenceMessage.isSynthesized` then called `hasPrefix:` with a nil argument and the process aborted about forty seconds in. The fake now sets a connection id and ids of the form `<connectionId>:<serial>`, as the server does.

**Not wired into CI, deliberately.** `swift test` runs every test target, and this one is a hundred connections for twenty minutes, so it is skipped unless `RUN_SOAK_TEST` is set. No workflow sets it — the test is a manual tool:

```sh
RUN_SOAK_TEST=1 swift test --filter 'AblySoakTests\.'
```

Verified both ways: `swift test` skips it, and a full `RUN_SOAK_TEST=1` run passes — `Executed 1 test, with 0 failures (0 unexpected) in 1200.018 seconds`.

### 2. Remove Ably.xcodeproj

The project built the three framework targets that the pod and the Carthage xcframework shipped, and hosted the soak test. Both reasons are now gone, and it had no way to resolve its dependencies since they came from Carthage.

`Ably.xcworkspace` stays. It already referenced the repository root, so it now contains only the package; `xcodebuild -list -workspace Ably.xcworkspace` still offers the `ably-cocoa` scheme the Fastlane lanes build. Also deleted are the files the project alone read: `Ably.xcconfig`, `Source/Info-{iOS,tvOS,macOS}.plist` (dropped from `Package.swift`'s `exclude` list with them) and `Source/Ably.modulemap`, which leaves `Source/include/module.modulemap` as the SDK's only module map.

Docs updated to match: `CONTRIBUTING.md` (no more "add it to all three SDK targets" steps, one module map, a Soak test section), `CLAUDE.md` (the no-pbxproj-edit rule has nothing left to protect), `Docs/plugins.md` and `Test/UTS/README.md`.


maratal:
Closed in favour of #2283, same two commits rebased onto the new `dist/drop-cocoapods`. Recreated rather than retargeted because a pull request in a tracked stack cannot change its base.

PR #2287, opened 2026-09-16
Clean up the public API for 2.0: deletions, renames, internal HTTP client, no ART prefix in Swift

maratal:
Stacked on #2285. PDR-091d decided on 2026-09-14 that the new major takes the renamed public surface and drops the deprecated one, on the grounds that the new packages are *"new artifacts with a new major version and no existing consumers… free of backward-compatibility obligation"*.

Ten commits, each of which builds on its own. 322 files.

## Deleted

The three members the headers marked for removal: `ARTClientOptions.fallbackHostsUseDefault`, `ARTConnection.recoveryKey` (a straight alias for `createRecoveryKey`) and `ARTStats.intervalGranularity`. This follows 091d's rule — *"anything already marked deprecated in the current release is a deletion candidate by default; keeping it requires the SDK owner to say why"*.

`ARTFallbackIncompatibleOptionsException` went with them. It was public but raised only by the two guards against combining the removed flag with a custom fallback-host list, so nothing can raise it now. Flagging it because it was not itself marked deprecated.

Eight tests covered the removed property and are gone. A ninth nearly went with them and shouldn't have: RSL1k4's idempotent-publishing test only set the flag to make fallback hosts available, so it now sets `fallbackHosts` directly.

**On the spec.** Removing `fallbackHostsUseDefault` also removes RTN17b3's unconditional fallback retry. That clause no longer specifies anything — `features.md:638` reads *"`(RTN17b3)` This clause has been replaced by `RSC17h` as of specification version 4.0.0"* — and the spec's preamble says support for the option *"is optional. Client libraries that have no requirement for backwards compatibility with these options (for example, new SDKs) may omit support for them entirely."* So no deviation needs recording.

**One behaviour change worth a second look.** The custom `setFallbackHosts:` existed only for one of those guards. Removing it lets the compiler synthesise the setter, which honours the property's `copy` attribute — the hand-written one assigned without copying.

## Renamed

`ARTRest` → `ARTHttpClient`, `ARTRestChannel(s)` → `ARTHttpChannel(s)`, `ARTRestPresence` → `ARTHttpPresence`, `ARTRestAnnotations` → `ARTHttpAnnotations`, with their internal counterparts, protocols, the sixteen files carrying those names, and the seven test suites that exercise them.

`ARTRealtime` → `ARTRealtimeClient`, together with its internal companion `ARTRealtimeInternal` → `ARTRealtimeClientInternal`. `ARTRealtimeChannel`, `ARTRealtimeChannels`, `ARTRealtimePresence` and `ARTRealtimeAnnotations` keep their names, which the spec also uses.

**The stated rationale for that one does not apply here.** 091d justifies it as consistency with a `createRealtimeClient(…)` factory, and per 091b2 this SDK has no such factory — ably-cocoa is device-only, with `PubSubDevice.createClient(options)` as its one door. What carries it instead: the spec already names the interface `RealtimeClient` (`features.md:2140`), so this moves toward the spec rather than away; and it restores the pair with `ARTHttpClient`, since both top-level entry objects now say what they are.

Separately, the private HTTP executor was `ARTHttp`, which said nothing about its job and sat one word from `ARTHttpClient` after the rename. It becomes `ARTHTTPExecutor`, and the protocol it implements becomes `ARTHTTPExecuting`.

## No longer public

**The HTTP client family.** `ARTHttpClient`, `ARTHttpChannel`, `ARTHttpChannels`, `ARTHttpPresence` and `ARTHttpAnnotations` move from `AblyPublic.h` to `AblyInternal.h`, and `ARTHttpClient`'s initializers move to `ARTHttpClient+Private.h`. Two shared query types were declared inside HTTP headers that public realtime headers had to import for them — `ARTPresenceQuery` and `ARTAnnotationsQuery` — so they move to `ARTPresence.h` and `ARTAnnotation.h` and stay documented.

`ARTPushProtocol` declared four class methods taking a `rest:` parameter of that type — the remote- and location-notification registration callbacks — each with a `realtime:` twin. An application has no way to obtain the argument any more, so those four are removed, along with the wrapper proxy's implementations of them. The `realtime:` twins are what an app uses; `Examples/AblyPush/App.swift` calls them. `ARTPushInternal` keeps its own `rest:` set, which is what the push tests call.

**Construction.** `ARTRealtimeClient`'s three initializers move to a Construction category in `ARTRealtimeClient+Private.h`, and the six `NS_SWIFT_UNAVAILABLE` class factories on both top-level classes go entirely. An application builds a client through `PubSubDevice.createClient(options:)`, which declares that the client runs on an end user's device; a client built any other way carries no such declaration. The class itself stays public, since the factory returns one, and tests reach the initializers through the `Private` submodule.

**The internal status types.** `ARTState`, `ARTStatus`, `ARTException` and `ARTClientCodeError` are internal machinery that nothing outside `Source/` uses, and no public declaration names them — `ARTStatusCallback` and two `ARTStatus` forward declarations had no use in the repository at all. They move to private headers. The public header keeps the error model and is renamed `ARTErrorInfo.h` after the class it declares.

**Naming where the internals come from.** Moving those types exposed something worth fixing on its own: twelve test files and five implementation files used SDK internals without importing the header or submodule that declares them. They compiled because neither Clang nor Swift enforces that strictly — a transitive include in Objective-C, and in Swift one `import AblyPubSubDevice.Private` anywhere in a module resolving the name everywhere in it. The last commit makes each of those files say what it depends on.

## No ART prefix in Swift

186 declarations in the public headers gain an `NS_SWIFT_NAME` without the prefix, so Swift now writes `RealtimeClient`, `ClientOptions`, `ErrorInfo` and `Message`. Objective-C is unchanged, and so is the private module: Swift that reaches internals through `import AblyPubSubDevice.Private` still writes `ARTProtocolMessage`.

Two things to know before approving this part.

`Annotation` and `Cancellable` collide with SwiftUI and Combine. Neither breaks anything — a caller disambiguates with `AblyPubSubDevice.Annotation` — but a SwiftUI file that writes the bare name gets an ambiguity error. `ARTState` would have been worse, because it shadows SwiftUI's property wrapper and would have broken every view in a file importing both; it is internal now, so the question does not arise.

Inside this repository, three names need a module qualifier, because `_AblyPluginSupportPrivate` already spells its mirror protocols without a prefix: `RealtimeChannel` in two LiveObjects files, `RealtimeChannel` and `RealtimeChannelState` in `PluginAPITests`, and `AblyLiveObjects`' own internal `PublishResult` in two test helpers. A consumer never sees this, since that module is not a product.

## Through the device door

The README, the LiveObjects README, the SPM example, `Examples/AblyPush`, the LiveObjects example app and the DocC example on `AblyLiveObjects.Plugin` all built clients with the initializer that is now internal. They use `PubSubDevice.createClient(options:)` instead, in the same commit that makes the initializer internal, so no commit in the stack leaves an example unbuildable.

## Deliberately not here

`Test/UTS` keeps its `rest` directories: that layout mirrors the `uts` modules in `ably/specification` (`objects`, `realtime`, `rest`), which `Test/UTS/README.md` keeps in sync with the Kotlin module. Test *method* names still carry the old vocabulary, since they encode the suite path — only the handful naming a renamed symbol were changed. Folding `ARTHttpClient` into `ARTHttpClientInternal` is a separate change, and so is doing anything about `ARTState` and `ARTClientCodeError` being non-spec error-code spaces that leak into `ARTErrorInfo.code`.

## Verification

Every one of the ten commits builds under `swift build --build-tests`. At the tip: clean under `-warnings-as-errors`; `build-for-testing` succeeds on iOS and tvOS simulators; `Examples/SPM` tests, `AblyPushExample` and the LiveObjects example app all build on every platform their CI jobs use; LiveObjects' `sortImports` lint is clean; EditorConfig rules hold across the changed files. Tests pass locally against the sandbox.

The iOS `build-for-testing` leg is the one that matters for the push changes: `PushTests.swift` sits entirely inside `#if os(iOS)`, so a macOS-only check never compiles it.

`Scripts/set-version.sh` lists one renamed test file by path with no existence check, so it is updated in step; otherwise every `make bump_*` would have hit a missing file.


Source/include/AblyPubSubDevice/ARTRealtimeChannel.h
```diff
@@ -1,7 +1,8 @@
 #import <Foundation/Foundation.h>
 
+#import <AblyPubSubDevice/ARTChannelProtocol.h>
+
```
maratal: why this empty line?

Source/include/AblyPubSubDevice/ARTStats.h
```diff
@@ -115,7 +115,7 @@ typedef NS_ENUM(NSUInteger, ARTStatsGranularity) {
 @property (readonly, nonatomic) ARTStatsMessageTypes *realtime;
 
 /**
- * A `ARTStatsMessageTypes` object containing a breakdown of usage by message type for messages transferred over a rest transport such as `ARTHttpClient`.
+ * A `ARTStatsMessageTypes` object containing a breakdown of usage by message type for messages transferred over a REST transport.
```
maratal: why this change? "REST"

Source/ARTHttpClient.m
maratal, line 1: I beleive these changes should be in a different commit

CLAUDE.md
```diff
@@ -58,6 +58,8 @@ All SDK source is in `Source/`, written entirely in Objective-C:
 
 Key classes follow the `ART` prefix convention: `ARTRealtimeClient`, `ARTHttpClient`, `ARTAuth`, `ARTChannel`, `ARTConnection`, `ARTPresence`, `ARTPush`.
 
+Swift sees these names without the prefix: `RealtimeClient`, `HttpClient`, `Auth` and so on. Every declaration in the module's public headers carries an `NS_SWIFT_NAME` that strips `ART`, so a new public declaration needs one too. The prefix stays in Objective-C and in the private headers, so Swift code that reaches internals through `import AblyPubSubDevice.Private` still writes `ARTProtocolMessage` and the like.
```
maratal: "so Swift code that reaches internals through `import AblyPubSubDevice.Private` still writes `ARTProtocolMessage` and the like" - not sure about this.

_AblyPluginSupportPrivate/include/APRealtimeClient.h
```diff
@@ -4,7 +4,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// The interface that plugins use to interact with a realtime client.
 ///
-/// This exists so that plugins do not need to make use of the public `ARTRealtime` class, which allows the internal components of ably-cocoa to continue the (existing before we introduced plugins) pattern of also not making use of this public class.
+/// This exists so that plugins do not need to make use of the public `ARTRealtimeClient` class, which allows the internal components of ably-cocoa to continue the (existing before we introduced plugins) pattern of also not making use of this public class.
```
sacOO7, line 7: The file is called `APRealtimeClient.h` ?

sacOO7, line 7:
```suggestion
/// This exists so that plugins do not need to make use of the public `RealtimeClient` class, which allows the internal components of ably-cocoa to continue the (existing before we introduced plugins) pattern of also not making use of this public class.
```

sacOO7: If we decided to remove `ART` from everywhere, then we should scan code for all `ART` prefixes and get rid of them. I can see `ART` prefix present at number of places

maratal: that's only true for swift sources, objc keeps this notation and would mislead reader without proper prefix

maratal: yes, not to confuse with another objc header - `ARTRealtimeClient.h`

sacOO7: @coderabbitai what do you think about this

sacOO7: @coderabbitai do you think we should rename it to something else like `LegacyRealtimeClient` or something like that?

ttypic (approved):

PR #2288, opened 2026-09-16
Serialize the LiveObjects integration suites

maratal:
Each case in `LiveObjects/Tests/AblyLiveObjectsTests/Integration` stands up its own realtime client, its own channel and its own object sync against the sandbox:

```swift
let realtime = try await ClientHelper.realtimeWithObjects(…)
let channel = realtime.channels.get(UUID().uuidString, …)
try await channel.attachAsync()
let root = try await channel.object.get()
```

swift-testing runs a suite's cases in parallel unless told otherwise, and none of these three suites said otherwise. `NumberPrecisionTests` alone is four test functions over three argument values, so twelve concurrent clients.

## What that does on CI

From the `Xcode, iOS (Xcode 16.4)` job on run 35081997799:

```
09:58:33  ◇ Test mapNumberSurvivesJSONRoundTrip started
09:58:33  ◇ Test counterValueSurvivesJSONRoundTrip started
09:58:33  ◇ Test mapNumberSurvivesMessagePackRoundTrip started
09:58:33  ◇ Test counterValueSurvivesMessagePackRoundTrip started
…
10:01:06  ✔ mapNumberSurvivesJSONRoundTrip passed after 127.618 seconds
10:01:09  ✘ counterValueSurvivesJSONRoundTrip … Error -1001 - The request timed out
```

A test that performs one `set` and one poll took over two minutes. The counter tests make twice the round trips — create, poll, increment, poll — so they were the ones to cross `SandboxEnvironment.provisioningTimeout` (30s) and fail with `NSURLErrorTimedOut`.

## It is not specific to any branch

Per-attempt data for the same job, which run-level conclusions hide because the failures were cleared by re-running:

| PR | Branch | Attempt | Failed |
| --- | --- | --- | --- |
| #2267 | `split/agent-identifiers` | 2 of 3 | Xcode, iOS |
| #2274 | `ci/retry-absorbs-flakes` | 2 of 3 | Xcode, iOS |
| #2275 | `split/uts-per-side` | 2 of 3 | Xcode, iOS and tvOS |
| #2284 | `deps/raise-platform-floor` | 1 of 2 | Xcode, iOS |
| #2286 | `fix/test-crash-on-timeout` | 1 | Xcode, iOS |

Those five branches have nothing in common, `NumberPrecisionTests` came from `main` in `fbb638b3` and none of them touch it, and tvOS is affected too — so this is contention, not a platform quirk. macOS has never failed this way, being the host rather than a simulator.

`LiveObjects/BuildTool` passes no `-retry-tests-on-failure`, so one timeout in one of twelve concurrent cases reddens the whole job and somebody has to press the button. Adding that flag is a separate, complementary change.

## The convention already exists

`Test/UTS/README.md` states it for the UTS integration tier — *"Suites are `@Suite(.serialized) final class … : UTSTestCase`"* — and `IntegrationTestCase` repeats it. These three suites never adopted it.

## The trade, measured

Serializing is **slower where there is no contention**, and I checked rather than assumed: locally `NumberPrecisionTests` goes from 3.6s to 11.9s. That is the cost. The benefit is on CI, where the parallel version currently spends minutes and then fails — twelve sequential cases at a few seconds each should beat twelve concurrent ones at 128 seconds each.

I cannot demonstrate the CI improvement from a local run, because the contention comes from the runner's latency to the sandbox. CI on this PR is the measurement.

## Verification

All three suites pass serialized: `NumberPrecisionTests` (4 tests, 11.4s), `AblyLiveObjectsTests` (418 tests in 109 suites, 11.7s), `ObjectLifetimesTests` (3 tests, 3.3s).

## Summary by CodeRabbit

- **Tests**
  - Updated integration test suites to run serially rather than in parallel.
  - No test logic or end-user functionality changed.


ttypic (approved):
LGTM

PR #2291, opened 2026-09-18
Test that the device factory carries over the plugins

maratal:
Stacked on #2265, and addresses the MAJOR finding on it (https://github.com/ably/ably-cocoa/pull/2265#discussion_r4026592298).

`PubSubDevice.createClient(options:)` works by copying the caller's options, and nothing checked that `ARTClientOptions.plugins` survives that copy. Losing it is invisible at construction and fatal later: a client built without the plugin traps on first access to `channel.object`, so every LiveObjects application reaching Ably through the factory would break.

`test__011__createClient__carries_over_the_plugins` puts a plugin in the options, builds through the factory, and reads the plugins back off the client. It sits with the five other contract tests for the same call and uses their idiom, `client.internal.options`.

No manifest change was needed. The finding expected `AblyTests` to gain a dependency on `AblyLiveObjects`; instead the test reuses `PluginAPITests.MockLiveObjectsPlugin`, which already exists in this target and already conforms to the real `LiveObjectsPluginProtocol`. So the core SDK's test target stays clear of the plugin, and the test needs no network — it runs in every existing lane on all three platforms.

Checked that it can fail: with `options.plugins = self.plugins;` removed from `ARTClientOptions.m`'s `copyWithZone:`, it fails with `XCTUnwrap failed: expected non-nil value of type "Dictionary<ARTPluginName, Any>"`.

The behavioural half of the finding — that `channel.object` works on a factory-built client — is #2292, at the top of the stack where the LiveObjects test target already depends on `AblyPubSubDevice`.

## Summary by CodeRabbit

- **Tests**
  - Added coverage to verify that client creation preserves the caller’s configured plugins.


sacOO7 (approved):
LGTM

ttypic (approved):

PR #2294, opened 2026-09-21
Rename ARTRealtimeClient to ARTPubSubClient

maratal:
The product is called Pub/Sub, so the client class says so.

`ARTRealtimeClient` becomes `ARTPubSubClient`, and its internal counterpart `ARTRealtimeClientInternal` becomes `ARTPubSubClientInternal`. Swift follows through `NS_SWIFT_NAME` and writes `PubSubClient`. Four files move with the class: `ARTPubSubClient.h`, `ARTPubSubClient.m`, `ARTPubSubClient+Private.h` and `ARTPubSubClient+WrapperSDKProxy.h`.

## Naming of parameters

Parameters typed as either class are named for it, with the casing rule that a name beginning with the product word spells it lowercase throughout and mid-identifier it is capitalised:

| Before | After |
|---|---|
| `realtime:(ARTPubSubClient *)realtime` | `pubsub:(ARTPubSubClient *)pubsub` |
| `initWithRealtime:(ARTPubSubClient[Internal] *)realtime` | `initWithPubSub:(…)pubsub` |
| `realtimeInternal:(ARTPubSubClientInternal *)realtimeInternal` | `pubsubInternal:(…)pubsubInternal` |

So `ARTPush`'s four class methods now read `pubsub:`, and from Swift that is `Push.didRegisterForRemoteNotifications(withDeviceToken:pubsub:)`.

The example apps say `client` instead. They hold one client and the type beside it already names the product, so `AblyHelper.client` and the view models' `init(client:)` read better than repeating it.

## Deliberately not renamed

- `ARTRealtimeChannel`, `ARTRealtimeChannels`, `ARTRealtimePresence`, `ARTRealtimeAnnotations` and their protocols.
- `ARTRealtimeProtocol` and `ARTRealtimeInstanceMethodsProtocol`, which this class conforms to.
- `ARTRealtimeConnectionState`, `ARTRealtimeChannelState`, `ARTRealtimeConnectionEvent` and their cases.
- The transport layer: `ARTRealtimeTransport` and friends describe the transport rather than the product, as does `ARTStats.realtime`, which the specification defines alongside `rest` and `webhook`.
- LiveObjects' `RealtimeObject`, which `objects-features.md` names 29 times.
- `_AblyPluginSupportPrivate`'s `APRealtimeClient`: a mirror protocol in a module of its own, keeping the Swift name `RealtimeClient`, so nothing becomes ambiguous.
- Four SDK properties of type `ARTPubSubClientInternal` still named `realtime`, and the five `RealtimeClient*Tests.swift` suites. Both are easy follow-ups if wanted.

**On the specification.** [`ably/specification`](https://github.com/ably/specification) at [`d9a04cac`](https://github.com/ably/specification/commit/d9a04cac2fbbb3847c940f67f5ad52d2d75eef26) (their `main` on 2026-09-17) still names this interface `RealtimeClient` in `features.md`, 27 times, and has no `PubSub*` type names. This is a deliberate divergence on product naming, not an oversight, and no deviation is recorded for it.

## Verification

`swift build --build-tests`, clean under `-warnings-as-errors`. `build-for-testing` succeeds on iOS and tvOS simulators. `Examples/SPM` tests build; `AblyPushExample` builds on iOS; the LiveObjects example app builds on iOS, macOS and tvOS. LiveObjects' `sortImports` is clean and SwiftLint reports its usual six pre-existing findings. 48 tests pass locally.

Worth noting one thing the package build could not catch: after the parameter rename, `ARTWrapperSDKProxyPush.m` still forwarded with the old label, which SPM reported only as a warning while the `AblyPushExample` Xcode project failed on it. The example builds are load-bearing here.

## Summary by CodeRabbit

* **Breaking Changes**
  * Renamed the public client from `RealtimeClient` to `PubSubClient`.
  * Updated client factory methods to return `PubSubClient`.
  * Updated push registration APIs to accept Pub/Sub clients and use `pubsub` parameters.

* **Documentation**
  * Updated guides, examples, and integration documentation to reflect the new client name and lifecycle behavior.

* **Tests**
  * Updated unit, integration, soak, and UTS tests to use `PubSubClient` while preserving existing coverage and behavior.


ttypic (approved):
LGTM

